PR #57, opened 2017-08-31
Fix typo in description default value

poum:


bobbydeveaux:
Not required as we have now removed this default value.

PR #59, opened 2017-10-02
[Improvement] Add multiple networks

dn1s:
Implemented functionality of adding multiple networks at once to a vapp/vm.

PR #74, opened 2017-12-05
[WIP] Merge in code as per v2Plan.md

bobbydeveaux:

PR #86, opened 2018-03-20
[Improvement] Add multiple networks (supersedes #59)

tlawrence:
ref #59 

This PR is to attempt to fix the issues therein.

### DIFF
--- a/vcd/resource_vcd_vapp.go
+++ b/vcd/resource_vcd_vapp.go
@@ -522,6 +522,11 @@ func getVAppNetwork(d *schema.ResourceData, meta interface{}) ([]map[string]inte
 		// 'first' one, and tests will fail sometimes (annoying huh?)
 		vm, err := vcdClient.OrgVdc.FindVMByName(vapp, d.Get("name").(string))
 
+		if err != nil {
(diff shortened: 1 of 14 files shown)


tlawrence:
Multi VM per vApp & Multi NIC are now working but with some bugs.

Logic needs adding to govcloudair vapp.go AddVM and RemoveVM function to rebuild the "NetworkConfigSection" when VMs are added or removed. Otherwise adding nice on network which are not in the list fails. 

Additionally, when I remove a NIC and run terraform apply, The vapp_vm resource is flagged for recreation but deletion of the existing VM fails. Not sure why. 


Possible stretch goal: Work out how to add and remove NICs without recreating VMs.

tlawrence:
@dn1s any comments on this?

dn1s:
Well biggest problem here with this pr and with mine as well is terraform state handling is broken.
Which results in onstant "changed" states when running `terraform plan` for ips of network adapters.

dn1s:
#74 is working for me please use this pr

spengilley:
Will this PR cover multiple networks for the VM?

Didainius:
I believe #177 should give even more than this PR and therefore am closing it. Do ping us if you think otherwise.

PR #104, opened 2018-10-26
Update vendor directory

dataclouder:
Refresh vendor directory with latest vmware/go-vcloud-director code
(include updated network creation function and catalog upload work)
Code imported using govendor.

Signed-off-by: Giuseppe Maxia

### DIFF
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -1,6 +1,6 @@
 {
 	"comment": "",
-	"ignore": "test",
+	"ignore": "appengine test github.com/hashicorp/nomad/ github.com/hashicorp/terraform/backend",
 	"package": [
 		{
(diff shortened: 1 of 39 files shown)


vendor/vendor.json
```diff
@@ -1,6 +1,6 @@
 {
 	"comment": "",
-	"ignore": "test",
+	"ignore": "appengine test github.com/hashicorp/nomad/ github.com/hashicorp/terraform/backend",
```
lvirbalas: Where are these these additional ignores coming from and why do we need them? I can't find any references of them in our current code.

dataclouder: They might be a leftover from a manual import that I then aborted.
I see that they don't affect the rest of the vendor directory.
I will remove them at my next import (Terraform upstream code)

Issue #96, opened 2018-10-01
Support Isolated Org VDC Network type


lvirbalas:
Feature request to support creating isolated VDC networks

### Description

Currently, you can create a network which is connected to the gateway. For example:

```hcl
# Org VDC network
resource "vcd_network" "TerraformNetwork" {
  name         = "TerraformNetwork"
  edge_gateway = "EdgeToCorporate"

  gateway = "192.168.0.1"

  static_ip_pool {
    start_address = "192.168.0.2"
    end_address   = "192.168.0.100"
  }

  dhcp_pool {
    start_address = "192.168.0.101"
    end_address   = "192.168.0.200"
  }
}
```

However, there is no way to create a network, which is **not** attached neither to Edge Gateway nor to External Network.

### Affected Resource(s)

Only one resource should be affected:
- vcd_network

### Terraform Configuration Files

The following configuration is suggested (**in addition** to the currently supported configuration):

```hcl
# Org VDC network
resource "vcd_network" "IsolatedNetwork" {
  name         = "IsolatedNetwork"

  gateway = "192.168.0.1"

  static_ip_pool {
    start_address = "192.168.0.2"
    end_address   = "192.168.0.100"
  }

  dhcp_pool {
    start_address = "192.168.0.101"
    end_address   = "192.168.0.200"
  }
}
```

The only difference in the above configuration is that the `edge_gateway` parameter has been removed. The lack of this parameter should be treated as an ask for an isolated network.

### Expected Behavior

Terraform vCD provider should create an isolated Org VDC Network. This is analogous to manually going to vCD GUI -> Org VDC Networks -> Add Network... -> select the first option "Create an isolated network within this virtual datacenter". 

### Important Factoids

This is additional feature. Current ability to create a network with an Edge Gateway should remain intact.

lvirbalas:
@dataclouder I'm hitting a bug in isolated network DHCP pool creation. I'm asking for DHCP to use 192.168.0.101-192.168.0.111 range, but when I check in vCD it uses 192.168.0.101-192.168.0.254 range. Here's my network configuration:
`
resource "vcd_network_isolated" "TerraformNetworkIsolated" {
  name       = "TerraformNetworkIsolated"
  gateway    = "192.168.0.1"
  dns1       = "192.168.0.1" // TODO: Why/do we need "1" in the key name?
  static_ip_pool {
    start_address = "192.168.0.2"
    end_address   = "192.168.0.100"
  }
[...]

Issue #95, opened 2018-10-01
Add Direct connection type to Org VDC Network for attaching to External Networks


lvirbalas:
Feature request to support creating networks, which are directly attached to External Networks at System level

### Description

Currently, you can create a network which is connected to the gateway. For example:

```hcl
# Org VDC network
resource "vcd_network" "TerraformNetwork" {
  name         = "TerraformNetwork"
  edge_gateway = "EdgeToCorporate"

  gateway = "192.168.0.1"

  static_ip_pool {
    start_address = "192.168.0.2"
    end_address   = "192.168.0.100"
  }

  dhcp_pool {
    start_address = "192.168.0.101"
    end_address   = "192.168.0.200"
  }
}
```

However, there is no way to create a network, that is directly attached to a System External Network. Let's implement that.

### Affected Resource(s)

Only one resource should be affected:
- vcd_network

### Terraform Configuration Files

The following configuration is suggested (**in addition** to the currently supported configuration):

```hcl
# Org VDC network
resource "vcd_network" "TerraformNetwork" {
  name         = "TerraformExtNetwork"

  # The new parameter which when defined connects to this network directly without router
  external_network = "CorporateNetwork"
}
```

### Expected Behavior

Terraform vCD provider should create a directly attached Org VDC Network. This is analogous to manually going to vCD GUI -> Org VDC Networks -> Add Network... -> select the third option "Connect directly to an external network". 

### Important Factoids

This is additional feature. Current ability to create a network with an Edge Gateway should remain intact.

PR #105, opened 2018-10-29
 Add OrgVDC network types routed, bridged, isolated

dataclouder:
* Deprecate current vcd_network, replaced by vcd_network_routed.
* Implement vcd_network_routed, as an Org VDC network with mandatory Edge
  Gateway.
* Implement vcd_network_direct, with a mandatory connection to an External
  network.
* Implement vcd_network_isolated, with no external connections.
* Implement tests for all the three network types.
* Implement an optional method to add the provider to every Terraform snippet
  used for testing. (Triggered by env variable VCD_ADD_PROVIDER).
* Update .gitignore with common test artifacts
* Remove deprecated property maxRetryTimeout.
* Add new network resources to vcd provider.
* Remove explicit calls to vdc.Refresh prior to vdc.Find* calls, which include
 a refresh in go-vcloud-director code.
* Replace vcd_network with vcd_network_routed in other tests.

Related Issues: #95 and #96 


vcd/config_test.go
```diff
@@ -207,6 +236,7 @@ func getConfigStruct() TestConfig {
 	if config_struct.Provider.SysOrg == "" {
 		config_struct.Provider.SysOrg = config_struct.VCD.Org
 	}
+	// fmt.Printf("[%s] config_struct %#v\n", callFuncName(), config_struct)
```
lvirbalas: Can we remove commented code here?

vcd/provider.go
```diff
@@ -76,16 +76,17 @@ func Provider() terraform.ResourceProvider {
...
-			"vcd_vapp":            resourceVcdVApp(),
-			"vcd_firewall_rules":  resourceVcdFirewallRules(),
-			"vcd_dnat":            resourceVcdDNAT(),
-			"vcd_snat":            resourceVcdSNAT(),
-			"vcd_edgegateway_vpn": resourceVcdEdgeGatewayVpn(),
-			"vcd_vapp_vm":         resourceVcdVAppVm(),
-			"vcd_org":             resourceOrg(),
-			"vcd_catalog":         resourceVcdCatalog(),
-			"vcd_catalog_item":    resourceVcdCatalogItem(),
+			"vcd_network":          resourceVcdNetwork(), // DEPRECATED
```
lvirbalas: Please expand comment to "DEPRECATED, replaced by vcd_network_routed"

vcd/resource_vcd_network.go
```diff
@@ -187,25 +208,28 @@ func resourceVcdNetworkCreate(d *schema.ResourceData, meta interface{}) error {
...
 	if err != nil {
-		return fmt.Errorf("Error: %#v", err)
+		return fmt.Errorf("error: %#v", err)
 	}
 
-	err = vdc.Refresh()
-	if err != nil {
-		return fmt.Errorf("error refreshing VDC: %#v", err)
-	}
+	// err = vdc.Refresh()
```
lvirbalas: Why Refresh() is not needed here anymore?

vcd/resource_vcd_vapp.go
```diff
@@ -452,10 +452,10 @@ func getVAppIPAddress(d *schema.ResourceData, meta interface{}, vdc govcd.Vdc, o
 	vcdClient := meta.(*VCDClient)
 	var ip string
 	err := retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
-		err := vdc.Refresh()
-		if err != nil {
-			return resource.RetryableError(fmt.Errorf("error refreshing VDC: %#v", err))
-		}
+		// err := vdc.Refresh()
```
lvirbalas: Why we don't need vdc.Refresh() here?

vcd/resource_vcd_network.go
```diff
@@ -151,7 +155,7 @@ func resourceVcdNetwork() *schema.Resource {
 	}
 }
 
-func resourceVcdNetworkCreate(d *schema.ResourceData, meta interface{}) error {
+func resourceVcdNetworkDeprecatedCreate(d *schema.ResourceData, meta interface{}) error {
```
lvirbalas: Do we really need to mark deprecation of this function? Having "Deprecated" inside the name looks nasty. We should be deprecating the whole resource instead.

Can we make resource_vcd_network.go and the resource_vcd_network_routed.go, which replaces it, be functionally identical? Meaning that the old resource_vcd_network.go would just call the new resource_vcd_network_routed? In that case, even if someone were using the deprecated resource_vcd_network.go, it would be fine. We would only want to print a warning.

vcd/resource_vcd_network.go
dataclouder: Because `vdc.FindVDCNetwork` already includes it.

vcd/resource_vcd_vapp.go
dataclouder: `vdc.FindVAppByName` already has an implicit Refresh

vcd/provider.go
```diff
@@ -95,10 +96,10 @@ func Provider() terraform.ResourceProvider {
...
 	maxRetryTimeout := d.Get("max_retry_timeout").(int)
 
-	// TODO: Deprecated, remove in next major release
-	// if v, ok := d.GetOk("maxRetryTimeout"); ok {
-	// 	maxRetryTimeout = v.(int)
-	// }
+	// It was deprecated, and now removed
+	//if v, ok := d.GetOk("maxRetryTimeout"); ok {
+	//	maxRetryTimeout = v.(int)
+	//}
```
vbauzysvmware: Could you elaborate here what happen?

vcd/resource_vcd_network.go
```diff
@@ -162,22 +166,39 @@ func resourceVcdNetworkCreate(d *schema.ResourceData, meta interface{}) error {
...
-	if err != nil {
-		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	// Collecting components
+	egName, ok := d.GetOk("edge_gateway")
+	var edgeGateway govcd.EdgeGateway
+	fenceMode := "natRouted"
+
+	edgeGatewayName := ""
+	if ok {
+		edgeGatewayName = egName.(string)
```
vbauzysvmware: I don't see usage 'edgeGatewayName' in code. Directly egName can be used.

vcd/resource_vcd_network.go
```diff
@@ -187,25 +208,28 @@ func resourceVcdNetworkCreate(d *schema.ResourceData, meta interface{}) error {
...
-		},
 		IsShared: d.Get("shared").(bool),
 	}
 
-	log.Printf("[INFO] NETWORK: %#v", newnetwork)
+	orgVDCNetwork.EdgeGateway = &types.Reference{
+		HREF: edgeGateway.EdgeGateway.HREF,
+	}
+
+	util.Logger.Printf("[INFO] NETWORK: %#v", orgVDCNetwork)
```
vbauzysvmware: Is this info really intended to be INFO level?

vcd/resource_vcd_network.go
vbauzysvmware: So shouldn't we remove commented out code?

vcd/resource_vcd_network.go
```diff
@@ -242,10 +266,10 @@ func resourceVcdNetworkRead(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf(errorRetrievingOrgAndVdc, err)
 	}
 
-	err = vdc.Refresh()
-	if err != nil {
-		return fmt.Errorf("error refreshing VDC: %#v", err)
-	}
+	// err = vdc.Refresh()
```
vbauzysvmware: Why commented out code isn't removed? In all places.

vcd/resource_vcd_network_direct.go
```diff
@@ -0,0 +1,115 @@
...
+				Type: externalNetwork.Type,
+				Name: externalNetwork.Name,
+			},
+			FenceMode:                 "bridged",
+			BackwardCompatibilityMode: true,
+		},
+		IsShared: d.Get("shared").(bool),
+	}
+
+	util.Logger.Printf("[INFO] NETWORK: %#v", orgVDCNetwork)
```
vbauzysvmware: This will look like random messages in info logs with strange information about what happend. "NETWORK:". Another question do we really want this in info level.

vcd/resource_vcd_network_routed.go
```diff
@@ -0,0 +1,245 @@
...
+					DNSSuffix:   d.Get("dns_suffix").(string),
+					IPRanges:    &ipRanges,
+				},
+			},
+			BackwardCompatibilityMode: true,
+		},
+		IsShared: d.Get("shared").(bool),
+	}
+
+	util.Logger.Printf("[INFO] NETWORK: %#v", orgVDCNetwork)
```
vbauzysvmware: This will look like random messages in info logs with strange information about what happend. "NETWORK:". Another question do we really want this in info level.

vcd/resource_vcd_network.go
lvirbalas: Thank you. We don't like too much code :)

vcd/config_test.go
```diff
@@ -207,6 +236,7 @@ func getConfigStruct() TestConfig {
 	if config_struct.Provider.SysOrg == "" {
```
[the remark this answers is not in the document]
dataclouder: Will change it

vcd/provider.go
dataclouder: A messed up merge, I assume. Will clean it up.

vcd/resource_vcd_network.go
dataclouder: This code will disappear. Anyway, `egName` is `interface{}`, while `edgeGatewayName` is a string.

vcd/resource_vcd_network.go
dataclouder: I reused the logging calls from existing code, changing them to use our logger. Given the better logging that we have in go-vcloud-director, I could even remove these lines.

vcd/resource_vcd_network_direct.go
dataclouder: I will remove it.

vbauzysvmware:
Manually merged here: https://github.com/terraform-providers/terraform-provider-vcd/pull/107

PR #119, opened 2018-11-22
Fix vApp/VM provisioning guest customization initscript

pasikarkkainen:
Guest Customization is a feature used by vCloud Director during vApp provisioning to configure various settings inside the VM guest operating system (via vmware-tools):
- Set hostname.
- Set/change local root/admin password.
- Set network settings (static IP, dhcp, etc).
- Run "initscript" inside the VM on next/first power-on, enabling automated ssh authorized keys configuration etc to get unattended remote management and devops tools working.
 
Currently terraform-provider-vcd enables guest customization for the vApp when configuring "initscript" for the vApp in Terraform settings. Unfortunately guest customization / initscript is enabled by terraform-provider-vcd in the wrong place/order. Currently it's enabled *after* vApp power-on is executed, which is wrong and results in the scenario where guest customization and initscript do not work automatically during vApp provisioning.. instead they get executed during the *next* vApp power-off/power-on cycle after provisioning, which clearly isn't ideal or intended, and results in Terraform provisioned vApps having wrong hostname, wrong root/admin passwords, wrong networking settings, and most importantly (because the initscript isn't executed) things like ssh authorized keys are not getting set inside the VMs, preventing unattended remote management by Terraform remote provisioners, ansible and other devops tools.

This PR changes the order of actions and calls vapp.RunCustomizationScript() before running vapp.PowerOn(), which makes Guest Customization and initscripts working properly during the vApp provisioning process.

I've tested the patch now against vCloud Director 9.1, and similar patch earlier against vCloud Director 8.20 and 9.0.

Fixes at least the following terraform-provider-vcd issues:
https://github.com/terraform-providers/terraform-provider-vcd/issues/62
https://github.com/terraform-providers/terraform-provider-vcd/issues/88

This supersedes the earlier PR (https://github.com/terraform-providers/terraform-provider-vcd/pull/92), which was against the old codebase in the master branch.

### DIFF
--- a/vcd/resource_vcd_vapp.go
+++ b/vcd/resource_vcd_vapp.go
@@ -211,20 +211,6 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
 				}
 			}
 
-			if d.Get("power_on").(bool) == true {
-				err = retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
-					task, err := vapp.PowerOn()
-					if err != nil {
-						return resource.RetryableError(fmt.Errorf("error powerOn machine: %#v", err))
-					}
-					return resource.RetryableError(task.WaitTaskCompletion())
-				})
-
-				if err != nil {
-					return fmt.Errorf("error completing powerOn tasks: %#v", err)
-				}
-			}
-
 			initscript, ok := d.GetOk("initscript")
 			if ok {
 				err = retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
@@ -237,6 +223,20 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
 				})
 				if err != nil {
 					return fmt.Errorf(errorCompletingTask, err)
+				}
+			}
+
+			if d.Get("power_on").(bool) == true {
+				err = retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
+					task, err := vapp.PowerOn()
+					if err != nil {
+						return resource.RetryableError(fmt.Errorf("error powerOn machine: %#v", err))
+					}
+					return resource.RetryableError(task.WaitTaskCompletion())
+				})
+
+				if err != nil {
+					return fmt.Errorf("error completing powerOn tasks: %#v", err)
 				}
 			}
 


pasikarkkainen:
@dataclouder : hmm, is it possible currently in the tests to connect to the provisioned guest vm over the network (ssh), to verify the initscript provisioned ssh authorized_keys, network settings, etc ?

I guess I'll have a look at the existing tests.

dataclouder:
I don't think we have SSH capabilities in the current library. Is there any external clues that we can leverage to check whether the operation was successful? 
I am not familiar (yet) with this particular code, as it was developed before we took over.

dn1s:
Please first check if this https://www.terraform.io/docs/provisioners/local-exec.html is what you need before implementing it yourself in this provider.

pasikarkkainen:
@dn1s : hmm, do you mean leveraging terraform local-exec for testing the remote connections to provisioned VMs by running a local script which uses ssh etc commands?

Here's an example how to test and use the existing terraform-provider-vcd "initscript" / vCloud Director Guest Customization feature. 

"initscript" setting is used to deploy ssh authorized_keys to the newly provisioned guest VM via vCloud Director Guest Customization (which uses vmware-tools in the guest VM) making it possible for Terraform to do an unattended ssh to the newly provisioned VM. Terraform "file" and "remote-exec" provisioners are then used to copy a local script to the remote VM and execute it there over unattended ssh as part of the provisioning process.

``
resource "vcd_vapp" "jumpbox" {
    name          = "${var.jumpbox_hostname}"
    catalog_name  = "${var.vcd_catalog}"
    template_name = "${var.vapp_centos7_template}"
    memory        = 1024
    cpus          = 1
    network_name  = "${vcd_network.mgt_net.name}"
    ip            = "${var.jumpbox_int_ip}"
    initscript    = "mkdir -p ${var.ssh_user_home}/.ssh; echo \"${data.null_data_source.ssh.inputs.pubkey}\" >> ${var.ssh_user_home}/.ssh/authorized_keys; chmod -R go-rwx ${var.ssh_user_home}/.ssh; restorecon -Rv ${var.ssh_user_home}/.ssh"
    power_on = true

    depends_on    = [ "vcd_dnat.jumpbox-ssh", "vcd_firewall_rules.jumpbox-fw", "vcd_snat.mgt-outbound" ]

    connection {
        host = "${var.jumpbox_ext_ip}"
        user = "${var.ssh_user}"
        private_key = "${data.null_data_source.ssh.inputs.privkey}"
        timeout = "7m"
    }

    provisioner "file" {
        source = "config.sh"
        destination = "/tmp/config.sh"
    }

    provisioner "remote-exec" {
        inline = [
          "chmod +x /tmp/config.sh",
          "/tmp/config.sh",
          "reboot"
        ]
    }

}
``

dn1s:
Yeah @pasikarkkainen I'm sorry I deleted my comment since it was not related to go tests. It was related to testing locally.

lvirbalas:
Hi @pasikarkkainen , would you be open to implementing the test for this as per Vaidotas' comment above?

lvirbalas:
@pasikarkkainen what about @vbauzysvmware suggestion above of verifying that initscript worked indirectly:

> Is it possible to with initscript assign IP address and then verify it by geting VM details using VCD API (go-vcd existing function FindVMByName) ?

vbauzysvmware:
Possible flow would be like this:

* Create VAPP/VM
* Create init script which create network with static ip adress
* Assign init script for VM
* Startup VM
* Wait for VM starting up
* Query VM and check NetworkConnection details.
* Verify NIC static ip adress matches expected one in init script

> @pasikarkkainen what about @vbauzysvmware suggestion above of verifying that initscript worked indirectly:
> 
> > Is it possible to with initscript assign IP address and then verify it by geting VM details using VCD API (go-vcd existing function FindVMByName) ?

vbauzysvmware:
@pasikarkkainen I haven't checked yet, but it is possible as you said, then only possible options is to use SSH which not very nice :)

lvirbalas:
@pasikarkkainen What about using `ping` for the test? That would be simpler than SSH. You can check whether the IP has been successfully changed with ping.

vbauzysvmware:
For ping possible to use: https://github.com/tatsushid/go-fastping as there isn't anything in standard library.

pasikarkkainen:
Thinking more about this: Changing the guest IP address / networking settings is part of the "standard" vcloud guest customization feature (which happens via vmware-tools aswell), there's no need to use the initscript for that.. 

initscript is used for *additional* guest customization. When using Terraform additional customization often means things like setting up the ssh authorized keys in the guest (like in my example script above) for unattended ssh access to work, or tweaking guest firewall rules, or tweaking some other system settings inside the guest VM.. whatever magic is required to happen in the provisioning workflow.

So the "standard" vcloud guest customization stuff could by tested with for example ping, but the initscript feature requires more advanced testing functionality, for example verifying that unattended ssh to the guest works using ssh keys.

lvirbalas:
Implement test for guest customization initscript #139

Issue #112, opened 2018-11-15
Document new features in vmware-takes-ownership branch


lvirbalas:
Now that the movements in vmware-takes-ownership branch stabilised, we need to bring website documentation up to date. Namely, this includes the following.

New resources:
- Org resource
- Catalog resource
- Catalog item resource
- Isolated network resource
- Directly attached network resource

Resource rename:
- vcd_network -> vcd_network_routed

Update on capabilities:
- A line that we now officially support vCD versions 8.20, 9.0, 9.1 and 9.5
- Ability to connect as sysadmin
- Default Org setting at `provider` configuration level:
```
provider "vcd" {
  sysorg               = "System"
  org                  = "Solpan"
  vdc                  = "SolpanVDC"
...
```
- Different org support at resource level:
```
resource ... {
  org = "Solpan"
  vdc = "SolpanVDC"
...
```

And, overall, compare the docs in master and see what else would be useful to add.

lvirbalas:
@vbauzysvmware thank you for `catalog_item.html.markdown` and `catalog_media.html.markdown` doc additions! Please also provide a doc file for catalog itself.

lvirbalas:
@vbauzysvmware please also update documentation for SNAT/DNAT rules with the changes you made earlier (protocol, etc.).

lvirbalas:
@dataclouder please write a documentation page for the new Org resource (`resource_vcd_org.go`). Hint: most of the parameters are in vCD API docs, e.g. here https://code.vmware.com/apis/287/vcloud

lvirbalas:
Docs are updated and in the master PR #143.

PR #124, opened 2018-12-07
Update doc entry page with features in vmware-takes-ownership branch

lvirbalas:
- Covers changes in "provider" config level
- "v2.0+" keyword is used to emphasize semantic differences of the new
  version

Part of issue #112

Signed-off-by: Linas Virbalas <lvirbalas@vmware.com>

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -6,15 +6,27 @@ description: |-
   The VMware vCloud Director provider is used to interact with the resources supported by VMware vCloud Director. The provider needs to be configured with the proper credentials before it can be used.
 ---
 
-# VMware vCloud Director Provider
+# VMware vCloud Director Provider 2.0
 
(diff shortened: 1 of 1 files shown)


lvirbalas:
This is the first take on documentation updates and covers only the `provider` and related config. Documentation for resources will come in a separate PR.

vbauzysvmware:
LGTM
Yet I am not sure about v2.0 cause we don't have release process/tags or system or downloads for previous version too. Seems mostly guys just checkout latest version and that's all. So v2.0 may misleading when you checkout latest code.

lvirbalas:
This new code has been referred to as v2.0 in some discussions with users and Hashicorp. It deserves the version bump, because:
- It's using a different Go vCD vendor repo.
- Many improvements with some semantic changes.

We will make sure it's "tagged" as such in the binary download process, as `terraform init` does recognise version numbers of the adapters.

PR #129, opened 2018-12-11
Update doc with catalog item feature

vbauzysvmware:
Part of issue #112
Signed-off-by: Vaidotas Bauzys <vbauzys@vmware.com>


website/vcd.erb
```diff
@@ -13,6 +13,9 @@
         <li<%= sidebar_current("docs-vcd-resource") %>>
           <a href="#">Resources</a>
           <ul class="nav nav-visible">
+            <li<%= sidebar_current("docs-vcd-resource-catalog-item") %>>
+              <a href="/docs/providers/vcd/r/catalog_item.html">vcd_catalog_item</a>
```
lvirbalas: After merge, please make sure that vcd_catalog_item and vcd_catalog_media entries are listed one after another.

website/docs/r/catalog_item.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+  upload_piece_size = 10 
+  show_upload_progress = true
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, isn't needed if defined/match at provider level
```
dataclouder: "The name of organization to use, isn't needed if defined/match at provider level" ==> 
"The name of organization to use. Not needed if defined at provider level"

website/docs/r/catalog_item.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+layout: "vcd"
+page_title: "vCloudDirector: vcd_catalog_item"
+sidebar_current: "docs-vcd-resource-catalog-item"
+description: |-
+Provides a vCloud Director Catalog item resource. This can be used to upload and delete OVA file in catalog.
+---
+
+# vcd\_catalog\_item
+
+Provides a vCloud Director Catalog item resource. This can be used to upload OVA to catalog and delete it.
```
lvirbalas: Please use consistent capitalisation: "Catalog Item" or "catalog item".

PR #131, opened 2018-12-11
Add catalog resource docs

vbauzysvmware:
Part of issue #112
Signed-off-by: Vaidotas Bauzys vbauzys@vmware.com

### DIFF
--- a/website/docs/r/catalog.html.markdown
+++ b/website/docs/r/catalog.html.markdown
@@ -32,5 +32,5 @@ The following arguments are supported:
 * `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
 * `name` - (Required) Catalog name
 * `description` - (Optional) - Description of catalog
-* `delete_recursive` - (Required) - When destroying use delete_recursive=True to remove the catalog and any objects it contains that are in a state that normally allows removal.
-* `delete_force` - (Required) - When destroying use delete_force=True with delete_recursive=True to remove a catalog and any objects it contains, regardless of their state.
+* `delete_recursive` - (Required) - When destroying use delete_recursive=True to remove the catalog and any objects it contains that are in a state that normally allows removal
+* `delete_force` -(Required) - When destroying use delete_force=True with delete_recursive=True to remove a catalog and any objects it contains, regardless of their state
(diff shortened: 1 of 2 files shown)


website/docs/r/catalog.html.markdown
```diff
@@ -0,0 +1,36 @@
+---
+layout: "vcd"
+page_title: "vCloudDirector: vcd_catalog"
+sidebar_current: "docs-vcd-resource-catalog"
+description: |-
+Provides a vCloud Director catalog resource. This can be used to create and delete catalog.
```
lvirbalas: Determiner missing: "... to create and delete **a** catalog."

website/docs/r/catalog.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+  delete_force      = "true"
+  delete_recursive  = "true"  
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use. Not needed if defined at provider level
```
lvirbalas: Let's make it consistent with other files:
`The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations`

website/docs/r/catalog.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+* `name` - (Required) Catalog name
+* `description` - (Optional) - Description of catalog
+* `delete_force` - (Required) - When destroying use delete_force=True with delete_recursive=True to remove a catalog and any objects it contains, regardless of their state.
```
lvirbalas: Please move both `delete_force` write-up and the parameter in the example above under the `delete_recursive` parameter. Then it will be enough for the new user to read the section once (as opposed to twice) to understand how it works :)

website/docs/r/catalog.html.markdown
```diff
@@ -29,8 +29,8 @@ resource "vcd_catalog" "myNewCatalog" {
...
 The following arguments are supported:
 
-* `org` - (Optional) The name of organization to use. Not needed if defined at provider level
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
 * `name` - (Required) Catalog name
 * `description` - (Optional) - Description of catalog
-* `delete_force` - (Required) - When destroying use delete_force=True with delete_recursive=True to remove a catalog and any objects it contains, regardless of their state.
-* `delete_recursive` - (Required) - When destroying use delete_recursive=True to remove the catalog and any objects it contains that are in a state that normally allows removal.
\ No newline at end of file
+* `delete_force` - (Required)
```
lvirbalas: You lost text here and order of parameters in the list was not changed either.

website/docs/r/catalog.html.markdown
lvirbalas: Looks great now, thanks!

PR #133, opened 2018-12-11
Add support for logging in the provider

dataclouder:
Add properties to enable logging and define the log file name directly in
the provider.

This addition wants to facilitate troubleshooting using terraform-provider-vcd, by offering an easy and documented way to enable a log of API operations.


dataclouder:
Everyone (except me) seemed to be afraid of the cache. I will gladly make it the default if the team approves. It saves a few minutes of testing time, and a few thousand lines of response text per call.

PR #135, opened 2018-12-13
Add org docs

dataclouder:
##  Add documentation for Org resource

* Add missing property to org resource
* Add definition for non self-explanatory resources
* Remove "use_server_boot_sequence" property, as the docs define it as "This value is ignored."
* Add missing error handling from several resources
* Update vendor directory

### DIFF
--- a/website/docs/r/org.html.markdown
+++ b/website/docs/r/org.html.markdown
@@ -9,7 +9,7 @@ Provides a vCloud Director Organization resource. This can be used to create and
 # vcd\_org
 
 Provides a vCloud Director Org resource. This can be used to create and delete an organization.
-Requires administrator privileges.
+Requires system administrator privileges.
 
 Supported in provider *v2.0+*
 


vcd/resource_vcd_org_test.go
```diff
@@ -77,9 +80,20 @@ func testAccCheckOrgDestroy(s *terraform.State) error {
 			continue
 		}
 
-		org, err := govcd.GetOrgByName(conn.VCDClient, rs.Primary.Attributes["name"])
-		if org != (govcd.Org{}) || err != nil {
-			return fmt.Errorf("org with name %s was found", rs.Primary.Attributes["name"])
+		deleted := false
+		var org govcd.Org
+		var err error
+		for N := 0; N < 10; N++ {
```
vbauzysvmware: I would say this isn't sustainable for all systems. If we need to wait 100 count? Couldn't we use while?

website/docs/r/org.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) Org name
+* `full_name` - (Required) Org full name
+* `delete_recursive` - (Required) - pass delete_recursive=true as query parameter to remove an organization or VDC and any objects it contains that are in a state that normally allows removal.
+* `delete_force` - (Required) - pass delete_force=true and delete_recursive=true to remove an organization or VDC and 
+   any objects it contains, regardless of their state.
+* `is_enabled` - (Optional) - True if this organization is enabled (allows login and all other operations).
```
vbauzysvmware: Mising default value

vcd/resource_vcd_org_test.go
dataclouder: Right. I was actually fishing for comments on this one, as I have seen similar methods in a few places.
But we need to define a few public timeout values for long term operations and then use them consistently. I have a related task in my TODO list.

dataclouder:
The use of "_" is idiomatic when we want to explicitly ignore errors, but at the same time we want to silence detection tools. If there is a particular error that we want to handle for `FPrintf`, please suggest it.

vcd/resource_vcd_org_test.go
vbauzysvmware: ok

vbauzysvmware:
Yes, but I don't see difference between _ = func and func .

dataclouder:
There is no substantial difference. The end result is that we don't check the error. 
However the `_ = func` syntax will signal the tools like [errcheck](https://github.com/kisielk/errcheck) (which we use in our repo) that we take responsibility for the lack of error checking, and therefore we don't get notified for this call.

vbauzysvmware:
ok. Got it.

vcd/resource_vcd_org.go
```diff
@@ -34,44 +39,53 @@ func resourceOrg() *schema.Resource {
...
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
 				ForceNew: false,
 			},
+			"is_enabled": &schema.Schema{
+				Type:        schema.TypeBool,
+				Optional:    true,
+				ForceNew:    false,
+				Default:     false,
```
lvirbalas: Shouln't organisation be enabled by default?

vcd/resource_vcd_org.go
```diff
@@ -34,44 +39,53 @@ func resourceOrg() *schema.Resource {
...
+				Description: "True if this organization is enabled (allows login and all other operations).",
+			},
 			"deployed_vm_quota": &schema.Schema{
-				Type:     schema.TypeInt,
-				Optional: true,
-				Default:  -1,
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     -1,
+				Description: "Maximum number of virtual machines that can be deployed simultaneously by a member of this organization.",
```
lvirbalas: Thanks for adding these descriptions to arguments!

vcd/resource_vcd_org_test.go
```diff
@@ -77,9 +80,20 @@ func testAccCheckOrgDestroy(s *terraform.State) error {
...
+			org, err = govcd.GetOrgByName(conn.VCDClient, rs.Primary.Attributes["name"])
+			deleted = (org == (govcd.Org{})) && (err != nil)
+			if deleted {
+				break
+			}
+			time.Sleep(time.Second)
```
lvirbalas: Would really like to see consistent handling of timeouts in the adapter. Would you think about it and file a separate issue for this? It should lay out the idea how can we resolve this, like what configurable parameter we need and which places in the code need to be refactored.

website/docs/r/org.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+## Example Usage
+
+```
+resource "vcd_org" "my-org" {
+  name             = "my-org"
+  full_name        = "My organization"
+  description      = "The pride of my work"
+  is_enabled       = "true"
+  delete_force     = "true"
+  delete_recursive = "true"
```
lvirbalas: Please switch places of delete_force and delete_recursive parameters to better match the logical relationship between them:

```
delete_recursive = "true"
delete_force     = "true"  
```

website/docs/r/org.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) Org name
+* `full_name` - (Required) Org full name
+* `delete_recursive` - (Required) - pass `delete_recursive`=true as query parameter to remove an organization or VDC and any objects it contains that are in a state that normally allows removal.
```
lvirbalas: Strange code quoting. What about this (include the true/false assignment in the code quote):
`delete_recursive=true`

website/docs/r/org.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+* `delete_force` - (Required) - pass `delete_force`=true and `delete_recursive`=true to remove an organization or VDC and any objects it contains, regardless of their state.
+* `is_enabled` - (Optional) - True if this organization is enabled (allows login and all other operations). Default is `false`.
+* `description` - (Optional) - Org description. There is no default value.
```
lvirbalas: "There is no default value" -> "Default is empty"?

website/docs/r/org.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+Provides a vCloud Director Organization resource. This can be used to create and delete an organization.
+---
+
+# vcd\_org
+
+Provides a vCloud Director Org resource. This can be used to create and delete an organization.
+
+Supported in provider *v2.0+*
+
+## Example Usage
```
lvirbalas: I believe creation of an Org is only possible with the sys admin user (correct me, if you know of a way to do it with other user type). Hence, we need a sentence about this and it wouldn't hurt to add a `provider` sample in the example below with the correct user.

vcd/resource_vcd_org.go
dataclouder: Originally, this was a mandatory property.  I can either revert it to `Required` or change the default as `true`. Which one would be preferred?

vcd/resource_vcd_org_test.go
dataclouder: I am working on it. But we need some discussion on this matter.

vcd/resource_vcd_org.go
lvirbalas: Let's make Org enabled by default, which will lead to a usable configuration with less parameters.

website/docs/r/org.html.markdown
```diff
@@ -9,19 +9,26 @@ Provides a vCloud Director Organization resource. This can be used to create and
 # vcd\_org
 
 Provides a vCloud Director Org resource. This can be used to create and delete an organization.
+Requires administrator privileges.
```
lvirbalas: For clarity I would use "system administrator", as org admin is also administrator.

dataclouder: done

PR #138, opened 2018-12-18
Add default timeout to test configuration

dataclouder:
Add option to define timeout for tests
Set a default value of 300 for timeout when option is blank
Add option to enable connection cache for tests
Add missing external network definition in sample file

This addition will allow users to run tests with less recourse to environment variables.

### DIFF
--- a/vcd/config_test.go
+++ b/vcd/config_test.go
@@ -32,6 +32,8 @@ type TestConfig struct {
 		SysOrg                   string `json:"sysOrg"`
 		AllowInsecure            bool   `json:"allowInsecure"`
 		TerraformAcceptanceTests bool   `json:"tfAcceptanceTests"`
+		UseVcdConnectionCache    bool   `json:"useVcdConnectionCache"`
+		MaxRetryTimeout          int    `json:"maxRetryTimeout"`
 	} `json:"provider"`
 	VCD struct {
 		Org     string `json:"org"`
@@ -190,8 +192,11 @@ func templateFill(tmpl string, data StringMap) string {
 		if err != nil || count == 0 {
 			panic(fmt.Errorf("error writing to file %s. Reported %d bytes written. %s", resourceFile, count, err))
 		}
-		writer.Flush()
-		file.Close()
+		err = writer.Flush()
+		if err != nil {
+			panic(fmt.Errorf("error writing to file %s. Reported %d bytes written. %s", resourceFile, count, err))
+		}
+		_ = file.Close()
 	}
 	// Returns the populated template
 	return string(writeStr)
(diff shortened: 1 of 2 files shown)


vcd/sample_vcd_test_config.json
```diff
@@ -11,7 +11,14 @@
     "allowInsecure": true,
     "//": "tfAcceptanceTests is needed to allow the whole test suite to run",
     "//": "Warning: it may cost time (from 14 to 40 min) and possibly money",
-    "tfAcceptanceTests": false
+    "tfAcceptanceTests": false,
+    "//": "If enabled, the test suite will cache the provider connection for up to 20 minutes",
+    "//": "and then renew it automatically. This will save some heavy network traffic.",
+    "//": "The default is false",
```
lvirbalas: You mean default if UseVcdConnectionCache is not set in the config file at all? It's a bit strange to read "default is false" and then see ""UseVcdConnectionCache": true" a line below. Better rephrase the comment to something like "Disabled by default if UseVcdConnectionCache is not set".

vcd/sample_vcd_test_config.json
```diff
@@ -11,7 +11,14 @@
...
+    "UseVcdConnectionCache": true,
+    "//": "The default for max retry timeout is 60, which often leads to premature test termination",
+    "//": "If no value is provided, the test suite will set the value to 300",
```
lvirbalas: Really confusing sentence about defaults. Imagine new user reading: ... "default for max retry timeout is 60" ... "the test suite will set the value to 300" ...
I suggest amending the comment that it's clear that 60 is default for _Terraform provider_, while tests set a higher default.

vcd/sample_vcd_test_config.json
```diff
@@ -11,7 +11,14 @@
     "allowInsecure": true,
     "//": "tfAcceptanceTests is needed to allow the whole test suite to run",
     "//": "Warning: it may cost time (from 14 to 40 min) and possibly money",
-    "tfAcceptanceTests": false
+    "tfAcceptanceTests": false,
```
lvirbalas: Why don't we set tfAcceptanceTests to true in the sample by default? Seems an extra hassle to change this and easy to miss. Actually, do we need to change this to true of you run `make testacc`?

vcd/sample_vcd_test_config.json
```diff
@@ -11,7 +11,14 @@
...
+    "//": "If enabled, the test suite will cache the provider connection for up to 20 minutes",
+    "//": "and then renew it automatically. This will save some heavy network traffic.",
+    "//": "The default is false",
+    "UseVcdConnectionCache": true,
```
lvirbalas: First letter case not consistent with other variables, should be `useVcdConnectionCache`

vcd/config_test.go
```diff
@@ -235,20 +240,30 @@ func getConfigStruct() TestConfig {
 	// Now we fill the environment variables that the library is using for its own initialization.
 	if configStruct.Provider.TerraformAcceptanceTests {
 		// defined in vendor/github.com/hashicorp/terraform/helper/resource/testing.go
-		os.Setenv("TF_ACC", "1")
+		_ = os.Setenv("TF_ACC", "1")
+	}
+	if configStruct.Provider.MaxRetryTimeout == 0 {
+		// Setting a default value that should be reasonable for these tests, as we run many heavy operations
+		_ = os.Setenv("VCD_MAX_RETRY_TIMEOUT", "300")
```
lvirbalas: Could you double check that we are consistent in that test config values from JSON take priority over env variables?

dataclouder: Already checked. We set the variables just before invoking the provider constructor.

vcd/sample_vcd_test_config.json
dataclouder: done

vcd/sample_vcd_test_config.json
dataclouder: fixed

vcd/sample_vcd_test_config.json
dataclouder: fixed

vcd/sample_vcd_test_config.json
dataclouder: done

vcd/sample_vcd_test_config.json
```diff
@@ -11,13 +11,13 @@
     "allowInsecure": true,
     "//": "tfAcceptanceTests is needed to allow the whole test suite to run",
     "//": "Warning: it may cost time (from 14 to 40 min) and possibly money",
-    "tfAcceptanceTests": false,
+    "tfAcceptanceTests": true,
```
lvirbalas: Thanks! The only remaining question is what happens if we set this to false, but run `make testacc` which sets `TF_ACC=1` - which setting takes priority?

dataclouder: if `TF_ACC` is set by `make`, it is used.
As it is now, this is a setting that can only be enabled, but not disabled. Having it in the configuration file is for convenience. 
I don't think it's a problem, because when the variable is not set the test is over in about 1 second and you can re-run with the appropriate setting.

Issue #137, opened 2018-12-18
Tests shouldn't require pre-existing catalog and should create it instead


lvirbalas:
Now that we have functionality to create catalogs, tests shouldn't have a prerequisite for existing catalog - they should just create it.

### Affected Files

- vcd/sample_vcd_test_config.json
- vcd/vcd_test_config.json
- make testacc

### Test Configuration Files

We need to  define the following section in `vcd_test_config.json`

```json
# ...
 "vcd": {
    "//": "This section contains the organization composition",
    "org": "Solpan",
    "vdc": "SolpanVDC",
    "catalog": {
      "name": "OperatingSystems",
      "catalogItem": "MicroCore-Linux.ova"
    }
  }
# ...
```

### Expected Behaviour

We would expect that the above catalog is created automatically.

### Actual Behavior

If catalog doesn't exist in vCD (like in a fresh instance), an error is thrown:

```
--- FAIL: TestAccVcdCatalogItemBasic (10.52s)
    testing.go:434: Step 0 error: Error applying: 1 error(s) occurred:

        * vcd_catalog_item.TestAccVcdCatalogItemBasic: 1 error(s) occurred:

        * vcd_catalog_item.TestAccVcdCatalogItemBasic: error finding Catalog: <nil>
```

### Important Factoids

We need to adjust the catalog tests so they create the catalog, upload the item and media and then destroy the catalog. This will remove one prerequisite from test environment and will make running tests easier.

lvirbalas:
> I think the catalog name should be pre-configured as a constant by the test suite, and have a name that reflects its origin, such as TestUploadCatalogItem.

I agree, only the name for this test case should not contain "Item" - only "Catalog", something like "TestCreateCatalog".

lvirbalas:
> Another thing to take into account is that the test suite will become slower when the OVA needs to be uploaded before every other operation starts. We may want to let in place the current setting, so that users may decide to run a quicker test with an existing catalog and item, rather than uploading a new one at every test run.

Not an issue. Default OVA for test is small and it takes around a minute to pass.

```
=== RUN   TestAccVcdCatalogItemBasic
vcd_catalog_item.TestAccVcdCatalogItemBasic: Upload progress 0.00%
vcd_catalog_item.TestAccVcdCatalogItemBasic: Upload progress 100.00%
vcd_catalog_item.TestAccVcdCatalogItemBasic: vCD import catalog item progress 1%
[...]

PR #148, opened 2019-01-09
Improve catalog tests

vbauzysvmware:
Reference: #137 

Add support for test suite to create own catalog and catalog item Ova

- update vendor govcd
- update make script to download OVA if needed
- change tests to reuse created catalog and OVA
- add capability to use user provided catalog and OVA


scripts/ovadownload.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+echo "==> Searching for OVA download URL..."
+ova_download_url=$(grep -w ovaDownloadUrl vcd/vcd_test_config.json | tr -d '",' | awk '{print $2}')
+ova_file_name=$(grep -w ovaTestFileName vcd/vcd_test_config.json | tr -d '",' | awk '{print $2}')
```
lvirbalas: 1.) Are you sure you did not forget to update vcd_test_config.json? I can't see "ovaDownloadUrl" and "ovaTestFileName" there. How this should work?
2.) Do we need "ova_file_name"? Can't we get the filename from the URL?
3.) Rename `ovadownload.sh` -> `download-ova.sh` to be more consistent with things like `install-plugin.sh`

GNUmakefile
```diff
@@ -12,7 +12,7 @@ test: fmtcheck
 	go test -i $(TEST) || exit 1
 	cd vcd ; VCD_SHORT_TEST=1 go test -v . -timeout 3m
 
-testacc: fmtcheck
+testacc: fmtcheck download-ova
```
dataclouder: Is it necessary to run this operation in the Makefile?
I think the right place would be in function `TestMain` (`config_test.go`).
The way it is configured now, the test will work with `make testacc` bit will fail when we run the test as `go test` inside the `vcd` directory.

scripts/ovadownload.sh
dataclouder: See my comment above. If we implement the OVA retrieval in the Go `TestMain`, we just use the fields in the configuration file, with more control on what we are doing.

vcd/config_test.go
```diff
@@ -314,9 +323,155 @@ func TestMain(m *testing.M) {
...
+	if testConfig.VCD.Catalog.Name == "" {
+		fmt.Printf("Creating catalog for test suite...\n")
+		catalog, err = org.CreateCatalog(testSuiteCatalogName, "Test suite purpose")
+		if err != nil || catalog == (govcd.Catalog{}) {
+			panic(err)
+		}
+		fmt.Printf("Catalog created successfuly\n")
+
+	} else {
+		fmt.Printf("Skiping catalog creation - found preconfigured one: %s \n", testConfig.VCD.Catalog.Name)
```
dataclouder: "Skiping" -> "Skipping"

vcd/config_test.go
```diff
@@ -314,9 +323,155 @@ func TestMain(m *testing.M) {
...
+
+		catalog, err = org.FindCatalog(testConfig.VCD.Catalog.Name)
+		if err != nil || catalog == (govcd.Catalog{}) {
+			fmt.Printf("Preconfigured catalog wasn't found \n")
+			panic(err)
+		}
+
+		fmt.Printf("Catalog found successfuly\n")
```
dataclouder: "successfuly" -> "successfully"

vcd/config_test.go
```diff
@@ -314,9 +323,155 @@ func TestMain(m *testing.M) {
...
+		testSuiteCatalogName = testConfig.VCD.Catalog.Name
+	}
+
+	if testConfig.VCD.Catalog.CatalogItem == "" {
+		fmt.Printf("Creating catalog item for test suite...\n")
+		task, err := catalog.UploadOvf("../test-resources/"+config.Ova.OvaTestFileName, testSuiteCatalogOVAItem, "Test suite purpose", 20*1024*1024)
```
dataclouder: The hard coded path `../test-resources` should be checked for existence. Even better, its path should be found as relative to the test file, as done in `getConfigStruct` (`config_test.go`)

vcd/config_test.go
```diff
@@ -314,9 +323,155 @@ func TestMain(m *testing.M) {
 		}
 	}
 
+	createSuiteCatalogAndItem(testConfig)
```
dataclouder: Add a comment about what we do when we find the item and when we don't.

vcd/config_test.go
```diff
@@ -314,9 +323,155 @@ func TestMain(m *testing.M) {
...
+func getTestVCDFromJson(testConfig TestConfig) (*govcd.VCDClient, error) {
+	configUrl, err := url.ParseRequestURI(testConfig.Provider.Url)
+	if err != nil {
+		return &govcd.VCDClient{}, fmt.Errorf("could not parse Url: %s", err)
+	}
+	vcdClient := govcd.NewVCDClient(*configUrl, true)
+	return vcdClient, nil
+}
+
+func destroySuiteCatalogAndItem(config TestConfig) {
```
dataclouder: Should we have an option for keeping such resources for another test?

vcd/config_test.go
```diff
@@ -314,9 +323,155 @@ func TestMain(m *testing.M) {
...
+			return
+		}
+		err = catalogItem.Delete()
+		if err != nil {
+			fmt.Errorf("error removing catalog item %#v", err)
+			return
+		}
+		fmt.Printf("Catalog %s item removed successfuly\n", catalogItem.CatalogItem.Name)
+	} else {
+		fmt.Printf("Catalog item deletion skiped as user defined resource is used or removed with catalog\n")
```
dataclouder: "skiped" -> "skipped"

scripts/ovadownload.sh
lvirbalas: Thanks for responses and config sample update.

> Well I used same naming convention which is are '''scripts''' folder. Are we changing all them?

Someone added "install-plugin.sh" and it looks just so much nicer with the dash :)

scripts/ovadownload.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
```
lvirbalas: Please add a comment explaining what this script is used for.

scripts/ovadownload.sh
vbauzysvmware: In current repo version there isn't such file :) 
> Someone added "install-plugin.sh" and it looks just so much nicer with the dash :)

vcd/config_test.go
vbauzysvmware: This destroy function called after all tests are finished.

vcd/config_test.go
vbauzysvmware: Fixed

vcd/config_test.go
vbauzysvmware: Fixed

vcd/config_test.go
vbauzysvmware: Fixed

vcd/config_test.go
vbauzysvmware: Fixed

GNUmakefile
vbauzysvmware: Well, that was Linas idea and discussed here https://github.com/terraform-providers/terraform-provider-vcd/issues/137. From your point of view maybe is better to have in code - yet mentioned issue is `go test` is solved now with 
`if vcdShortTest {
	t.Skip(acceptanceTestsSkipped)
	return
}`
@lvirbalas We need common agreement should I invest time and move code from make files to go `config_test.go`

lvirbalas: @dataclouder the download logic is in the script right now, is your suggestion to rewrite it in Golang?

dataclouder: Yes. Or else we have a test that can only run through `make` and not using `go test`.

lvirbalas: My bad. I didn't realise we're making the tests only `make` compatible with this.

@vbauzysvmware Yes, please invest time to refactoring this. The middle way would be to call the download script from Go, but you decide which is more elegant.

vbauzysvmware: Changes are done.

vcd/config_test.go
vbauzysvmware: Fixed

vcd/config_test.go
```diff
@@ -314,9 +329,204 @@ func TestMain(m *testing.M) {
...
+
+		fmt.Printf("Catalog found successfully\n")
+		testSuiteCatalogName = testConfig.VCD.Catalog.Name
+	}
+
+	if testConfig.VCD.Catalog.CatalogItem == "" {
+		fmt.Printf("Creating catalog item for test suite...\n")
+		task, err := catalog.UploadOvf(ovaFilePath, testSuiteCatalogOVAItem, "Test suite purpose", 20*1024*1024)
+		if err != nil {
+			fmt.Errorf("error uploading new catalog item: %#v", err)
```
dataclouder: `fmt.Errorf`result  is not used

vcd/config_test.go
```diff
@@ -314,9 +329,204 @@ func TestMain(m *testing.M) {
...
+			panic(err)
+		}
+
+		err = task.ShowUploadProgress()
+		if err != nil {
+			fmt.Errorf("error waiting from task to complete: %+v", err)
```
dataclouder: `fmt.Errorf` is not used

vcd/config_test.go
```diff
@@ -314,9 +329,204 @@ func TestMain(m *testing.M) {
...
+			panic(err)
+		}
+
+		err = task.WaitTaskCompletion()
+		if err != nil {
+			fmt.Errorf("error waiting from task to complete: %+v", err)
```
dataclouder: `fmt.Errorf` is  not used

vcd/config_test.go
dataclouder: I mean: to use it with the next test suite run.

run n. 1: create default catalog and item. If we have an option to preserve them, they are not deleted.
run n.2: if the default catalog and item already exist, we reuse them, without another download/upload setup operation

Currently, if the catalog already exists (for a failure of the previous test) we get an unrecoverable error.

scripts/ovadownload.sh
vbauzysvmware: No script - no issue here anymore :)

vcd/config_test.go
vbauzysvmware: During the call we agreed on some additional behaviour. Now json configuration holds `preserve` which means not to destroy test suite created resources and on next run will be reused. TEST_SUITE_CLEANUP env variable will allow to force resource deletion before testsuite running - in case aborted or errors with previous test run.

scripts/ovadownload.sh
vbauzysvmware: Script removed.

vcd/config_test.go
```diff
@@ -76,6 +82,12 @@ type TestConfig struct {
 	} `json:"media"`
 }
 
+// names for created resources names for all the tests
+var (
```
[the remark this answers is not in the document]
vbauzysvmware: Can't do cause the name changed then catalog or item is preconfigured.

vcd/config_test.go
vbauzysvmware: fixed

vcd/config_test.go
vbauzysvmware: Printf

vcd/config_test.go
vbauzysvmware: fixed

lvirbalas:
@dataclouder can you add a TESTING.md which explains the overall concept on how the tests work now and how to run them?

PR #149, opened 2019-01-09
Define resource for vdc.

megalord:
This is very much a work in progress at the moment, but I have been able to successfully `terraform apply` and `terraform destroy` a `vcd_vdc` resource.  I am creating the PR at this stage so folks know it's being worked on.

The `Update` function is intentionally blank right now because I did not add it in https://github.com/vmware/go-vcloud-director/pull/117.  Either I will need to go back to that before this gets too far, or we'll hide this feature until it all gets added in subsequent PRs.  I'll aim for the former option but naturally it will take a bit longer.

I haven't started on the tests yet but they will be written :smile: .   I think this is the right branch to target, at least for now such that the diff is correct, but I figure it'll change based on the status of the Christmas present.

### DIFF
--- a/vcd/resource_vcd_vdc.go
+++ b/vcd/resource_vcd_vdc.go
@@ -421,7 +421,7 @@ func getVcdVdcInput(d *schema.ResourceData, vcdClient *VCDClient) (*types.VdcCon
 	}
 
 	if vCpuInMhz, ok := d.GetOk("v_cpu_in_mhz"); ok {
-		params.VCpuInMhz = vCpuInMhz.(int64)
+		params.VCpuInMhz = int64(vCpuInMhz.(int))
 	}
 
 	if isThinProvision, ok := d.GetOk("is_thin_provision"); ok {
--- a/vcd/resource_vcd_vdc.go
+++ b/vcd/resource_vcd_vdc.go
@@ -1,0 +1,423 @@
+package vcd
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/vmware/go-vcloud-director/govcd"
+	types "github.com/vmware/go-vcloud-director/types/v56"
+)
+
+func resourceVcdVdc() *schema.Resource {
+	capacityWithUsage := schema.Schema{
+		Type:     schema.TypeSet,
+		Required: true,
+		ForceNew: false,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"units": {
+					Type:     schema.TypeString,
+					Required: true,
+				},
+				"allocated": {
+					Type:     schema.TypeInt,
+					Required: false,
+					Optional: true,
+				},
+				"limit": {
+					Type:     schema.TypeInt,
+					Required: false,
+					Optional: true,
+				},
+				"reserved": {
+					Type:     schema.TypeInt,
+					Required: false,
+					Optional: true,
+				},
+				"used": {
+					Type:     schema.TypeInt,
+					Required: false,
+					Optional: true,
+				},
+				"overhead": {
+					Type:     schema.TypeInt,
+					Required: false,
+					Optional: true,
+				},
+			},
+		},
+	}
+
+	return &schema.Resource{
+		Create: resourceVcdVdcCreate,
+		Delete: resourceVcdVdcDelete,
+		Read:   resourceVcdVdcRead,
+		Update: resourceVcdVdcUpdate,
+
+		Schema: map[string]*schema.Schema{
+			"org": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: false,
+			},
+
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: false,
+				Optional: true,
+				ForceNew: false,
+			},
+			"allocation_model": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: false,
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
+					v := val.(string)
+					switch v {
+					case
+						"AllocationVApp",
+						"AllocationPool",
+						"ReservationPool":
+						return
+					default:
+						errs = append(errs, fmt.Errorf("%q must be one of {AllocationVApp, AllocationPool, ReservationPool}, got: %s", key, v))
+					}
+					return
+				},
+			},
+			"compute_capacity": &schema.Schema{
+				Required: true,
+				ForceNew: false,
+				Type:     schema.TypeSet,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"cpu":    &capacityWithUsage,
+						"memory": &capacityWithUsage,
+					},
+				},
+			},
+			"nic_quota": &schema.Schema{
+				Type:     schema.TypeInt,
+				Required: false,
+				Optional: true,
+				ForceNew: false,
+			},
+			"network_quota": &schema.Schema{
+				Type:     schema.TypeInt,
+				Required: false,
+				Optional: true,
+				ForceNew: false,
+			},
+			"vm_quota": &schema.Schema{
+				Type:     schema.TypeInt,
+				Required: false,
+				Optional: true,
+				ForceNew: false,
+			},
+			"is_enabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Required: false,
+				Optional: true,
+				ForceNew: false,
+			},
+			"storage_profile": &schema.Schema{
+				Type:     schema.TypeSet,
+				Required: true,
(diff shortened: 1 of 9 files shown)


vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,423 @@
...
+				Type:        schema.TypeBool,
+				Required:    true,
+				ForceNew:    false,
+				Description: "When destroying use delete_recursive=True to remove the vdc and any objects it contains that are in a state that normally allows removal.",
+			},
+		},
+	}
+}
+
+func resourceVcdVdcCreate(d *schema.ResourceData, meta interface{}) error {
```
dataclouder: Please add a description of the function purpose

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,423 @@
...
+		log.Printf("[DEBUG] Error creating vdc: %#v", err)
+		return fmt.Errorf("error creating vdc: %#v", err)
+	}
+
+	d.SetId(d.Get("name").(string))
+	log.Printf("[TRACE] vdc created: %#v", task)
+	return resourceVcdVdcRead(d, meta)
+}
+
+func resourceVcdVdcRead(d *schema.ResourceData, meta interface{}) error {
```
dataclouder: Please add a description of the function purpose

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,423 @@
...
+	if err != nil {
+		log.Printf("[DEBUG] Error removing vdc %#v", err)
+		return fmt.Errorf("error removing vdc %#v", err)
+	}
+
+	log.Printf("[TRACE] vdc delete completed: %#v", vdc.Vdc)
+	return nil
+}
+
+func capacityWithUsage(d map[string]interface{}) *types.CapacityWithUsage {
```
dataclouder: Please add a description of the function purpose

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,423 @@
...
+	v := &types.CapacityWithUsage{
```
dataclouder: Please use a more descriptive name for `v`

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+
+	if overhead, ok := d["overhead"]; ok {
+		v.Overhead = int64(overhead.(int))
+	}
+
+	return v
+}
+
+func getVcdVdcInput(d *schema.ResourceData, vcdClient *VCDClient) (*types.VdcConfiguration, error) {
```
dataclouder: Some comments are needed here.
What are we extracting from the resource data and what do we want to achieve?

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,423 @@
...
+	computeCapacity := d.Get("compute_capacity").(*schema.Set).List()[0].(map[string]interface{})
```
dataclouder: There should be a check about the `[0]` element being set.

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,423 @@
...
+	storageProfileMap := d.Get("storage_profile").(*schema.Set).List()[0].(map[string]interface{})
+
+	params := &types.VdcConfiguration{
+		Name:            d.Get("name").(string),
+		Xmlns:           "http://www.vmware.com/vcloud/v1.5",
+		AllocationModel: d.Get("allocation_model").(string),
+		ComputeCapacity: []*types.ComputeCapacity{
+			&types.ComputeCapacity{
+				CPU:    capacityWithUsage(computeCapacity["cpu"].(*schema.Set).List()[0].(map[string]interface{})),
```
dataclouder: See above comment

megalord:
Ok I think I addressed all the feedback in the first round of review, thanks!  As I'm getting more familiar with the go-vcloud repo and this one, I see a mix of referring to resources by name vs id.  Seems like many of the retrieval methods want a name passed rather than id, and even here I see  `d.SetId(d.Get("name").(string))`.  Is one preferred over the other?  It naively seems like ids would be better so that we don't have to make as many query calls (relationships in the api use id everywhere I have seen), but I'll defer to your judgement.

You'll notice that I used ids in the test config, which I suppose is technically a separate question, since the test itself could turn names into ids before rendering the template.  If we were to use names in the config, I think I'd have to parse the `HREF` field since the query endpoints do not return ids.

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+  delete_force           = true
+  delete_recursive       = true
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Required) Organization to create the VDC in
```
lvirbalas: Please add the "v2.0+" flag to this option. Given the amount of changes recently, we're trying to make it easier for users to know what's available when. Please see `vcd_vapp_vm` doc for an example.

vcd/sample_vcd_test_config.json
```diff
@@ -24,6 +24,11 @@
     "//": "This section contains the organization composition",
     "org": "myorg",
     "vdc": "myvdc",
+    "providerVdc": {
+      "id": "<uuid>",
```
Didainius: Agree for names here vs UUIDs. Just for consistency reasons with other parts of project.

megalord:
In regard to `vmware-stash-while-master-pr-open`, how stable are the module and terraform version changes?  I see some huge changes are getting merged as recently as two days ago, and unfortunately I don't have the time to rebase regularly against a rapidly moving target.  So I'm a bit hesitant but perhaps that's just my ignorance with what is going on there.

Otherwise, for the UUID vs name feedback, I'm totally with y'all that the UUIDs are cumbersome to deal with.  I was imagining that eventually there would be something like:
```
data "vcd_provider_vdc" "app" {
  name = "app"
}

resource "vcd_vdc" "app" {
  provider_vdc = "${vcd_provider_vdc.app.id}"
}
```
This way, the provider code is based on passing IDs, but the user gets to work with names.  This should mean the provider is making less calls to the API to translate names to IDs, because it only gets done once overall instead of once per use (e.g. 3 VDCs from the same provider).  Of course the `data` elements don't exist yet, but you can avoid a breaking change in the long run if you like the idea.

lvirbalas:
> In regard to vmware-stash-while-master-pr-open, how stable are the module and terraform version changes?

We're more or less out of the woods there. I'll let @Didainius to confirm latest news.

Didainius:
> > In regard to vmware-stash-while-master-pr-open, how stable are the module and terraform version changes?
> 
> We're more or less out of the woods there. I'll let @Didainius to confirm latest news.

AFAIK it is stable. I haven't observed any problems. The test suite works fine for me on multiple environments. Some of the merges may look huge, but that's mainly because of `vendor` directory rebuilds for `go mod vendor`.
If you try to rebase on `vmware-stash-while-master-pr-open` please have a look at [README.md ](https://github.com/terraform-providers/terraform-provider-vcd/blob/vmware-stash-while-master-pr-open/README.md) and the `Developing the Provider` section. You'll have to start using Go modules and you may find `replace` directive valuable in `go.mod` to work on your own local version of `go-vcloud-director`.
If you hit any problems - I'm here :)

vbauzysvmware:
Yes, it seems good idea when you create provider vdc and organization vdc on same time or already created by terraform. This saves additional search by name call. But if you created by other means and want reference by name? I recommend to do by referencing by name and be safe for all cases.

> Otherwise, for the UUID vs name feedback, I'm totally with y'all that the UUIDs are cumbersome to deal with. I was imagining that eventually there would be something like:
> 
> ```
> data "vcd_provider_vdc" "app" {
>   name = "app"
> }
> 
> resource "vcd_vdc" "app" {
>   provider_vdc = "${vcd_provider_vdc.app.id}"
> }
> ```
> 
> This way, the provider code is based on passing IDs, but the user gets to work with names. This should mean the provider is making less calls to the API to translate names to IDs, because it only gets done once overall instead of once per use (e.g. 3 VDCs from the same provider). Of course the `data` elements don't exist yet, but you can avoid a breaking change in the long run if you like the idea.

megalord:
> But if you created by other means and want reference by name? I recommend to do by referencing by name and be safe for all cases.

That's exactly what `data` sources are for.  I'd encourage y'all to look into how other providers have used them (I can suggest AWS and DO from personal experience).  Most resources that rely on other resources require the unfriendly id (specifically an arn in the AWS case).

I'll go ahead and switch to the names, but you may keep this in mind.  The more resources the provider defines, the worse this problem will get.

vbauzysvmware:
> `data` sources

Thank you for the hint! Yes `data` sources looks very meaningful to have. I will add to our backlog, until then we have to live with names.

Wittionary:
@megalord Thanks so so much for all your work on this massive contribution! This is the killer resource I've been waiting for. I'm looking forward to this getting merged. ❤️

vbauzysvmware:
Thank you @megalord for huge work. We continue on #236. Hope we can release with next version.

PR #154, opened 2019-01-30
Improve generation of test snippets

dataclouder:
When we run the test suite, we generate Terraform scripts that
can later be used for manual tests. The changes in this commit
add elements to the provider portion of the script that make the
manual test setup faster.

To use the improvement, run the tests with the environment variable `VCD_ADD_PROVIDER=1`, and then look into `./vcd/test-artifacts`

### DIFF
--- a/vcd/config_test.go
+++ b/vcd/config_test.go
@@ -74,19 +74,24 @@ type TestConfig struct {
 		UploadPieceSize int64  `json:"uploadPieceSize,omitempty"`
 		UploadProgress  bool   `json:"uploadProgress,omitempty"`
 	} `json:"media"`
+	EnvVariables map[string]string `json:"envVariables,omitempty"`
 }
 
 const (
+	// This library major version
+	currentProviderVersion   = "2.0"
 	// Warning message used for all tests
 	acceptanceTestsSkipped = "Acceptance tests skipped unless env 'TF_ACC' set"
 	// This template will be added to test resource snippets on demand
 	providerTemplate = `
 provider "vcd" {
-  user     = "{{.User}}"
-  password = "{{.Password}}"
-  url      = "{{.Url}}"
-  sysorg   = "{{.SysOrg}}"
-  org      = "{{.Org}}"
+  user                 = "{{.User}}"
+  password             = "{{.Password}}"
+  url                  = "{{.Url}}"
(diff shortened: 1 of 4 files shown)


vcd/config_test.go
```diff
@@ -74,19 +74,24 @@ type TestConfig struct {
 		UploadPieceSize int64  `json:"uploadPieceSize,omitempty"`
 		UploadProgress  bool   `json:"uploadProgress,omitempty"`
 	} `json:"media"`
+	EnvVariables map[string]string `json:"envVariables,omitempty"`
 }
 
 const (
+	// This library major version
+	currentProviderVersion = "2.0"
```
lvirbalas: We have a VERSION file - it should be reused here.

vcd/sample_vcd_test_config.json
```diff
@@ -72,5 +72,12 @@
     "//": "Size in megabytes",
     "uploadPieceSize": 5,
     "uploadProgress": true
+  },
+  "envVariables" : {
+    "//" : "Environment variables that we want to set (or unset)",
```
lvirbalas: It's fully clear, I suggest rephrasing to: "Environment variables that we want the test to set (or unset)"

vcd/sample_vcd_test_config.json
```diff
@@ -72,5 +72,12 @@
...
+    "GOVCD_DEBUG": "",
+    "VCD_SKIP_TEMPLATE_WRITING": "",
+    "VCD_ADD_PROVIDER": ""
```
lvirbalas: We are at a point where we need to document environment variables in the Terraform project. README.md needs to be updated or even better a separate TESTING.md created with all the nitty details of test setup.

vcd/config_test.go
dataclouder: Done. Changed the code to get the version from there

vcd/sample_vcd_test_config.json
dataclouder: Done

vcd/sample_vcd_test_config.json
dataclouder: Removed testing instructions from `README.md` and added a `TESTING.md` file with explanation of the environment variables.

PR #166, opened 2019-02-06
Improve generation of test snippets

dataclouder:
When we run the test suite, we generate Terraform scripts that
can later be used for manual tests. The changes in this commit
add elements to the provider portion of the script that make the
manual test setup faster.

To use the improvement, run the tests with the environment variable `VCD_ADD_PROVIDER=1`, and then look into `./vcd/test-artifacts`

* Add usability options to provider snippets
* Add environment variable section to configuration file
* Fix resource creation with short test
* Remove testing instructions from README.md
* Add file TESTING.md
* Improve configuration file detection in GNUmakefile

Note: this is the same as PR #154 (with an additional bug fix), but set against `vwmare-stash-while-pr-is-open` instead of `master`.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -20,7 +20,7 @@ test: fmtcheck
 	cd vcd ; VCD_SHORT_TEST=1 go test -v . -timeout 3m
 
 testacc: fmtcheck
-	if [ ! -f vcd/vcd_test_config.json ] ; then \
+	if [ ! -f vcd/vcd_test_config.jsoni -a -z "${VCD_CONFIG}" ] ; then \
 		echo "ERROR: test configuration file vcd/vcd_test_config.json is missing"; \
 		exit 1; \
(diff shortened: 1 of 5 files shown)


GNUmakefile
```diff
@@ -20,7 +20,7 @@ test: fmtcheck
 	cd vcd ; VCD_SHORT_TEST=1 go test -v . -timeout 3m
 
 testacc: fmtcheck
-	if [ ! -f vcd/vcd_test_config.json ] ; then \
+	if [ ! -f vcd/vcd_test_config.jsoni -a -z "${VCD_CONFIG}" ] ; then \
```
lvirbalas: Is there a typo bug here? The file extension is `jsoni`.

dataclouder: Thanks for catching it. It was a typo

Issue #162, opened 2019-02-04
[PROPOSAL] Switch to Go Modules


appilon:
As part of the preparation for Terraform v0.12, we would like to migrate all providers to use [Go Modules](https://github.com/golang/go/wiki/Modules). We plan to continue checking dependencies into vendor/ to remain compatible with existing tooling/CI for a period of time, however go modules will be used for management. Go Modules is the official solution for the go programming language, we understand some providers might not want this change yet, however we encourage providers to begin looking towards the switch as this is how we will be managing all Go projects in the future. Would maintainers please react with :+1: for support, or :-1: if you wish to have this provider omitted from the first wave of pull requests. If your provider is in support, we would ask that you avoid merging any pull requests that mutate the dependencies while the Go Modules PR is open (in fact a total codefreeze would be even more helpful), otherwise we will need to close that PR and re-run `go mod init`. Once merged, dependencies can be added or updated as follows:

```
$ GO111MODULE=on go get github.com/some/module@master
$ GO111MODULE=on go mod tidy
$ GO111MODULE=on go mod vendor
```

GO111MODULE=on might be unnecessary depending on your environment, this example will fetch a module @ master and record it in your project's go.mod and go.sum files. It's a good idea to tidy up afterward and then copy the dependencies into vendor/. To remove dependencies from your project, simply remove all usage from your codebase and run:

```
$ GO111MODULE=on go mody tidy
$ GO111MODULE=on go mod vendor
```

Thank you sincerely for all your time, contributions, and cooperation!

lvirbalas:
Hi @appilon 

We are planning to switch to Go Modules. @dataclouder is working on this. We even have an issue for this :)
#146 

However, the timing is of essence as we need to align with vCloud Director releases. Please help to understand:

> Would maintainers please react with 👍 for support, or 👎 if you wish to have this provider omitted from the first wave of pull requests.

1. When is the first wave planned?
2. When should we switch to Go Modules to react with 👍 to your message?

> If your provider is in support, we would ask that you avoid merging any pull requests that mutate the dependencies while the Go Modules PR is open

3. Are you referring to pull requests to master in our provider?
4. Which Go Modules PR?

Thanks for helping out!

CC: @dataclouder @vbauzysvmware

dataclouder:
Hi @appilon,
I would like a clarification on a point that you have mentioned.

> ﻿﻿﻿﻿copy the dependencies into vendor

[...]

PR #168, opened 2019-02-08
Update repository to use Go modules

dataclouder:
This is a preliminary change to support go modules and Terraform 0.11 (current master uses 0.10).

It is not intended to be merged soon, but rather to provide a playground for the team to try this setup before adopting it on a stable basis.

Related Issue #162 

How to try the same setup on your own without using this branch:

1. Upgrade Go to 1.11.4+
2. clone your repository **outside the $GOPATH**
3. ` git checkout vmware-stash-while-master-pr-open`
4. `git checkout -b try-modules` # or any branch name that you like
5. `git rm -r vendor`  # You will see a bunch of files being deleted here
6. create the file `go.mod` with the following contents:
```go
module github.com/terraform-providers/terraform-provider-vcd

require (
	github.com/hashicorp/terraform v0.11.11
	github.com/vmware/go-vcloud-director v2.1.0-alpha.1
)
```
7. `make test`

You will see that Go will collect the repositories needed to satisfy `go.mod`, and will also update the file with more items and create a file named `go.sum` with the checksums for the repositories being loaded.
The repositories referred in `go.mod` will end up in `$HOME/go/pkg/mod`: both source and compiled objects will be there.

**IMPORTANT**: I have tried to leave the `vendors` folder in place, but its contents are so confusing that I did not manage to get a clean compilation. 

NOTE for reviewers: `go.sum` is generated during `go build`. If you want to change it, you need to change `go.mod`.

**TO DO**: 
* Update documentation about minimum Go version required.
* Insert deprecation messages in resources (supported by Terraform 0.11)


vbauzysvmware:
I followed your guideline and everything worked. And when tried `make testacc` I got issues that modules ins't found in the path. I tried various ways but didnt' find correct `export GOPATH=/home/...` variation to work. Could you help me ?
Also GNU make file needs update cause using still vendor stuff.

vbauzysvmware:
Additional thing: we need to plan moving govcd together

vbauzysvmware:
Please investigate `go.sum` if it should be commited or personal generated env file.

dataclouder:
> Please investigate `go.sum` if it should be commited or personal generated env file.

[Yes. We should keep it in the repo](https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file)

dataclouder:
I am closing this PR and creating a new one after the go-vcloud-director switch.

Issue #50, opened 2017-08-22
self.ip in provisioner returns allocated


harmjanblok:
Hello,

I'm using the variable `${self.ip}` in a provisioner, the server uses DHCP to get it's address. In the previous terraform version `0.9.8` I didn't specify the `ip` argument as part of the `vcd_vapp`. However the IP isn't available anymore during the provisioning phase.

### Terraform Version
- terraform v0.10.2
- terraform-provider-vcd from 2b6e626

### Affected Resource(s)
- vcd_vapp

### Terraform Configuration Files
```hcl
resource "vcd_vapp" "worker_a" {
  catalog_name = "foo"
  cpus = 12
  count = 2
  initscript = "echo 'set ComputerName'"
  memory = 32768
  name = "worker-${element(split("-", uuid()), 0)}"
  network_name = "PRODUCTION-workers"
  provider = "vcd.ZONE-A"
  template_name = "foo-0.0.8"
}

provisioner "local-exec" {
  command = "echo '${self.ip}'"
}
```

### Expected Behavior
Echo's the DHCP assigned IP address of the `vapp`.

### Actual Behavior
Echo's `allocated`.

harmjanblok:
When I configure the `ip` argument and set it to `dhcp` it returns the actual IP during provisioning.

```hcl
resource "vcd_vapp" "worker_a" {
  catalog_name = "foo"
  cpus = 12
  count = 2
  initscript = "echo 'set ComputerName'"
  ip = "dhcp"
  memory = 32768
  name = "worker-${element(split("-", uuid()), 0)}"
  network_name = "PRODUCTION-workers"
  provider = "vcd.ZONE-A"
  template_name = "foo-0.0.8"
}
```

bobbydeveaux:
Potential fix needed around the else statement here:


```
if _, ok := d.GetOk("ip"); ok {
		ip := "allocated"

		oldIp, newIp := d.GetChange("ip")

		log.Printf("[DEBUG] IP has changes, old: %s - new: %s", oldIp, newIp)

		if newIp != "allocated" {
			log.Printf("[DEBUG] IP is assigned. Lets get it (%s)", d.Get("ip"))
			ip, err = getVAppIPAddress(d, meta)
			if err != nil {
				return err
			}
		} else {
			log.Printf("[DEBUG] IP is 'allocated'")
		}

		d.Set("ip", ip)
	} else {
		d.Set("ip", "allocated")
	}

```

[...]

PR #176, opened 2019-02-27
Retrieve IP for vApps using static-ip-pool

harmjanblok:
Fix for #50. Maybe someone has more details on the use case for the `if newIp != "allocated"` construct, as it's unclear to me. Getting the IP should at least be done when using `dhcp` or `allocated` when defining the `vcd_vapp`.

### DIFF
--- a/vcd/resource_vcd_vapp.go
+++ b/vcd/resource_vcd_vapp.go
@@ -425,14 +425,10 @@ func resourceVcdVAppRead(d *schema.ResourceData, meta interface{}) error {
 
 		log.Printf("[DEBUG] IP has changes, old: %s - new: %s", oldIp, newIp)
 
-		if newIp != "allocated" {
-			log.Printf("[DEBUG] IP is assigned. Lets get it (%s)", d.Get("ip"))
-			ip, err = getVAppIPAddress(d, meta, vdc, org)
-			if err != nil {
-				return err
-			}
-		} else {
-			log.Printf("[DEBUG] IP is 'allocated'")
+		log.Printf("[DEBUG] IP is assigned. Lets get it (%s)", d.Get("ip"))
+		ip, err = getVAppIPAddress(d, meta, vdc, org)
+		if err != nil {
+			return fmt.Errorf("Error retrieve IP: %#v", err)
 		}
 
 		d.Set("ip", ip)


lvirbalas:
Hi @harmjanblok , welcome back.

I suspect there may be an issue here with the update workflow. Historically, we have the "ip" parameter used for both (a) IP address and (b) allocation type. I believe this mix is the ultimate cause of bug here. Nevertheless, if we just remove returning of "allocated" and return the assigned-from-static-pool IP instead, won't we overwrite that this is an "allocated" IP address with the next update operation?

vbauzysvmware:
I am not keen regarding this change. This will create other issues as will be not possible to track what ip assignment was used. I need to spend more time to think and try if small work a round possible with using additional value where to set ip address internally or we need to split up property.

Didainius:
I agree with above comments. This seems very confusing for both vApps and VMs. 
I suspect your requirements would be satisfied with #177, wouldn't they?

lvirbalas:
@harmjanblok please respond to the above comments

harmjanblok:
> I suspect there may be an issue here with the update workflow. Historically, we have the "ip" parameter used for both (a) IP address and (b) allocation type. I believe this mix is the ultimate cause of bug here. Nevertheless, if we just remove returning of "allocated" and return the assigned-from-static-pool IP instead, won't we overwrite that this is an "allocated" IP address with the next update operation?

@lvirbalas when using 'allocated' or 'dhcp' for `ip` the next time I run `terraform plan` it reports the IP as changed an therefore wants to recreate the recourse. My workaround is to `ignore_changes = ["ip"]`. So this might indicate this mix (a) and (b) might cause some conflicts.

So the best option would be to split allocation type from actuall assignment.

lvirbalas:
> when using 'allocated' or 'dhcp' for ip the next time I run terraform plan it reports the IP as changed an therefore wants to recreate the recourse
@Didainius and @vbauzysvmware are you observing same behaviour?

Didainius:
@lvirbalas ,
I do see that, but it does not want to recreate it because it doesn't have `ForceNew`. It reports `update` and it happens because of Computed: true:
`"ip": {
				Type:     schema.TypeString,
				Optional: true,
				Computed: true,
			},`

What usually happens is that it reports as in-place update.

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

  ~ vcd_vapp_vm.insert_name
      ip: "192.168.1.111" => "allocated"
```

As far as fixing it,  #177 should be solving this and adding much more functionality. @harmjanblok, what do you think about that one?

I'm a bit afraid of altering the functionality of `ip` field now (although I agree it is flawed). People may have various hacks done on top of it.

Didainius:
@harmjanblok, I'm closing this as the new networks functionality coming with #177 will solve the issues you have.

Issue #118, opened 2018-11-21
Configuration of multiple NICs


WildWon:
Hi there,

When deploying a vm from template currently only one NIC can be configured. Vms with multiple NICs have to be manually reconfigured after deployment.

### Terraform Version
Terraform v0.11.10
+ provider.vcd (unversioned) built from (git commit 704012e) of the "vmware-takes-ownership" branch

### Affected Resource(s)
resource "vcd_vapp_vm"

### Terraform Configuration Files
```hcl
resource "vcd_vapp_vm" "my_vm01" {
	vapp_name = "${var.vappname}"
	name = "${var.vmnamewin}"
	catalog_name = "${var.catalognamewin}"
	template_name = "${var.templatenamewin}"
	memory = 4096
	cpus = 2
	ip = ""
	power_on = false
	network_name = "${var.networkname}"
}
```

### Debug Output
n/a

### Panic Output
n/a

### Expected Behavior
Being able to either configure all NICs of a template or alternatively add NICs incl. configuration to a vm at deployment.

### Actual Behavior
Only one NIC can be configured

### Steps to Reproduce
This can be easily reproduced by deploying a new vm to a vCloud instance

### Important Factoids
n/a

### References
n/a

lvirbalas:
Thanks @WildWon for this issue!

@vbauzysvmware please have a crack at it. One caveat is that we need to maintain backwards compatibility with existing behaviour.

dn1s:
I'm working on this feature. Will probably issue a pull request in a couple of days.

lvirbalas:
@dn1s, that's great. Please take assignment of this task if you can (for some reason GitHub doesn't allow me to select you from the list).

dn1s:
I have a question about this feature.
Can I deprecate network_name like this: https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-removal

In favor of a new parameter networks with a structure like this:

```
"networks": {
				Type:          schema.TypeList,
				Optional:      true,
				ForceNew:      true,
				ConflictsWith: []string{"ip"},
				Elem: &schema.Resource{
					Schema: map[string]*schema.Schema{
						"orgnetwork": {
							Type:     schema.TypeString,
							Required: true,
						},
						"ip": {
							Type:             schema.TypeString,
							Optional:         true,
							Computed:         true,
							DiffSuppressFunc: suppressIfIpIsOneOf(),
						},
						"ip_address_allocation_mode": {
[...]

PR #177, opened 2019-02-28
Add multi nic support

dn1s:
Issue #118

- Adds multi nic support
- Deprecates old network attributes (Code not yet removed see: https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-removal)
- Updates documentation how to use new network attribute


vcd/resource_vcd_vapp_vm.go
```diff
@@ -86,14 +87,66 @@ func resourceVcdVAppVm() *schema.Resource {
...
+							Type:     schema.TypeString,
+						},
+						"ip": {
+							Computed:         true,
+							DiffSuppressFunc: suppressIfIpIsOneOf(),
+							ForceNew:         true,
+							Optional:         true,
+							Type:             schema.TypeString,
+						},
+						"ip_address_allocation_mode": {
```
lvirbalas: Suggesting shorted name: `ip_allocation_mode`

vcd/resource_vcd_vapp_vm.go
```diff
@@ -86,14 +87,66 @@ func resourceVcdVAppVm() *schema.Resource {
...
+							Type:         schema.TypeString,
+							ValidateFunc: checkIpAddressAllocationMode(),
+						},
+						"is_primary": {
+							Default:  false,
+							ForceNew: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						/*
```
lvirbalas: Nit ;) Maybe it's just me, but these in-line comments between `/*` and `*/` poking out, especially because the paragraph is indented. What about just using `//`?

website/docs/r/vapp_vm.html.markdown
```diff
@@ -61,21 +73,28 @@ resource "vcd_vapp_vm" "web2" {
...
+one of dhcp, allocated or none. If given the address must be within the
+`static_ip_pool` set for the network. If left blank, and the network has
+`dhcp_pool` set with at least one available IP then this will be set with
+DHCP. **Conflicts with** with networks.
+* `memory` - (Optional) The amount of RAM (in MB) to allocate to the vApp
+* `network_name` - (**Deprecated**) Name of the network this VM should connect to. **Conflicts with** with networks.
+* `networks` - (Optional; *v2.1+*) List of network interfaces to attach to vm. **Conflicts with**: networks, ip. **Note**: all params of this parameter and itself do force recreation of vms!
+  * `orgnetwork` (Required) Name of the network this VM should connect to.
+  * `ip` (Optional) One of: dhcp, allocated, none or an ip.
+  * `ip_address_allocation_mode` (Optional) One of: DHCP, POOL, MANUAL or NONE. Defaults to "POOL"
```
lvirbalas: Need more explanation what POOL and NONE really mean.

lvirbalas:
BTW, there are failures in Travis which need to be fixed.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -228,9 +311,7 @@ func resourceVcdVAppVmCreate(d *schema.ResourceData, meta interface{}) error {
 
 // Adds existing org VDC network to VM network configuration
 // Returns configured OrgVDCNetwork for Vm, networkName, error if any occur
-func addVdcNetwork(d *schema.ResourceData, vdc govcd.Vdc, vapp govcd.VApp, vcdClient *VCDClient) (*types.OrgVDCNetwork, error) {
-
-	networkNameToAdd := d.Get("network_name").(string)
+func addVdcNetwork(networkNameToAdd string, vdc govcd.Vdc, vapp govcd.VApp, vcdClient *VCDClient) (*types.OrgVDCNetwork, error) {
```
Didainius: missing `d *schema.ResourceData` down below the code

vcd/resource_vcd_vapp_vm.go
```diff
@@ -390,8 +471,22 @@ func resourceVcdVAppVmRead(d *schema.ResourceData, meta interface{}) error {
...
 
 	d.Set("name", vm.VM.Name)
-	if len(vm.VM.NetworkConnectionSection.NetworkConnection) > 0 {
+	networks := d.Get("networks").([]interface{})
+	network := d.Get("network_name").(string)
+	switch {
+	// network_name is not set. networks is set in config
+	case network != "" && len(networks) == 0:
 		d.Set("ip", vm.VM.NetworkConnectionSection.NetworkConnection[0].IPAddress)
+		d.Set("mac", vm.VM.NetworkConnectionSection.NetworkConnection[0].MACAddress)
```
Didainius: Setting the `mac` on root schema, but there is no such attribute

vcd/resource_vcd_vapp_vm.go
dn1s: Can you tell me exactly where? I can't find it.

vcd/resource_vcd_vapp_vm.go
dn1s: Fixed.

vcd/resource_vcd_vapp_vm.go
dn1s: Fixed.

vcd/resource_vcd_vapp_vm.go
dn1s: Fixed.

website/docs/r/vapp_vm.html.markdown
dn1s: Fixed.

vcd/resource_vcd_vapp_vm.go
Didainius: Disregard. Sorry - misread it in code preview.

Didainius:
@dn1s, a quick note while I'm diving deeper. Could you change the PR to point to `master` (and rebase on master) rather than `vmware-stash-while-master...`. We're going to stick to using master therefore latest code is there  after 2.1 release.
Thanks

website/docs/r/vapp_vm.html.markdown
```diff
@@ -51,20 +55,28 @@ resource "vcd_vapp_vm" "web2" {
...
-  
+  networks = [{
+    orgnetwork                 = "net"
+    ip                         = "10.10.104.162"
+    ip_address_allocation_mode = "MANUAL"
+    is_primary                 = true
+  },
+  {
+    orgnetwork                 = "net"
+    ip_address_allocation_mode = "POOL"
```
Didainius:
```suggestion
    ip_allocation_mode = "POOL"
```

dn1s: Fixed.

website/docs/r/vapp_vm.html.markdown
```diff
@@ -77,18 +89,30 @@ The following arguments are supported:
 * `vapp_name` - (Required) The vApp this VM should belong to.
 * `name` - (Required) A unique name for the VM
 * `catalog_name` - (Required) The catalog name in which to find the given vApp Template
+* `name` - (Required) A unique name for the vApp
```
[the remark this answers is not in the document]
dn1s: Fixed.

website/docs/r/vapp_vm.html.markdown
```diff
@@ -77,18 +89,30 @@ The following arguments are supported:
...
   `dhcp_pool` set with at least one available IP then this will be set with
-  DHCP.
+DHCP. **Conflicts with** with networks.
+* `networks` - (Optional; *v2.2+*) List of network interfaces to attach to vm. **Conflicts with**: networks, ip. **Note**: all params of this parameter and itself do force recreation of vms!
+  * `orgnetwork` (Required) Name of the network this VM should connect to.
+  * `ip` (Optional) One of: dhcp, allocated, none or an ip.
+  * `ip_allocation_mode` (Optional) IP address allocation mode. Defaults to "POOL". One of:
+    * `POOL` (A static IP address is allocated automatically from a pool of addresses.)
+    * `DHCP` (The IP address is obtained from a DHCP service.)
+    * `MANUAL` (The IP address is assigned manually in the IpAddress element.)
```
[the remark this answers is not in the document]
dn1s: What exactly is unclear?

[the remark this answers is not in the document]
Didainius: >  * `MANUAL` (The IP address is assigned manually in the **IpAddress element**.)

Maybe it's just me, but it sounds as `ip attribute` would be better replacement.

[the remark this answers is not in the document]
dn1s: Oh that's taken directly from https://code.vmware.com/apis/72/vcloud/doc/doc/types/NetworkConnectionType.html

vcd/resource_vcd_vapp_vm.go
```diff
@@ -616,8 +707,22 @@ func resourceVcdVAppVmRead(d *schema.ResourceData, meta interface{}) error {
...
+	case network == "" && len(networks) > 0:
+		var networks []map[string]interface{}
+		for index, net := range d.Get("networks").([]interface{}) {
+			n := net.(map[string]interface{})
+			n["ip"] = vm.VM.NetworkConnectionSection.NetworkConnection[index].IPAddress
```
Didainius: This will crash even with `terraform refresh` if the nic was removed (or even not provisioned successfully due to misconfig) via some other mechanism. `panic: runtime error: index out of range`
Simple example:
* Provision the code as per sample tf  code
* Go into vCD and remove 2nd nic (but can also happen if provisioning failed for any NIC for whatever reason)
* `terraform refresh`
```
* vcd_vapp_vm.web1: 1 error(s) occurred:

* vcd_vapp_vm.web1: vcd_vapp_vm.web1: unexpected EOF


panic: runtime error: index out of range
2019-04-03T10:02:59.538+0300 [DEBUG] plugin.terraform-provider-vcd_v2.1.0:
2019-04-03T10:02:59.538+0300 [DEBUG] plugin.terraform-provider-vcd_v2.1.0: goroutine 98 [running]:
2019-04-03T10:02:59.538+0300 [DEBUG] plugin.terraform-provider-vcd_v2.1.0: github.com/terraform-providers/terraform-provider-vcd/v2/vcd.resourceVcdVAppVmRead(0xc000384000, 0x1ae6820, 0xc0004ba410, 0xc000384000, 0x0)
2019-04-03T10:02:59.538+0300 [DEBUG] plugin.terraform-provider-vcd_v2.1.0:      /Users/dserplis/Documents/vmware/terraform-provider-vcd/vcd/resource_vcd_vapp_vm.go:685 +0xc4e
```


```hcl
resource "vcd_vapp_vm" "web1" {
  vapp_name     = "${vcd_vapp.web.name}"
  name          = "web1"
  catalog_name  = "my-catalog97"
  template_name = "photon-os"
  memory        = 512
  cpus          = 2
  cpu_cores     = 1

  networks = [{
    orgnetwork                 = "${vcd_network_routed.net.name}"
    ip                         = "11.10.0.153"
    ip_allocation_mode         = "MANUAL"
    is_primary                 = true
  },{
    orgnetwork                 = "${vcd_network_routed.net.name}"
    ip                         = "11.10.0.154"
    ip_allocation_mode         = "MANUAL"
  }]
}
```

website/docs/r/vapp_vm.html.markdown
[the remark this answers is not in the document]
Didainius: Yep, but imagine a user who does not read API docs. I think it's better to write doc in Terraform context as majority of people will not be be familiar with API.

vcd/resource_vcd_vapp_vm.go
dn1s: I fixed that issue in the other branch ^^
I'll add my fix from there here as well.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -163,12 +251,31 @@ func resourceVcdVAppVmCreate(d *schema.ResourceData, meta interface{}) error {
...
 			return err
 		}
+		if network != nil {
+			networks = append(networks, network)
+		}
+		networksChanges = append(networksChanges, map[string]interface{}{
+			"ip":         d.Get("ip").(string),
+			"orgnetwork": network.Name,
+		})
+	} else if d.Get("networks").([]interface{}) != nil {
```
Didainius: Can be simplified to a simple if because it is guarded by ConflictsWith?

```suggestion
	}
	if d.Get("networks").([]interface{}) != nil {
```

vcd/resource_vcd_vapp_vm.go
```diff
@@ -616,8 +707,22 @@ func resourceVcdVAppVmRead(d *schema.ResourceData, meta interface{}) error {
...
 		d.Set("ip", vm.VM.NetworkConnectionSection.NetworkConnection[0].IPAddress)
+		d.Set("mac", vm.VM.NetworkConnectionSection.NetworkConnection[0].MACAddress)
+	case network == "" && len(networks) > 0:
+		var networks []map[string]interface{}
+		for index, net := range d.Get("networks").([]interface{}) {
+			n := net.(map[string]interface{})
+			n["ip"] = vm.VM.NetworkConnectionSection.NetworkConnection[index].IPAddress
+			n["mac"] = vm.VM.NetworkConnectionSection.NetworkConnection[index].MACAddress
+			networks = append(networks, n)
+			d.Set("networks", networks)
```
Didainius: Are you sure you need to set it inside the loop?
If I understand correctly this will trigger statefile updates everytime.

Didainius:
Probably a harder ask but could you cover these bigger problems with tests. I'm afraid we'll run into these sooner or later.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -96,13 +107,65 @@ func resourceVcdVAppVm() *schema.Resource {
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"orgnetwork": {
+							ForceNew: true,
+							Required: true,
+							Type:     schema.TypeString,
+						},
+						"ip": {
+							Computed:         true,
+							DiffSuppressFunc: suppressIfIPIsOneOf(),
```
Didainius: I'm thinking if we could fix this odd behavior when text attribute is overriden with IP address (like in the old case). What are your thoughts if we only allow to enter IP address here for manual case and return computed in other cases?

dn1s: I would like to go for that. I tried to recreate current behavior.

Didainius: I think as this is new functionality we can make it proper (no one will get a surprise) here (but preserve it in the original implementation until we remove it in future).

vcd/resource_vcd_vapp_vm.go
dn1s: Fixed.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -132,6 +195,31 @@ func resourceVcdVAppVm() *schema.Resource {
 	}
 }
 
+func checkIPAddressAllocationMode() schema.SchemaValidateFunc {
+	return func(v interface{}, key string) (warns []string, errs []error) {
+		if v == "POOL" || v == "DHCP" || v == "MANUAL" || v == "NONE" {
+			return
+		}
+		errs = append(errs, fmt.Errorf("ip_address_allocation_mode must be one of POOL, DHCP, MANUAL or NONE got: %v", v))
```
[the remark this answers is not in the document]
dn1s: Fixed.

vcd/resource_vcd_vapp_vm.go
dn1s: Fixed.

vcd/resource_vcd_vapp_vm.go
dn1s: Added validation for attribute "ip"

vcd/resource_vcd_vapp_vm.go
```diff
@@ -616,8 +732,25 @@ func resourceVcdVAppVmRead(d *schema.ResourceData, meta interface{}) error {
...
+	switch {
+	// network_name is not set. networks is set in config
+	case network != "":
 		d.Set("ip", vm.VM.NetworkConnectionSection.NetworkConnection[0].IPAddress)
+		d.Set("mac", vm.VM.NetworkConnectionSection.NetworkConnection[0].MACAddress)
+	case len(networks) > 0:
+		var nets []map[string]interface{}
+		for index, net := range d.Get("networks").([]interface{}) {
+			n := net.(map[string]interface{})
+			if len(vm.VM.NetworkConnectionSection.NetworkConnection) > 0 {
```
Didainius: Imagine I've got 1 NIC in VM and 2 NICs in .tf. It would still hit an `index out of range` panic.
I think in `case len(networks) > 0:` (which means we're using the newer functionality) it is worth to loop over `vm.VM.NetworkConnectionSection.NetworkConnection` rather than `d.Get("networks").([]interface{})`. That way you will save the real state in vCD into the `terraform.tfstate` file.
What are your thoughts on doing it this way round?

vcd/resource_vcd_vapp_vm.go
```diff
@@ -96,13 +107,70 @@ func resourceVcdVAppVm() *schema.Resource {
...
+							Type:     schema.TypeString,
+						},
+						"ip": {
+							Computed: true,
+							ForceNew: true,
+							Optional: true,
+							Type:     schema.TypeString,
+							ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
+								if net.ParseIP(val.(string)) == nil {
+									errs = append(errs, fmt.Errorf("IP must be valid, got: %d", val.(string)))
```
Didainius:
```suggestion
									errs = append(errs, fmt.Errorf("IP must be valid, got: %s", val.(string)))
```

vcd/resource_vcd_vapp_vm.go
dn1s: I do understand your concern and I'll change it.

dn1s:
vm.go +262
```
if networkConnection.Network == network["orgnetwork"] {
```
This is the culprit. If network name is the same it just sets the ip for both network connections.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -96,13 +107,70 @@ func resourceVcdVAppVm() *schema.Resource {
...
+							ForceNew: true,
+							Required: true,
+							Type:     schema.TypeString,
+						},
+						"ip": {
+							Computed: true,
+							ForceNew: true,
+							Optional: true,
+							Type:     schema.TypeString,
+							ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
```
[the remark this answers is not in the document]
dn1s: Fixed.

vcd/resource_vcd_vapp_vm.go
dn1s: Fixed as in replaced it with SingleIP() function

vcd/resource_vcd_vapp_vm.go
```diff
@@ -182,18 +270,74 @@ func resourceVcdVAppVmCreate(d *schema.ResourceData, meta interface{}) error {
...
+		for i := 0; i < len(vAppNetworkNames); i++ {
+			m[vAppNetworkNames[i]] = true
+		}
+		for _, netName := range netNames {
+			// if the network is not assigned, assigne it to vapp
+			if _, ok := m[netName]; !ok {
+				//err = retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
+				n, err := vdc.FindVDCNetwork(netName)
+				_, err = addVdcNetwork(n.OrgVDCNetwork.Name, vdc, vapp, vcdClient)
+				// task, err := vapp.AppendNetworkConfig(n.OrgVDCNetwork)
```
Didainius: Cleanup these commented out lines if it's handled inside `addVdcNetwork`

vcd/resource_vcd_vapp_vm.go
```diff
@@ -65,9 +67,19 @@ func resourceVcdVAppVm() *schema.Resource {
...
+				ConflictsWith:    []string{"networks"},
+				Deprecated:       "In favor of networks",
+				DiffSuppressFunc: suppressIfIPIsOneOf(),
+				ForceNew:         true,
+				Optional:         true,
+				Type:             schema.TypeString,
+			},
+			"mac": {
+				Computed:      true,
+				ConflictsWith: []string{"networks"},
```
Didainius: Add 'Deprecated'

vcd/resource_vcd_vapp_vm.go
```diff
@@ -96,13 +108,65 @@ func resourceVcdVAppVm() *schema.Resource {
...
 			},
 			"network_href": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
+				Type:       schema.TypeString,
+				Optional:   true,
+				Deprecated: "In favor of networks",
+			},
+			"networks": {
+				ConflictsWith: []string{"ip", "network_name"},
```
Didainius: `mac` as well

Didainius:
Can you temporarily point to your fork in go.mod using replace directive so that Travis is also happy.

website/docs/r/vapp_vm.html.markdown
```diff
@@ -97,13 +109,24 @@ The following arguments are supported:
...
+    * `NONE` (No IP addressing mode specified.)
```
Didainius: Is there still a point to keep "NONE"? One could probably omit defining NIC at all? Or am I missing something?

Didainius: I tested it and see defaulted to "DHCP" in the end. To me it looks as if this would leave source of confusion and wouldn't allow to enter such mode via validator in the new config.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -616,9 +778,33 @@ func resourceVcdVAppVmRead(d *schema.ResourceData, meta interface{}) error {
...
+			singleNIC["orgnetwork"] = vmNet.Network
+			singleNIC["ip_allocation_mode"] = vmNet.IPAddressAllocationMode
+
+			singleNIC["is_primary"] = false
+			if i == vm.VM.NetworkConnectionSection.PrimaryNetworkConnectionIndex {
+				singleNIC["is_primary"] = true
+			}
+			nets = append(nets, singleNIC)
+		}
+		d.Set("networks", nets)
```
Didainius: I know I proposed this, but I spotted it is missing one more feature.
Could you sort the `nets` by NetworkConnectionIndex. What happens now is that the statefile does not necessarily gets update with the right order of network interfaces.

image
 
```
"networks.#": "4",
                            "networks.0.ip": "11.10.0.153",
                            "networks.0.ip_allocation_mode": "POOL",
                            "networks.0.is_primary": "false",
                            "networks.0.mac": "00:50:56:01:03:d2",
                            "networks.0.orgnetwork": "multinic-net",
                            "networks.1.ip": "",
                            "networks.1.ip_allocation_mode": "DHCP",
                            "networks.1.is_primary": "true",
                            "networks.1.mac": "00:50:56:01:03:d3",
                            "networks.1.orgnetwork": "multinic-net",
                            "networks.2.ip": "12.10.0.152",
                            "networks.2.ip_allocation_mode": "POOL",
                            "networks.2.is_primary": "false",
                            "networks.2.mac": "00:50:56:01:03:d4",
                            "networks.2.orgnetwork": "multinic-net2",
                            "networks.3.ip": "11.10.0.152",
                            "networks.3.ip_allocation_mode": "POOL",
                            "networks.3.is_primary": "false",
                            "networks.3.mac": "00:50:56:01:03:d1",
                            "networks.3.orgnetwork": "multinic-net",
```

If I do a `terraform apply` with zero it then suggests to recreate NICs because of incorrect order.

website/docs/r/vapp_vm.html.markdown
dn1s: As far as I remember "NONE" wasn't working for me anyways. I think the trick would have been to not connect the network adapter if you use "NONE" but the api documentation didn't help too much either so I dropped fixing it. So we can just remove it completely.

Didainius: None is fixed and should be working properly now.

Didainius:
Work continues in #233

PR #180, opened 2019-03-05
fix 'make vet' and resolve vet issues

### LINKED ISSUE #179: make vet does not work as it should in GNUmakefile
GNUmakefile has a directive called `vet` and currently it is not doing it's duty.

`make vet` checks if there are any suspicious constructs in the code, but at the moment we're only reporting error if it's exit code is equal to 1.

As per the Go team discussion this is not so: https://go-review.googlesource.com/c/go/+/101918 
> cmd/vet: only zero/non-zero exit status matters

> "go vet" exits with 2 on type-checking errors (because the compiler
> itself does so), but with 1 on vet errors.
>
> But we're really not trying to convey any nuance in vet's exit status.
> It's really just one bit: success or failure. Update the documentation
> so it doesn't claim anything subtler than this.

We currently have a few errors but `make vet` does not fail:
```
go vet .
# github.com/terraform-providers/terraform-provider-vcd/v2/vcd
vcd/resource_vcd_vapp_vm.go:153:14: result of fmt.Errorf call not used
vcd/resource_vcd_vapp_vm.go:240:13: result of fmt.Errorf call not used
```

I'm about to fix this.

Didainius:
Fixes #179

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -28,7 +28,7 @@ testacc: fmtcheck
 
 vet:
 	@echo "go vet ."
-	@go vet $$(go list ./... | grep -v vendor/) ; if [ $$? -eq 1 ]; then \
+	@go vet $$(go list ./... | grep -v vendor/) ; if [ $$? -ne 0 ]; then \
 		echo ""; \
 		echo "Vet found suspicious constructs. Please check the reported constructs"; \
 		echo "and fix them if necessary before submitting the code for review."; \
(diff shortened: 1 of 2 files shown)


lvirbalas:
Please always add @vbauzysvmware to reviewers as well.

Didainius:
> Please always add @vbauzysvmware to reviewers as well.

Ok. I've just added it as a draft at first and will move to review. Running some tests for now although this is very edgy and wouldn't be hit anyway.

Issue #68, opened 2017-11-08
vApp error: The requested operation on vApp is not supported in its current state


pkleanthous:
I'm trying to create a vApp and a VM inside the vApp and it fails as vApp is not ready when terraform tries to execute the next resource.

### Terraform Version
Terraform v0.10.8

### Affected Resource(s)
- vcd_vapp
- vcd_vapp_vm

### Terraform Configuration Files
```hcl
resource "vcd_vapp" "test-demo" {
  name = "test-demo"
}

resource "vcd_vapp_vm" "demo-vm-1" {
  vapp_name     = "${vcd_vapp.test-demo.name}"
  name          = "demo-vm-1"
  catalog_name  = "magic_catalog"
  template_name = "ubuntu-xenial"
  memory        = 2048
  cpus          = 1
  network_name  = "Magic Network"
  depends_on    = ["vcd_vapp.test-demo"]
}
```

### Debug Output
```
2017/11/08 13:19:32 [DEBUG] apply: vcd_vapp.test-demo: executing Apply
2017-11-08T13:19:32.095Z [DEBUG] plugin.terraform-provider-vcd_v1.0.0_x4: 2017/11/08 13:19:32 [DEBUG] Waiting for state to become: [success]
2017/11/08 13:19:33 [TRACE] root: eval: *terraform.EvalWriteState
2017/11/08 13:19:33 [TRACE] root: eval: *terraform.EvalApplyProvisioners
2017/11/08 13:19:33 [TRACE] root: eval: *terraform.EvalIf
2017/11/08 13:19:33 [TRACE] root: eval: *terraform.EvalWriteState
2017/11/08 13:19:33 [TRACE] root: eval: *terraform.EvalWriteDiff
2017/11/08 13:19:33 [TRACE] root: eval: *terraform.EvalApplyPost
2017/11/08 13:19:33 [ERROR] root: eval: *terraform.EvalApplyPost, err: 1 error(s) occurred:

* vcd_vapp.test-demo: Error Powering Up: &errors.errorString{s:"error powering on vApp: API Error: 400: [ 619fb0ef-167e-4684-9e58-b972cd120ace ] The requested operation on vApp \"test-demo\" is not supported in its current state."}
2017/11/08 13:19:33 [ERROR] root: eval: *terraform.EvalSequence, err: 1 error(s) occurred:

[...]

PR #186, opened 2019-03-12
Check if vApp is RESOLVED before powering it on

Didainius:
This PR should solve #68 and also solves crashing acceptance test problem when run on a fast local environment.
It should come together with https://github.com/vmware/go-vcloud-director/pull/169 to satisfy all dependencies.

## The problem
The basis of it is that one cannot trigger power on when the vApp is in "UNRESOLVED" (0) state. It happens for a very short time after a new empty vApp is being created. I'm introducing a blocking method to wait for state change before a power on task.

I have also considered embedding this check into `vapp.PowerOn()` in `go-vcloud-director` to guard any poweron. Do let me know your thoughts.

### DIFF
--- a/vcd/resource_vcd_vapp.go
+++ b/vcd/resource_vcd_vapp.go
@@ -374,6 +374,13 @@ func resourceVcdVAppUpdate(d *schema.ResourceData, meta interface{}) error {
 		}
 
 		if d.Get("power_on").(bool) {
+			// Ensure the vApp is not in UNRESOLVED status before powering on. This happens for
+			// less than a second after it is created.
+			err := vapp.StatusWaitNot("UNRESOLVED", vcdClient.MaxRetryTimeout)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


Didainius:
> I agree that this function should be implicitly included in crucial operations that users expect to succeed on fast environments :)

The more I think, the more I'm keen to do this. If everybody agrees then I probably abandon this PR and will add the `vapp.BlockWhileStatus(unwantedStatus string, timeOutAfterSeconds int)` into `vapp.PowerOn()` in `go-vcloud-director` with https://github.com/vmware/go-vcloud-director/pull/169
This would guard all powerOns from errors, not only the one I hit with tests

lvirbalas:
I agree. @vbauzysvmware ?

The only remaining question is what timeout value you'll use.

Didainius:
Agree. Investigating how to pass on the timeout value.

Issue #172, opened 2019-02-19
vApp + VM creation fails when multiple VMs are present in vApp


ukcloud-rubeon:
### Terraform Version

```
$ terraform version
Terraform v0.11.7
+ provider.vcd v2.0.0
```

### Affected Resource(s)

- `vcd_vapp`
- `vcd_vapp_vm`

### Terraform Configuration Files
```hcl
variable "catalog" {
  default = "MyVCD Catalogue"
}

variable "public_ip" {
  default = "1.2.3.4"
}

variable "vcd_max_retry_timeout" {
    default = 600
}

variable "centos" {
  default = "MyVCD_CentOS_7_3_x64_60GB_Small_v1.0.2"
}

variable "vcd_edgegateway" {
    default = "Internet_02(xxxx)"
}

variable "cidr" {
    default = {
        external = "172.16.3.0/24"
    }
}

variable "vcd_allow_unverified_ssl" {
  default = false
}

variable "vcd_logging" {
    default = true
}


variable "external" {
  default = {
      gw = "172.16.3.1"
      static_start_address = "172.16.3.2"
      static_end_address = "172.16.3.100"
      dhcp_start_address = "172.16.3.101"
      dhcp_end_address = "172.16.3.200"

  }
}

provider "vcd"  {
    user = "${var.vcd_creds["username"]}"
    password = "${var.vcd_creds["password"]}"
    org = "${var.vcd_creds["org"]}"
    vdc = "${var.vcd_creds["vdc"]}"
    url = "${var.vcd_creds["url"]}"
    allow_unverified_ssl = "${var.vcd_allow_unverified_ssl}"
    max_retry_timeout    = "${var.vcd_max_retry_timeout}"
    logging = "${var.vcd_logging}"
}


resource "vcd_network_routed" "external" {
    org = "${var.vcd_creds["org"]}"
    vdc = "${var.vcd_creds["vdc"]}"

    name = "external-network"
    edge_gateway = "${var.vcd_edgegateway}"
    gateway = "${var.external["gw"]}"

    static_ip_pool = {
        start_address = "${var.external["static_start_address"]}"
        end_address = "${var.external["static_end_address"]}"
    }

    dns1 = "8.8.8.8"
    dns2 = "8.8.4.4"

    dhcp_pool = {
        start_address = "${var.external["dhcp_start_address"]}"
        end_address = "${var.external["dhcp_end_address"]}"
    }

}

resource "vcd_vapp" "web" {
[...]

PR #195, opened 2019-03-19
Fix multi vm issue

vbauzysvmware:
Ref: #172

For this release fix is using retryable approach. In next release when we check with vCD team is using mutex lock. Seems that vApp can't handle multi requests on same time. So multi vm creation should be linear.


vcd/resource_vcd_vapp_vm.go
```diff
@@ -450,11 +450,14 @@ func resourceVcdVAppVmUpdate(d *schema.ResourceData, meta interface{}) error {
 		}
 
 		if d.Get("power_on").(bool) {
-			task, err := vm.PowerOn()
-			if err != nil {
-				return fmt.Errorf("error Powering Up: %#v", err)
-			}
-			err = task.WaitTaskCompletion()
+			err = retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
```
lvirbalas: Please add a comment on why this retry is needed.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -648,14 +651,17 @@ func resourceVcdVAppVmDelete(d *schema.ResourceData, meta interface{}) error {
 	log.Printf("[TRACE] vApp Status:: %s", status)
 	if status != "POWERED_OFF" {
 		log.Printf("[TRACE] Undeploying vApp: %s", vapp.VApp.Name)
-		task, err := vapp.Undeploy()
+		err = retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
```
lvirbalas: Please add a comment on why this retry is needed.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -680,13 +686,16 @@ func resourceVcdVAppVmDelete(d *schema.ResourceData, meta interface{}) error {
 		}
 
 		log.Printf("[TRACE] Powering on vApp: %s", vapp.VApp.Name)
-		task, err = vapp.PowerOn()
-		if err != nil {
-			return fmt.Errorf("error Powering on vApp: %#v", err)
-		}
-		err = task.WaitTaskCompletion()
+		err = retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
```
lvirbalas: Please add a comment on why this retry is needed.

BTW, for the future we should seriously think about encapsulating such behaviour inside the vapp.PowerOn() function, so that all callers benefit from it.

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: I added generic comment for all retryCall. 
```
// all functionality using retryCall as multi requests to same vAPP isn't handled and errors occur
```

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: I added generic comment for all retryCall.

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: This isn't PowerOn() issue it's global vApp issue. We will need some kind vApp scoped lock.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -397,6 +397,7 @@ func resourceVcdVAppVmUpdate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("error getting VM status: %#v", err)
 	}
 
+	// all functionality using retryCall as multi requests to same vAPP isn't handled and errors occur
```
lvirbalas:
```suggestion
	// When there is more then one VM in a vApp Terraform will try to parallelise their creation.
	// However, vApp throws errors when simultaneous requests are executed. To avoid them, below block is using retryCall in multiple places as a workaround, so that the VMs are created regardless of parallelisation.
```
(not fmt'ed though)

vbauzysvmware: Fixed

PR #197, opened 2019-03-20
Review changelog pre 2.1

Didainius:


website/docs/r/independent_disk.html.markdown
```diff
@@ -45,8 +45,8 @@ resource "vcd_vapp_vm" "web2" {
 
 The following arguments are supported:
 
-* `org` - (Optional; *v2.0+*) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
-* `vdc` - (Optional; *v2.0+*) The name of VDC to use, optional if defined at provider level
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
```
lvirbalas: Why have you removed the `v2.0+` tag? Given the amount of changes happening in the 2.0, 2.1 and next release, I'd like to keep using these tags in documentation, so for the users, who have not yet upgraded, it's easy to spot the new features.

Didainius: But independent disk resource wasn't there before 2.1 at all.

website/docs/r/vapp.html.markdown
```diff
@@ -24,12 +24,6 @@ resource "vcd_network_direct" "net" {
 resource "vcd_vapp" "web" {
   name = "web"
 
-  metadata {
```
lvirbalas: First I thought why this has been deleted, but then I realised that you removed what is not working yet (until @dn1s metadata PRs are merged). The doc for the old way of defining metadata in a vApp with VM in a _single_ resource (which we want to deprecate and which is described in the doc below) has been left.

Issue #157, opened 2019-02-01
Index out of range when creating new vm


spengilley:
### Terraform Version
0.11.11

### Affected Resource(s)
vcd_~vapp_vm

### Terraform Configuration Files
```hcl
# * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
# Create a vApp
# * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
resource "vcd_vapp" "master" {
    name         = "master"
    depends_on   = ["vcd_network_routed.INTERNET"]
}

# * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
# Create a Vm
# * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
resource "vcd_vapp_vm" "master" {
    vapp_name     = "${vcd_vapp.master.name}"
    name          = "master"
    
    # Define the catalog and template
    catalog_name  = "VANILLA"
    template_name = "VANILLA"

    # Give the app internet access so that it can be provisioned
    # This network resource is defined in networks.tf
    network_name  = "${vcd_network_routed.INTERNET.name}"
    ip            = "192.168.30.2"
    
    # Do enough to be able to run puppet agent
    # These steps are specifically for a puppet agent
    # provisioner "remote-exec" {
    #     inline    = [
    #         "apt-get update && apt-get install -y puppet",
    #         "echo ${local.flavour}:01 > /flavour.txt"
    #     ]
    # } 

    # These steps are specifically for a puppet master
    provisioner "remote-exec" {
        inline    = [
            "apt-get update && apt-get install -y unzip"
        ]
    }
}

```

### Panic Output
https://gist.github.com/spengilley/5e6b0210a23b6d269c0e13d253510005

### Expected Behavior
Create vm

### Actual Behavior
crashes

### Steps to Reproduce
1. `terraform apply`

- GH-1234

### Comment from Developer

In issue real error is that vApp template returned something but not fully what was expected and govcd panicked. According user issue was that vApp template(OVA) wasn't uploaded successfully. I couldn't recreate same error but got different null pointer exception when ova is partly uploaded. I added check that before using vApp template to check it status. Status 8 is parsed from responses - didn't find explanations in documentations. When vApp template was with error then status was 0.
-> Add check for vApp template to be valid #151

vbauzysvmware:
Which branch or version of provider you are using?
[...]

Issue #151, opened 2019-01-14
Resource vcd_firewall_rules protocol setting is ignored


vbauzysvmware:
vcd_firewall_rules resource ignores protocol setting and creates rule using "any".

Example of used tf script:

resource "vcd_firewall_rules" "rule1" {
  edge_gateway   = "myEdge"
  default_action = "allow"

  rule {
    description      = "Test rule3"
    policy           = "allow"
    protocol         = "TCP"
    destination_port = "80"
    destination_ip   = "x.x.x.x"
    source_port      = "8080"
    source_ip        = "y.y.y.y"
  }
}

PR #203, opened 2019-03-25
Documentation fixes before 2.1 release

Didainius:


lvirbalas:
Thanks, this will help the newcomers, who just want to clone the repo without prior SSH key setup.

Just piggy backing on this commit, could you also improve this doc page:

http://localhost:4567/docs/providers/vcd/r/vapp_vm.html#vapp_network_name
> ip - (Optional) The IP to assign to this vApp. Must be an IP address or one of dhcp, allocated or none. If given the address must be within the static_ip_pool set for the network. If left blank, and the network has dhcp_pool set with at least one available IP then this will be set with DHCP.

I suggest quoting the IP types, because it's not totally clear that these are keywords. Suggestion:

> Must be an IP address or one of "dhcp", "allocated" or "none".

p.s. The URL below is in the Docker webserver that is created with `make website`.

lvirbalas:
One more thing, @vbauzysvmware README.md says:
> * resource/vcd_vapp_vm add additional validation to check if vApp template is OK [GH-157] 

But #157 is this:
https://github.com/terraform-providers/terraform-provider-vcd/issues/157

vbauzysvmware:
@lvirbalas and what is issue with #157 ?

lvirbalas:
It is not clear that this link is correct, because the issue name is different and it is still open. I'd suggest relinking to the PR as per guidelines. The PR is clearer and references issue in the first line for those who want more history.

vbauzysvmware:
@Didainius Could update to 157 to #151 Add check for vApp template to be valid

Didainius:
> I suggest quoting the IP types, because it's not totally clear that these are keywords.

Agreed. I tried to emphasize using back ticks only, but it was still purely visible so left those quoted.

CHANGELOG.md
```diff
@@ -22,7 +22,7 @@ IMPROVEMENTS:
 
 BUG FIXES:
 * resource/vcd_vapp, resource/vcd_vapp_vm add vApp status handling when environment is very fast [GH-68]
-* resource/vcd_vapp_vm add additional validation to check if vApp template is OK [GH-157] 
+* resource/vcd_vapp_vm add additional validation to check if vApp template is OK [GH-151]
```
lvirbalas: Oops! 151 is a PR in go-vcloud-director project, so the above turned out linking to a wrong *issue* in terraform-provider-vcd

Didainius: Done. force pushed without the changes at all.

lvirbalas:
@Didainius let's add a comment to CHANGELOG.md similar to the one we had in v2.0.0:

`Please look for "v2.1+" keyword in documentation which is used to emphasize new features.`

PR #212, opened 2019-03-26
Upgrade govcd to 2.1.1 and change vapp network parameter

vbauzysvmware:
Upgrade govcd to 2.1.1 and change vapp network parameter to be not sent if not provided. This allows to support vCD 8.2 version

### DIFF
--- a/vcd/resource_vcd_vapp_network.go
+++ b/vcd/resource_vcd_vapp_network.go
@@ -72,7 +72,6 @@ func resourceVcdVappNetwork() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 				ForceNew: true,
-				Default:  false,
 			},
 
 			"dhcp_pool": &schema.Schema{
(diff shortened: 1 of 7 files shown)


vcd/resource_vcd_vapp_network.go
```diff
@@ -233,7 +236,7 @@ func resourceVappNetworkRead(d *schema.ResourceData, meta interface{}) error {
 			d.Set("dns2", c.IPScopes.IPScope.DNS2)
 			d.Set("dnsSuffix", c.IPScopes.IPScope.DNSSuffix)
 		}
-		d.Set("guestVlanAllowed", c.GuestVlanAllowed)
```
lvirbalas: How did this work before? Was there a different bug?

vbauzysvmware: Nothing was overriden, nothing was querring.

PR #220, opened 2019-04-11
Add hardware_assisted_virtualization for vcd_vapp_vm resource

### LINKED ISSUE #219: Introduce hypervisor nesting option for VMs
This community requested feature will allow to setup nested hypervisors in vCD (which is very useful for various LAB scenarios).

This issue is set to track work on both repos. It was originally driven in PR #71, but is abandoned and does not cover `go-vcloud-director` related work.

In a nutshell it allows to run ESXi (and other hypervisors) on ESXi. More details can be found in [documentation](https://docs.vmware.com/en/VMware-vSphere/6.7/com.vmware.vsphere.vm_admin.doc/GUID-2A98801C-68E8-47AF-99ED-00C63E4857F6.html).

GUI-wise the below checkbox will get enabled:
image

Didainius:
Closes #219.
It allows to enable [hardware assisted CPU virtualization](https://docs.vmware.com/en/VMware-vSphere/6.5/com.vmware.vsphere.vm_admin.doc/GUID-2A98801C-68E8-47AF-99ED-00C63E4857F6.html) which is mainly used for hypervisor nesting (ESXi on ESXi) scenarios. It is a very much needed feature for establishing environment LAB.

In addition to this there are few more tidy-ups added:
* ~~`type StringMap` has two attached methods `Copy()` and `CopyMutateKey(key string, value interface{})` for convenience while writing multistep acceptance tests while altering single key in configParams.~~
* A few code cleanups in 1b230e3, 114dbb1,  6efc369
* Changed checked functions `testAccCheckVcdVAppVmExists` and `testAccCheckVcdVAppVmDestroy` to accept vApp and VM names instead of hardcoded ones.

~~One issue which is spotted is that if a test has multiple steps the `test-artifacts` file will get overwritten with the latest version (which can mask error while troubleshooting multistep acceptance tests). I did not change it there, but my proposal would be to:~~
~~*Flush `test-artifacts` directory on new test run~~
~~*During test run while writing the `.tf` file check if it exists. If it does then increment it's name with something like `testName-stepX.tf`. That way the original file for step0 would be called as it was called, but the subsequent steps would be clearly distinguishable.~~
Using `FuncName` to split test-artifact files.

Full test suite run on vCD 9.5

**Note**. go.mod is temporary pointed to my fork to make Travis work. It will be changed before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 FEATURES:
 
 * `vcd_vapp_vm` - Ability to add metadata to a VM. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
+* `vcd_vapp_vm` - Ability to enable hardware assisted CPU virtualization for VM. It allows hypervisor nesting. [#219]
 
 BUG FIXES:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,16 +2,16 @@
 
 FEATURES:
 
-* `vcd_vapp_vm` - Ability to add metadata to a VM. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
-* `vcd_vapp_vm` - Ability to enable hardware assisted CPU virtualization for VM. It allows hypervisor nesting. [#219]
+* `vcd_vapp_vm` - Ability to add metadata to a VM. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` [GH-158]
+* `vcd_vapp_vm` - Ability to enable hardware assisted CPU virtualization for VM. It allows hypervisor nesting. [GH-219]
 
 BUG FIXES:
 
-* `vcd_vapp` - Ability to add metadata to empty vApp. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
+* `vcd_vapp` - Ability to add metadata to empty vApp. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` [GH-158]
 
 BACKWARDS INCOMPATIBILITIES / NOTES:
 
-* `vcd_vapp` - Metadata is no longer added to first VM in vApp it will be added to vApp directly instead. ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
+* `vcd_vapp` - Metadata is no longer added to first VM in vApp it will be added to vApp directly instead. [GH-158]
 
 ## 2.1.0 (March 27, 2019)
 
--- a/go.mod
+++ b/go.mod
@@ -7,4 +7,4 @@ require (
 	github.com/vmware/go-vcloud-director/v2 v2.2.0-alpha.1
 )
 
-replace github.com/vmware/go-vcloud-director/v2 => github.com/Didainius/go-vcloud-director/v2 v2.2.0-alpha.1.0.20190416112520-dcfd6747a9ce
+replace github.com/vmware/go-vcloud-director/v2 => github.com/Didainius/go-vcloud-director/v2 v2.2.0-alpha.1.0.20190417085158-6cbca97d9053
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,5 @@ go 1.12
 
 require (
 	github.com/hashicorp/terraform v0.11.13
-	github.com/vmware/go-vcloud-director/v2 v2.2.0-alpha.1
+	github.com/vmware/go-vcloud-director/v2 v2.2.0-alpha.2
 )
-
-replace github.com/vmware/go-vcloud-director/v2 => github.com/Didainius/go-vcloud-director/v2 v2.2.0-alpha.1.0.20190419085123-885ddab6bd61
--- a/go.mod
+++ b/go.mod
@@ -6,3 +6,5 @@ require (
 	github.com/hashicorp/terraform v0.11.13
 	github.com/vmware/go-vcloud-director/v2 v2.2.0-alpha.1
 )
+
+replace github.com/vmware/go-vcloud-director/v2 => github.com/Didainius/go-vcloud-director/v2 v2.2.0-alpha.1.0.20190411133523-6676eb189c4d
--- a/go.mod
+++ b/go.mod
@@ -7,4 +7,4 @@ require (
 	github.com/vmware/go-vcloud-director/v2 v2.2.0-alpha.1
 )
 
-replace github.com/vmware/go-vcloud-director/v2 => github.com/Didainius/go-vcloud-director/v2 v2.2.0-alpha.1.0.20190417085158-6cbca97d9053
+replace github.com/vmware/go-vcloud-director/v2 => github.com/Didainius/go-vcloud-director/v2 v2.2.0-alpha.1.0.20190419085123-885ddab6bd61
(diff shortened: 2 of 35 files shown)


vcd/resource_vcd_vapp_vm_hw_virtualization_test.go
```diff
@@ -0,0 +1,104 @@
...
+					resource.TestCheckResourceAttr(
+						"vcd_vapp_vm."+vmNameHwVirt, "expose_hardware_virtualization", "true"),
+				),
+			},
+		},
+	})
+}
+
+const testAccCheckVcdVAppVm_hardwareVirtualization = `
+resource "vcd_network_routed" "{{.NetworkName}}" {
```
vbauzysvmware: To save time for test execution - maybe network part isn't needed?

Didainius:
> To save time for test execution - maybe network part isn't needed?
Very nice spot. Removed.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 FEATURES:
 
 * `vcd_vapp_vm` - Ability to add metadata to a VM. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
+* `vcd_vapp_vm` - Ability to enable hardware assisted CPU virtualization for VM. It allows hypervisor nesting. [#219]
```
lvirbalas: Shouldn't we use "GH-" format which is then parsed by the build bot?
https://www.terraform.io/docs/extend/best-practices/versioning.html

go.mod
```diff
@@ -6,3 +6,5 @@ require (
 	github.com/hashicorp/terraform v0.11.13
 	github.com/vmware/go-vcloud-director/v2 v2.2.0-alpha.1
 )
+
+replace github.com/vmware/go-vcloud-director/v2 => github.com/Didainius/go-vcloud-director/v2 v2.2.0-alpha.1.0.20190417085158-6cbca97d9053
```
lvirbalas: What about merging the govcd PR, bumping up govcd to v2.2.0-alpha.2 and then dropping this replace from this PR?

go.sum
```diff
@@ -3,6 +3,8 @@ github.com/Azure/azure-sdk-for-go v10.3.0-beta+incompatible/go.mod h1:9XXNKU+eRn
 github.com/Azure/go-autorest v9.10.0+incompatible/go.mod h1:r+4oMnoxhatjLLJ6zxSWATqVooLgysK6ZNox3g/xq24=
 github.com/Azure/go-ntlmssp v0.0.0-20170803034930-c92175d54006/go.mod h1:chxPXzSsl7ZWRAuOIE23GDNzjWuZquvFlgA8xmpunjU=
 github.com/ChrisTrenkamp/goxpath v0.0.0-20170625215350-4fe035839290/go.mod h1:nuWgzSkT5PnyOd+272uUmV0dnAnAn42Mk7PiQC5VzN4=
+github.com/Didainius/go-vcloud-director/v2 v2.2.0-alpha.1.0.20190417085158-6cbca97d9053 h1:xzzeqQY7gih0ZxS8h0ubOwmeosv+XsCtdbiHDlovLHc=
+github.com/Didainius/go-vcloud-director/v2 v2.2.0-alpha.1.0.20190417085158-6cbca97d9053/go.mod h1:HonlGxbjJ1NAibWh99eE4/S2l6ZOZ5KJzKK1rh2a9vc=
```
lvirbalas: I believe we shouldn't be putting personal repo references to master. Better to prepare a new tag in govcd.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -128,6 +128,12 @@ func resourceVcdVAppVm() *schema.Resource {
 				Optional: true,
 				Set:      resourceVcdVmIndependentDiskHash,
 			},
+			"expose_hardware_virtualization": &schema.Schema{
```
lvirbalas: Why not just call it `hardware_virtualization` or `nested_hypervisor`? Just looking for a clear, but shorter name here.

dataclouder: We discussed this item, and the outcome was that we need a name that says we are just setting a switch. We don't want to convey the idea that this operation does more than changing metadata in the entity.

go.mod
dataclouder: If we do that, we always need to review PRs separately. The "replace" method, which is made for this purpose, allows developers to submit the whole code for review at once.
I suggest we leave it in place

go.sum
dataclouder: This is a consequence of using the "replace" method. It will go away after we remove the replacement

go.mod
Didainius: tldr; This is done only to make TravisCI happy.

This is probably a wider discussion how we handle PRs when they depend on govcd changes. 
This setting allows to see if Travis CI is ok or not (and in future hopefully test suites) without having a PR merged into govcd.

However the drawback as you just spotted is that there's a lot of changed files (rebuilt vendor directory) and constant madness with go.mod and go.sum.

Before real merge go.mod **must** be reset to newly tagged govcd commit as I've stated it in the PR message:
> 
> **Note**. go.mod is temporary pointed to my fork to make Travis work. It will be changed before merge.

CHANGELOG.md
Didainius: Sorry. Stepped on my own proposal. Of course I will change it and fix the above comment as well.

vcd/resource_vcd_vapp_vm.go
Didainius: Yeah. I wanted to make this shorter as it is long, but otherwise probably it's not fully clear.

vbauzysvmware: I agree that purposed shorter names aren't fully clear.

lvirbalas: After a long discussion... I agree that pure `hardware_virtualization` may be intuitively mixed with virtualization in general and that we should find a keyword that instantly makes it clear what it is, but at the same time is short enough. Hence, new suggestion: `hw_assisted_virtualization`
- It's probably safe to assume that `hw` is a well known abbreviation in our field
- Impact: we can use `hw` abbreviation in other fields in the future too

lvirbalas: Another approach is to follow vCD API and call it by use case as opposed to by what it's doing:
```
NestedHypervisorEnabled bool   `xml:"nestedHypervisorEnabled,attr,omitempty"` // True if hardware-assisted CPU virtualization capabilities in the host should be exposed to the guest operating system.
```
So, in our case that could be: `nested_hypervisor`

lvirbalas: Morning is brighter than the evening... After all, I'm OK with current `expose_hardware_virtualization` implementation. It's longest field name we have, but at least it makes things clear!

vcd/resource_vcd_vapp_vm_hw_virtualization_test.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const testAccCheckVcdVAppVm_hardwareVirtualization = `
+resource "vcd_vapp" "{{.VappName}}" {
+  name = "{{.VappName}}"
+  org  = "{{.Org}}"
+  vdc  = "{{.Vdc}}"
+}
+
+resource "vcd_vapp_vm" "{{.VmName}}" {
```
dataclouder: This sample is not correctly formatted for Terraform

Didainius: Not sure I follow you. Could you be more specific?

vcd/resource_vcd_vapp_vm.go
Didainius: Agreed

vcd/resource_vcd_vapp_vm_hw_virtualization_test.go
Didainius: I've applied terraform fmt on a test snippet.

PR #224, opened 2019-04-18
Add support for network name property in vcd_snat and vcd_dnat resources

vbauzysvmware:
This functionality is derived from PR https://github.com/terraform-providers/terraform-provider-vcd/pull/61 which will be closed.

Current NAT resource implementation apply new rules to first network of edge gtw. This change add new required property `network_name` which lets you select org vDC network for which this rule will be applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 FEATURES:
 
 * `vcd_vapp_vm` - Ability to add metadata to a VM. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
-* `vcd_snat` and `vcd_dnat` - now supports organization network name. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` 
+* `vcd_snat` and `vcd_dnat` - now supports organization network name. Until major version increase it's optional, later will be required field.  
 
 BUG FIXES:
 
@@ -12,7 +12,6 @@ BUG FIXES:
 BACKWARDS INCOMPATIBILITIES / NOTES:
 
 * `vcd_vapp` - Metadata is no longer added to first VM in vApp it will be added to vApp directly instead. ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
-* `vcd_snat` and `vcd_dnat` network is no longer selected as first one in edge gateway, but now explicitly required to be defined.
 
 ## 2.1.0 (March 27, 2019)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 FEATURES:
 
 * `vcd_vapp_vm` - Ability to add metadata to a VM. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
+* `vcd_snat` and `vcd_dnat` - now supports organization network name. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` 
 
 BUG FIXES:
 
@@ -11,6 +12,7 @@ BUG FIXES:
 BACKWARDS INCOMPATIBILITIES / NOTES:
 
 * `vcd_vapp` - Metadata is no longer added to first VM in vApp it will be added to vApp directly instead. ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
+* `vcd_snat` and `vcd_dnat` network is no longer selected as first one in edge gateway, but now explicitly required to be defined.
 
 ## 2.1.0 (March 27, 2019)
 
--- a/vcd/resource_vcd_dnat.go
+++ b/vcd/resource_vcd_dnat.go
@@ -106,7 +106,7 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
 	if nil != providedNetworkName && "" != providedNetworkName {
 		orgVdcnetwork, err = getNetwork(d, vcdClient, providedNetworkName.(string))
 	} else {
-		_, _ = fmt.Fprint(GetTerraformStdout(), "This resource will require network_name in the next major version %\n")
+		_, _ = fmt.Fprint(GetTerraformStdout(), "WARNING: This resource will require network_name in the next major version \n")
 	}
 	if err != nil {
 		return fmt.Errorf("unable to find orgVdcnetwork: %s, err: %s", providedNetworkName.(string), err)
--- a/vcd/resource_vcd_dnat.go
+++ b/vcd/resource_vcd_dnat.go
@@ -106,6 +106,8 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
 	providedNetworkName := d.Get("network_name")
 	if nil != providedNetworkName && "" != providedNetworkName {
 		orgVdcnetwork, err = getNetwork(d, vcdClient, providedNetworkName.(string))
+	} else {
+		_, _ = fmt.Fprint(GetTerraformStdout(), "This resource will require network_name in the next major version %\n")
 	}
 	if err != nil {
 		return fmt.Errorf("unable to find orgVdcnetwork: %s, err: %s", providedNetworkName.(string), err)
--- a/vcd/resource_vcd_dnat.go
+++ b/vcd/resource_vcd_dnat.go
@@ -103,7 +103,7 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
 
 	var orgVdcnetwork *types.OrgVDCNetwork
 	providedNetworkName := d.Get("network_name")
-	if nil != providedNetworkName && "" != providedNetworkName {
+	if nil != providedNetworkName && "" != providedNetworkName.(string) {
 		orgVdcnetwork, err = getNetwork(d, vcdClient, providedNetworkName.(string))
 	} else {
 		_, _ = fmt.Fprint(GetTerraformStdout(), "WARNING: This resource will require network_name in the next major version \n")
@@ -174,7 +174,7 @@ func resourceVcdDNATRead(d *schema.ResourceData, meta interface{}) error {
 	var found bool
 
 	providedNetworkName := d.Get("network_name")
-	if nil != providedNetworkName && providedNetworkName != "" {
+	if nil != providedNetworkName && providedNetworkName.(string) != "" {
 		for _, r := range e.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule {
 			if r.RuleType == "DNAT" && r.GatewayNatRule.Interface.Name == d.Get("network_name") &&
 				r.GatewayNatRule.OriginalIP == d.Get("external_ip").(string) &&
--- a/vcd/resource_vcd_dnat.go
+++ b/vcd/resource_vcd_dnat.go
@@ -103,11 +103,12 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	var orgVdcnetwork *types.OrgVDCNetwork
-	if networkName := d.Get("network_name").(string); networkName != "" {
-		orgVdcnetwork, err = getNetwork(d, vcdClient, networkName)
+	providedNetworkName := d.Get("network_name")
+	if nil != providedNetworkName && "" != providedNetworkName {
+		orgVdcnetwork, err = getNetwork(d, vcdClient, providedNetworkName.(string))
 	}
-	if orgVdcnetwork == nil || orgVdcnetwork == (&types.OrgVDCNetwork{}) {
-		return fmt.Errorf("unable to find orgVdcnetwork: %s, err: %s", d.Get("network_name").(string), err)
+	if err != nil {
+		return fmt.Errorf("unable to find orgVdcnetwork: %s, err: %s", providedNetworkName.(string), err)
 	}
 
 	// Creating a loop to offer further protection from the edge gateway erroring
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 FEATURES:
 
 * `vcd_vapp_vm` - Ability to add metadata to a VM. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
+* `vcd_snat` and `vcd_dnat` - now supports organization network name. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES`
```
dataclouder: This is not "support".
We're changing the API in an incompatible way. If we apply this change, we need to increase the major version number (go to 3.x) because existing configurations will be broken.

vbauzysvmware: Well regard back compatibility - What are options? Change to optional? even though current implementation kind'a is one big bug as it works only as expected if you have one organization. What are everyone's ideas?

dataclouder: Two options:
1. Deprecate the current resource, leaving it as it is, and create a new one with the mandatory network. property.
2. Add the property as optional, and default to the first network if no property is provided

lvirbalas: Absolutely, this new option should be optional and the current behaviour kept in tact. We can't break all the configurations out there.

lvirbalas: BTW, is there any way to show deprecation warning when configuration will **not** be using this new `network_name` field? This way we could warn the users in 2.2.0 and then make the field required later.

dataclouder: Yes. We can add the field `DeprecationMessage` to the resource.

vbauzysvmware: Done. Change to optional, added deprecation message, update documentation

vcd/resource_vcd_dnat.go
```diff
@@ -120,7 +134,7 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("error completing tasks: %#v", err)
 	}
 
-	d.SetId(d.Get("external_ip").(string) + ":" + portString + " > " + d.Get("internal_ip").(string) + ":" + translatedPortString)
+	d.SetId(orgVdcnetwork.Name + ":" + d.Get("external_ip").(string) + ":" + portString + " > " + d.Get("internal_ip").(string) + ":" + translatedPortString)
```
Didainius: Not now, but we need to revise this for 3.0. NAT rules do have proper IDs.

vcd/resource_vcd_snat.go
```diff
@@ -93,7 +107,7 @@ func resourceVcdSNATRead(d *schema.ResourceData, meta interface{}) error {
 	var found bool
 
 	for _, r := range e.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule {
-		if r.RuleType == "SNAT" &&
+		if r.RuleType == "SNAT" && r.GatewayNatRule.Interface.Name == d.Get("network_name") &&
```
Didainius: But the `network_name` is optional for now. It will start skipping this for who is not using new functionality.

vcd/resource_vcd_snat_test.go
```diff
@@ -125,11 +135,27 @@ func testAccCheckVcdSNATDestroy(s *terraform.State) error {
...
+    end_address   = "{{.EndIpAddress}}"
+  }
+}
+
+
 resource "vcd_snat" "{{.SnatName}}" {
   org          = "{{.Org}}"
   vdc          = "{{.Vdc}}"
   edge_gateway = "{{.EdgeGateway}}"
+  network_name = "{{.OrgVdcNetworkName}}"
```
Didainius: Do we still have a test without `network_name` after that? We may easily break the older functionality until 3.0 if we don't.

vcd/resource_vcd_dnat.go
```diff
@@ -136,7 +150,7 @@ func resourceVcdDNATRead(d *schema.ResourceData, meta interface{}) error {
 	var found bool
 
 	for _, r := range e.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule {
-		if r.RuleType == "DNAT" &&
+		if r.RuleType == "DNAT" && r.GatewayNatRule.Interface.Name == d.Get("network_name") &&
```
Didainius: `network_name` may be ommited because it is optional.

CHANGELOG.md
```diff
@@ -11,6 +12,7 @@ BUG FIXES:
 BACKWARDS INCOMPATIBILITIES / NOTES:
 
 * `vcd_vapp` - Metadata is no longer added to first VM in vApp it will be added to vApp directly instead. ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
+* `vcd_snat` and `vcd_dnat` network is no longer selected as first one in edge gateway, but now explicitly required to be defined.
```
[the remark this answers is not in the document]
vbauzysvmware: Updated, good spot

vcd/resource_vcd_dnat.go
```diff
@@ -68,6 +73,7 @@ func resourceVcdDNAT() *schema.Resource {
 				Default:  "any",
 			},
 		},
+		DeprecationMessage: "This resource will require network_name in the next major version",
```
lvirbalas: Does this mean the user will get this message even if he specifie `network_name`? Is there a way to make it smarter, so it only shows up when it's not defined?

vbauzysvmware: It will be showed always. We are not deprecate but add property. I don't know other way.

dataclouder: This is the wrong way of using deprecation.
The field `DeprecationMessage`should be added to a resource that we leave unchanged, and we use the message to suggest using the new resource.
However, if we modify the resource without creating a new one, the deprecation message does not have sense and should be removed.

Terraform can deprecate either a resource or a property, but not the lack of a property.

vbauzysvmware: Well - this is discussion matter, there was intention to have message. If there isn't right way - so we can tell only in documentation, or have message which isn't used properly.

website/docs/r/snat.html.markdown
```diff
@@ -26,6 +27,7 @@ resource "vcd_snat" "outbound" {
 The following arguments are supported:
 
 * `edge_gateway` - (Required) The name of the edge gateway on which to apply the SNAT
+* `network_name` - (Optional *v2.2+*) The name of the organization network name on which to apply the SNAT. This will be required field in the next major version.
```
dataclouder: Redundant: 
"The name of the organization network name" should be "The name of the organization network"

website/docs/r/dnat.html.markdown
```diff
@@ -43,6 +44,7 @@ resource "vcd_dnat" "forIcmp" {
 The following arguments are supported:
 
 * `edge_gateway` - (Required) The name of the edge gateway on which to apply the DNAT
+* `network_name` - (Optional *v2.2+*) The name of the organization network name on which to apply the DNAT. This will be required field in the next major version.
```
dataclouder: Redundant: 
"The name of the organization network name" should be "The name of the organization network"

vcd/resource_vcd_dnat.go
lvirbalas: Given that Terraform doesn't provide a way to deprecate the lack of property, we should print a warning message on our own instead during apply.

vcd/resource_vcd_dnat.go
vbauzysvmware: Addressed

vcd/resource_vcd_snat.go
vbauzysvmware: Addressed

vcd/resource_vcd_snat_test.go
vbauzysvmware: Addressed

vcd/resource_vcd_dnat.go
vbauzysvmware: Addressed

vcd/resource_vcd_dnat.go
vbauzysvmware: Removed this one and added simple message when network name is not provided.

website/docs/r/dnat.html.markdown
vbauzysvmware: Fixed

vcd/resource_vcd_dnat_test.go
```diff
@@ -212,16 +226,175 @@ func testAccCheckVcdDNATDestroy(s *terraform.State) error {
...
+  network_name = "{{.OrgVdcNetworkName}}"
   edge_gateway = "{{.EdgeGateway}}"
   external_ip  = "{{.ExternalIp}}"
   port         = 7777
   internal_ip  = "10.10.102.60"
+  depends_on   = ["vcd_network_routed.{{.OrgVdcNetworkName}}"]
 }
 `
 const testAccCheckVcdDnat_tlate = `
+resource "vcd_network_routed" "{{.OrgVdcNetworkName}}" {
```
dataclouder: By Terraform style, the test snippets should be all at the end of the file, after all the functions.

website/docs/r/dnat.html.markdown
dataclouder: It's fixed on snat file but not on dnat.

website/docs/r/snat.html.markdown
vbauzysvmware: fixed

website/docs/r/dnat.html.markdown
vbauzysvmware: oh, yes. Fixed

vcd/resource_vcd_dnat_test.go
vbauzysvmware: Moved to bottom

vcd/resource_vcd_dnat.go
```diff
@@ -96,31 +101,64 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
 		icmpSubType = ""
 	}
 
+	var orgVdcnetwork *types.OrgVDCNetwork
+	providedNetworkName := d.Get("network_name")
+	if nil != providedNetworkName && "" != providedNetworkName {
+		orgVdcnetwork, err = getNetwork(d, vcdClient, providedNetworkName.(string))
+	} else {
+		_, _ = fmt.Fprint(GetTerraformStdout(), "This resource will require network_name in the next major version %\n")
```
dataclouder: There is an extra `%` in the print statement.
The statement should be the same for both snat and dnat. Please add a `WARNING:` at the beginning.

vcd/resource_vcd_snat.go
```diff
@@ -66,19 +69,53 @@ func resourceVcdSNATCreate(d *schema.ResourceData, meta interface{}) error {
...
 
-	err = retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
-		task, err := edgeGateway.AddNATMapping("SNAT", d.Get("internal_ip").(string),
-			d.Get("external_ip").(string))
+	var orgVdcnetwork *types.OrgVDCNetwork
+	providedNetworkName := d.Get("network_name")
+	if nil != providedNetworkName && "" != providedNetworkName {
+		orgVdcnetwork, err = getNetwork(d, vcdClient, providedNetworkName.(string))
+	} else {
+		_, _ = fmt.Fprint(GetTerraformStdout(), "Warning: this resource will require network_name in the next major version \n")
```
dataclouder: Please change `Warning` to `WARNING`

vcd/resource_vcd_dnat.go
vbauzysvmware: fixed

vcd/resource_vcd_snat.go
vbauzysvmware: fixed

website/docs/r/dnat.html.markdown
```diff
@@ -43,6 +44,7 @@ resource "vcd_dnat" "forIcmp" {
 The following arguments are supported:
 
 * `edge_gateway` - (Required) The name of the edge gateway on which to apply the DNAT
+* `network_name` - (Optional *v2.2+*) The name of the organization network on which to apply the DNAT. This will be required field in the next major version.
```
lvirbalas: Looks good. Only please add a semicolon after Optional and emphasis to the warning:

`network_name` - (Optional; *v2.2+*) The name of the organization network on which to apply the DNAT. *`network_name` will be required field in the next major version*

website/docs/r/snat.html.markdown
```diff
@@ -26,6 +27,7 @@ resource "vcd_snat" "outbound" {
 The following arguments are supported:
 
 * `edge_gateway` - (Required) The name of the edge gateway on which to apply the SNAT
+* `network_name` - (Optional *v2.2+*) The name of the organization network on which to apply the SNAT. This will be required field in the next major version.
```
lvirbalas: Please add a semicolon after Optional and an emphasis to the warning as in DNAT rule description request above.

website/docs/r/dnat.html.markdown
vbauzysvmware: fixed

website/docs/r/snat.html.markdown
vbauzysvmware: fixed

vcd/resource_vcd_dnat.go
```diff
@@ -96,31 +101,64 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
...
+			return resource.RetryableError(task.WaitTaskCompletion())
+		})
+	}
 
 	if err != nil {
 		return fmt.Errorf("error completing tasks: %#v", err)
 	}
 
-	d.SetId(d.Get("external_ip").(string) + ":" + portString + " > " + d.Get("internal_ip").(string) + ":" + translatedPortString)
+	if nil != providedNetworkName && "" != providedNetworkName {
```
Didainius: It hold until 3.0, but real IDs are something like:
image

Then it is easy to do lookups.

vbauzysvmware: It's common approach we have now , that resources have their own Id's - not Vcd's. I think option to add to TODO's to analyse that.

Didainius: Yeah. I don't mind being like that now. IMHO it's just for future and lookups would be way more convenient than with current ID in this case.

vbauzysvmware: I agree

PR #226, opened 2019-04-25
Add tags to all test files

dataclouder:
Test files are now all built conditionally using tags.
Running the tests without tags will trigger an error message
with a list of available tags (`cd vcd && go test -v .`).
The tags allow users to run only a subset of the tests, or all of them.
The `Makefile` uses the tags to run the same tests as before (tag
`functional`), and a few tags to run tests related to several features
(`make testcatalog`, `make testvapp`, `make testnetwork`, `make testgateway`).

The split by tags allows also running several sets of tests, such as one
using a live vCD (our current default) and one using mocks.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -24,6 +24,20 @@ testacc: fmtcheck
 testmulti: fmtcheck
 	@sh -c "'$(CURDIR)/scripts/runtest.sh' multiple"
 
+testcatalog: fmtcheck
+	@sh -c "'$(CURDIR)/scripts/runtest.sh' catalog"
+
+testvapp: fmtcheck
+	@sh -c "'$(CURDIR)/scripts/runtest.sh' vapp"
+
+testvm: fmtcheck
+	@sh -c "'$(CURDIR)/scripts/runtest.sh' vm"
+
+testgateway: fmtcheck
+	@sh -c "'$(CURDIR)/scripts/runtest.sh' gateway"
+
+testnetwork: fmtcheck
+	@sh -c "'$(CURDIR)/scripts/runtest.sh' network"
 
 vet:
 	@echo "go vet ."
@@ -54,7 +68,7 @@ test-compile:
 		echo "  make test-compile TEST=./$(PKG_NAME)"; \
 		exit 1; \
 	fi
-	go test -c $(TEST) $(TESTARGS)
+	go test -tags ALL -c $(TEST) $(TESTARGS)
 
 website:
 ifeq (,$(wildcard $(GOPATH)/src/$(WEBSITE_REPO)))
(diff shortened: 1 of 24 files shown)


dataclouder:
> Before releasing HashiCorp runs `make testacc`. However, after these changes `testacc` only runs `functional`. Though we don't seem to have a single test that is without `functional` right now, how can HashiCorp (and anyone) run all the tests with a simple `make` call?

As mentioned in the similar PR for go-vcloud-director, this choice is intentional.
When we have tests that are in a different framework, we can decide whether it is safe to run them together (meaning use two tags instead of one in the same command) or separately (meaning run two separate commands, possibly in parallel).
As we don't have different tests in this repo at the moment, we don't know if we need to run the additional tests together with the acceptance tests or separately, or not at all (meaning we only keep them for internal use).

dataclouder:
> Before releasing HashiCorp runs `make testacc`. However, after these changes `testacc` only runs `functional`. Though we don't seem to have a single test that is without `functional` right now, how can HashiCorp (and anyone) run all the tests with a simple `make` call?

As mentioned in the similar PR for go-vcloud-director, this choice is intentional.
When we have tests that are in a different framework, we can decide whether it is safe to run them together (meaning use two tags instead of one in the same command) or separately (meaning run two separate commands, possibly in parallel).
As we don't have different tests in this repo at the moment, we don't know if we need to run the additional tests together with the acceptance tests or separately, or not at all (meaning we only keep them for internal use).

lvirbalas:
OK, as long as we agree that `make test` will run all the tests, it's fine.

GNUmakefile
```diff
@@ -54,7 +68,7 @@ test-compile:
 		echo "  make test-compile TEST=./$(PKG_NAME)"; \
 		exit 1; \
 	fi
-	go test -c $(TEST) $(TESTARGS)
+	go test -tags ALL -c $(TEST) $(TESTARGS)
```
lvirbalas: Might this be a mistake? Is this running the whole test suite in `test-compile`, while it should just run compilation + unit tests?

TESTING.md
```diff
@@ -25,6 +25,59 @@ template (`catalogItem`) are defined in the configuration file, new ones will be
...
+    api_test.go:66: # No tags were defined
+    api_test.go:41:
+        # -----------------------------------------------------
+        # Tags are required to run the tests
+        # -----------------------------------------------------
+
+        At least one of the following tags should be defined:
+
+           * ALL :       Runs all the tests
+           * functional: Runs all the acceptance tests
```
lvirbalas: Empty line above would be nice to differentiate ALL from other two set types.

GNUmakefile
dataclouder: Indeed. I ported the assumption from `go-vcloud-director`.
Simplified and fixed

TESTING.md
dataclouder: done

Issue #234, opened 2019-05-07
~Feature request: Load balancer


spengilley:
Hi there,

It would be wonderful if this provider gave the ability to configure the load balancer in the Edge gateway

Didainius:
Hi @spengilley,
Not yet, but there is a draft and work should commence pretty soon. #223

Didainius:
Any comments on the design or in general are appreciated in #223  post. I will close this if you don't mind.

PR #230, opened 2019-05-02
Add external network resource

vbauzysvmware:
Fixes #173

Based on Megalord work: https://github.com/terraform-providers/terraform-provider-vcd/pull/202

Depends on https://github.com/vmware/go-vcloud-director/pull/185

This PR allows to create external network in vCD.
Note: Until govcd PR pushed and tagged, you need manually reference govcd with that PR in go.mod file

Resource creation example:
```hcl
resource "vcd_external_network" "myTerraformTest1" {
  name = "myTerraformTest1"
  description = "ExtNet-101"

  vsphere_networks {
    vcenter = "vC1"
    vsphere_network = "ForTestingPG"
    type = "DV_PORTGROUP"
  }
  ip_scope {
    is_inherited = "true"
    gateway = "192.168.30.49"
    netmask = "255.255.255.0"
    dns1 = "192.168.0.164"
    dns2 = "192.168.0.196"
    dns_suffix = "my.biz"
    ip_range {
      start = "192.168.30.51"
      end = "192.168.30.60"
    }
    ip_range {
      start = "192.168.30.31"
      end = "192.168.30.40"
    }
  }

  ip_scope {
    is_inherited = "true"
    gateway = "192.168.31.49"
    netmask = "255.255.255.240"
    dns1 = "192.168.1.164"
    dns2 = "192.168.1.196"
    dns_suffix = "my.biz"
    ip_range {
      start = "192.168.31.51"
      end = "192.168.31.60"
    }
  }

  fence_mode = "isolated"
  retain_net_info_across_deployments = "false"
  #parent_network = "${ext_net_test_parent.id}"
}
```
Things to consider: type (NETWORK and DV_PORTGROUP) may be simplified. Any ideas?

PR #236, opened 2019-05-10
Add VdC resource

vbauzysvmware:
Based on Megalord work: #149

Depends on vmware/go-vcloud-director#190

This PR allows to create VDC resource
Note: Until govcd PR pushed and tagged, you need manually reference govcd with that PR in go.mod file

Resource creation example:
```
resource "vcd_vdc" "newOne" {
  name = "newOne"
  org  = "my1"

  allocation_model = "AllocationPool"
  network_pool_name = "providerVbVdc-VXLAN-NP"
  provider_vdc_name = "providerVbVdc"

  compute_capacity {
    cpu {
      units     = "MHz"
      allocated = 2048
      limit     = 2048
      reserved  = 2048
      used      = 0
      overhead  = 0
    }

    memory {
      units     = "MB"
      allocated = 2048
      limit     = 2048
      reserved  = 2048
      used      = 0
      overhead  = 0
    }
  }

  storage_profile {
    enabled  = true
    units    = "MB"
    limit    = 10240
    default  = true
    name = "*"
  }

  storage_profile {
    enabled  = true
    units    = "MB"
    limit    = 20480
    default  = false
    name = "Development"
  }

  is_enabled             = true
  is_thin_provision      = true
  uses_fast_provisioning = true
  delete_force           = true
  delete_recursive       = true
}
```


vcd/sample_vcd_test_config.json
```diff
@@ -29,6 +29,11 @@
     "//catalog": {
       "//name": "existingCatalogName",
       "//catalogItem": "existingOvaNameInCatalog"
+    },
+    "providerVdc": {
```
lvirbalas: This isn't provider VDC, it's **organization** VDC. Better not use the name "provider" together with "VDC", because we will add provider VDC support some time in the future.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ FEATURES:
 * `vcd_vapp_vm` - Ability to add metadata to a VM. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` [GH-158]
 * `vcd_vapp_vm` - Ability to enable hardware assisted CPU virtualization for VM. It allows hypervisor nesting. [GH-219]
 * **New Resource:** external network - `vcd_external_network` - [GH-230]
+* **New Resource:** VDC resource `vcd_vdc` - [GH-234]
```
lvirbalas: I believe we should name this resource "Organization VDC resource `vcd_org_vdc`". Reasoning:
- Reflect clearly that this is **Org** VDC and _not_ **PVDC**.
- Keep the name for PVDC implementation in the future. Even though we may call it `vcd_provider_vdc`, we need to clearly separate both.
- We have "Organization VDCs" in the GUI.
- `vcd_vdc` sounds to me highest VDC level, which intuitively implies that this is PVDC, which is not.
- We already have `vcd_org` resource, so having `vcd_org_vdc` sounds as a natural extension.

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,499 @@
...
+)
+
+func resourceVcdVdc() *schema.Resource {
+	capacityWithUsage := schema.Schema{
+		Type:     schema.TypeSet,
+		Required: true,
+		ForceNew: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"units": {
```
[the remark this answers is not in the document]
vbauzysvmware: This one is asked by VCD API. Would be a additional work remove that and caculate by ourselves, also UI asks the same.

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
+---
+layout: "vcd"
+page_title: "vCloudDirector: vcd_vdc"
+sidebar_current: "docs-vcd-resource-vdc"
+description: |-
+  Provides a vCloud Director Provider VDC resource. This can be used to create and delete an VDC.
```
lvirbalas: 1. "Provider VDC"? Should be "Organization VDC".
2. "an VDC" -> "a VDC"

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+
+The following arguments are supported:
+
+* `org` - (Optional; *v2.0+*) Organization to create the VDC in, optional if defined at provider level
+* `name` - (Required) VDC name
+* `description` - (Optional) VDC friendly description
+* `allocation_model` - (Required) The allocation model used by this VDC; must be one of {AllocationVApp, AllocationPool, ReservationPool}
+* `compute_capacity` - (Required) The compute capacity allocated to this VDC.  See [Compute Capacity](#computecapacity) below for details.
+* `nic_quota` - (Optional) Maximum number of virtual NICs allowed in this VDC. Defaults to 0, which specifies an unlimited number.
+* `network_quota` - (Optional) Maximum number of network objects that can be deployed in this VDC. Defaults to 0, which means no networks can be deployed.
```
lvirbalas: Just double checking, does 0 really mean no networks as opposed to unlimited?

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+* `vm_quota` - (Optional) The maximum number of VMs that can be created in this VDC. Includes deployed and undeployed VMs in vApps and vApp templates. Defaults to 0, which specifies an unlimited number.
+* `is_enabled` - (Optional) True if this VDC is enabled for use by the organization VDCs. A VDC is always enabled on creation.
```
lvirbalas: This sentence sounds strange:
- Shouldn't "VDC is enabled for use by the organization VDCs" be just "VDC is enabled for use by the organization"?
- What does it mean "A VDC is always enabled on creation"? That it's enabled even if `is_enabled` is false? If not, then consider just saying that by default VDC is enabled.

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+* `storage_profile` - (Required) Storage profiles supported by this VDC.  See [Storage Profile](#storageprofile) below for details.
+* `resource_guaranteed_memory` - (Optional) Percentage of allocated memory resources guaranteed to vApps deployed in this VDC. For example, if this value is 0.75, then 75% of allocated resources are guaranteed. Required when AllocationModel is AllocationVApp or AllocationPool. Value defaults to 1.0 if the element is empty.
```
lvirbalas: Why are we defining percentage as float? "0.75" is quite counter-intuitive compared to just "75%" when writing and reading config.

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+* `is_thin_provision` - (Optional) Boolean to request thin provisioning. Request will be honored only if the underlying datastore supports it. Thin provisioning saves storage space by committing it on demand. This allows over-allocation of storage.
+* `network_pool_name` - (Optional) Reference to a network pool in the Provider VDC. Required if this VDC will contain routed or isolated networks.
+* `provider_vdc_name` - (Required) A name of the Provider VDC from which this organization VDC is provisioned.
+* `uses_fast_provisioning` - (Optional) Boolean to request fast provisioning. Request will be honored only if the underlying datastore supports it. Fast provisioning can reduce the time it takes to create virtual machines by using vSphere linked clones. If you disable fast provisioning, all provisioning operations will result in full clones.
```
lvirbalas: What about some boolean params to begin with "enable"? Here would be `enable_fast_provisioning`

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+* `over_commit_allowed` - (Optional) Set to false to disallow creation of the VDC if the AllocationModel is AllocationPool or ReservationPool and the ComputeCapacity you specified is greater than what the backing Provider VDC can supply. Defaults to true if empty or missing.
+* `vm_discovery_enabled` - (Optional) True if discovery of vCenter VMs is enabled for resource pools backing this VDC. If left unspecified, the actual behaviour depends on enablement at the organization level and at the system level.
```
lvirbalas: What about some boolean params to begin with "enable"? Here would be `enable_vm_discovery`.

BTW, which GUI element represents this field?

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+* `is_thin_provision` - (Optional) Boolean to request thin provisioning. Request will be honored only if the underlying datastore supports it. Thin provisioning saves storage space by committing it on demand. This allows over-allocation of storage.
+* `network_pool_name` - (Optional) Reference to a network pool in the Provider VDC. Required if this VDC will contain routed or isolated networks.
+* `provider_vdc_name` - (Required) A name of the Provider VDC from which this organization VDC is provisioned.
+* `uses_fast_provisioning` - (Optional) Boolean to request fast provisioning. Request will be honored only if the underlying datastore supports it. Fast provisioning can reduce the time it takes to create virtual machines by using vSphere linked clones. If you disable fast provisioning, all provisioning operations will result in full clones.
+* `over_commit_allowed` - (Optional) Set to false to disallow creation of the VDC if the AllocationModel is AllocationPool or ReservationPool and the ComputeCapacity you specified is greater than what the backing Provider VDC can supply. Defaults to true if empty or missing.
```
lvirbalas: Suggesting: `allow_over_commit` like in `provider.allow_unverified_ssl`

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+* `resource_guaranteed_cpu` - (Optional) Percentage of allocated CPU resources guaranteed to vApps deployed in this VDC. For example, if this value is 0.75, then 75% of allocated resources are guaranteed. Required when AllocationModel is AllocationVApp or AllocationPool. Value defaults to 1.0 if the element is empty.
+* `v_cpu_in_mhz` - (Optional) Specifies the clock frequency, in Megahertz, for any virtual CPU that is allocated to a VM. A VM with 2 vCPUs will consume twice as much of this value. Ignored for ReservationPool. Required when AllocationModel is AllocationVApp or AllocationPool, and may not be less than 256 MHz. Defaults to 1000 MHz if the element is empty or missing.
+* `is_thin_provision` - (Optional) Boolean to request thin provisioning. Request will be honored only if the underlying datastore supports it. Thin provisioning saves storage space by committing it on demand. This allows over-allocation of storage.
```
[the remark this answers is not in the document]
lvirbalas: Alternative suggestion is to use "enable_" prefix before booleans where appropriate (see below too), so it would be: `enable_thin_provisioning`

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,499 @@
...
+				ForceNew: true,
+			},
+			"allocation_model": &schema.Schema{
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringInSlice([]string{"AllocationVApp", "AllocationPool", "ReservationPool"}, false),
+				Description:  "The allocation model used by this VDC; must be one of {AllocationVApp, AllocationPool, ReservationPool}",
+			},
+			"compute_capacity": &schema.Schema{
```
dataclouder: should this be "compute**d**_capacity" ?

vcd/provider.go
```diff
@@ -95,6 +95,7 @@ func Provider() terraform.ResourceProvider {
 			"vcd_edgegateway_vpn":  resourceVcdEdgeGatewayVpn(),
 			"vcd_vapp_vm":          resourceVcdVAppVm(),
 			"vcd_org":              resourceOrg(),
+			"vcd_vdc":              resourceVcdVdc(),
```
dataclouder: Should we call it "vcd_org_vdc" to avoid future misunderstandings?

CHANGELOG.md
dataclouder: I vote for "vcd_org_vdc"

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,499 @@
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Boolean to request thin provisioning. Request will be honored only if the underlying datastore supports it. Thin provisioning saves storage space by committing it on demand. This allows over-allocation of storage.",
+			},
+			"network_pool_name": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "A name of a network pool in the Provider VDC. Required if this VDC will contain routed or isolated networks.",
```
dataclouder: "A name" -> "The name"

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,499 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "A reference to the Provider VDC from which this organization VDC is provisioned.",
+			},
+			"uses_fast_provisioning": &schema.Schema{
+				Type:        schema.TypeBool,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Boolean to request fast provisioning. Request will be honored only if the underlying datastore supports it. Fast provisioning can reduce the time it takes to create virtual machines by using vSphere linked clones. If you disable fast provisioning, all provisioning operations will result in full clones.",
```
dataclouder: "Boolean to request fast provisioning. Request will ..." -> "Request for fast provisioning. It will ..."

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,499 @@
...
+			},
+			"over_commit_allowed": &schema.Schema{
```
dataclouder: I have a problem with the semantics here:

"allowed" in the property name is a positive value
"Set false to disallow" in the description is a double negative that doesn't match the meaning of the name.
Moreover, an optional boolean property is false by default. How do we set it to false? By leaving it blank? But that is a silent specification, which we don't want to allow.
I think we need to rename the property in a way that reflects the action we can perform, such as "over_commit_disallowed" meaning that by default it is allowed, and if we set this property we will disallow it.

website/docs/r/vdc.html.markdown
dataclouder: See comments in the resource definition

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ FEATURES:
 * `vcd_vapp_vm` - Ability to add metadata to a VM. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` [GH-158]
 * `vcd_vapp_vm` - Ability to enable hardware assisted CPU virtualization for VM. It allows hypervisor nesting. [GH-219]
 * **New Resource:** external network - `vcd_external_network` - [GH-230]
```
dataclouder: Shall we have something like this?

 * **New Resources:** 
    * external network - `vcd_external_network` - [GH-230]
    * VDC `vcd_org_vdc` - [GH-234]

It will avoid conflicts in the change log, and it looks more readable

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,499 @@
...
+				ForceNew:    true,
+				Description: "Percentage of allocated CPU resources guaranteed to vApps deployed in this VDC. For example, if this value is 0.75, then 75% of allocated resources are guaranteed. Required when AllocationModel is AllocationVApp or AllocationPool. Value defaults to 1.0 if the element is empty.",
+			},
+			"v_cpu_in_mhz": &schema.Schema{
+				Type:        schema.TypeInt,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Specifies the clock frequency, in Megahertz, for any virtual CPU that is allocated to a VM. A VM with 2 vCPUs will consume twice as much of this value. Ignored for ReservationPool. Required when AllocationModel is AllocationVApp or AllocationPool, and may not be less than 256 MHz. Defaults to 1000 MHz if the element is empty or missing.",
+			},
+			"is_thin_provision": &schema.Schema{
```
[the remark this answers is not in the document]
vbauzysvmware: Changed

website/docs/r/vdc.html.markdown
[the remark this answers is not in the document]
vbauzysvmware: Changed

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+  storage_profile {
+    name     = "storage-name"
+    enabled  = true
+    units    = "MB"
+    limit    = 10240
+    default  = true    
+  }
+
+  is_enabled             = true
+  is_thin_provision      = true
```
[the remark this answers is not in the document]
vbauzysvmware: Changed

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+* `allocated` - (Optional) Capacity that is committed to be available.
+* `limit` - (Optional) Capacity limit relative to the value specified for Allocation. It must not be less than that value. If it is greater than that value, it implies overprovisioning.
+* `reserved` - (Optional) Capacity reserved
+* `used` - (Optional) Capacity used. If the VDC AllocationModel is ReservationPool, this number represents the percentage of the reservation that is in use. For all other allocation models, it represents the percentage of the allocation that is in use.
+* `overhead` - (Optional) Number of Units allocated to system resources such as vShield Manager virtual machines and shadow virtual machines provisioned from this Provider VDC.
+
+
+## Sources
+
+* [CreateVdcParamsType](https://code.vmware.com/apis/287/vcloud#/doc/doc/types/CreateVdcParamsType.html)
```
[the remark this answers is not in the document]
vbauzysvmware: It's not me point. I'll check

vcd/sample_vcd_test_config.json
dataclouder: The provider Vdc information is needed when creating an Org Vdc.
I think it should stay as it is. Perhaps we can add a comment to explain why we need it in the configuration

vcd/resource_vcd_vdc_test.go
```diff
@@ -0,0 +1,197 @@
...
+	if testConfig.VCD.ProviderVdc.StorageProfile == "" {
+		t.Skip("Variable networking.StorageProfile must be set to run VDC tests")
+		return
+	}
+
+	var vdc govcd.Vdc
+	var params = StringMap{
+		"VdcName":                   TestAccVcdVdc,
+		"OrgName":                   testConfig.VCD.Org,
+		"AllocationModel":           "ReservationPool",
```
dataclouder: Can we have a parametrized test for allocation models?

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+  delete_force             = true
+  delete_recursive         = true
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional; *v2.0+*) Organization to create the VDC in, optional if defined at provider level
```
dataclouder: The version in this definition is not needed, since the whole resource is available in 2.2+

website/docs/r/vdc.html.markdown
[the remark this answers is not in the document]
lvirbalas: Thanks! Hoping for the same approach for the fast provisioning.

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,499 @@
...
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "Name of Provider VDC storage profile.",
+						},
+						"enabled": {
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Description: "True if this storage profile is enabled for use in the VDC.",
+						},
+						"units": {
```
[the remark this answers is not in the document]
vbauzysvmware: After few trying figure out that only valuable accepted is `MB`. So `units` removed

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,505 @@
...
+				ForceNew:    true,
+				Description: "Maximum number of network objects that can be deployed in this VDC. Defaults to 0, which means no networks can be deployed.",
+			},
+			"vm_quota": &schema.Schema{
+				Type:        schema.TypeInt,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "The maximum number of VMs that can be created in this VDC. Includes deployed and undeployed VMs in vApps and vApp templates. Defaults to 0, which specifies an unlimited number.",
+			},
+			"is_enabled": &schema.Schema{
```
dataclouder: If enabled is the normal condition of a newly created VDC, we need a property to disable it.
What happens is we omit this field? It is not clear. Perhaps adding a field `disable_on_creation` would be better.
It's the same problem we have with `over_commit_allowed` below

vcd/resource_vcd_vdc.go
lvirbalas: I find the suggested "over_commit_disallowed" more confusing that what we have here now, because then you do have to play with double inverse logic. It's always more confusing when the parameter has "not" ("dis" in this case) in the name.

Description says that "Defaults to true if empty or missing". Is this actual behaviour? If it is, I suggest renaming to `allow_over_commit`, which is simpler and consistent with `allow_(thin/fast)_provisioning` and if someone wants to disallow this behaviour, they will set `allow_over_commit = "false"`.

Looking from a wider perspective, is there such parameter in the GUI (I can't find it)? What is the added value of having it and how come the GUI chose not to expose it?

Thanks!

vcd/resource_vcd_vdc.go
lvirbalas: Having negative logic in the parameter names is even more confusing. The description says that "A VDC is always enabled on creation". Is this actual behaviour? If it is, then can we just use the "Default" word before to make it really clear and consistent with other descriptions? Something like `Default is true, which ebables VDC on creation`.

> What happens is we omit this field?

Should be enabled by default.

>  we need a property to disable it

`is_enabled = "false"` should work.

But let's see what @vbauzysvmware replies.

dataclouder: My point here is that if we omit `is_enabled`, its value will be false, but the VDC will be enabled by default (i.e. the behavior when we don't set an optional property).

website/docs/r/vdc.html.markdown
vbauzysvmware: fixed

website/docs/r/vdc.html.markdown
vbauzysvmware: This message is copied from here https://code.vmware.com/apis/287/vcloud/doc/doc/types/CreateVdcParamsType.html

vbauzysvmware: In UI you can provide more than 0 when choose network pool

website/docs/r/vdc.html.markdown
vbauzysvmware: fixed

vcd/sample_vcd_test_config.json
vbauzysvmware: Yes, Giuseppe is right. This provider VDC information to create organization VDC in unit tests.

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,499 @@
...
+				ForceNew:    true,
+				Description: "Percentage of allocated memory resources guaranteed to vApps deployed in this VDC. For example, if this value is 0.75, then 75% of allocated resources are guaranteed. Required when AllocationModel is AllocationVApp or AllocationPool. Value defaults to 1.0 if the element is empty.",
+			},
+			"resource_guaranteed_cpu": &schema.Schema{
+				Type:        schema.TypeFloat,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Percentage of allocated CPU resources guaranteed to vApps deployed in this VDC. For example, if this value is 0.75, then 75% of allocated resources are guaranteed. Required when AllocationModel is AllocationVApp or AllocationPool. Value defaults to 1.0 if the element is empty.",
+			},
+			"v_cpu_in_mhz": &schema.Schema{
```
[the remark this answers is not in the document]
vbauzysvmware: changed

CHANGELOG.md
vbauzysvmware: Changed

website/docs/r/vdc.html.markdown
vbauzysvmware: Default value provided. Comments improved.

website/docs/r/vdc.html.markdown
vbauzysvmware: changed

website/docs/r/vdc.html.markdown
vbauzysvmware: Changed. Seems this property isn't in GUI

vcd/resource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+		vdc = &newVdc
+		return nil
+	}
+}
+
+func testAccCheckVdcDestroy(s *terraform.State) error {
+	conn := testAccProvider.Meta().(*VCDClient)
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "vcd_vdc" && rs.Primary.Attributes["name"] != TestAccVcdVdc {
```
dataclouder: This should be `vcd_org_vdc`

vcd/resource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,196 @@
...
+			return fmt.Errorf("vdc %s still exists or other error: %#v", rs.Primary.ID, err)
+		}
+
+	}
+
+	return nil
+}
+
+func init() {
+	testingTags["vdc"] = "resource_vcd_vdc_test.go"
```
dataclouder: It should be `resource_vcd_org_vdc_test.go`

website/docs/r/vdc.html.markdown
vbauzysvmware: Renamed

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,499 @@
+package vcd
+
+import (
+	"errors"
+	"fmt"
+	"github.com/hashicorp/terraform/helper/validation"
```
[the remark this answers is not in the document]
vbauzysvmware: Fixed

vcd/resource_vcd_vdc.go
vbauzysvmware: We agree to leave it as it is

vcd/provider.go
vbauzysvmware: Renamed.

vcd/resource_vcd_vdc.go
vbauzysvmware: fixed

vcd/resource_vcd_vdc.go
vbauzysvmware: Changed

vcd/resource_vcd_vdc.go
vbauzysvmware: Changed to allow_over_commit and default value true. Comment is improved.

website/docs/r/vdc.html.markdown
vbauzysvmware: Renamed and comments improved

CHANGELOG.md
vbauzysvmware: @lvirbalas what do you think ?

vbauzysvmware: In general, I think we copied format from Harshicorp recommended format.

vcd/resource_vcd_vdc_test.go
vbauzysvmware: Changed

vcd/resource_vcd_vdc.go
vbauzysvmware: we agreed to change to `enabled`, default value configured, comments improved.

vcd/resource_vcd_org_vdc_test.go
vbauzysvmware: done

vcd/resource_vcd_org_vdc_test.go
vbauzysvmware: Fixed

vcd/resource_vcd_vdc.go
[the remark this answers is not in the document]
vbauzysvmware: We agree to keep it for now

vbauzysvmware:
>     * It would be nice to see `metadata` setting ability as we now have it in other places (vApp and VMs). It eases lives for people when used in conjunction with other tools (like Ansible).

This one happens not with creation request, additional request needed. So additional services needed. I don't see happen in current time frame. I suggest move it to separate Issue.
 
>     * The `resourceVcdVdcRead` does not actually read real attributes to update statefile. It just checks if the object exists. We can probably live with that if it did so one would be able to catch of something was changed outside of Terraform.

This always is challenging questions. I am not sure at the moment should we reread values or not. Currently major our resources don't reread - just checks if resource exist.

vcd/config_test.go
```diff
@@ -43,8 +43,14 @@ type TestConfig struct {
 		MaxRetryTimeout          int    `json:"maxRetryTimeout"`
 	} `json:"provider"`
 	VCD struct {
-		Org     string `json:"org"`
-		Vdc     string `json:"vdc"`
+		Org             string `json:"org"`
+		Vdc             string `json:"vdc"`
+		AllocationModel string `json:"allocationModel"`
```
Didainius: IMHO we don't need to hardcode one `AllocationModel` setting in the config. Acceptance tests can just include multiple options to test as I wouldn't want to change config manually for running test for each case.

vcd/resource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+func TestAccVcdVdcBasic(t *testing.T) {
+
+	if !usingSysAdmin() {
+		t.Skip("TestAccVcdVdcBasic requires system admin privileges")
+		return
+	}
+
+	if testConfig.VCD.ProviderVdc.Name == "" {
+		t.Skip("Variable networking.Name must be set to run VDC tests")
```
dataclouder: "networking.Name" -> "providerVdc.name"

website/docs/r/vdc.html.markdown
lvirbalas: Thanks! Follow up: does changing `enabled` value force recreation of the resource?

vcd/resource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,196 @@
...
+		return
+	}
+
+	if testConfig.VCD.ProviderVdc.NetworkPool == "" {
+		t.Skip("Variable networking.NetworkPool must be set to run VDC tests")
```
dataclouder: "networking.NetworkPool" -> "providerVdc.networkPool"

website/docs/r/org_vdc.html.markdown
```diff
@@ -0,0 +1,115 @@
+---
+layout: "vcd"
+page_title: "vCloudDirector: vcd_org_vdc"
+sidebar_current: "docs-vcd-resource-org-vdc"
+description: |-
+  Provides a vCloud Director Organization VDC resource. This can be used to create and delete a Organization VDC.
+---
+
+# vcd\_vdc
```
lvirbalas: Page title left old.

vcd/resource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,196 @@
...
+		return
+	}
+
+	if testConfig.VCD.ProviderVdc.StorageProfile == "" {
+		t.Skip("Variable networking.StorageProfile must be set to run VDC tests")
```
dataclouder: "networking.StorageProfile" -> "providerVdc.storageProfile"

website/docs/r/vdc.html.markdown
lvirbalas: Thanks, please also move the `enable_fast_provisioning` section below the `enable_thin_provisioning`, so they are both one after another.

vcd/resource_vcd_org_vdc.go
```diff
@@ -0,0 +1,502 @@
...
+	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/vmware/go-vcloud-director/v2/govcd"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+)
+
+func resourceVcdOrgVdc() *schema.Resource {
+	capacityWithUsage := schema.Schema{
+		Type:     schema.TypeSet,
+		Required: true,
+		ForceNew: true,
```
Didainius: Worth adding for better validation and avoiding index out of range panics.
```
MaxItems     1,
MinItems      1
```

vcd/resource_vcd_org_vdc.go
```diff
@@ -0,0 +1,502 @@
...
+			},
+			"enabled": &schema.Schema{
+				Type:        schema.TypeBool,
+				Optional:    true,
+				ForceNew:    true,
+				Default:     true,
+				Description: "True if this VDC is enabled for use by the organization VDCs. Default is true.",
+			},
+			"storage_profile": &schema.Schema{
+				Type:     schema.TypeList,
```
Didainius: Worth adding for better validation and avoiding index out of range panics.
```
MaxItems     1,
MinItems      1
```

vcd/resource_vcd_org_vdc.go
```diff
@@ -0,0 +1,502 @@
...
+			"allocation_model": &schema.Schema{
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringInSlice([]string{"AllocationVApp", "AllocationPool", "ReservationPool"}, false),
+				Description:  "The allocation model used by this VDC; must be one of {AllocationVApp, AllocationPool, ReservationPool}",
+			},
+			"compute_capacity": &schema.Schema{
+				Required: true,
+				ForceNew: true,
```
Didainius: Worth adding for better validation and avoiding index out of range panics.
```
MaxItems     1,
MinItems      1
```

vcd/resource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,196 @@
...
+		return
+	}
+
+	var vdc govcd.Vdc
+	var params = StringMap{
+		"VdcName":                   TestAccVcdVdc,
+		"OrgName":                   testConfig.VCD.Org,
+		"AllocationModel":           testConfig.VCD.AllocationModel,
```
dataclouder: We need to check that allocation model was set in the configuration file.

website/docs/r/vdc.html.markdown
lvirbalas: Please also add default value description - from current one it's not clear.

CHANGELOG.md
lvirbalas: Yes, current implementation is good as it matches Terraform guidelines.
https://www.terraform.io/docs/extend/best-practices/versioning.html

vcd/sample_vcd_test_config.json
```diff
@@ -29,6 +29,13 @@
     "//catalog": {
       "//name": "existingCatalogName",
       "//catalogItem": "existingOvaNameInCatalog"
+    },
+    "allocationModel": "ReservationPool",
```
lvirbalas: It looks an overkill on adding allocation model to test config, which leads to one more parameter... Can we test all models automatically?

website/docs/r/org_vdc.html.markdown
```diff
@@ -0,0 +1,115 @@
...
+* `provider_vdc_name` - (Required) A name of the Provider VDC from which this organization VDC is provisioned.
+* `enable_fast_provisioning` - (Optional) Request fast provisioning. Request will be honored only if the underlying datastore supports it. Fast provisioning can reduce the time it takes to create virtual machines by using vSphere linked clones. If you disable fast provisioning, all provisioning operations will result in full clones.
+* `allow_over_commit` - (Optional) Set to false to disallow creation of the VDC if the AllocationModel is AllocationPool or ReservationPool and the ComputeCapacity you specified is greater than what the backing Provider VDC can supply. Defaults is true.
+* `enable_vm_discovery` - (Optional) True if discovery of vCenter VMs is enabled for resource pools backing this VDC. If left unspecified, the actual behaviour depends on enablement at the organization level and at the system level.
+* `delete_force` - (Required) When destroying use `delete_force=True` to remove a vdc and any objects it contains, regardless of their state.
```
lvirbalas: Please capitalise "vdc" to "VDC" across all doc page.

website/docs/r/org_vdc.html.markdown
```diff
@@ -0,0 +1,115 @@
...
+* `enable_thin_provisioning` - (Optional) Boolean to request thin provisioning. Request will be honored only if the underlying data store supports it. Thin provisioning saves storage space by committing it on demand. This allows over-allocation of storage.
+* `network_pool_name` - (Optional) Reference to a network pool in the Provider VDC. Required if this VDC will contain routed or isolated networks.
+* `provider_vdc_name` - (Required) A name of the Provider VDC from which this organization VDC is provisioned.
+* `enable_fast_provisioning` - (Optional) Request fast provisioning. Request will be honored only if the underlying datastore supports it. Fast provisioning can reduce the time it takes to create virtual machines by using vSphere linked clones. If you disable fast provisioning, all provisioning operations will result in full clones.
+* `allow_over_commit` - (Optional) Set to false to disallow creation of the VDC if the AllocationModel is AllocationPool or ReservationPool and the ComputeCapacity you specified is greater than what the backing Provider VDC can supply. Defaults is true.
```
lvirbalas: "Defaults" -> "Default"

vcd/resource_vcd_org_vdc.go
vbauzysvmware: Added. And yet we have inside check too for that

vcd/resource_vcd_org_vdc.go
vbauzysvmware: Added `MinItems` only. And yet we have inside check too for that

vcd/resource_vcd_org_vdc.go
vbauzysvmware: Added

website/docs/r/org_vdc.html.markdown
vbauzysvmware: fixed

vcd/resource_vcd_org_vdc_test.go
vbauzysvmware: fixed

vcd/resource_vcd_org_vdc.go
Didainius: Yeah, but that should allow `terraform validate`. Great!

website/docs/r/vdc.html.markdown
```diff
@@ -0,0 +1,122 @@
...
+* `delete_force` - (Required) When destroying use `delete_force=True` to remove a vdc and any objects it contains, regardless of their state.
+* `delete_recursive` - (Required) When destroying use `delete_recursive=True` to remove the vdc and any objects it contains that are in a state that normally allows removal.
+
+
+<a id="storageprofile"></a>
+## Storage Profile
+
+* `name` - (Required) Name of Provider VDC storage profile.
+* `enabled` - (Optional) True if this storage profile is enabled for use in the VDC.
+* `limit` - (Required) Maximum number of Units allocated for this storage profile. A value of 0 specifies unlimited Units.
```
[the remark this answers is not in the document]
vbauzysvmware: fixed

vcd/resource_vcd_org_vdc.go
Didainius: Great. I initially forgot multiple storage profiles possible initially, but that makes sense. Thanks!

website/docs/r/org_vdc.html.markdown
vbauzysvmware: fixed

website/docs/r/vdc.html.markdown
lvirbalas: Also, please move `provider_vdc_name` to the top below `description`.

Didainius:
> > ```
> > * It would be nice to see `metadata` setting ability as we now have it in other places (vApp and VMs). It eases lives for people when used in conjunction with other tools (like Ansible).
> > ```
> 
> This one happens not with creation request, additional request needed. So additional services needed. I don't see happen in current time frame. I suggest move it to separate Issue.
> 
> > ```
> > * The `resourceVcdVdcRead` does not actually read real attributes to update statefile. It just checks if the object exists. We can probably live with that if it did so one would be able to catch of something was changed outside of Terraform.
> > ```
> 
> This always is challenging questions. I am not sure at the moment should we reread values or not. Currently major our resources don't reread - just checks if resource exist.

Agreed with both. I do like including read everywhere (because it allows not only environment configuration, but validation as well), but again - that would consume more time and probably this wouldn't add as much value as reading NIC configuration.

website/docs/r/org_vdc.html.markdown
vbauzysvmware: done

vcd/sample_vcd_test_config.json
vbauzysvmware: allocation model removed, test improved to cover all cases

vcd/resource_vcd_org_vdc_test.go
vbauzysvmware: allocation model removed, test improved to cover all cases

vcd/config_test.go
vbauzysvmware: allocation model removed, test improved to cover all cases

website/docs/r/vdc.html.markdown
vbauzysvmware: Currently we don't support any updates. So answer - yes, resource will be recreated.

vcd/sample_vcd_test_config.json
lvirbalas: Thank you! @Didainius please review the change and close this conversation if it fits your ideas as well.

website/docs/r/vdc.html.markdown
vbauzysvmware: done

website/docs/r/vdc.html.markdown
vbauzysvmware: There isn't default. If provide true means one thing if provide false means another thing. If you don't configure means `behaviour depends on enablement at the organization level and at the system level.`

website/docs/r/vdc.html.markdown
vbauzysvmware: Cause it vCD API approach. If we different design we need agree and write converter.

website/docs/r/vdc.html.markdown
lvirbalas: This particular `enabled` parameter is quite exceptional in the perspective of resource recreation. Especially for this one the user may expect that it is a simple switch: true - resource is enabled, toggle to false - and it's disabled. In comparison, the `is_enabled` of `vcd_org` resource has `ForceNew:    false`, so it would be consistent to have this behaviour here.

vcd/sample_vcd_test_config.json
Didainius: Thanks! I like it -  we cover more and it's not that expensive in time:
```
=== RUN   TestAccVcdOrgVdcReservationPool
--- PASS: TestAccVcdOrgVdcReservationPool (18.30s)
=== RUN   TestAccVcdOrgVdcAllocationPool
--- PASS: TestAccVcdOrgVdcAllocationPool (17.10s)
=== RUN   TestAccVcdOrgVdcAllocationVApp
--- PASS: TestAccVcdOrgVdcAllocationVApp (20.04s)
PASS
ok      github.com/terraform-providers/terraform-provider-vcd/v2/vcd    57.514s
```

website/docs/r/vdc.html.markdown
vbauzysvmware: I agree and yet dead line is here.

vbauzysvmware: And similar logic applies to bunch of various parameters. Cpu, memory and so on.

website/docs/r/vdc.html.markdown
lvirbalas: GUI has percentages. It's a matter of convenience, but I may be biased here. @Didainius and @dataclouder what are your preferences?

website/docs/r/vdc.html.markdown
lvirbalas: Sorry, but still not clear. What does happen if you define it to false?

lvirbalas: Got it, please expand the description to reflect all three possible states so it's really clear for the users, something like this:
"If true, discovery of vCenter VMs is enabled for <...>. If false, discovery is disabled. If left unspecified, <...>."

website/docs/r/vdc.html.markdown
lvirbalas: OK, I filed an issue for this. We'll get back to it on a better timing.
https://github.com/terraform-providers/terraform-provider-vcd/issues/237

website/docs/r/org_vdc.html.markdown
```diff
@@ -0,0 +1,115 @@
...
+* `storage_profile` - (Required) Storage profiles supported by this VDC.  See [Storage Profile](#storageprofile) below for details.
+* `memory_guaranteed` - (Optional) Percentage of allocated memory resources guaranteed to vApps deployed in this VDC. For example, if this value is 0.75, then 75% of allocated resources are guaranteed. Required when AllocationModel is AllocationVApp or AllocationPool. Value defaults to 1.0 if the element is empty.
+* `cpu_guaranteed` - (Optional) Percentage of allocated CPU resources guaranteed to vApps deployed in this VDC. For example, if this value is 0.75, then 75% of allocated resources are guaranteed. Required when AllocationModel is AllocationVApp or AllocationPool. Value defaults to 1.0 if the element is empty.
+* `cpu_frequency` - (Optional) Specifies the clock frequency, in Megahertz, for any virtual CPU that is allocated to a VM. A VM with 2 vCPUs will consume twice as much of this value. Ignored for ReservationPool. Required when AllocationModel is AllocationVApp or AllocationPool, and may not be less than 256 MHz. Defaults to 1000 MHz if value isn't provided.
```
lvirbalas: GUI would be matched with `cpu_speed`. What would be your and @Didainius and @dataclouder vote?

dataclouder: cpu_speed sounds good

website/docs/r/vdc.html.markdown
vbauzysvmware: Changed

website/docs/r/org_vdc.html.markdown
vbauzysvmware: changed

website/docs/r/vdc.html.markdown
lvirbalas: Decided to leave it as float.

Wittionary:
> 
> 
> Thanks for all the effort @vbauzysvmware and @megalord !

Ditto! 🥇 🥇

vcd/resource_vcd_vdc.go
```diff
@@ -0,0 +1,499 @@
...
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+			"allocation_model": &schema.Schema{
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringInSlice([]string{"AllocationVApp", "AllocationPool", "ReservationPool"}, false),
```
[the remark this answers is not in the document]
vbauzysvmware: Decision was made to leave this functionality of this release, due complexity and changes it requires.

PR #240, opened 2019-05-14
Added check for existing org VDC in Org VDC resource

vbauzysvmware:
Change allows to provide meaningful message when Org VDC with such name already exist.


vcd/resource_vcd_org_vdc.go
```diff
@@ -248,6 +248,12 @@ func resourceVcdVdcCreate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf(errorRetrievingOrg, err)
 	}
 
+	orgVdcName := d.Get("name").(string)
+	orgVdc, _ := adminOrg.GetVdcByName(orgVdcName)
+	if orgVdc != (govcd.Vdc{}) {
+		return fmt.Errorf("org VDC with such name already exis: %s", orgVdcName)
```
dataclouder: "exis" -> "exists"

vbauzysvmware: fixed

PR #246, opened 2019-05-16
Remove new attributes from DNAT/SNAT

vbauzysvmware:
Due a few issues, we remove network_name and network_type until fixed.

### DIFF
--- a/vcd/resource_vcd_snat.go
+++ b/vcd/resource_vcd_snat.go
@@ -38,7 +38,7 @@ func resourceVcdSNAT() *schema.Resource {
 			"network_type": &schema.Schema{
 				Type:         schema.TypeString,
 				Optional:     true,
-				Default:      "org",
+				Default:      "ext",
 				ForceNew:     true,
 				ValidateFunc: validation.StringInSlice([]string{"ext", "org"}, false),
 			},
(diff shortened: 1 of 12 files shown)


vcd/resource_vcd_snat.go
```diff
@@ -40,7 +40,7 @@ func resourceVcdSNAT() *schema.Resource {
 				Optional:     true,
 				Default:      "org",
 				ForceNew:     true,
-				ValidateFunc: validation.StringInSlice([]string{"org"}, false),
+				ValidateFunc: validation.StringInSlice([]string{"ext"}, false),
```
lvirbalas: Let's allow both "org" and "ext". We're leaving "org" undocumented for now, but we can do some useful tests even with existing code.

lvirbalas: And what about DNAT resource?

vcd/resource_vcd_snat.go
```diff
@@ -40,7 +40,7 @@ func resourceVcdSNAT() *schema.Resource {
 				Optional:     true,
 				Default:      "org",
```
lvirbalas: Default should be "ext".

lvirbalas: And what about DNAT resource?

PR #248, opened 2019-05-17
Add check for version in docs header

dataclouder:
Add check for version in index.html.markdown
Move general purpose functions to api_test.go

The main reason for moving functions is that the version check will then run always, regardless of any tags being defined.

### DIFF
--- a/vcd/api_test.go
+++ b/vcd/api_test.go
@@ -13,8 +13,11 @@ import (
 	"testing"
 )
 
-// These variables are needed by tests running under any tags
+const (
+	testVerbose = "TEST_VERBOSE"
+)
 
+// These variables are needed by tests running under any tags
 var (
 	// Collection of defined tags in the current test run
 	testingTags = make(map[string]string)
@@ -33,7 +36,7 @@ At least one of the following tags should be defined:
 
    * ALL :       Runs all the tests
    * functional: Runs all the acceptance tests
-   * unit:       Runs unit tests that don't need a live vCD (currently unused, but we plan to)
+   * unit:       Runs unit tests that don't need a live vCD
 
    * catalog:    Runs catalog related tests (also catalog_item, media)
    * disk:       Runs disk related tests
@@ -46,9 +49,15 @@ At least one of the following tags should be defined:
 
 Examples:
 
-go test -tags functional -v -timeout=45m .
-go test -tags catalog -v -timeout=15m .
-go test -tags "org vdc" -v -timeout=5m .
+  go test -tags unit -v -timeout=45m .
+  go test -tags functional -v -timeout=45m .
+  go test -tags catalog -v -timeout=15m .
+  go test -tags "org vdc" -v -timeout=5m .
+
+Tagged tests can also run using make
+  make testunit
+  make testacc
+  make testcatalog
 `
 	t.Logf(helpText)
 }
@@ -80,7 +89,7 @@ func TestTags(t *testing.T) {
 		t.Fail()
 		return
 	}
-	if os.Getenv("SHOW_TAGS") != "" {
+	if os.Getenv(testVerbose) != "" {
 		showTags()
 	}
 }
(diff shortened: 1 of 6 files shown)


vcd/api_test.go
```diff
@@ -72,3 +84,77 @@ func TestTags(t *testing.T) {
...
+	if err != nil {
+		panic(fmt.Errorf("could not read index file %s: %v", indexFile, err))
+	}
+
+	vcdHeader := `# VMware vCloud Director Provider`
+	expectedText := vcdHeader + ` ` + currentProviderVersion
+	reExpectedVersion := regexp.MustCompile(`(?m)^` + expectedText)
+	reFoundVersion := regexp.MustCompile(`(?m)^` + vcdHeader + ` \d+\.\d+`)
+	if reExpectedVersion.MatchString(string(indexText)) {
+		t.Logf("Found expected version <%s> in index.html.markdown", currentProviderVersion)
```
Didainius: I'd say we don't need to be too verbose when expected version is found:
```
=== RUN   TestProviderVersion
--- PASS: TestProviderVersion (0.00s)
    api_test.go:147: Found expected version <2.2> in index.html.markdown
```

vcd/api_test.go
```diff
@@ -72,3 +84,77 @@ func TestTags(t *testing.T) {
...
+	}
+	// A successful match will look like
+	// [][]string{[]string{"v2.0.0", "2.0"}}
+	// Where the first element is the full text matched, and the second one is the first captured text
+	return versionList[0][1]
+}
+
+// Checks that the provider header in index.html.markdown
+// has the version defined in the VERSION file
+func TestProviderVersion(t *testing.T) {
```
Didainius: I am thinking if this test should be guarded with a separate tag (means in different file).
Currently this kind of ruins your nice `go test -v .` message. I think I wouldn't want to see this test being run as no others are run at that time.

```
$  go test -v .
=== RUN   TestTags
--- FAIL: TestTags (0.00s)
    api_test.go:78: # No tags were defined
    api_test.go:53:
        # -----------------------------------------------------
        # Tags are required to run the tests
        # -----------------------------------------------------

        At least one of the following tags should be defined:

           * ALL :       Runs all the tests
           * functional: Runs all the acceptance tests
           * unit:       Runs unit tests that don't need a live vCD (currently unused, but we plan to)

           * catalog:    Runs catalog related tests (also catalog_item, media)
           * disk:       Runs disk related tests
           * network:    Runs network related tests
           * gateway:    Runs edge gateway related tests
           * org:        Runs org related tests
           * vapp:       Runs vapp related tests
           * vdc:        Runs vdc related tests
           * vm:         Runs vm related tests

        Examples:

        go test -tags functional -v -timeout=45m .
        go test -tags catalog -v -timeout=15m .
        go test -tags "org vdc" -v -timeout=5m .
=== RUN   TestProviderVersion
--- PASS: TestProviderVersion (0.00s)
    api_test.go:147: Found expected version <2.2> in index.html.markdown
FAIL
FAIL    github.com/terraform-providers/terraform-provider-vcd/v2/vcd    0.017s
```

dataclouder: I did that on purpose. This check is so important that we want to run it regardless.

But if the team disagrees, I can either include this test in the `api` tag (it will run with `make test`) or create a new file and start using the `unit` tag, as it doesn't require a connection to run.

vcd/api_test.go
dataclouder: Normally, I would let the test to pass silently. In this case I think we should see the version at every test run, to make us aware of what we are running with.

Alternatively, I can show this message depending on an environment variable. I can't tie the message to the test verbosity, because it is mandatory for Terraform test.

vcd/api_test.go
lvirbalas: I agree with @Didainius . And, yes, `unit` tag sounds good.

vcd/api_test.go
```diff
@@ -72,3 +84,77 @@ func TestTags(t *testing.T) {
...
 }
+
+// Finds the current directory, through the path of this running test
+func getCurrentDir() string {
+	_, currentFilename, _, _ := runtime.Caller(0)
+	return filepath.Dir(currentFilename)
+}
+
+// Reads the version from the VERSION file in the root directory
+func getMajorVersion() string {
```
lvirbalas: We are duplicating code here. We already have a getMajorVersion() in config_test_.go

dataclouder: Not a duplication.
I moved the function to be visible outside tags.
When I implement the `unit` tag, this function should remain in this file to allow visibility across tags.

lvirbalas: Oops! Missed it. Sorry.

vcd/api_test.go
dataclouder: I have added an environment variable to show details.
The test is now mute by default.

vcd/api_test.go
dataclouder: Created the first unit test with `unit` tag.
Also available as `make testunit`

Running `make test` and `make testacc` will also run the unit test as a preliminary step

vcd/api_test.go
lvirbalas: Thanks, TESTING.md needs updates about the new unit group and addition for the new env variable.

dataclouder: TESTING.md updated

PR #254, opened 2019-05-28
Add Terraform 0.12 support

Didainius:
This PR is meant to introduce Terraform 0.12 support. It addresses the following:
* Add Terraform 0.12 SDK dependency
* Update Acceptance test code snippets to make Terraform 0.12 testing pass
* Update `vcd_vapp_network` resource to avoid setting guest_vlan_allowed in `terraform refresh` when the actual value is not returned
* Update version file
* Add message to changelog
* Adjust website examples to be 0.12 compatible

Note. `make testacc` and `make binary-test` (with 0.11.4 and 0.12.0 both passed).


CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
 ## 2.3.0 (Unreleased)
+
+IMPROVEMENTS:
+
+* Switch to Terraform 0.12 SDK which is required for Terraform 0.12 support. HCL parsing behaviour may have changed as a result of changes made by the new SDK version [GH-254]
```
vbauzysvmware: Would add in (...) what `HCL` means

Didainius: Added

Didainius:
> @lgtm, one doc improve request. And also needs README updated `Requirements` paragrapgh

I removed the version from there at all. It would still work with 0.11 so I guess there's no need to explicitly state version.

dataclouder:
> And also needs README updated `Requirements` paragraph

Not needed. As noted in the change log, our code works with 0.11.
We'll change that when we break compatibility.

vcd/resource_vcd_vapp_network.go
```diff
@@ -236,7 +236,12 @@ func resourceVappNetworkRead(d *schema.ResourceData, meta interface{}) error {
 			d.Set("dns2", c.IPScopes.IPScope[0].DNS2)
 			d.Set("dnsSuffix", c.IPScopes.IPScope[0].DNSSuffix)
 		}
-		d.Set("guest_vlan_allowed", &c.GuestVlanAllowed)
+		// TODO adjust when we have option to switch between API versions or upgrade the default version
```
lvirbalas: Is this TODO still actual? I'm reading the comment below and the code and it looks like it works on all versions, no?

website/docs/r/vapp_vm.html.markdown
```diff
@@ -63,11 +63,11 @@ resource "vcd_vapp_vm" "web2" {
...
-  metadata {
-    role         = "web"
-    env          = "staging"
-    version      = "v2"
-    my_extra_key = "My extra value"
+  metadata = {
+    role    = "web"
+    env     = "staging"
+    version = "v1"
+    bla     = "foo"
```
lvirbalas: Nit: "bla" may have negative associations for someone, and it's nice to have a sample with underscores. Maybe better:
`my_key = "my value"`

vcd/resource_vcd_vapp_network.go
Didainius: It does not. This was the problem I discovered. Currently we are using v27 and the response is not being returned.

In this case we're not explicitly setting it if the response was not returned - we just rely that it was set as in the configuration. I did check in UI and enabling it actually works, but retrieving with v27 doesn't.

lvirbalas: In other words, we can't retrieve it from the newer vCDs only because we're handshaking with the older v27?

website/docs/r/vapp_vm.html.markdown
Didainius: changed

vcd/resource_vcd_vapp_network.go
Didainius: Yes. I have tested this on vCD 9.5 and it was the case. If you use API v31.0 - this field is returned (and that's the version HTML5 UI is using). If you use v27 - this field is not returned.

lvirbalas: Got it, thanks!

PR #255, opened 2019-05-31
Mutex change

vbauzysvmware:
This initial change which allows us come from retrycall approach to mutex one.

Notes:
* Remove rectrycall's
* Add k,v mutex
* Add locking/unlocking helper functions
* remove locking where isn't needed - now isolated/direct networks works faster

Spotted issues:
* for a moment vm.changeName fails without rectrycall - looks like function initiated to soon, though vm creation task completed
* Test TestAccVcdVApp_PowerOff failing on my env. Seems that in test's terraform dependencies isn't working - if HCL runned with binary - everything is fine. Still looking for solution

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 ## 2.4.0 (Unreleased)
+IMPROVEMENTS:
+
+* Change resource handling to use locking mechanism when resource parallel handling is not supported by vDC. [#255] 
+
 ## 2.3.0 (May 29, 2019)
 
 IMPROVEMENTS:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 2.4.0 (Unreleased)
 IMPROVEMENTS:
 
-* Change resource handling to use locking mechanism when resource parallel handling is not supported by vDC. [#255] 
+* Change resource handling to use locking mechanism when resource parallel handling is not supported by vCD. [GH-#255] 
 
 ## 2.3.0 (May 29, 2019)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ FEATURES:
 * `vcd_vapp_vm` - Ability to add metadata to a VM. For previous behaviour please see `BACKWARDS INCOMPATIBILITIES` ([#158](https://github.com/terraform-providers/terraform-provider-vcd/issues/158))
 * `vcd_vapp_vm` - Ability to enable hardware assisted CPU virtualization for VM. It allows hypervisor nesting. ([#219](https://github.com/terraform-providers/terraform-provider-vcd/issues/219))
 * **New Resource:** external network - `vcd_external_network` - ([#230](https://github.com/terraform-providers/terraform-provider-vcd/issues/230))
-* **New Resource:** VDC resource `vcd_org_vdc` - ([#234](https://github.com/terraform-providers/terraform-provider-vcd/issues/234))
+* **New Resource:** VDC resource `vcd_org_vdc` - ([#236](https://github.com/terraform-providers/terraform-provider-vcd/issues/236))
 * resource/vcd_vapp_vm: Add `network` argument for multiple NIC support and more flexible configuration ([#233](https://github.com/terraform-providers/terraform-provider-vcd/issues/233))
 * resource/vcd_vapp_vm: Add `mac` argument to store MAC address in state file ([#233](https://github.com/terraform-providers/terraform-provider-vcd/issues/233))
 
--- a/vcd/config.go
+++ b/vcd/config.go
@@ -9,6 +9,7 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/hashicorp/terraform/helper/mutexkv"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/vmware/go-vcloud-director/v2/govcd"
 )
@@ -88,6 +89,55 @@ func debugPrintf(format string, args ...interface{}) {
 	if enableDebug {
 		fmt.Printf(format, args...)
 	}
+}
+
+// This is a global MutexKV for all resources
+var vcdMutexKV = mutexkv.NewMutexKV()
+
+func (cli *VCDClient) lockvApp(d *schema.ResourceData) {
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("name").(string))
+	vcdMutexKV.Lock(key)
+}
+
+func (cli *VCDClient) unLockvApp(d *schema.ResourceData) {
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("name").(string))
+	vcdMutexKV.Unlock(key)
+}
+
+func (cli *VCDClient) lockParentvApp(d *schema.ResourceData) {
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("vapp_name").(string))
+	vcdMutexKV.Lock(key)
+}
+
+func (cli *VCDClient) unLockParentvApp(d *schema.ResourceData) {
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("vapp_name").(string))
+	vcdMutexKV.Unlock(key)
+}
+
+func (cli *VCDClient) lockParentEdgeGtw(d *schema.ResourceData) {
+	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("edge_gateway").(string))
+	vcdMutexKV.Lock(key)
+}
+
+func (cli *VCDClient) unLockParentEdgeGtw(d *schema.ResourceData) {
+	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("edge_gateway").(string))
+	vcdMutexKV.Unlock(key)
+}
+
+func (cli *VCDClient) getOrgName(d *schema.ResourceData) string {
+	orgName := d.Get("org").(string)
+	if orgName == "" {
+		orgName = cli.Org
+	}
+	return orgName
+}
+
+func (cli *VCDClient) getVdcName(d *schema.ResourceData) string {
+	orgName := d.Get("vdc").(string)
+	if orgName == "" {
+		orgName = cli.Vdc
+	}
+	return orgName
 }
 
 // GetOrgAndVdc finds a pair of org and vdc using the names provided
--- a/vcd/config.go
+++ b/vcd/config.go
@@ -97,7 +97,7 @@ var vcdMutexKV = mutexkv.NewMutexKV()
 func (cli *VCDClient) lockVapp(d *schema.ResourceData) {
 	vappName := d.Get("name").(string)
 	if vappName == "" {
-		panic("vapp name isn't found")
+		panic("vApp name not found")
 	}
 	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), vappName)
 	vcdMutexKV.Lock(key)
@@ -106,16 +106,18 @@ func (cli *VCDClient) lockVapp(d *schema.ResourceData) {
 func (cli *VCDClient) unLockVapp(d *schema.ResourceData) {
 	vappName := d.Get("name").(string)
 	if vappName == "" {
-		panic("vapp name isn't found")
+		panic("vApp name not found")
 	}
 	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), vappName)
 	vcdMutexKV.Unlock(key)
 }
 
+// function lockParentVapp locks using vapp_name name existing in resource parameters. Parent means resource needs
+// to lock vApp it depends
 func (cli *VCDClient) lockParentVapp(d *schema.ResourceData) {
 	vappName := d.Get("vapp_name").(string)
 	if vappName == "" {
-		panic("vapp name isn't found")
+		panic("vApp name not found")
 	}
 	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), vappName)
 	vcdMutexKV.Lock(key)
@@ -124,16 +126,18 @@ func (cli *VCDClient) lockParentVapp(d *schema.ResourceData) {
 func (cli *VCDClient) unLockParentVapp(d *schema.ResourceData) {
 	vappName := d.Get("vapp_name").(string)
 	if vappName == "" {
-		panic("vapp name isn't found")
+		panic("vApp name not found")
 	}
 	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), vappName)
 	vcdMutexKV.Unlock(key)
 }
 
+// function lockParentEdgeGtw locks using edge_gateway name existing in resource parameters. Parent means resource needs
+// to lock edge gtw it depends
 func (cli *VCDClient) lockParentEdgeGtw(d *schema.ResourceData) {
 	edgeGtwName := d.Get("edge_gateway").(string)
 	if edgeGtwName == "" {
-		panic("edge gtw name isn't found")
+		panic("edge gateway not found")
 	}
 	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", cli.getOrgName(d), cli.getVdcName(d), edgeGtwName)
 	vcdMutexKV.Lock(key)
@@ -142,7 +146,7 @@ func (cli *VCDClient) lockParentEdgeGtw(d *schema.ResourceData) {
 func (cli *VCDClient) unLockParentEdgeGtw(d *schema.ResourceData) {
 	edgeGtwName := d.Get("edge_gateway").(string)
 	if edgeGtwName == "" {
-		panic("edge gtw name isn't found")
+		panic("edge gateway not found")
 	}
 	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", cli.getOrgName(d), cli.getVdcName(d), edgeGtwName)
 	vcdMutexKV.Unlock(key)
--- a/vcd/config.go
+++ b/vcd/config.go
@@ -112,8 +112,8 @@ func (cli *VCDClient) unLockVapp(d *schema.ResourceData) {
 	vcdMutexKV.Unlock(key)
 }
 
-// function lockParentVapp locks using vapp_name name existing in resource parameters. Parent means resource needs
-// to lock vApp it depends
+// function lockParentVapp locks using vapp_name name existing in resource parameters.
+// Parent means the resource belongs to the vApp being locked
 func (cli *VCDClient) lockParentVapp(d *schema.ResourceData) {
 	vappName := d.Get("vapp_name").(string)
 	if vappName == "" {
@@ -132,8 +132,8 @@ func (cli *VCDClient) unLockParentVapp(d *schema.ResourceData) {
 	vcdMutexKV.Unlock(key)
 }
 
-// function lockParentEdgeGtw locks using edge_gateway name existing in resource parameters. Parent means resource needs
-// to lock edge gtw it depends
+// function lockParentEdgeGtw locks using edge_gateway name existing in resource parameters.
+// Parent means the resource belongs to the edge gateway being locked
 func (cli *VCDClient) lockParentEdgeGtw(d *schema.ResourceData) {
 	edgeGtwName := d.Get("edge_gateway").(string)
 	if edgeGtwName == "" {
--- a/vcd/config.go
+++ b/vcd/config.go
@@ -95,32 +95,56 @@ func debugPrintf(format string, args ...interface{}) {
 var vcdMutexKV = mutexkv.NewMutexKV()
 
 func (cli *VCDClient) lockVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("name").(string))
+	vappName := d.Get("name").(string)
+	if vappName == "" {
+		panic("vapp name isn't found")
+	}
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), vappName)
 	vcdMutexKV.Lock(key)
 }
 
 func (cli *VCDClient) unLockVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("name").(string))
+	vappName := d.Get("name").(string)
+	if vappName == "" {
+		panic("vapp name isn't found")
+	}
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), vappName)
 	vcdMutexKV.Unlock(key)
 }
 
 func (cli *VCDClient) lockParentVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("vapp_name").(string))
+	vappName := d.Get("vapp_name").(string)
+	if vappName == "" {
+		panic("vapp name isn't found")
+	}
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), vappName)
 	vcdMutexKV.Lock(key)
 }
 
 func (cli *VCDClient) unLockParentVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("vapp_name").(string))
+	vappName := d.Get("vapp_name").(string)
+	if vappName == "" {
+		panic("vapp name isn't found")
+	}
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), vappName)
 	vcdMutexKV.Unlock(key)
 }
 
 func (cli *VCDClient) lockParentEdgeGtw(d *schema.ResourceData) {
-	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("edge_gateway").(string))
+	edgeGtwName := d.Get("edge_gateway").(string)
+	if edgeGtwName == "" {
+		panic("edge gtw name isn't found")
+	}
+	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", cli.getOrgName(d), cli.getVdcName(d), edgeGtwName)
 	vcdMutexKV.Lock(key)
 }
 
 func (cli *VCDClient) unLockParentEdgeGtw(d *schema.ResourceData) {
-	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("edge_gateway").(string))
+	edgeGtwName := d.Get("edge_gateway").(string)
+	if edgeGtwName == "" {
+		panic("edge gtw name isn't found")
+	}
+	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", cli.getOrgName(d), cli.getVdcName(d), edgeGtwName)
 	vcdMutexKV.Unlock(key)
 }
 
--- a/vcd/config.go
+++ b/vcd/config.go
@@ -94,22 +94,22 @@ func debugPrintf(format string, args ...interface{}) {
 // This is a global MutexKV for all resources
 var vcdMutexKV = mutexkv.NewMutexKV()
 
-func (cli *VCDClient) lockvApp(d *schema.ResourceData) {
+func (cli *VCDClient) lockVapp(d *schema.ResourceData) {
 	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("name").(string))
 	vcdMutexKV.Lock(key)
 }
 
-func (cli *VCDClient) unLockvApp(d *schema.ResourceData) {
+func (cli *VCDClient) unLockVapp(d *schema.ResourceData) {
 	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("name").(string))
 	vcdMutexKV.Unlock(key)
 }
 
-func (cli *VCDClient) lockParentvApp(d *schema.ResourceData) {
+func (cli *VCDClient) lockParentVapp(d *schema.ResourceData) {
 	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("vapp_name").(string))
 	vcdMutexKV.Lock(key)
 }
 
-func (cli *VCDClient) unLockParentvApp(d *schema.ResourceData) {
+func (cli *VCDClient) unLockParentVapp(d *schema.ResourceData) {
 	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("vapp_name").(string))
 	vcdMutexKV.Unlock(key)
 }
--- a/vcd/provider.go
+++ b/vcd/provider.go
@@ -116,37 +116,37 @@ func Provider() terraform.ResourceProvider {
 var vcdMutexKV = mutexkv.NewMutexKV()
 
 func lockVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("name").(string))
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("name").(string))
 	log.Printf("[TRACE] Locked vapp with key %s.", key)
 	vcdMutexKV.Lock(key)
 }
 
 func unLockVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("name").(string))
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("name").(string))
 	log.Printf("[TRACE] Unlocked vapp with key %s.", key)
 	vcdMutexKV.Unlock(key)
 }
 
 func lockParentVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("vapp_name").(string))
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("vapp_name").(string))
 	log.Printf("[TRACE] Locked parent vapp with key %s.", key)
 	vcdMutexKV.Lock(key)
 }
 
 func unLockParentVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("vapp_name").(string))
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("vapp_name").(string))
 	log.Printf("[TRACE] Unlocked parent vapp with key %s.", key)
 	vcdMutexKV.Unlock(key)
 }
 
 func lockParentEdgeGtw(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("edge_gateway").(string))
+	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("edge_gateway").(string))
 	log.Printf("[TRACE] Locked parent edge gtw with key %s.", key)
 	vcdMutexKV.Lock(key)
 }
 
 func unLockParentEdgeGtw(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("edge_gateway").(string))
+	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("edge_gateway").(string))
 	log.Printf("[TRACE] Unlocked parent edge gtw with key %s.", key)
 	vcdMutexKV.Unlock(key)
 }
--- a/vcd/provider.go
+++ b/vcd/provider.go
@@ -116,37 +116,37 @@ func Provider() terraform.ResourceProvider {
 var vcdMutexKV = mutexkv.NewMutexKV()
 
 func lockVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("vdc").(string), d.Get("org").(string), d.Get("name").(string))
+	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("name").(string))
 	log.Printf("[TRACE] Locked vapp with key %s.", key)
 	vcdMutexKV.Lock(key)
 }
 
 func unLockVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("vdc").(string), d.Get("org").(string), d.Get("name").(string))
+	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("name").(string))
 	log.Printf("[TRACE] Unlocked vapp with key %s.", key)
 	vcdMutexKV.Unlock(key)
 }
 
 func lockParentVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("vdc").(string), d.Get("org").(string), d.Get("vapp_name").(string))
+	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("vapp_name").(string))
 	log.Printf("[TRACE] Locked parent vapp with key %s.", key)
 	vcdMutexKV.Lock(key)
 }
 
 func unLockParentVapp(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("vdc").(string), d.Get("org").(string), d.Get("vapp_name").(string))
+	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("vapp_name").(string))
 	log.Printf("[TRACE] Unlocked parent vapp with key %s.", key)
 	vcdMutexKV.Unlock(key)
 }
 
 func lockParentEdgeGtw(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("vdc").(string), d.Get("org").(string), d.Get("edge_gateway").(string))
+	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("edge_gateway").(string))
 	log.Printf("[TRACE] Locked parent edge gtw with key %s.", key)
 	vcdMutexKV.Lock(key)
 }
 
 func unLockParentEdgeGtw(d *schema.ResourceData) {
-	key := fmt.Sprintf("%s|%s|%s", d.Get("vdc").(string), d.Get("org").(string), d.Get("edge_gateway").(string))
+	key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("edge_gateway").(string))
 	log.Printf("[TRACE] Unlocked parent edge gtw with key %s.", key)
 	vcdMutexKV.Lock(key)
 }
--- a/vcd/provider.go
+++ b/vcd/provider.go
@@ -2,11 +2,12 @@ package vcd
 
 import (
 	"fmt"
+	"log"
+
 	"github.com/hashicorp/terraform/helper/mutexkv"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/vmware/go-vcloud-director/v2/util"
-	"log"
 )
 
 // Provider returns a terraform.ResourceProvider.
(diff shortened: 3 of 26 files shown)


vcd/provider.go
```diff
@@ -108,6 +109,33 @@ func Provider() terraform.ResourceProvider {
 	}
 }
 
+// This is a global MutexKV for all resources
+var vcdMutexKV = mutexkv.NewMutexKV()
+
+func lockVapp(d *schema.ResourceData) {
+	vcdMutexKV.Lock(d.Get("vdc").(string) + d.Get("name").(string))
```
Didainius: I think we should add TRACE/DEBUG level logging to each of these function with as much details as we can access (maybe a caller function or a resource `name` which are mandatory everywhere anyway?

This should be valuable if we hit any bad situations

vcd/resource_vcd_vapp_vm.go
```diff
@@ -563,63 +557,58 @@ func resourceVcdVAppVmUpdate(d *schema.ResourceData, meta interface{}) error {
 		}
```
Didainius: Can't comment on the `resourceVcdVAppVmUpdate` signature, but it needs a lock. I was able to get such error: `Error: error Powering Off: &errors.errorString{s:"error powering off VM: API Error: 400: [ 4001d52e-3028-4474-921e-215f8a201348 ] The entity web1 is busy completing an operation VAPP_POWER_OFF. VAPP_POWER_OFF(com.vmware.vcloud.entity.task:1a66a737-99bc-4a0c-89e7-572e4f107069)"}`

vcd/resource_vcd_vapp_vm.go
```diff
@@ -563,63 +557,58 @@ func resourceVcdVAppVmUpdate(d *schema.ResourceData, meta interface{}) error {
...
-				task, err := vm.PowerOn()
-				if err != nil {
-					return resource.RetryableError(fmt.Errorf("error Powering Up: %#v", err))
-				}
 
-				return resource.RetryableError(task.WaitTaskCompletion())
-			})
+			task, err := vm.PowerOn()
+			if err != nil {
+				fmt.Errorf("error Powering Up: %#v", err)
```
Didainius:
```suggestion
				return fmt.Errorf("error Powering Up: %#v", err)
```

vcd/provider.go
dataclouder: There should be a separator character between names to minimize the risk of collision. For example, VDC "pant" and entity "ear" generates the same lock name as "pan" + "tear". If we create the names as "pant#ear" and "pan#tear", we avoid this situation. While this case is unlikely to occur in production, it is possible during testing, when users come up with not-well-thought short names.

vcd/resource_vcd_dnat.go
```diff
@@ -128,36 +128,30 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
...
+			d.Get("internal_ip").(string),
+			translatedPortString, protocol,
+			icmpSubType)
+		if err != nil {
+			return fmt.Errorf("error setting DNAT rules: %#v", err)
+		}
 
-			return resource.RetryableError(task.WaitTaskCompletion())
-		})
+		err = task.WaitTaskCompletion()
```
dataclouder: Missing error handling.

vcd/resource_vcd_snat.go
```diff
@@ -91,25 +90,20 @@ func resourceVcdSNATCreate(d *schema.ResourceData, meta interface{}) error {
...
-				return resource.RetryableError(fmt.Errorf("error setting SNAT rules: %#v", err))
-			}
-			return resource.RetryableError(task.WaitTaskCompletion())
-		})
+		task, err := edgeGateway.AddNATRule(orgVdcnetwork, "SNAT",
+			d.Get("external_ip").(string), d.Get("internal_ip").(string))
+		if err != nil {
+			return fmt.Errorf("error setting SNAT rules: %#v", err)
+		}
+		err = task.WaitTaskCompletion()
```
dataclouder: missing error handling

vcd/resource_vcd_vapp.go
```diff
@@ -184,59 +181,52 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
...
-					task, err := vapp.RunCustomizationScript(d.Get("name").(string), initscript.(string))
-					if err != nil {
-						return resource.RetryableError(fmt.Errorf("error with setting init script: %#v", err))
-					}
-					return resource.RetryableError(task.WaitTaskCompletion())
-				})
+				log.Printf("running customisation script")
+				task, err := vapp.RunCustomizationScript(d.Get("name").(string), initscript.(string))
+				if err != nil {
+					fmt.Errorf("error with setting init script: %#v", err)
```
dataclouder: There should be a `return` statement here

vcd/resource_vcd_vapp.go
```diff
@@ -356,28 +339,25 @@ func resourceVcdVAppUpdate(d *schema.ResourceData, meta interface{}) error {
...
 
 		if d.HasChange("cpus") {
-			err = retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
-				task, err := vapp.ChangeCPUCount(d.Get("cpus").(int))
-				if err != nil {
-					return resource.RetryableError(fmt.Errorf("error changing cpu count: %#v", err))
-				}
+			task, err := vapp.ChangeCPUCount(d.Get("cpus").(int))
+			if err != nil {
+				fmt.Errorf("error changing cpu count: %#v", err)
```
dataclouder: There should be a `return` statement here

vcd/resource_vcd_vapp.go
```diff
@@ -454,35 +432,35 @@ func resourceVcdVAppRead(d *schema.ResourceData, meta interface{}) error {
...
-		// getting the IP of the specific Vm, rather than index zero.
-		// Required as once we add more VM's, index zero doesn't guarantee the
-		// 'first' one, and tests will fail sometimes (annoying huh?)
-		vm, err := vdc.FindVMByName(vapp, d.Get("name").(string))
-		if err != nil {
-			return resource.RetryableError(fmt.Errorf("unable to find vm: %s", err))
-		}
+	vapp, err := vdc.FindVAppByName(d.Id())
+	if err != nil {
+		fmt.Errorf("unable to find vapp")
```
dataclouder: There should be a `return` statement here

vcd/provider.go
lvirbalas: Good catch! I'd like to see the separator for another reason - when printing an error/debug message it will be a lot easier to read when there's a separator. I just hope that no-one is using the hash sign in the names, like "Obj \#1", "Obj \#2"... Maybe let's use comma as in "pan,tear" or pipe "pan|tear"?

lvirbalas: +1

dataclouder: > I just hope that no-one is using the hash sign in the names

We are. In function `(c *Config) Client()` (`config.go`)

vcd/provider.go
```diff
@@ -108,6 +109,33 @@ func Provider() terraform.ResourceProvider {
 	}
 }
 
+// This is a global MutexKV for all resources
+var vcdMutexKV = mutexkv.NewMutexKV()
+
+func lockVapp(d *schema.ResourceData) {
```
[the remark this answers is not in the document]
lvirbalas: Good question, indeed! Also, how do we handle different locking levels with a single global mutex map? There are the very least three of them:
1. vApp operations - on Org name (?), VDC name and vApp name
2. Edge Gateway operations - on Org name and EG name
3. Org VDC creation/destroy/update - on Org name alone? Or is it possible to create Org VDCs in parallel and no locking is needed at all?

vcd/provider.go
lvirbalas: That's a bit different though (used for connection cache pointers). What I meant is that you can name your VDC like "Test VDC #1" or vApp "My vApp #2" and the hash separator would result in a crash.

Screen Shot 2019-06-03 at 10 38 11

dataclouder: I misunderstood the remark. Using commas (or dashes, colons, or pipe signs)  is fine

vcd/provider.go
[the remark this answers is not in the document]
vbauzysvmware: 1. yes, locking on specific VAPP
2. yes we could for Edge gtw operations, but not creation, and yet no other operation at the moment. VDC networks create/delete locks on parent edge gtw.
3. Org VDC - dunno, haven't test or anyone complain about that.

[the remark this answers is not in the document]
vbauzysvmware: P.s. to be on safe side, will add org name to `key`

vcd/resource_vcd_network_routed.go
```diff
@@ -279,14 +283,11 @@ func resourceVcdNetworkDelete(d *schema.ResourceData, meta interface{}) error {
...
-		task, err := network.Delete()
-		if err != nil {
-			return resource.RetryableError(
-				fmt.Errorf("error Deleting Network: %#v", err))
-		}
-		return resource.RetryableError(task.WaitTaskCompletion())
-	})
+	task, err := network.Delete()
+	if err != nil {
+		return fmt.Errorf("error DDeleting Network: %#v", err)
```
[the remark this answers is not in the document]
vbauzysvmware: Fixed

vcd/resource_vcd_snat.go
```diff
@@ -63,8 +62,8 @@ func resourceVcdSNATCreate(d *schema.ResourceData, meta interface{}) error {
 	// Multiple VCD components need to run operations on the Edge Gateway, as
 	// the edge gatway will throw back an error if it is already performing an
 	// operation we must wait until we can aquire a lock on the client
-	vcdClient.Mutex.Lock()
-	defer vcdClient.Mutex.Unlock()
+	lockParentEdgeGtw(d)
+	defer unLockParentEdgeGtw(d)
 
 	// Creating a loop to offer further protection from the edge gateway erroring
```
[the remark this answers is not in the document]
vbauzysvmware: Removed

vcd/resource_vcd_vapp.go
```diff
@@ -493,32 +471,38 @@ func resourceVcdVAppDelete(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("error finding vapp: %s", err)
 	}
 
-	_ = retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
-		task, err := vapp.Undeploy()
-		// Very often the vApp is powered off at this point and Undeploy() would wait for as long as
```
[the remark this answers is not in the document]
vbauzysvmware: fixed

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp_vm.go
```diff
@@ -866,14 +852,12 @@ func resourceVcdVAppVmDelete(d *schema.ResourceData, meta interface{}) error {
...
+		if err != nil {
+			return fmt.Errorf("error Powering Up vApp: %#v", err)
+		}
 
-			return resource.RetryableError(task.WaitTaskCompletion())
-		})
+		err = task.WaitTaskCompletion()
 		if err != nil {
 			return fmt.Errorf("error Powering Up vApp: %#v", err)
 		}
```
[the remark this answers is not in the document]
vbauzysvmware: Fixed

vcd/resource_vcd_dnat.go
vbauzysvmware: It was, moved to be more obvious

vcd/resource_vcd_snat.go
vbauzysvmware: It was, moved to be more obvious

vcd/resource_vcd_vapp.go
vbauzysvmware: fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: fixed

vcd/provider.go
vbauzysvmware: Will use PIPE

vbauzysvmware: Changed

vbauzysvmware: Added

vcd/provider.go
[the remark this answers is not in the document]
vbauzysvmware: > Please also add an antry in CHANGELOG.md for this feature.

Added

vcd/resource_vcd_network_direct.go
```diff
@@ -93,9 +90,7 @@ func resourceVcdNetworkDirectCreate(d *schema.ResourceData, meta interface{}) er
 		IsShared: d.Get("shared").(bool),
 	}
 
-	err = retryCall(vcdClient.MaxRetryTimeout, func() *resource.RetryError {
-		return resource.RetryableError(vdc.CreateOrgVDCNetworkWait(orgVDCNetwork))
-	})
+	err = vdc.CreateOrgVDCNetworkWait(orgVDCNetwork)
```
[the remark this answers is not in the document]
vbauzysvmware: As I mentioned in comment - isolated and direct can be created in parallel.

vcd/provider.go
```diff
@@ -1,9 +1,12 @@
 package vcd
 
 import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/mutexkv"
```
dataclouder: Imports aren't properly sorted

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: ok. I cannot add directly cause it will create deadlock - will create wrapper.

vcd/provider.go
```diff
@@ -108,6 +111,45 @@ func Provider() terraform.ResourceProvider {
 	}
 }
 
+// This is a global MutexKV for all resources
+var vcdMutexKV = mutexkv.NewMutexKV()
```
dataclouder: Is this the best place to put the mutex functions?
I didn't notice before that they are in Provider. Should we leave `provider.go` clean and add the mutex functions in `config.go`?
@Didainius , @lvirbalas , please chime in.

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: Added

vcd/provider.go
vbauzysvmware: Fixed

vcd/provider.go
vbauzysvmware: AWS was holding in provider, so I also choose there. Though could agree on config too.

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 ## 2.4.0 (Unreleased)
+IMPROVEMENTS:
+
+* Change resource handling to use locking mechanism when resource parallel handling is not supported by vDC. [#255]
```
lvirbalas: - "vDC" -> "vCD"
- `[#255]` should be in "GH-" format for the build system to parse

vcd/provider.go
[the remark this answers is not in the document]
lvirbalas: Thanks. One nuance: I suggest having "org" first, "vdc" second and object name last in accordance with the hierarchy in vCD. It would be more intuitive.

vcd/resource_vcd_network_routed.go
```diff
@@ -263,10 +260,17 @@ func resourceVcdNetworkRead(d *schema.ResourceData, meta interface{}) error {
...
+
+	return resourceVcdNetworkDelete(d, meta)
+}
+
 func resourceVcdNetworkDelete(d *schema.ResourceData, meta interface{}) error {
 	vcdClient := meta.(*VCDClient)
-	vcdClient.Mutex.Lock()
-	defer vcdClient.Mutex.Unlock()
+
+	time.Sleep(10 * time.Second)
```
lvirbalas: Why this ten second sleep here and even at the very beginning of the function??

vbauzysvmware: Leftover from my investigation, removed

CHANGELOG.md

vcd/provider.go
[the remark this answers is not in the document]
vbauzysvmware: Changed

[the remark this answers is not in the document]
vbauzysvmware: regarding `3` point - I made check and it is possible to create org VDC in parallel.

vcd/resource_vcd_vapp_vm.go
Didainius: The same test I had failing works fine now 👍

vcd/provider.go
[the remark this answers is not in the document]
Didainius: Regarding the logging. Need others to share their thoughts, but while debugging such messages can be confusing. Here's why:


`[TRACE] Locked vapp with key my-org|my-vdc||my-vapp`
You get this message, but actually the lock may still not be acquired, just stuck at waiting for it to be acquired. IMHO this can be very confusing when debugging logs.

I would propose something like.
```
key := fmt.Sprintf("%s|%s|%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("name").(string))
log.Printf("[TRACE] Acquiring lock for vapp with key %s.", key)
vcdMutexKV.Lock(key)
log.Printf("[TRACE] Locked vapp with key %s.", key)
```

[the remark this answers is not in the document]
lvirbalas: One more thing, we discussed with Dainius that there's a risk to collide if vApp name and EG name are the same. This is realistic - for example a name "Test". Hence, we're suggesting to add type to the string:
`org:myOrg|vdc:myVdc|edge:myTest`
`org:myOrg|vdc:myVdc|vapp:myTest`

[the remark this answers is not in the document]
vbauzysvmware: Changed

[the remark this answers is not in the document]
lvirbalas: Nit: please capitalise "vApp" in the logs as well. It's just so much easier to scan the text when it's everywhere a "vApp" instead of "vapp".

vcd/provider.go
```diff
@@ -108,6 +112,48 @@ func Provider() terraform.ResourceProvider {
...
+func lockParentEdgeGtw(d *schema.ResourceData) {
+	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("edge_gateway").(string))
+	log.Printf("[TRACE] Acquiring lock for edge gtw with key %s.", key)
+	vcdMutexKV.Lock(key)
+	log.Printf("[TRACE] Locked parent edge gtw with key %s.", key)
+}
+
+func unLockParentEdgeGtw(d *schema.ResourceData) {
+	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", d.Get("org").(string), d.Get("vdc").(string), d.Get("edge_gateway").(string))
+	log.Printf("[TRACE] Unlocked parent edge gtw with key %s.", key)
```
lvirbalas: Nit: please use the full word "gateway", so it's easier to find things in the log.

vcd/resource_vcd_edgegateway_vpn.go
```diff
@@ -206,22 +205,18 @@ func resourceVcdEdgeGatewayVpnCreate(d *schema.ResourceData, meta interface{}) e
...
-		task, err := edgeGateway.AddIpsecVPN(ipsecVPNConfig)
-		if err != nil {
-			log.Printf("[INFO] Error setting ipsecVPNConfig rules: %s", err)
-			return resource.RetryableError(
-				fmt.Errorf("error setting ipsecVPNConfig rules: %#v", err))
-		}
+	err = edgeGateway.Refresh()
+	if err != nil {
+		log.Printf("[INFO] Error refreshing edge gateway: %#v", err)
+		return fmt.Errorf("error error refreshing edge gateway: %#v", err)
```
lvirbalas: Typo: "error error"

vcd/resource_vcd_edgegateway_vpn.go
```diff
@@ -253,22 +248,18 @@ func resourceVcdEdgeGatewayVpnDelete(d *schema.ResourceData, meta interface{}) e
...
```
lvirbalas: Typo: "error error"

vcd/resource_vcd_firewall_rules.go
```diff
@@ -106,23 +105,19 @@ func resourceVcdFirewallRulesCreate(d *schema.ResourceData, meta interface{}) er
...
-		task, err := edgeGateway.CreateFirewallRules(d.Get("default_action").(string), firewallRules)
-		if err != nil {
-			log.Printf("[INFO] Error setting firewall rules: %s", err)
-			return resource.RetryableError(
-				fmt.Errorf("error setting firewall rules: %#v", err))
-		}
+	err = edgeGateway.Refresh()
+	if err != nil {
+		log.Printf("[INFO] Error refreshing edge gateway: %#v", err)
+		return fmt.Errorf("error error refreshing edge gateway: %#v", err)
```
lvirbalas: Typo.

vcd/provider.go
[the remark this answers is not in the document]
Didainius: I was too early to complain about log messages... Seems like the `mutexkv` library logs itself. Here's the output. You will see the `DEBUG` level messages are coming from library itself while `TRACE` ones are coming from @vbauzysvmware  code:
https://github.com/hashicorp/terraform/blob/master/helper/mutexkv/mutexkv.go#L21
```
2019-06-04T11:08:14.283+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:14 [TRACE] Acquiring lock for vapp with key org:my-org|vdc:my-org-vdc|vapp:TestAccVcdVAppVapp-other.
2019-06-04T11:08:14.283+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:14 [DEBUG] Locking "org:my-org|vdc:my-org-vdc|vapp:TestAccVcdVAppVapp-other"
2019-06-04T11:08:14.283+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:14 [DEBUG] Locked "org:my-org|vdc:my-org-vdc|vapp:TestAccVcdVAppVapp-other"
2019-06-04T11:08:14.283+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:14 [TRACE] Locked vapp with key org:my-org|vdc:my-org-vdc|vapp:TestAccVcdVAppVapp-other.
2019-06-04T11:08:17.811+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:17 [TRACE] Unlocked vapp with key org:my-org|vdc:my-org-vdc|vapp:TestAccVcdVAppVapp-other.
2019-06-04T11:08:17.811+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:17 [DEBUG] Unlocking "org:my-org|vdc:my-org-vdc|vapp:TestAccVcdVAppVapp-other"
2019-06-04T11:08:17.811+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:17 [DEBUG] Unlocked "org:my-org|vdc:my-org-vdc|vapp:TestAccVcdVAppVapp-other"
2019-06-04T11:08:17.824+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:17 [TRACE] Acquiring lock for edge gtw with key org:my-org|vdc:my-org-vdc|edge:my-edge-gw.
2019-06-04T11:08:17.824+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:17 [DEBUG] Locking "org:my-org|vdc:my-org-vdc|edge:my-edge-gw"
2019-06-04T11:08:17.824+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:17 [DEBUG] Locked "org:my-org|vdc:my-org-vdc|edge:my-edge-gw"
2019-06-04T11:08:17.824+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:17 [TRACE] Locked parent edge gtw with key org:my-org|vdc:my-org-vdc|edge:my-edge-gw.
2019-06-04T11:08:18.317+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:18 [TRACE] Unlocked parent edge gtw with key org:my-org|vdc:my-org-vdc|edge:my-edge-gw.
2019-06-04T11:08:18.317+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:18 [DEBUG] Unlocking "org:my-org|vdc:my-org-vdc|edge:my-edge-gw"
2019-06-04T11:08:18.317+0300 [DEBUG] plugin.terraform-provider-vcd_v2.3.0: 2019/06/04 11:08:18 [DEBUG] Unlocked "org:my-org|vdc:my-org-vdc|edge:my-edge-gw"
```

Probably logging from two places is overkill. What do others think?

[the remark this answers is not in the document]
lvirbalas: Ouch! Yes, now it's overkill.

BTW, it looks clear what's being locked with the new string format with types!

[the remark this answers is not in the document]
vbauzysvmware: Removed logs

vcd/resource_vcd_edgegateway_vpn.go
vbauzysvmware: Changed

vcd/resource_vcd_edgegateway_vpn.go
vbauzysvmware: Changed

vcd/resource_vcd_firewall_rules.go
vbauzysvmware: Changed

vcd/provider.go
vbauzysvmware: logs removed

vbauzysvmware:
I added change which solved powerOff test failing

vcd/resource_vcd_vapp.go
```diff
@@ -493,32 +473,50 @@ func resourceVcdVAppDelete(d *schema.ResourceData, meta interface{}) error {
...
+	task, err = vapp.Delete()
+	if err != nil {
+		return fmt.Errorf("error deleting: %#v", err)
+	}
 
-		return resource.RetryableError(task.WaitTaskCompletion())
-	})
+	err = task.WaitTaskCompletion()
+	if err != nil {
+		return fmt.Errorf("error with deleting VAPP task: %#v", err)
```
lvirbalas: Nit, but please use "vApp" instead of "VAPP" in the messages and comments, there are more occurrences like these in this PR.

vcd/resource_vcd_vapp.go
```diff
@@ -493,32 +473,50 @@ func resourceVcdVAppDelete(d *schema.ResourceData, meta interface{}) error {
...
+	return nil
+}
 
-		return resource.RetryableError(task.WaitTaskCompletion())
-	})
+func tryUndeploy(vapp govcd.VApp) error {
+	task, err := vapp.Undeploy()
+	// Very often the vApp is powered off at this point and Undeploy() would fail with error:
+	// "The requested operation could not be executed since vApp vApp_name is not running"
+	// So, if the error matches we just ingore and fast forward to vapp.Delete()
```
lvirbalas: After refactoring this code block to a separate function, this comment is assuming that this function "knows" how it's being called and what's being done after it. Instead, please add a comment above the function definition, which describes the interfacing and how it works. For example:

// **Try to undeploy a vApp, but do not throw an error if the vApp is powered off.**
// Very often the vApp is powered off at this point and Undeploy() would fail with error:
// "The requested operation could not be executed since vApp vApp_name is not running"
// So, if the error matches we just **ignore it** and **the caller may** fast forward to vapp.Delete()

vbauzysvmware: fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: fixed

vcd/provider.go
lvirbalas: Both places look fine.

Didainius:
PASS
ok      github.com/terraform-providers/terraform-provider-vcd/v2/vcd    1996.660s

vcd/provider.go
vbauzysvmware: At the end I moved to config, cause function type changed and more functions were added.

vcd/config.go
```diff
@@ -90,6 +91,55 @@ func debugPrintf(format string, args ...interface{}) {
...
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("name").(string))
+	vcdMutexKV.Lock(key)
+}
+
+func (cli *VCDClient) unLockvApp(d *schema.ResourceData) {
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), d.Get("name").(string))
+	vcdMutexKV.Unlock(key)
+}
+
+func (cli *VCDClient) lockParentvApp(d *schema.ResourceData) {
```
lvirbalas: What is the reason of having "Parent" in these functions? Why not just `lockVApp`, `lockEdgeGateway` and so on?

BTW, we use "VApp" in function names (like FindVAppByName), while "vApp" in comments and messages.

vbauzysvmware: Linas, if you look to content of functions you will see difference what resource names are used to construct key.

Regarding function names - reverting back, seems you forgot that you asked to change to current ones.

lvirbalas: I did not forget, I asked to use "vApp" "in comments and messages" where they are used as stand-alone words, in function names it's a different story.

lvirbalas: Thanks for addressing feedback and adding checks to the functions that the values would exist in the maps.

vcd/resource_vcd_network_routed.go
```diff
@@ -263,10 +258,16 @@ func resourceVcdNetworkRead(d *schema.ResourceData, meta interface{}) error {
 	return nil
 }
 
+func resourceVcdNetworkDeleteLocked(d *schema.ResourceData, meta interface{}) error {
+	vcdClient := meta.(*VCDClient)
+	vcdClient.lockParentEdgeGtw(d)
+	defer vcdClient.unLockParentEdgeGtw(d)
+
+	return resourceVcdNetworkDelete(d, meta)
```
lvirbalas: Just to be on a safe side, I double checked that the order of defer execution is as we are expecting even with a function in return statement, and it is :) Small illustration:

```
package main

import "fmt"

func resourceVcdNetworkDeleteLocked() error {
    fmt.Println("lock")
	defer fmt.Println("unlock")

    return resourceVcdNetworkDelete()
}

func resourceVcdNetworkDelete() error {
	fmt.Println("delete")

	return nil
}

func main() {
	resourceVcdNetworkDeleteLocked()
}
```

Output:
```
lock
delete
unlock
```

vendor/modules.txt
```diff
@@ -215,7 +216,7 @@ github.com/ulikunitz/xz/internal/hash
 # github.com/vmihailenco/msgpack v4.0.1+incompatible
 github.com/vmihailenco/msgpack
 github.com/vmihailenco/msgpack/codes
-# github.com/vmware/go-vcloud-director/v2 v2.2.0
+# github.com/vmware/go-vcloud-director/v2 v2.3.0-alpha.1
```
lvirbalas: The tag is incorrect as we're now working on v2.4.0. I know it's easy to miss, given that we released v2.3.0 so quickly :)

vcd/resource_vcd_network_routed.go
vbauzysvmware: Yes, I confirmed that too :) before using

vendor/modules.txt
vbauzysvmware: govcd tags a little diff story now: https://github.com/vmware/go-vcloud-director/tags

vbauzysvmware:
> resource_vcd_vapp_vm.g

Comment updated to moved to better place

vcd/config.go
```diff
@@ -90,6 +91,79 @@ func debugPrintf(format string, args ...interface{}) {
 	}
 }
 
+// This is a global MutexKV for all resources
+var vcdMutexKV = mutexkv.NewMutexKV()
+
+func (cli *VCDClient) lockVapp(d *schema.ResourceData) {
+	vappName := d.Get("name").(string)
+	if vappName == "" {
+		panic("vapp name isn't found")
```
dataclouder: Please change all occurrences also in the other functions below
"vapp name isn't found" -> "vApp name not found"

vcd/config.go
```diff
@@ -90,6 +91,79 @@ func debugPrintf(format string, args ...interface{}) {
...
+func (cli *VCDClient) unLockVapp(d *schema.ResourceData) {
+	vappName := d.Get("name").(string)
+	if vappName == "" {
+		panic("vapp name isn't found")
+	}
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), vappName)
+	vcdMutexKV.Unlock(key)
+}
+
+func (cli *VCDClient) lockParentVapp(d *schema.ResourceData) {
```
dataclouder: Please add a comment to explain what parent we are talking about

vcd/config.go
```diff
@@ -90,6 +91,79 @@ func debugPrintf(format string, args ...interface{}) {
...
+		panic("vapp name isn't found")
+	}
+	key := fmt.Sprintf("org:%s|vdc:%s|vapp:%s", cli.getOrgName(d), cli.getVdcName(d), vappName)
+	vcdMutexKV.Unlock(key)
+}
+
+func (cli *VCDClient) lockParentEdgeGtw(d *schema.ResourceData) {
+	edgeGtwName := d.Get("edge_gateway").(string)
+	if edgeGtwName == "" {
+		panic("edge gtw name isn't found")
```
dataclouder: "edge gtw name isn't found" -> "edge gateway not found"

vcd/config.go
```diff
@@ -90,6 +91,79 @@ func debugPrintf(format string, args ...interface{}) {
...
+	}
+	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", cli.getOrgName(d), cli.getVdcName(d), edgeGtwName)
+	vcdMutexKV.Lock(key)
+}
+
+func (cli *VCDClient) unLockParentEdgeGtw(d *schema.ResourceData) {
```
dataclouder: Please add a comment about which parent we are dealing with

vcd/config.go
```diff
@@ -90,6 +91,79 @@ func debugPrintf(format string, args ...interface{}) {
...
+	edgeGtwName := d.Get("edge_gateway").(string)
+	if edgeGtwName == "" {
+		panic("edge gtw name isn't found")
+	}
+	key := fmt.Sprintf("org:%s|vdc:%s|edge:%s", cli.getOrgName(d), cli.getVdcName(d), edgeGtwName)
+	vcdMutexKV.Unlock(key)
+}
+
+func (cli *VCDClient) getOrgName(d *schema.ResourceData) string {
```
dataclouder: What's the purpose of this function?
If the resource doesn't have the Org, we get it from the client? In which cases?
What if the client doesn't have it either?

vcd/config.go
```diff
@@ -90,6 +91,79 @@ func debugPrintf(format string, args ...interface{}) {
...
+	orgName := d.Get("org").(string)
+	if orgName == "" {
+		orgName = cli.Org
+	}
+	return orgName
+}
+
+func (cli *VCDClient) getVdcName(d *schema.ResourceData) string {
```
dataclouder: Same as what mentioned for `getOrgName`

vcd/resource_vcd_network_routed.go
```diff
@@ -279,14 +280,11 @@ func resourceVcdNetworkDelete(d *schema.ResourceData, meta interface{}) error {
...
-		task, err := network.Delete()
-		if err != nil {
-			return resource.RetryableError(
-				fmt.Errorf("error Deleting Network: %#v", err))
-		}
-		return resource.RetryableError(task.WaitTaskCompletion())
-	})
+	task, err := network.Delete()
+	if err != nil {
+		return fmt.Errorf("error Deleting Network: %#v", err)
```
dataclouder: ""error Deleting Network" -> ""error deleting network"

vcd/resource_vcd_vapp.go
```diff
@@ -151,18 +152,13 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
...
-
-					if err != nil {
-						return resource.RetryableError(fmt.Errorf("error creating vapp: %#v", err))
-					}
-
-					return resource.RetryableError(task.WaitTaskCompletion())
-				})
+				task, err := vdc.ComposeVApp(nets, vapptemplate, storage_profile_reference, d.Get("name").(string), d.Get("description").(string), d.Get("accept_all_eulas").(bool))
+				if err != nil {
+					return fmt.Errorf("error creating vapp: %#v", err)
```
dataclouder: "vapp" -> "vApp"

vcd/resource_vcd_vapp.go
```diff
@@ -172,94 +168,84 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
...
-				return resource.RetryableError(task.WaitTaskCompletion())
-			})
+			networks := []map[string]interface{}{map[string]interface{}{
+				"ip":         d.Get("ip").(string),
+				"is_primary": true,
+				"orgnetwork": d.Get("network_name").(string),
+			}}
+			task, err = vapp.ChangeNetworkConfig(networks, d.Get("ip").(string))
+			if err != nil {
+				return fmt.Errorf("error with Networking change: %#v", err)
```
dataclouder: "Networking" -> "networking"

vcd/resource_vcd_vapp.go
```diff
@@ -172,94 +168,84 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
...
-
-					if err != nil {
-						return resource.RetryableError(fmt.Errorf("error set ovf: %#v", err))
-					}
-					return resource.RetryableError(task.WaitTaskCompletion())
-				})
+				task, err := vapp.SetOvf(convertToStringMap(ovf.(map[string]interface{})))
+
+				if err != nil {
+					return fmt.Errorf("error set ovf: %#v", err)
```
dataclouder: "error set ovf" -> ""error setting the ovf"

vcd/resource_vcd_vapp.go
```diff
@@ -172,94 +168,84 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
...
-					if err != nil {
-						return resource.RetryableError(fmt.Errorf("error with setting init script: %#v", err))
-					}
-					return resource.RetryableError(task.WaitTaskCompletion())
-				})
+				log.Printf("running customisation script")
+				task, err := vapp.RunCustomizationScript(d.Get("name").(string), initscript.(string))
+				if err != nil {
+					return fmt.Errorf("error with setting init script: %#v", err)
+				}
```
dataclouder: ""error with setting init script" -> "error with init script setting"

vcd/resource_vcd_vapp.go
```diff
@@ -172,94 +168,84 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
...
-					task, err := vapp.PowerOn()
-					if err != nil {
-						return resource.RetryableError(fmt.Errorf("error powerOn machine: %#v", err))
-					}
-					return resource.RetryableError(task.WaitTaskCompletion())
-				})
+
+				task, err := vapp.PowerOn()
+				if err != nil {
+					return fmt.Errorf("error powerOn machine: %#v", err)
```
dataclouder: "error powerOn machine" -> ""error power on the machine"

vcd/resource_vcd_vapp.go
```diff
@@ -172,94 +168,84 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
...
-			e = vdc.Refresh()
-			if e != nil {
-				return resource.RetryableError(fmt.Errorf("Error: %#v", e))
-			}
-			return nil
-		})
-		if err != nil {
-			return err
+		if e != nil {
+			return fmt.Errorf("Error: %#v", e)
```
dataclouder: "Error" -> "error"

vcd/resource_vcd_vapp.go
```diff
@@ -172,94 +168,84 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
...
+		}
+
+		e = vdc.Refresh()
+		if e != nil {
+			return fmt.Errorf("Error: %#v", e)
```
dataclouder: "Error" -> "error"

vcd/resource_vcd_vapp.go
```diff
@@ -395,14 +376,12 @@ func resourceVcdVAppUpdate(d *schema.ResourceData, meta interface{}) error {
...
-				task, err := vapp.SetOvf(convertToStringMap(ovf.(map[string]interface{})))
+			task, err := vapp.SetOvf(convertToStringMap(ovf.(map[string]interface{})))
 
-				if err != nil {
-					return resource.RetryableError(fmt.Errorf("Error set ovf: %#v", err))
-				}
-				return resource.RetryableError(task.WaitTaskCompletion())
-			})
+			if err != nil {
+				return fmt.Errorf("error set ovf: %#v", err)
```
dataclouder: "error set ovf" -> "error setting the ovf"

vcd/resource_vcd_vapp.go
```diff
@@ -454,35 +433,35 @@ func resourceVcdVAppRead(d *schema.ResourceData, meta interface{}) error {
...
-		// getting the IP of the specific Vm, rather than index zero.
-		// Required as once we add more VM's, index zero doesn't guarantee the
-		// 'first' one, and tests will fail sometimes (annoying huh?)
-		vm, err := vdc.FindVMByName(vapp, d.Get("name").(string))
-		if err != nil {
-			return resource.RetryableError(fmt.Errorf("unable to find vm: %s", err))
-		}
+	vapp, err := vdc.FindVAppByName(d.Id())
+	if err != nil {
+		return "", fmt.Errorf("unable to find vapp")
```
dataclouder: "vapp" -> "vApp"

vcd/resource_vcd_vapp.go
```diff
@@ -493,32 +472,51 @@ func resourceVcdVAppDelete(d *schema.ResourceData, meta interface{}) error {
...
-		// "The requested operation could not be executed since vApp vApp_name is not running"
-		// So, if the error matches we just quit retrying and fast forward to vapp.Delete()
-		var reErr = regexp.MustCompile(`.*The requested operation could not be executed since vApp.*is not running.*`)
-		if err != nil && reErr.MatchString(err.Error()) {
-			return &resource.RetryError{Err: err, Retryable: false}
-		}
+	// to avoid network destroy issues - detach networks from vApp
+	task, err := vapp.RemoveAllNetworks()
+	if err != nil {
+		return fmt.Errorf("error with Networking change: %#v", err)
```
dataclouder: "Networking" -> "networking"

vcd/resource_vcd_vapp.go
```diff
@@ -172,94 +168,84 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
...
+			err = vapp.BlockWhileStatus("UNRESOLVED", vcdClient.MaxRetryTimeout)
+			if err != nil {
+				return fmt.Errorf("error composing vApp: %s", err)
+			}
 
-				return resource.RetryableError(task.WaitTaskCompletion())
-			})
+			task, err := vapp.ChangeVMName(d.Get("name").(string))
+			if err != nil {
+				return fmt.Errorf("error with vm name change: %#v", err)
```
dataclouder: "vm" -> "VM"

vcd/resource_vcd_vapp.go
```diff
@@ -454,35 +433,35 @@ func resourceVcdVAppRead(d *schema.ResourceData, meta interface{}) error {
...
-			return resource.RetryableError(fmt.Errorf("timeout: VM did not acquire IP address"))
-		}
-		return nil
-	})
+	// getting the IP of the specific Vm, rather than index zero.
+	// Required as once we add more VM's, index zero doesn't guarantee the
+	// 'first' one, and tests will fail sometimes (annoying huh?)
+	vm, err := vdc.FindVMByName(vapp, d.Get("name").(string))
+	if err != nil {
+		return "", fmt.Errorf("unable to find vm: %s", err)
```
dataclouder: "vm" -> "VM"

vcd/config.go
vbauzysvmware: This is the same we have everywhere. Or provider has to have org or resource explicitly state it.
"Inheritance" is that provider org is added to client and that's how we access when resource don't have it. If provider and resource won't define it, then we have errors everywhere. Need to test which part will fail first. It would depend on provider wrong configuration

vcd/config.go
vbauzysvmware: Replyed

vcd/config.go
vbauzysvmware: Fixed

vcd/config.go
vbauzysvmware: Added

vcd/config.go
vbauzysvmware: Fixed

vcd/config.go
vbauzysvmware: Added

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/resource_vcd_network_routed.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/resource_vcd_vapp.go
vbauzysvmware: Fixed

vcd/config.go
dataclouder: Please change:
"Parent means resource needs to lock vApp it depends" -> "Parent means  the resource belong to the vApp being locked"

vcd/config.go
dataclouder: "Parent means resource needs to lock edge gtw it depends" -> "Parent means  the resource belongs to the edge gateway being locked"

vcd/resource_vcd_network_routed.go
dataclouder: Still not fixed. `Network` -> `network`

vbauzysvmware: Fixed

vcd/config.go
vbauzysvmware: Changed

vcd/config.go
vbauzysvmware: Changed

vendor/modules.txt
lvirbalas: All right, let's officially decouple the versions of these two projects and let them live their own lives (versions). And, hopefully, this will lead to many new users of go-vcloud-director! :)

Issue #223, opened 2019-04-18
Introduce edge gateway load balancers


Didainius:
## Why
We have a need (#182) to introduce edge gateway load balancing capabilities.
### Details
This is a master issue for work required to introduce Edge gateway load balancer (LB) support. Because this is a complex task and multiple resources will have to be created there will be child issues for each of the planned resources.

## Prerequisites
* Edge gateway up (with Load balancer enabled in `Edge gateway services -> Load balancer -> Global configuration`). At the current state it is manual. Once we have ability to create it this would be a feature toggle attribute.
* Existing vApp and VMs  (to be used as load balancer pool members)

## Considerations
* Have to check if it supports API 27.0 on all vCDs (does it depend on NSX edge gateway version?)

## Proposal
The plan is to mimic Edge gateway UI in vCD. It has 6 tabs for configuration (one of them being just `enable/disable` which should happen on edge gateway resource once we have it). The other five are listed below. To make a working solution without manual intervention (provided edge gateway is created already) we need at least 4 of them.
* Application profiles
* Service monitoring
* Pools
* Application rules (Not a mandatory feature to have working LB)
* Virtual servers

These objects reference each other to finally make a `virtual server` which is the main object for load balancer.

Each of them is described well in the vCD [Load Balancing documentation](https://docs.vmware.com/en/vCloud-Director/9.7/com.vmware.vcloud.spportal.doc/GUID-2E6696DB-5F03-411C-A7EA-358E50EBB992.html).

## Sample HCL configuration
```hcl
# Load balancer -> Service Monitoring (in Edge gateway services)
resource "vcd_lb_service_monitor" "service-monitor" {    
    org = "my-org"     # Optional
    vdc = "my-org-vdc" # Optional

    # Dependencies
    edge_gateway = ""

    name = ""
    interval = ""
    timeout = ""
    max_retries = ""
    type = ""               # Better name could be ?protocol? or "protocol_type". HTTP, HTTPS, TCP, ICMP, or UDP
    expected_response = ""  # if type="HTTP" || type="HTTPS" 
    http_method = ""        # if type="HTTP" || type="HTTPS" GET, POST, PUT, etc....
    url = ""                # if type="HTTP" || type="HTTPS"
    send = ""               # if type="HTTP" || type="HTTPS" || type="UDP"
    receive = ""            # if send != "" .Response from send. String to be matched in the response content.
[...]

PR #256, opened 2019-06-03
Add load balancer service monitor resource and datasource

Didainius:
Ref #223 

~~**Note** Because load balancers depend on edge gateway having none other operations this PR will adopt locking functionality introduced in  #255 before merge once it is in master.~~

This is the first PR to start supporting edge gateway load balancers.

It includes resource and datasource `vcd_lb_service_monitor`. All actions should be possible including `import`.


vcd/resource_vcd_lb_service_monitor_test.go
```diff
@@ -0,0 +1,147 @@
+// +build gateway lb lbsm ALL functional
```
lvirbalas: I thought I posted this, but looks like the comment got lost between the tabs. "lb" abbreviation is OK, but "lbsm" is too much for my taste :) Maybe let's expand into full "service monitor" wording?
CC: @dataclouder

vcd/provider.go
```diff
@@ -102,6 +102,12 @@ func Provider() terraform.ResourceProvider {
 			"vcd_inserted_media":   resourceVcdInsertedMedia(),
 			"vcd_independent_disk": resourceVcdIndependentDisk(),
 			"vcd_external_network": resourceVcdExternalNetwork(),
+			// "vcd_lb_pool":            resourceVcdLbPool(),
+			"vcd_lb_service_monitor": ResourceVcdLbServiceMonitor(),
```
Didainius: Note for myself.
```suggestion
			"vcd_lb_service_monitor": resourceVcdLbServiceMonitor(),
```

vcd/resource_vcd_lb_service_monitor_test.go
vbauzysvmware: Agree, let's avoid to much acronyms

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,339 @@
...
+		return []*schema.ResourceData{}, fmt.Errorf("unable to find load balancer service monitor with ID %s: %s", d.Id(), err)
+	}
+
+	d.Set("org", orgName)
+	d.Set("vdc", vdcName)
+	d.Set("edge_gateway", edgeName)
+	d.Set("name", monitorName)
+
+	d.SetId(readLBMonitor.ID)
+	return []*schema.ResourceData{d}, nil
```
vbauzysvmware: We add only this values cause, user will have to add other manually?

website/docs/d/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+~> **Note:** The vCloud Director API for NSX supports a subset of the operations and objects defined in the NSX vSphere 
+API Guide. The API supports NSX 6.2, 6.3, and 6.4.
+
+Supported in provider *v2.4+*
+
+## Example Usage
+
+```hcl
+data "vcd_lb_service_monitor" "sm-ds" {
```
vbauzysvmware:
```suggestion
data "vcd_lb_service_monitor" "my-monitor" {
```

website/docs/d/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+The following arguments are supported:
+
+* `edge_gateway` - (Required) The name of the edge gateway on which the service monitor is defined
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `name` - (Required) Service Monitor name for identifying the exact service monitor
+
+## Attribute Reference
+
+All the attributes defined in `vcd_lb_service_monitor` resource are be available.
```
vbauzysvmware: Shouldn't we state explicitly that state file will be filled only with 4 values and others have to be added manually and verified?

website/docs/d/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,47 @@
+---
+layout: "vcd"
+page_title: "vCloudDirector: vcd_lb_service_monitor"
```
vbauzysvmware: maybe add `data`

website/docs/d/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+sidebar_current: "docs-vcd-data-source-lb-service-monitor"
+description: |-
+  Provides a NSX load balancer service monitor data source.
+---
+
+# vcd\_lb\_service\_monitor
```
vbauzysvmware: maybe add `data`

website/docs/r/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+
+```hcl
+provider "vcd" {
+  user     = "${var.admin_user}"
+  password = "${var.admin_password}"
+  org      = "System"
+  url      = "https://AcmeVcd/api"
+}
+
+resource "vcd_lb_service_monitor" "monitor" {
```
vbauzysvmware: `terraform fmt` seems needed :)

website/docs/r/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+  org                 = "my-org"
+  vdc                 = "my-org-vdc"
+  edge_gateway        = "my-edge-gw"
+
+  name                = "testing-monitor-yy"
+  interval = "5"
+  timeout = "20"
+  max_retries = "3"
+  type = "http"
```
vbauzysvmware: can we have monitor without url/api ?

website/docs/r/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+
+## Importing
+
+An existing load balancer service monitor can be [imported][docs-import] into this resource
+via supplying the full dot separated path for load balancer service monitor. An example is below:
+
+[docs-import]: /docs/import/index.html
+
+```
+terraform import vcd_lb_service_monitor.imported my-org.my-org-vdc.my-edge-gw.existing-sm
```
vbauzysvmware: `existing-sm` - should be improved and agreed as in code commented.

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,339 @@
...
+		Receive:    d.Get("receive").(string),
+		Extension:  expandLBMonitorExtension(d),
+	}
+
+	return lbMonitor, nil
+}
+
+// expandLBMonitorExtension expands the specified map for sending via API. It appends newline to every extension as
+// per API requirement. Based on the research the underlying structure should not cause problems because duplicate keys
+// are not needed and order of the keys does not matter for API.
```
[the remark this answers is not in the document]
lvirbalas: For a fresh reader.. :) The purpose is a bit unclear - maybe add a sentence _why_ is this needed?

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,339 @@
...
+	vcdClient := meta.(*VCDClient)
+	vcdClient.lockParentEdgeGtw(d)
+	defer vcdClient.unLockParentEdgeGtw(d)
+
+	edgeGateway, err := vcdClient.GetEdgeGatewayFromResource(d)
+	if err != nil {
+		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	}
+
+	lbMonitor, err := expandLBMonitor(d)
```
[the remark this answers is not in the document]
lvirbalas: For me it's a bit unclear as well. If it's formatting the structure to be API edible, maybe the function name could contain "API" too? `getLBMonitorAPIStruct`? Just thinking out loud.

Then again, first question I get when I see this is "why is this needed?". Would be nice to have a comment on that.

vcd/resource_vcd_lb_service_monitor_test.go
Didainius: It's the same as in govcd now. `lbServiceMonitor`

vcd/datasource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,90 @@
...
+				Description: "Edge gateway name",
+			},
+			"name": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Service Monitor name",
+			},
+			"interval": &schema.Schema{
+				Type:     schema.TypeInt,
+				Computed: true,
```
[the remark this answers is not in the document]
Didainius: There is no point in having them for datasource as you would never see them.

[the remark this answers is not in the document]
Didainius: Yes because datasource is meant to have set only the mandatory things for lookup org, vdc, edgegw and name. Everything else will then be returned for you.

vcd/datasource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,90 @@
...
+			},
+			"extension": {
+				Type:     schema.TypeMap,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func datasourceVcdLbServiceMonitorRead(d *schema.ResourceData, meta interface{}) error {
```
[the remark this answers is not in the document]
Didainius: It's behaving about IDs a bit differently.

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,339 @@
...
+			},
+			"expected": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "String that the monitor expects to match in the status line of the HTTP or HTTPS response (for example, HTTP/1.1)",
+			},
+			"method": &schema.Schema{
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "Method to be used to detect server status",
```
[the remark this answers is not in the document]
Didainius: I was thinking of that but then I kind of refrained as the documentation states one list while the real API validator gives another one. Instead I made the `govcd` part so that errors are parsed and returned in nice format from NSX API which usually proposes possible options as well.

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
Didainius: I could change. I actually used naming  inspiration from https://www.terraform.io/docs/extend/writing-custom-providers.html.
The point is that other load balancer parts will have very similar way and ones have a "flatter" struct, others have some nesting. I tried to apply some consistency for them.

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,339 @@
...
+	updatedLBMonitor, err := edgeGateway.UpdateLBServiceMonitor(updateLBMonitorConfig)
+	if err != nil {
+		return fmt.Errorf("unable to update load balancer service monitor with ID %s: %s", d.Id(), err)
+	}
+
+	if err := flattenLBMonitor(d, updatedLBMonitor); err != nil {
+		return err
+	}
+
+	return nil
```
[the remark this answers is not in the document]
Didainius: Good spot. Done.

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,339 @@
...
+	if err != nil {
+		return fmt.Errorf("error deleting load balancer service monitor: %s", err)
+	}
+
+	d.SetId("")
+	return nil
+}
+
+// resourceVcdLbServiceMonitorImport expects dot formatted path to Load Balancer Service Monitor
+// i.e. my-org.my-org-vdc.my-edge-gw.existing-sm
```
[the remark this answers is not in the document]
Didainius: Done

vcd/resource_vcd_lb_service_monitor.go
Didainius: No. The point of import is that you lookup the object ID and it calls `terraform refresh` implicitly which populates all the other data using the `read` method as far as I've seen.

website/docs/d/lb_service_monitor.html.markdown
Didainius: Done

website/docs/d/lb_service_monitor.html.markdown
Didainius: No. The point of having a datasource is that allows to access all the fields without being created via Terraform. You only specify the fields needed to lookup object.

website/docs/d/lb_service_monitor.html.markdown
Didainius: I think it is not needed as the `data` is more of terraform keyword like `resource` and website menu clearly separates these objects. Open to discussion.

website/docs/d/lb_service_monitor.html.markdown
Didainius: I think it is not needed as the `data` is more of terraform keyword like `resource` and website menu clearly separates these objects.

website/docs/r/lb_service_monitor.html.markdown
Didainius: True, done.

website/docs/r/lb_service_monitor.html.markdown
Didainius: Done

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
Didainius: I mentioned that this is `API requirement` in description. Is it getting any better after I attached XML example?

website/docs/r/lb_service_monitor.html.markdown
Didainius: Yes. This is allowed and works. I think it simply queries `/` URL then.

vcd/datasource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
vbauzysvmware: Yes, it's only one line difference and yet, at the moment didn't find easy solution to avoid that. Only one read function and id set and flatten in wrapper functions.

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
vbauzysvmware: I still recommend to write down possible values today - cause otherwise I have to find another documentation or guess and read errors when I run. I think current approach don't validation values though in terraform.
cc: @lvirbalas

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
vbauzysvmware: I also see expand/flatten in other provider code, yet I am not so keen to use that naming convention.

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
vbauzysvmware: cc: @lvirbalas @dataclouder

vcd/resource_vcd_lb_service_monitor.go
vbauzysvmware: Why not import automatically without refresh? - cause in terraform documentation I don't `refresh` mentioned as process.

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
vbauzysvmware: Yes

vcd/resource_vcd_lb_service_monitor.go
Didainius: I could of course set them here as well however what I saw while investigating other providers is they're setting only defaults (which are not populated by read/refresh. However I aim to read all fields via import.

As far as the refresh behaviour -  you can't change that behaviour. That's how import looks like (You will see it is refreshing):
```bash
$ terraform import vcd_lb_service_monitor.import my-org.my-org-vdc.my-edge-gw.service-monitor-1
vcd_lb_service_monitor.import: Importing from ID "my-org.my-org-vdc.my-edge-gw.service-monitor-1"...
vcd_lb_service_monitor.import: Import complete!
  Imported vcd_lb_service_monitor
vcd_lb_service_monitor.import: Refreshing state... [id=monitor-140]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```


After this operation the statefile has such fields:
```json
{
      "mode": "managed",
      "type": "vcd_lb_service_monitor",
      "name": "import",
      "provider": "provider.vcd",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "edge_gateway": "my-edge-gw",
            "expected": "HTTP/1.1",
            "extension": {
              "content-type": "application/json",
              "no-body": ""
            },
            "id": "monitor-140",
            "interval": 5,
            "max_retries": 3,
            "method": "POST",
            "name": "service-monitor-1",
            "org": "my-org",
            "receive": "OK",
            "send": "{\"key\": \"value\"}",
            "timeout": 5,
            "type": "http",
            "url": "",
            "vdc": "my-org-vdc"
          }
        }
      ]
    },
```

website/docs/r/lb_service_monitor.html.markdown
vbauzysvmware: doesn't example should be more understandable?:) I suggest add URL.

website/docs/d/lb_service_monitor.html.markdown
vbauzysvmware: cc: @lvirbalas

website/docs/d/lb_service_monitor.html.markdown
vbauzysvmware: cc: @lvirbalas

vcd/resource_vcd_lb_service_monitor.go
vbauzysvmware: ok. Tx for info

website/docs/r/lb_service_monitor.html.markdown
Didainius: Agreed. Done

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
Didainius: Ok. I've added at least some, but thinking to skip validation. If you miss the value - NSX API error is quite nicely formatted:
```
vcd_lb_service_monitor.lb-service-monitor[0]: Modifying... [id=monitor-297]

Error: unable to update load balancer service monitor with ID monitor-297: error while updating load balancer service monitor : vShield Edge [LoadBalancer] Invalid monitor type asd. Valid
monitor types are: HTTP|HTTPS|TCP. (API error: 14557)
```

**P.S.** You will probably notice that this list contradicts to list in documentation which I used - https://docs.vmware.com/en/vCloud-Director/9.0/com.vmware.vcloud.tenantportal.doc/GUID-DE3EDF95-1A8E-4D2F-B7B1-1254E9420868.html

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,339 @@
...
+		if strings.ToUpper(v) != v {
+			es = append(es, fmt.Errorf(
+				"expected string to be upper cased, got: %s", v))
+		}
+		return
+	}
+}
+
+// validateUpperCase checks that a string is upper cased
+func validateLowerCase() schema.SchemaValidateFunc {
```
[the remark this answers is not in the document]
Didainius: Done.

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
vbauzysvmware: Thank you. Anyways I think @lvirbalas thinks to avoid specific validation for these cases even they longer to catch. As API could change in future, thouth we still could no support that until make API version more flexible.

website/docs/r/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+  vdc          = "my-org-vdc"
+  edge_gateway = "my-edge-gw"
+
+  name        = "testing-monitor-yy"
+  interval    = "5"
+  timeout     = "20"
+  max_retries = "3"
+  type        = "http"
+  http_method = "GET"
+  url         = "health"
```
vbauzysvmware: more meaningful? e.g. http://192.168.10.10/health - Just suggesting :)

Didainius: The healthcheck should not imply hostname because load balancer server pool will specify list of servers and they will be checked. I can probably add `/health` for clarity.

lvirbalas: Yes, that would be nice :)

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
lvirbalas: To me NSX error here is very clear. However, I can't match *our* documentation to what it writes. We have:

> method - (Optional) For types http and https. Select http method to be used to detect server status

Two issues here:
- No "TCP"
- Values are lower case, while NSX error shows upper case. As such as a user, I'm confused which to use.

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
lvirbalas: Sorry, is the question what's nicer to read? I'm a bit confused, as Dainius wrote "Done", but I still see old code here:
https://github.com/Didainius/terraform-provider-vcd/blob/lb1/vcd/resource_vcd_lb_service_monitor.go

I would vote for something in the middle (shorter to the last suggestion and more similar to the first params):
`// i.e. my-org.my-org-vdc.my-edge-gw.my-lb-service-monitor`

website/docs/d/lb_service_monitor.html.markdown
lvirbalas: Sorry, don't get where you're suggesting to put `data`. Looking at the latest doc version it looks very nice, but I'd like to ask one addition to example (I'll add the ask to the corresponding line).

website/docs/d/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+## Example Usage
+
+```hcl
+data "vcd_lb_service_monitor" "my-monitor" {
+  org                 = "my-org"
+  vdc                 = "my-org-vdc"
+  edge_gateway        = "my-edge-gw"
+
+  name = "not-managed"
+}
```
lvirbalas: What about expanding example to not only show how to define this datasource, but also how to use it in some other resource?

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
Didainius: Changed to `my-org.my-org-vdc.my-edge-gw.my-lb-service-monitor` everywhere

website/docs/d/lb_service_monitor.html.markdown
Didainius: We can, but not yet. The next resource later on (vcd_lb_server_pool) will be able to consume it. Nothing else can use it now.

website/docs/r/lb_service_monitor.html.markdown
Didainius: Done

vbauzysvmware: It's still showin  `url         = "health"` for me

vbauzysvmware: seems valid :)

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
Didainius: I'm open to naming suggestions then.
`getXStruct` and `setXState` ?

It's actually very common to use `flatten` and `expand` naming across providers:

`terraform-provider-vsphere`
```
$ grep -Ri "func expand" * | wc -l
     114
$ grep -Ri "func flatten" * | wc -l
      96
```

`terraform-provider-aws`
```
$ grep -Ri "func expand" * | wc -l
     381
$ grep -Ri "func flatten" * | wc -l
     428
```

`terraform-provider-azure`
```
$ grep -Ri "func expand" * | wc -l
     340

$ grep -Ri "func flatten" * | wc -l
     376
```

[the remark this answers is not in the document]
lvirbalas: Works for me then. @vbauzysvmware ?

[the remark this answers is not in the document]
vbauzysvmware: Maybe `updateXState` and fillX or getX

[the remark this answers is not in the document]
datacharmer: `getXStruct` and `setXState` are OK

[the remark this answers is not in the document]
lvirbalas: Just noticed more discussion in GitHub about this. Growing on proposals from Vaidotas and Giuseppe, how about `getXType` and `setXData`, e.g.:
`expandLBMonitor` -> `getLBMonitorType` (as in "types.go")
`flattenLBMonitor` -> `setLBMonitorData` (as in "ResourceData")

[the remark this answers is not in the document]
Didainius: Works fine for me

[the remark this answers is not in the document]
vbauzysvmware: I am ok, cause better than expand, flatten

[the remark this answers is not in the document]
Didainius: Have a look at how it is now :)

website/docs/d/lb_service_monitor.html.markdown
lvirbalas: Where to add `data` @vbauzysvmware ?

vcd/resource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
vbauzysvmware: Better ;-)

website/docs/r/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+* `edge_gateway` - (Required) The name of the edge gateway on which the service monitor is to be created
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `name` - (Required) Service Monitor name
+
+* `interval` - (Required) Interval at which a server is to be monitored using the specified Method. 
+* `timeout` - (Required) Maximum time in seconds within which a response from the server must be received 
+* `max_retries` - (Required) Number of times the specified monitoring Method must fail sequentially before the server is declared down
+* `type` - (Required) Select the way in which you want to send the health check request to the server — `http`, `https`, 
+`tcp`, `icmp`, or `udp`. Depending on the type selected, the remaining attributes are allowed or not
```
lvirbalas: Is this the field, if specified incorrectly, we return an error from NSX directly and NSX shows these values in upper case? How can we make sure the user knows to use lower case instead?

website/docs/d/lb_service_monitor.html.markdown
vbauzysvmware: I am not sure: `"vCloudDirector: data source vcd_lb_service_monitor"`, "vCloudDirector: data vcd_lb_service_monitor", "vCloudDirector: data_vcd_lb_service_monitor"

website/docs/r/lb_service_monitor.html.markdown
Didainius: It is a bit of dumb situation inside the API. API accepts both cases for `type` and `method`, but returns in specific cases. The `type` is always being returned in lower case (doesn't matter in which case you submit it) while the `method` always returns uppercase. I didn't to allow confusion about settings state, casing.

My idea is that I have put in terraform validators on these fields to verify case (`type` and `method`) so that you are unable to submit incorrectly cased string.

website/docs/d/lb_service_monitor.html.markdown
Didainius: But is it for like SEO reasons? This field sets only page title (name of browser tab). Rarely people see it I guess. While browsing you see the location quite clearly (or so it looks to me)

image

website/docs/d/lb_service_monitor.html.markdown
lvirbalas: Different providers do it differently. vSphere uses same "silent" approach as us, AWS to the contrary has "Data Source: " or "Resource: " before each title.

website/docs/d/lb_service_monitor.html.markdown
lvirbalas: Thanks, would be great to have that when possible.

website/docs/r/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+## Attribute Reference
+
+The following attributes are exported on the base level of this resource:
+
+* `id` - The ID of the load balancer service monitor
+
+## Importing
+
+An existing load balancer service monitor can be [imported][docs-import] into this resource
+via supplying the full dot separated path for load balancer service monitor. An example is below:
```
lvirbalas: I'm thinking about a warning here that after import you need to do manual operations. As it stands right now, the user may be tempted to run the import command and think that all is done.

go.mod
```diff
@@ -6,3 +6,5 @@ require (
 	github.com/hashicorp/terraform v0.12.0
 	github.com/vmware/go-vcloud-director/v2 v2.3.0-alpha.1
 )
+
+replace github.com/vmware/go-vcloud-director/v2 => github.com/Didainius/go-vcloud-director/v2 v2.3.0-alpha.1.0.20190607105159-e74197f7f588
```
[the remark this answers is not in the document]
Didainius: Already done.

website/docs/r/lb_service_monitor.html.markdown
Didainius: Done.

vcd/resource_vcd_lb_service_monitor_test.go
```diff
@@ -0,0 +1,167 @@
...
+  vdc          = "{{.Vdc}}"
+  edge_gateway = "{{.EdgeGateway}}"
+
+  name        = "{{.ServiceMonitorName}}"
+  interval    = {{.Interval}}
+  timeout     = {{.Timeout}}
+  max_retries = {{.MaxRetries}}
+  type        = "http"
+  method      = "{{.Method}}"
+  send        = "{\"key\": \"value\"}"
```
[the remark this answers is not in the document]
Didainius: Done

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,332 @@
...
+		if k != "" && v != "" { // When key and value are given it must look like "content-type=STRING"
+			extensionString += k + "=" + v.(string) + "\n"
+		} else { // If only key is specified it does not need equals sign. Like "no-body" extension
+			extensionString += k + "\n"
+		}
+	}
+	return extensionString
+}
+
+func setLBMonitorData(d *schema.ResourceData, lBmonitor *types.LBMonitor) error {
```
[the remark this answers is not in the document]
Didainius: Done

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,332 @@
...
+	d.Set("org", orgName)
+	d.Set("vdc", vdcName)
+	d.Set("edge_gateway", edgeName)
+	d.Set("name", monitorName)
+
+	d.SetId(readLBMonitor.ID)
+	return []*schema.ResourceData{d}, nil
+}
+
+func getLBMonitorType(d *schema.ResourceData) (*types.LBMonitor, error) {
```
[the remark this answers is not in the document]
Didainius: Done

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,332 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Unique Service Monitor name",
+			},
+			"interval": &schema.Schema{
+				Type:        schema.TypeInt,
+				Default:     10,
+				Optional:    true,
+				Description: "Interval at which a server is to be monitored",
```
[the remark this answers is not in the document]
Didainius: Done

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,335 @@
...
+			State: resourceVcdLbServiceMonitorImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"vdc": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+			"org": {
```
dataclouder: Please put org before vdc

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,335 @@
...
+	return nil
+}
+
+// resourceVcdLbServiceMonitorImport expects dot formatted path to Load Balancer Service Monitor
+// i.e. my-org.my-org-vdc.my-edge-gw.my-lb-service-monitor
+func resourceVcdLbServiceMonitorImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+
+	resourceURI := strings.Split(d.Id(), ".")
+	if len(resourceURI) != 4 {
+		return nil, fmt.Errorf("resource name must be specified in such way my-org.my-org-vdc.my-edge-gw.my-lb-service-monitor")
```
dataclouder: Suggestion: make the error message shorter and easier to read: "resource name must be specified as org.VDC.edge-gw.lb-service-monitor"

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,335 @@
...
+	return extensionString
+}
+
+// setLBMonitorData sets object state from *types.LBMonitor
+func setLBMonitorData(d *schema.ResourceData, lBmonitor *types.LBMonitor) error {
+	d.Set("interval", lBmonitor.Interval)
+	d.Set("timeout", lBmonitor.Timeout)
+	d.Set("max_retries", lBmonitor.MaxRetries)
+	d.Set("type", lBmonitor.Type)
+	// Optional attributes may not necessarilly
```
dataclouder: "may not necessarilly" -> "may not be necessary"

vcd/resource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,335 @@
...
+	d.Set("expected", lBmonitor.Expected)
+
+	if err := setLBMonitorExtensionData(d, lBmonitor); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// setLBMonitorExtensionData is responsive to parse response extension field from API and
```
dataclouder: "responsive to" -> "responsible for"

vcd/test-templates/misc3.tf
```diff
@@ -0,0 +1,45 @@
+# Remove the first '#' from the next two lines to enable options for terraform executable
```
dataclouder: This sample file is specific to LB. It's probably better to call it load-balancer1.tf

website/docs/d/lb_service_monitor.html.markdown
Didainius: Agreed to go with naked names so far.

website/docs/d/lb_service_monitor.html.markdown
Didainius: Agreed to go with naked names so far.

vcd/resource_vcd_lb_service_monitor_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+
+	"github.com/hashicorp/terraform/terraform"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccVcdLbServiceMonitor(t *testing.T) {
+
+	if vcdShortTest {
```
dataclouder: This check should be moved after the call to `templateFill`, or else the test won't be generated with `make test-binary-prepare`

website/docs/d/lb_service_monitor.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+  Provides a NSX load balancer service monitor data source.
+---
+
+# vcd\_lb\_service\_monitor
+
+Provides a vCloud Director Edge Gateway Load Balancer Service Monitor data source. A service monitor 
+defines health check parameters for a particular type of network traffic. It can be associated with
+a pool. Pool members are monitored according to the service monitor parameters. 
+
+~> **Note:** To make load balancing work one must ensure that load balancing is enabled on edge gateway (edge gateway must be advanced).
```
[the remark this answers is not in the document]
Didainius: Put a link.

vcd/datasource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,90 @@
...
+
+func datasourceVcdLbServiceMonitor() *schema.Resource {
+	return &schema.Resource{
+		Read: datasourceVcdLbServiceMonitorRead,
+		Schema: map[string]*schema.Schema{
+			"vdc": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"org": {
```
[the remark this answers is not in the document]
Didainius: Done

vcd/datasource_vcd_lb_service_monitor.go
```diff
@@ -0,0 +1,90 @@
...
+			"name": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Service Monitor name",
+			},
+			"interval": &schema.Schema{
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"timeout": &schema.Schema{
```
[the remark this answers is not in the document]
Didainius: I can add them but you will never see descriptions in datasource while using the GUI. Resource does have them.

vcd/resource_vcd_lb_service_monitor.go
Didainius: Done

vcd/resource_vcd_lb_service_monitor.go
Didainius: Done.

vcd/resource_vcd_lb_service_monitor.go
Didainius: Done.

vcd/resource_vcd_lb_service_monitor.go
Didainius: Done.

vcd/test-templates/misc3.tf
Didainius: Agree. Renamed

vcd/resource_vcd_lb_service_monitor_test.go
Didainius: Done.

vcd/datasource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
dataclouder: I asked because two properties have descriptions. 
Either we duplicate the descriptions for all, or we add a comment saying that descriptions are listed under the resource definition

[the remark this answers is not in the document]
Didainius: Ok. That's true for these two. Can we agree I add for the "non-computed" ones (then I'll add for `org` and `vdc`)? As these make sense, they're required and if missed the description may be printed.
On the other hand I can easily copy over all of them, it's just that I don't expect them to be seen.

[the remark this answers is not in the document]
dataclouder: Agreed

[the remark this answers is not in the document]
Didainius: Here's my take at it.

vcd/datasource_vcd_lb_service_monitor.go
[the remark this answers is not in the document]
Didainius: It's actually even more than ID. It also includes different search patterns (`readLBMonitor, err := edgeGateway.ReadLBServiceMonitor(&types.LBMonitor{Name: d.Get("name").(string)})` searches by Name, inside resource it searched by ID.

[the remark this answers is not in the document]
vbauzysvmware: ok. Let's leave it as it is then

Issue #69, opened 2017-11-12
Scaling down VMs in vApp restarts vApp


MrMorice:
Hello Terraform Community! We have have a suggestion for optimising scale down scenarios! Unfortunately I'm not that fluent in Go to implement it myself. Thanks for your help!

### Terraform Version
Terraform v0.10.4
terraform-provider-vcd_v1.0.0_x4

### Affected Resource(s)
- vcd_vapp_vm

### Expected Behavior
When scaling down (destroying one or more VMs) the other VMs in the vApp should keep running.

### Actual Behavior
When scaling down (destroying one or more VMs) other VMs get shut down and after the deletion restarted. This leads to a disruption of the service.

### Steps to Reproduce
1. Create a vApp with two VM count of 2
2. Reduce the count to 1 for scaling down
3. The vApp gets shut down before deleting the VMs

pkleanthous:
Hi @MrMorice can you add a code snippet of how are you trying to achieve the scale down?

MrMorice:
Hi @pkleanthous thanks for your response. We basically just change the count of the machines via an input variable. Here is a short example:

resource "vcd_vapp_vm" "example" {
  vapp_name       = "${var.vapp}"
  [...]
  count = "${var.examplecount}"
}

Increasing the "examplecount" input variable works fine and the vApp keeps running while we scale up. Decreasing the input variable however will result in a shutdown of all the VMs and then a restart after the machines are destroyed.

The reason for the behaviour is this line in the resource_vcd_vapp_vm.go file that undeployes the vApp before destroying one VM. This is probably legacy functionality from the time where the provider only supported one VM per vApp. But it should not be needed anymore.
https://github.com/terraform-providers/terraform-provider-vcd/blob/master/vcd/resource_vcd_vapp_vm.go#L332

Thanks for your help!

Kind regards,
MrMorice

pkleanthous:
I can confirm the reported behaviour and when you remove a VM resource from a TF file

```hcl
resource "vcd_vapp_vm" "demo-vm-1" {
  vapp_name     = "${vcd_vapp.test-demo.name}"
  name          = "demo-vm-1"
  catalog_name  = "Catalogue"
  template_name = "magic_template"
  network_name  = "${vcd_network.net.name}"
}

resource "vcd_vapp_vm" "demo-vm-2" {
  vapp_name     = "${vcd_vapp.test-demo.name}"
  name          = "demo-vm-2"
[...]

PR #261, opened 2019-06-10
fix typo that undeploys vapp instead of vm while deleting a vm

vkhacharia:
Found the pointer in #69 Updated the code and tested locally.


vbauzysvmware:
Nice catch, still we need to investigate how it related to `vapp.Deploy()` part which start from 875 line.

vbauzysvmware:
If everyone would approve I suggest together with this fix remove 873-894 case. I don't see any more value for deploy if don't undepoy vAPP. My testing seem to verify that. I would like other to verify that.

vkhacharia:
Do I update the changelog and make new pull request?

Didainius:
@vkhacharia NP, I will put in a note with the next PR I have in the oven :)
Thank you again.

PR #272, opened 2019-06-28
Update independent disk size type

dataclouder:
Using int64, as required in the API, instead of int, which was
   mistakenly used in go-vcloud-director


Didainius:
It is missing `go mod tidy` to pass tests.

dataclouder:
> It is missing `go mod tidy` to Travis pass tests.

Yep. Done

PR #278, opened 2019-07-08
Add load balancer application rule resource and data source

Didainius:
Continuing work on terraform-providers/terraform-provider-vcd#223

This one introduces resource and data source for Load Balancer Application Rule.


go.mod
```diff
@@ -6,3 +6,5 @@ require (
 	github.com/hashicorp/terraform v0.12.0
 	github.com/vmware/go-vcloud-director/v2 v2.3.0-alpha.8
 )
+
+replace github.com/vmware/go-vcloud-director/v2 => github.com/Didainius/go-vcloud-director/v2 v2.3.0-alpha.3.0.20190708140304-330dcdc2ba13
```
vbauzysvmware: reminder to update

vcd/resource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	}
+
+	LBRule, err := getLBAppRuleType(d)
+	if err != nil {
+		return fmt.Errorf("unable to expand load balancer app rule: %s", err)
+	}
+
+	createdPool, err := edgeGateway.CreateLBAppRule(LBRule)
```
vbauzysvmware: `CreateLBAppRule(LBRule)` - IMHO would be much better readable as `CreateLBAppRule(name, script)`. as it's now it shows that LBRULE ir complex, but it isn't

vcd/resource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		return fmt.Errorf("error creating new load balancer app rule: %s", err)
+	}
+
+	// We store the values once again because response include pool member IDs
+	err = setLBAppRuleData(d, createdPool)
```
vbauzysvmware: but ID settted not in this function. Seems comment isn't correct

vcd/resource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	}
+
+	updateLBRuleConfig, err := getLBAppRuleType(d)
+	if err != nil {
+		return fmt.Errorf("could not expand load balancer app rule for update: %s", err)
+	}
+
+	updatedLBRule, err := edgeGateway.UpdateLBAppRule(updateLBRuleConfig)
```
vbauzysvmware: Again, Struct for name and script

vcd/resource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,220 @@
...
+		return []*schema.ResourceData{}, fmt.Errorf("unable to find load balancer app rule with name %s: %s",
+			d.Id(), err)
+	}
+
+	d.Set("org", orgName)
+	d.Set("vdc", vdcName)
+	d.Set("edge_gateway", edgeName)
+	d.Set("name", appProfileName)
+
+	d.SetId(readLBRule.ID)
```
vbauzysvmware: Why for Rule `appProfileName' is saved and why script isn't saved ?

vcd/resource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,220 @@
...
+		Script: scriptString,
+	}
+
+	return lbAppRule, nil
+}
+
+// setLBAppRuleData sets name and script API fields. API output returns a single string separated by
+// newline ("\n") for each line of script. To store it in Terraform's TypeList we must convert it
+// into []interface{} before calling d.Set(). API response must be split by newline ("\n") and
+// then typecast to []interface{}
```
vbauzysvmware: E.g. would be valuable

vcd/resource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,220 @@
...
+	defer vcdClient.unLockParentEdgeGtw(d)
+
+	edgeGateway, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
+	if err != nil {
+		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	}
+
+	LBRule, err := getLBAppRuleType(d)
+	if err != nil {
+		return fmt.Errorf("unable to expand load balancer app rule: %s", err)
```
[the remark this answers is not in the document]
Didainius: Very good catch. It actually left in many places over the errors. Retrofitted.

vcd/resource_vcd_lb_app_rule.go
Didainius: Indeed. This is coming from the "pancake" copy/paste work...
Removed it at all as it doesn't have much meaning in this case.

vcd/resource_vcd_lb_app_rule.go
Didainius: Because as we spoke before we only care to get the ID in `import` (other fields org, vdc, edge_gateway and name are required anyway. Could even avoid setting them. Import automatically triggers `refresh` and it fills in the missing pieces. (script in this case)

vcd/resource_vcd_lb_app_rule.go
Didainius: Added.

vcd/resource_vcd_lb_app_rule.go
Didainius: I could create wrapper, but I would like to keep this consistent and transparent to the struct we have in `types`.
The reason I dislike when we introduce our own "simplified types" for structs is that we end up translating between structs. Then if a new field appears (or we missed some) in the configuration our APIs become deprecated and a new method must be created.
Again if this is a team decision I can change :)
@lvirbalas, @dataclouder, your thoughts about it?

vcd/resource_vcd_lb_app_rule.go
Didainius: Put a comment above.

vcd/resource_vcd_lb_app_rule.go
Didainius: One more thought. What if I added more explanation to the type fields we have in Go fashion. That way any modern IDE would show more context of what is expected in the field but would still keep the transparency.

These are mockups comments, not exactly the ones I intend to add.
image

lvirbalas: > The reason I dislike when we introduce our own "simplified types" for structs is that we end up translating between structs. Then if a new field appears (or we missed some) in the configuration our APIs become deprecated and a new method must be created.
Again if this is a team decision I can change :)

I dislike the simplified types as well, as they add another layer of abstraction.

vbauzysvmware: Well, seems we have diff reading and API usage experience. For me as terraform and API reader/user it's easier to pass two string, than Structure with four fields which I have to check which to fill. From Java world there is `rule` if method with 4-5 parameters when refactor to pass object.

vcd/resource_vcd_lb_app_rule.go
vbauzysvmware: Don't see the change

vcd/resource_vcd_lb_app_rule.go
vbauzysvmware: don't see the change

vcd/resource_vcd_lb_app_rule.go
Didainius: Failed to push. Now it shouldn't be there.

vcd/resource_vcd_lb_app_rule.go
Didainius: Sorry. I meant a comment in the above discussion we have. https://github.com/terraform-providers/terraform-provider-vcd/pull/278#discussion_r301886652

vcd/resource_vcd_lb_app_rule.go
vbauzysvmware: Why appProfleName and not RuleName ?

vcd/resource_vcd_lb_app_rule.go
vbauzysvmware: Oh, Go ti now:)

vcd/resource_vcd_lb_app_rule.go
Didainius: It may be so. But I have already burnt with such methods in this project. I had to change quite a few details on VM networking just because there were "simplified" fields. The public API gets too many changes this way.

vcd/resource_vcd_lb_app_rule.go
Didainius: Another one. And I thought I have read it a few times after changing.
Changed. Thanks

vcd/resource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,227 @@
...
+		return fmt.Errorf("error deleting load balancer app rule: %s", err)
+	}
+
+	d.SetId("")
+	return nil
+}
+
+// resourceVcdLBAppRuleImport is responsible for importing the resource.
+// The d.Id() field as being passed from `terraform import _resource_name_ _the_id_string_ requires
+// a name based dot-formatted path to the object to lookup the object and sets the id of object.
```
lvirbalas: A bit confused when reading this without context for the first time. Consider rephrasing. Maybe in a more declarative form and with a numbered list (1, 2, 3 ...)? I.e. what happens after what with some sample values.

website/docs/d/lb_app_rule.html.markdown
```diff
@@ -0,0 +1,42 @@
...
+layout: "vcd"
+page_title: "vCloudDirector: vcd_lb_app_rule"
+sidebar_current: "docs-vcd-data-source-lb-app-rule"
+description: |-
+  Provides an NSX edge gateway load balancer application rule data source.
+---
+
+# vcd\_lb\_app\_rule
+
+Provides a vCloud Director Edge Gateway Load Balancer Application Rule resource. An application rule
```
lvirbalas: "resource" -> "data source"?

website/docs/d/lb_app_rule.html.markdown
```diff
@@ -0,0 +1,42 @@
...
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+* `edge_gateway` - (Required) The name of the edge gateway on which the service monitor is defined
+* `name` - (Required) Application rule name for identifying the exact application rule
+
+## Attribute Reference
+
+All the attributes defined in `vcd_lb_app_rule` resource are be available.
```
lvirbalas: Grammar mistake: "are be"

website/docs/r/lb_app_rule.html.markdown
```diff
@@ -0,0 +1,104 @@
...
+
+resource "vcd_lb_app_rule" "example-two" {
+  edge_gateway = "my-edge-gw"
+  org          = "my-org"
+  vdc          = "my-org-vdc"
+  name         = "script1"
+  script = [
+    "acl vmware_page url_beg / vmware redirect location https://www.vmware.com/ ifvmware_page",
+    "acl other_page2 url_beg / other2 redirect location https://www.other2.com/ ifother_page2",
+    "acl hello payload(0,6) -m bin 48656c6c6f0a"
```
lvirbalas: Interesting approach to divide lines. Why such? Does this come from below API? We have one place where we define scripts - the `initscript` - is its way of any use here?

vcd/resource_vcd_lb_app_rule.go
Didainius: In the light of current discussion - what if I add a wrapper for it now and we have both options available?

website/docs/r/lb_app_rule.html.markdown
Didainius: I might have overengineered with this. I'll have a look if Terraform could accept raw string (with naturally including newlines) and what `initscript` does.

website/docs/d/lb_app_rule.html.markdown
Didainius: Done.

website/docs/d/lb_app_rule.html.markdown
Didainius: Thanks. Done.

vcd/resource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,227 @@
...
+	defer vcdClient.unLockParentEdgeGtw(d)
+
+	edgeGateway, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
+	if err != nil {
+		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	}
+
+	LBRule, err := getLBAppRuleType(d)
+	if err != nil {
+		return fmt.Errorf("unable to create load balancer app rule type: %s", err)
```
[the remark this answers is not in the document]
Didainius: Should be better unless I missed something.

vcd/resource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,227 @@
...
+		Importer: &schema.ResourceImporter{
+			State: resourceVcdLBAppRuleImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"org": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "vCD organization in which the Application Rule is located",
```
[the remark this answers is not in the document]
Didainius: Should be added.

vcd/resource_vcd_lb_app_rule.go
lvirbalas: I think you already have this approach for all other previous LB resources. Also, there are many other instances of this approach in EG already. Hence, I suggest leaving this as is in this PR, but  @dataclouder could add this to refactoring consideration together with https://github.com/vmware/go-vcloud-director/issues/211

vcd/resource_vcd_lb_app_rule.go
Didainius: Changed. Any better?

lvirbalas: A lot! Thanks!

website/docs/r/lb_app_rule.html.markdown
Didainius: Here's another take at it. Apparently Terraform has a HEREDOC syntax which plays nicely here.

lvirbalas: Fantastic!

vcd/datasource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,61 @@
...
+			},
+			"name": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "LB Application Rule name for lookup",
+			},
+			"script": {
+				Computed: true,
+				Type:     schema.TypeString,
+				Description: "The script for the LB Application Rule. Each line will be separated " +
```
dataclouder: technically, one line is "_terminated_ by a newline", while "many lines are _separated_ by newlines".
I prefer using "newline" instead of "\n character".

vcd/resource_vcd_lb_app_rule_test.go
```diff
@@ -0,0 +1,194 @@
...
+	"testing"
+
+	"github.com/vmware/go-vcloud-director/v2/govcd"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccVcdLBAppRule(t *testing.T) {
+	// The Script parameter must be sent as multiline string separated by \n characters.
```
dataclouder: can we say "newlines" instead of "\n characters"?

vcd/resource_vcd_lb_app_rule.go
dataclouder: I am OK with @Didainius approach. 
Simplified structures should be used only when: (I) we have more than 3 fields to pass by position, (II) the original structures are not flat, (III) require internal structure manipulation, (IV) and can't be passed to update methods without changes. A clear example is Edge Gateway, where the structure contains a huge amount of details that we may not intend to deal with in every operation. But this structure is manageable as it is.

vcd/resource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,212 @@
...
+	d.SetId("")
+	return nil
+}
+
+// resourceVcdLBAppRuleImport is responsible for importing the resource.
+// The following steps happen as part of import
+// 1. The user supplies `terraform import _resource_name_ _the_id_string_` command
+// 2. `_the_id_string_` contains a dot formatted path to resource as in the example below
+// 3. The functions splits the dot-formatted path and tries to lookup the object
+// 4. If the lookup succeeds it set's the ID field for `_resource_name_` resource in statefile
```
dataclouder: "set's" -> "sets"

vcd/resource_vcd_lb_app_rule.go
```diff
@@ -0,0 +1,212 @@
...
+func getLBAppRuleType(d *schema.ResourceData) (*types.LBAppRule, error) {
+	lbAppRule := &types.LBAppRule{
+		Name:   d.Get("name").(string),
+		Script: d.Get("script").(string),
+	}
+
+	return lbAppRule, nil
+}
+
+// setLBAppRuleData sets name and script API fields. API output returns a single string separated by
```
dataclouder: "string separated" -> "string terminated"

vcd/resource_vcd_lb_app_rule_test.go
```diff
@@ -0,0 +1,194 @@
...
+	configText := templateFill(testAccVcdLBAppRule_OneLine, params)
+	debugPrintf("#[DEBUG] CONFIGURATION for step 0: %s", configText)
+
+	params["FuncName"] = t.Name() + "-step1"
+	configText1 := templateFill(testAccVcdLBAppRule_MultiLine, params)
+	debugPrintf("#[DEBUG] CONFIGURATION for step 1: %s", configText1)
+
+	params["FuncName"] = t.Name() + "-step3"
+	// This test must fail with invalid rule script so we avoid running it in `make test-binary`
+	params["SkipTest"] = "# skip-test"
```
dataclouder: The way skip test is implemented, it allows you to add a short message after "skip-test" and until the end of the line, so that you can state the reason for skipping, and `test-binary.sh` will report such reason.
For example: `# skip-test: it will fail on purpose`

website/docs/r/lb_app_rule.html.markdown
```diff
@@ -0,0 +1,104 @@
...
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+* `edge_gateway` - (Required) The name of the edge gateway on which the application rule is to be created
+* `name` - (Required) Application rule name
+* `script` - (Required) A multiline application rule script.
+Terraform's [HEREDOC syntax](https://www.terraform.io/docs/configuration/expressions.html#string-literals)
+may be usefull for multiline scripts. **Note:** For information on
```
dataclouder: "usefull" -> "useful"

website/docs/r/lb_app_rule.html.markdown
```diff
@@ -0,0 +1,104 @@
...
+## Importing
+
+~> **Note:** The current implementation of Terraform import can only import resources into the state.
+It does not generate configuration. [More information.](https://www.terraform.io/docs/import/)
+
+An existing load balancer application rule can be [imported][docs-import] into this resource
+via supplying the full dot separated path for load balancer application rule. An example is
+below:
+
+[docs-import]: /docs/import/index.html
```
dataclouder: This address results in a broken link. If it is the same as `https://www.terraform.io/docs/import/`, the same address should be used for both cases.

vcd/resource_vcd_lb_app_rule_test.go
lvirbalas: Let's leave `\n` as it's instantly clear, while newline has some blurry aspects to it (you need to digest that this is \n first, and even then you may think (if you come from Windows background) does it mean \n or \r\n...

Didainius: I can mention both

website/docs/r/lb_app_rule.html.markdown
Didainius: Yes. Thanks. I remember I will have to replace everywhere else. Replaced here.

website/docs/r/lb_app_rule.html.markdown
Didainius: Done.

vcd/resource_vcd_lb_app_rule_test.go
Didainius: Very nice. Using it.

vcd/resource_vcd_lb_app_rule.go
Didainius: Done.

vcd/resource_vcd_lb_app_rule.go
Didainius: Done.

vcd/resource_vcd_lb_app_rule_test.go
Didainius: Done.

vcd/datasource_vcd_lb_app_rule.go
Didainius: Mentioned both. Hopefully this makes it clearer.

PR #279, opened 2019-07-10
Add org user

dataclouder:
* Add resource for Org User
    * Add create, update, read, delete functionalities for Org User
    * Add importing for Org User
    * Add tests for create, update, delete, import
* Add Org User section to environment build script
* Add testing tags for lb and user to GNUmakefile
* Add feature to skip a script from comments in test-binary.sh
* Add documentation for Org User resource

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,20 +3,20 @@
 FEATURES:
 
 * **New Resource:** `vcd_lb_service_monitor`  Load Balancer Service Monitor - [GH-256]
-* **New Data Source:** `vcd_lb_service_monitor` Load Balancer Service Monitor  - [GH-256]
 * **New Resource:** `vcd_edgegateway` creates and deletes edge gateways - [GH-262]
 * **New Resource:** `vcd_lb_server_pool` Load Balancer Server Pool - [GH-268]
-* **New Data Source:** `vcd_lb_server_pool`  Load Balancer Server Pool - [GH-268]
 * **New Resource:** `vcd_lb_app_profile` Load Balancer Application profile - [GH-274]
-* **New Data Source:** `vcd_lb_app_profile` Load Balancer Application profile - [GH-274]
 * **New Resource:** `vcd_org_user`  Organization User - [GH-279]
-* resource/vcd_org_vdc added Org VDC update and full state read - [GH-275]
-* resource/vcd_org_vdc added Org VDC metadata support - [GH-276]
+* **New Data Source:** `vcd_lb_service_monitor` Load Balancer Service Monitor  - [GH-256]
+* **New Data Source:** `vcd_lb_server_pool`  Load Balancer Server Pool - [GH-268]
+* **New Data Source:** `vcd_lb_app_profile` Load Balancer Application profile - [GH-274]
 * **New build commands** `make test-env-init` and `make test-env-apply` can configure an empty vCD to run the test suite. See `TESTING.md` for details.
+* `resource/vcd_org_vdc` added Org VDC update and full state read - [GH-275]
+* `resource/vcd_org_vdc` added Org VDC metadata support - [GH-276]
 
 IMPROVEMENTS:
-* resource/vcd_org_vdc: Fix ignoring of resource guarantee values - [GH-265]
-* resource/vcd_org_vdc: Org VDC state ID changed from name to vCD ID - [GH-275]
+* `resource/vcd_org_vdc`: Fix ignoring of resource guarantee values - [GH-265]
+* `resource/vcd_org_vdc`: Org VDC state ID changed from name to vCD ID - [GH-275]
 * Change resource handling to use locking mechanism when resource parallel handling is not supported by vCD. [GH-255] 
 * Fix issue when vApp is power cycled during member VM deletion. [GH-261]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ FEATURES:
 * **New Resource:** `vcd_lb_service_monitor`  Load Balancer Service Monitor - [GH-256]
 * **New Data Source:** `vcd_lb_service_monitor` Load Balancer Service Monitor  - [GH-256]
 * **New Resource:** `vcd_edgegateway` creates and deletes edge gateways - [GH-262]
-* **New build commands** `make test-env-init` and `make test-env-apply` can configure an empty vCD to run the test suite. See `TESTING.md` for details.
 * **New Resource:** `vcd_lb_server_pool` Load Balancer Server Pool - [GH-268]
 * **New Data Source:** `vcd_lb_server_pool`  Load Balancer Server Pool - [GH-268]
 * **New Resource:** `vcd_lb_app_profile` Load Balancer Application profile - [GH-274]
@@ -13,6 +12,7 @@ FEATURES:
 * **New Resource:** `vcd_org_user`  Organization User - [GH-279]
 * resource/vcd_org_vdc added Org VDC update and full state read - [GH-275]
 * resource/vcd_org_vdc added Org VDC metadata support - [GH-276]
+* **New build commands** `make test-env-init` and `make test-env-apply` can configure an empty vCD to run the test suite. See `TESTING.md` for details.
 
 IMPROVEMENTS:
 * resource/vcd_org_vdc: Fix ignoring of resource guarantee values - [GH-265]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,12 @@ FEATURES:
 * **New Resource:** `vcd_lb_service_monitor`  Load Balancer Service Monitor - [GH-256]
 * **New Data Source:** `vcd_lb_service_monitor` Load Balancer Service Monitor  - [GH-256]
 * **New Resource:** `vcd_edgegateway` creates and deletes edge gateways - [GH-262]
-* **New build command** `make test-env-build` can configure an empty vCD to run the test suite. See `TESTING.md` for details.
+* **New build commands** `make test-env-init` and `make test-env-apply` can configure an empty vCD to run the test suite. See `TESTING.md` for details.
 * **New Resource:** `vcd_lb_server_pool` Load Balancer Server Pool - [GH-268]
 * **New Data Source:** `vcd_lb_server_pool`  Load Balancer Server Pool - [GH-268]
 * **New Resource:** `vcd_lb_app_profile` Load Balancer Application profile - [GH-274]
 * **New Data Source:** `vcd_lb_app_profile` Load Balancer Application profile - [GH-274]
+* **New Resource:** `vcd_org_user`  Organization User - [GH-279]
 * resource/vcd_org_vdc added Org VDC update and full state read - [GH-275]
 * resource/vcd_org_vdc added Org VDC metadata support - [GH-276]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## 2.4.0 (Unreleased)
+
 FEATURES:
 
 * **New Resource:** `vcd_lb_service_monitor`  Load Balancer Service Monitor - [GH-256]
@@ -7,14 +8,16 @@ FEATURES:
 * **New build command** `make test-env-build` can configure an empty vCD to run the test suite. See `TESTING.md` for details.
 * **New Resource:** `vcd_lb_server_pool` Load Balancer Server Pool - [GH-268]
 * **New Data Source:** `vcd_lb_server_pool`  Load Balancer Server Pool - [GH-268]
+* **New Resource:** `vcd_lb_app_profile` Load Balancer Application profile - [GH-274]
+* **New Data Source:** `vcd_lb_app_profile` Load Balancer Application profile - [GH-274]
 * resource/vcd_org_vdc added Org VDC update and full state read - [GH-275]
+* resource/vcd_org_vdc added Org VDC metadata support - [GH-276]
 
 IMPROVEMENTS:
 * resource/vcd_org_vdc: Fix ignoring of resource guarantee values - [GH-265]
 * resource/vcd_org_vdc: Org VDC state ID changed from name to vCD ID - [GH-275]
 * Change resource handling to use locking mechanism when resource parallel handling is not supported by vCD. [GH-255] 
 * Fix issue when vApp is power cycled during member VM deletion. [GH-261]
-
 
 ## 2.3.0 (May 29, 2019)
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -88,8 +88,8 @@ integrate such data with additional information that is only used for the enviro
     "routedNetwork": "net_datacloud_r",
     "isolatedNetwork": "net_datacloud_i",
     "directNetwork": "net_datacloud_d"
-	"orgUser": "company-admin",
-	"orgUserPassword": "secretpwd"
+    "orgUser": "company-admin",
+    "orgUserPassword": "secretpwd"
   }
 }
 ```
--- a/TESTING.md
+++ b/TESTING.md
@@ -204,6 +204,7 @@ $ go test -v .
             * network:    Runs network related tests
             * gateway:    Runs edge gateway related tests
             * org:        Runs org related tests
+            * user:       Runs user related tests
             * vapp:       Runs vapp related tests
             * vdc:        Runs vdc related tests
             * vm:         Runs vm related tests
--- a/TESTING.md
+++ b/TESTING.md
@@ -88,12 +88,14 @@ integrate such data with additional information that is only used for the enviro
     "routedNetwork": "net_datacloud_r",
     "isolatedNetwork": "net_datacloud_i",
     "directNetwork": "net_datacloud_d"
+	"orgUser": "company-admin",
+	"orgUserPassword": "secretpwd"
   }
 }
 ```
 
-When this section is filled, the test system has all the information to create all the elements (except, for now, the
-org admin user). When you run the command `make test-binary-prepare`, among the files ready to run there will be one named
+When this section is filled, the test system has all the information to create all the elements. When you run the
+command `make test-binary-prepare`, among the files ready to run there will be one named
 `cust.full-env.tf`, containing all the information to populate your vCD with the test resources.
 **IMPORTANT**: this procedure assumes that **the regular part of the configuration file is filled with all the correct
 information**. The list of fields given as mandatory above is not sufficient to create a working environment, if
--- a/vcd/config_test.go
+++ b/vcd/config_test.go
@@ -460,13 +460,13 @@ func TestMain(m *testing.M) {
 	// If VCD_SHORT_TEST is defined, it means that "make test" is called,
 	// and we won't really run any tests involving vcd connections.
 	configFile := getConfigFileName()
-	if configFile == "" {
-		fmt.Println("No configuration file found")
-		os.Exit(1)
-	}
 	testConfig = getConfigStruct(configFile)
 	if !vcdShortTest {
 
+		if configFile == "" {
+			fmt.Println("No configuration file found")
+			os.Exit(1)
+		}
 		fmt.Printf("Connecting to %s\n", testConfig.Provider.Url)
 		fmt.Printf("as user %s@%s\n", testConfig.Provider.User, testConfig.Provider.SysOrg)
 		// Provider initialization moved here from provider_test.init
--- a/vcd/config_test.go
+++ b/vcd/config_test.go
@@ -460,11 +460,15 @@ func TestMain(m *testing.M) {
 	// If VCD_SHORT_TEST is defined, it means that "make test" is called,
 	// and we won't really run any tests involving vcd connections.
 	configFile := getConfigFileName()
-	if configFile != "" {
-		testConfig = getConfigStruct(configFile)
+	if configFile == "" {
+		fmt.Println("No configuration file found")
+		os.Exit(1)
 	}
+	testConfig = getConfigStruct(configFile)
 	if !vcdShortTest {
 
+		fmt.Printf("Connecting to %s\n", testConfig.Provider.Url)
+		fmt.Printf("as user %s@%s\n", testConfig.Provider.User, testConfig.Provider.SysOrg)
 		// Provider initialization moved here from provider_test.init
 		testAccProvider = Provider().(*schema.Provider)
 		testAccProviders = map[string]terraform.ResourceProvider{
--- a/vcd/config_test.go
+++ b/vcd/config_test.go
@@ -1,4 +1,4 @@
-// +build api functional catalog vapp network extnetwork org query vm vdc gateway disk binary lb lbServiceMonitor lbServerPool ALL
+// +build api functional catalog vapp network extnetwork org query vm vdc gateway disk binary lb lbServiceMonitor lbServerPool user ALL
 
 package vcd
 
@@ -110,6 +110,8 @@ type TestConfig struct {
 		MediaPath                    string `json:"mediaPath"`                    // Media path, if different from Media.MediaPath
 		MediaName                    string `json:"mediaName"`                    // Media name to create
 		OvaPath                      string `json:"ovaPath"`                      // Ova Path, if different from Ova.OvaPath
+		OrgUser                      string `json:"orgUser"`                      // Org User to be created within the organization
+		OrgUserPassword              string `json:"orgUserPassword"`              // Password for the Org User to be created within the organization
 	} `json:"testEnvBuild"`
 	EnvVariables map[string]string `json:"envVariables,omitempty"`
 }
@@ -148,6 +150,7 @@ provider "vcd" {
   max_retry_timeout    = {{.MaxRetryTimeout}}
   version              = "~> {{.VersionRequired}}"
   logging              = {{.Logging}}
+  logging_file         = "{{.LoggingFile}}"
 }
 `
 )
@@ -245,6 +248,11 @@ func templateFill(tmpl string, data StringMap) string {
 		data["MaxRetryTimeout"] = testConfig.Provider.MaxRetryTimeout
 		data["VersionRequired"] = currentProviderVersion
 		data["Logging"] = testConfig.Logging.Enabled
+		if testConfig.Logging.LogFileName != "" {
+			data["LoggingFile"] = testConfig.Logging.LogFileName
+		} else {
+			data["LoggingFile"] = util.ApiLogFileName
+		}
 	}
 	if _, ok := data["Tags"]; !ok {
 		data["Tags"] = "ALL"
@@ -315,7 +323,7 @@ func templateFill(tmpl string, data StringMap) string {
 		}
 		err = writer.Flush()
 		if err != nil {
-			panic(fmt.Errorf("error writing to file %s. Reported %d bytes written. %s", resourceFile, count, err))
+			panic(fmt.Errorf("error flushing file %s. %s", resourceFile, err))
 		}
 		_ = file.Close()
 	}
--- a/vcd/config_test.go
+++ b/vcd/config_test.go
@@ -460,7 +460,9 @@ func TestMain(m *testing.M) {
 	// If VCD_SHORT_TEST is defined, it means that "make test" is called,
 	// and we won't really run any tests involving vcd connections.
 	configFile := getConfigFileName()
-	testConfig = getConfigStruct(configFile)
+	if configFile != "" {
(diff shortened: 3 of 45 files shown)


vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,351 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+
+				ConflictsWith: []string{"password_file"},
+				Description: "The user's password. This value is never returned on read. " +
+					"It is inspected on create and modify. " +
+					"On modify, the absence of this element indicates that the password should not be changed.",
+			},
+			"password_file": &schema.Schema{
```
Didainius: One thing I see about the password file is that it is hard to change the password (you must change file name).

Don't know how to make it better other than state including timestamp of when the file was modified  in a separate Computed field and comparing against it. Up to you if it's worth it :)

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,351 @@
...
+				Optional:    true,
+				ForceNew:    false,
+				Description: "The Org User's description",
+			},
+			"provider_type": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     govcd.OrgUserProviderIntegrated,
+				ForceNew:    false,
+				Description: "Identity provider type for this this user. One of: 'INTEGRATED', 'SAML', 'OAUTH'",
```
Didainius: Worth noting default value?

website/docs/r/org_user.html.markdown
```diff
@@ -0,0 +1,165 @@
...
+
+* `id` - The ID of the Organization user
+
+
+## Importing
+
+~> **Note:** The current implementation of Terraform import can only import resources into the state. It does not generate
+configuration. [More information.][docs-import]
+
+An existing load balancer server pool can be [imported][docs-import] into this resource
```
Didainius:
```suggestion
An existing user can be [imported][docs-import] into this resource
```

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,351 @@
...
+	userName := d.Get("name").(string)
+	orgUser, err := adminOrg.FetchUserByName(userName, false)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return orgUser, &adminOrg, nil
+}
+
+func setOrgUserResource(d *schema.ResourceData, orgUser *govcd.OrgUser, adminOrg *govcd.AdminOrg) error {
```
Didainius: I think we had agreement on naming these functions so:
getXType
setXData

vcd/config_test.go
```diff
@@ -110,6 +110,8 @@ type TestConfig struct {
 		MediaPath                    string `json:"mediaPath"`                    // Media path, if different from Media.MediaPath
 		MediaName                    string `json:"mediaName"`                    // Media name to create
 		OvaPath                      string `json:"ovaPath"`                      // Ova Path, if different from Ova.OvaPath
+		OrgUser                      string `json:"orgUser"`                      // Org User to be created within the organization
```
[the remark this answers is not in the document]
dataclouder: I considered that, but then I thought that the purpose of this field is to allow us to create the environment the way we want it, because after that this is the user that we connect with when running tests as non-sysAdmin.
Moreover, you know that we have a tool that creates the environment for you. If you leave it to de defaults, the user will be `org_name-admin`.

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,351 @@
...
+			},
+			"password": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+
+				ConflictsWith: []string{"password_file"},
+				Description: "The user's password. This value is never returned on read. " +
+					"It is inspected on create and modify. " +
+					"On modify, the absence of this element indicates that the password should not be changed.",
```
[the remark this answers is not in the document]
dataclouder: The wording is copied verbatim from the API definition, and I thought that it defines well what is happening.
But I can find a better wording.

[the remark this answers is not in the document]
Didainius: > The wording is copied verbatim from the API definition

Yes. Exactly. But the user is not dealing with API so he does not have to skip field.
That's my main idea.

vcd/resource_vcd_org_user.go
dataclouder: I have thought of adding a "change_password" Boolean field for this purpose.
Both approaches ("password" and "password_file") have pros and cons. For automation, I think using a password file that can be generated and removed in a script is the easy approach.

The *easiest* approach would be to update using the "password" field but then you would have to apply again with an empty password to remove it from the state file.

dataclouder:
> CHANGELOG is waiting for you :)
Yes, thanks. And `sample_vcd_test_config.json` and `TESTING.md` need to be updated.

vcd/resource_vcd_org_user.go
dataclouder: The `Default` line above the description is not enough?

vcd/resource_vcd_org_user.go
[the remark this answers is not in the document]
dataclouder: I left out the unnecessary wording.

website/docs/r/org_user.html.markdown
dataclouder: changed

vcd/resource_vcd_org_user.go
dataclouder: Changed

vcd/resource_vcd_org_user.go
Didainius: Probably you are right. In the end the user simply won't see this description if he/she skips this field.

CHANGELOG.md
```diff
@@ -5,11 +5,12 @@ FEATURES:
 * **New Resource:** `vcd_lb_service_monitor`  Load Balancer Service Monitor - [GH-256]
 * **New Data Source:** `vcd_lb_service_monitor` Load Balancer Service Monitor  - [GH-256]
 * **New Resource:** `vcd_edgegateway` creates and deletes edge gateways - [GH-262]
-* **New build command** `make test-env-build` can configure an empty vCD to run the test suite. See `TESTING.md` for details.
+* **New build commands** `make test-env-init` and `make test-env-apply` can configure an empty vCD to run the test suite. See `TESTING.md` for details.
```
Didainius: Not sure about positioning this one (could probably go down in the list due to orderding) and could use (if we have) a [GH-XXX] reference.

lvirbalas: Please move this section to the bottom of the list, so it stands out on its own (as opposed to in the group of LB additions).

TESTING.md
```diff
@@ -88,12 +88,14 @@ integrate such data with additional information that is only used for the enviro
     "routedNetwork": "net_datacloud_r",
     "isolatedNetwork": "net_datacloud_i",
     "directNetwork": "net_datacloud_d"
+	"orgUser": "company-admin",
+	"orgUserPassword": "secretpwd"
```
lvirbalas: Maybe GitHub thing, but shows not indented compared to other lines above.

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,352 @@
...
+				ForceNew:    false,
+				Description: "Role within the organization",
+			},
+			"password": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+
+				ConflictsWith: []string{"password_file"},
+				Description: "The user's password. This value is never returned on read. " +
```
lvirbalas: Please elaborate, if value is not returned on read, is it reporting that there's a change with each succeeding `plan`?

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,352 @@
...
+				ForceNew:      false,
+				ConflictsWith: []string{"password"},
+				Description: "Name of a file containing the user's password. " +
+					`Either "password_file" or "password" must be included on creation.`,
+			},
+			"description": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    false,
+				Description: "The Org User's description",
```
lvirbalas: Capitalization: you're using "user" or "User" in different places. Please choose one and use everywhere.

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,352 @@
...
+			"instant_messaging": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    false,
+				Description: "The Org User's telephone",
+			},
+			"is_enabled": &schema.Schema{
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
```
lvirbalas: We had this somewhere else and decided that `is_enabled` should be `true` by default and user can explicitly disable with `is_enabled="false"` when needed. This translates to one parameter less to worry about for the easy use cases.

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,352 @@
...
+				ForceNew:    false,
+				Description: "True if this user has a group role.",
+			},
+			"is_locked": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: false,
+				Description: "True if the user account has been locked due to too many invalid login attempts. " +
+					"A locked user account can be re-enabled by updating the user with this flag set to false. " +
+					"Only the system can set the value to true.",
```
lvirbalas: > "A locked user account can be re-enabled by updating the user with this flag set to false. " 

It sounds like in order to re-enable such user one would need to set the flag to **false**...

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,352 @@
...
+				Optional:    true,
+				Default:     10,
+				ForceNew:    false,
+				Description: "Quota of vApps that this user can store. A value of 0 specifies an unlimited quota.",
+			},
+		},
+	}
+}
+
+func resourceToUserData(d *schema.ResourceData, meta interface{}) (*govcd.OrgUserConfiguration, *govcd.AdminOrg, error) {
```
lvirbalas: Short comment above the function's purpose would be nice.

Also, why we started to need this kind of function? (I don't believe we have it in other places or it's named differently).

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,352 @@
...
+		}
+	}
+
+	if userData.Password == "" {
+		return nil, nil, fmt.Errorf(`no password provided in either "password"" or "password_file"`)
+	}
+	return &userData, &adminOrg, nil
+}
+
+func resourceToOrgUser(d *schema.ResourceData, meta interface{}) (*govcd.OrgUser, *govcd.AdminOrg, error) {
```
lvirbalas: Short comment above the function's purpose would be nice.

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,352 @@
...
+		return err
+	}
+	err = orgUser.UpdateSimple(*userData)
+	if err != nil {
+		return err
+	}
+	return resourceVcdOrgUserRead(d, meta)
+}
+
+func resourceVcdOrgUserImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
lvirbalas: Please add a beefy function comment - similar to the ones we have for LB imports.

vcd/resource_vcd_org_user_test.go
```diff
@@ -0,0 +1,315 @@
...
+		t.Skip(acceptanceTestsSkipped)
+		return
+	}
+	cleanUserData(t)
+}
+
+func testAccCheckVcdUserDestroy(userName string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		conn := testAccProvider.Meta().(*VCDClient)
+		adminOrg, err := conn.GetAdminOrg(testConfig.VCD.Org)
```
lvirbalas: Can we run this test with both sys admin and org admin?

vcd/sample_vcd_test_config.json
```diff
@@ -120,11 +120,14 @@
 
     "//": "Leave the network names, the second storage profile, and the media name empty to skip their creation",
     "//": "They are not needed for terraform-provider-vcd testing, but are useful for go-vcloud-director",
+    "//": "Likewise, leaving the user or password empty will skip the org user creation",
```
lvirbalas: Comment became like a book right now (long). It's better to move it just above the respective lines ("orgUser" one), and make it more declarative, like "Leave the user or password empty will skip the org user creation".

website/docs/r/org_user.html.markdown
```diff
@@ -0,0 +1,165 @@
...
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to which the VDC belongs. Optional if defined at provider level.
+* `name` - (Required) A unique name for the user.
+* `password` - (Optional, but required if `password_file` was not given) The user password. This value is never returned 
+  on read. It is inspected on create and modify. On modify, the absence of this element indicates that the password 
+  should not be changed.
+* `password_file` (Optional, but required if `password` was not given). A text file containing the password. 
+  Using this property instead of `password` has the advantage that the sensitive data is not saved into Terraform state
```
lvirbalas: Need a comment on how to update after changing the password in file.

website/docs/r/org_user.html.markdown
```diff
@@ -0,0 +1,165 @@
...
+    * `Defer to Identity Provider`
+* `full_name` - (Optional) The full name of the user.
+* `description` - (Optional) An optional description of the user.
+* `telephone` - (Optional) The Org User telephone number.
+* `email_address` - (Optional) The Org User email address. Needs to be a properly formatted email address.
+* `instant_messaging` - (Optional) The Org User instant messaging.
+* `is_enabled` - (Optional) True if the user is enabled and can log in. The default is `false`.
+* `is_group_role` - (Optional) True if this user has a group role.. The default is `false`.
+* `is_locked` - (Optional) True if the user account has been locked due to too many invalid login attempts. A locked 
+  user account can be re-enabled by updating the user with this flag set to false. Only the system can set the value to
```
lvirbalas: > re-enabled by updating the user with this flag set to false

Had a comment above somewhere that this phrasing is confusing.

vcd/resource_vcd_org_user.go
dataclouder: The value is not reported by vCD, but the plan is based on the values in the configuration file.
If you change the value of the password from an existing string to empty text, the plan reports a change that will not correspond to a real change in vCD. 
The way of using a password (as opposed to a password file) is to set the password, then re-run with an empty password.
From that point on, until you change the password using Terraform, the plan will mean what it should.

Summing up: when changing the password to a different text, there will be a real change and a plan change. When changing a text to empty string (or removing the password field) there will be a change in the plan, but it will be a dud in vCD (meaning an update without changes and without errors).

vcd/resource_vcd_org_user.go
dataclouder: That case was different. We took that decision because vCD would set that property (field `advanced` in Edge Gateway) to `true` by default in 9.7+, while leaving the Boolean field out would translate to `false`. 
In this case, the default is always `false`.
I feel more comfortable mimicking vCD behavior.

vcd/resource_vcd_org_user.go
dataclouder: Correct.
Suggestions?

vcd/resource_vcd_org_user.go
dataclouder: We need this function for both Create and Update. Rather than repeating a lot of `Set` functions in two places, this function allows to keep the main functions cleaner.

vcd/resource_vcd_org_user_test.go
dataclouder: Yes. Same as we did in go-vcloud-director

vcd/sample_vcd_test_config.json
dataclouder: done

website/docs/r/org_user.html.markdown
dataclouder: done

website/docs/r/org_user.html.markdown
dataclouder: waiting for suggestions

CHANGELOG.md
dataclouder: moved

dataclouder: moved

TESTING.md
dataclouder: fixed

vcd/resource_vcd_org_user.go
dataclouder: fixed

vcd/resource_vcd_org_user.go
dataclouder: comments added

vcd/resource_vcd_org_user.go
dataclouder: done

vcd/resource_vcd_org_user.go
dataclouder: done

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,364 @@
...
+	if err != nil || adminOrg == (govcd.AdminOrg{}) {
+		return nil, fmt.Errorf(errorRetrievingOrg, orgName)
+	}
+
+	user, err := adminOrg.FetchUserByName(userName, false)
+	if err != nil {
+		return nil, govcd.ErrorEntityNotFound
+	}
+
+	err = setOrgUserData(d, user, &adminOrg)
```
vbauzysvmware: @Didainius can comment more, but according his imports it should set only a few details and user will fetch all latter details with `Refresh`

website/docs/r/org_user.html.markdown
```diff
@@ -0,0 +1,165 @@
...
+  org  = "my-org"
+  name = "my-org-admin"
+  role = "Organization Administrator"
+}
+```
+
+You can import such user into terraform state using this command
+
+```
+terraform import vcd_org_user.my-org-admin my-org.my-org-admin
```
vbauzysvmware: I am kind'a lost with this example. @lvirbalas  @Didainius  Are understanding this?

Didainius: I do, but I can be a bad example as I've played with this too much :)

vcd/resource_vcd_external_network_test.go
```diff
@@ -29,11 +29,6 @@ func TestAccVcdExternalNetworkBasic(t *testing.T) {
 		"Tags":                "network extnetwork",
 	}
 
-	if !usingSysAdmin() {
```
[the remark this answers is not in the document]
dataclouder: It was in the same function twice

vcd/resource_vcd_org_user.go
Didainius: Yeah. It's probably not necessary if`read` sets all fields as import automatically triggers `refresh` however this probably can't make it worse.

Here's example of `import` triggering `read` 
```bash
$  terraform import vcd_lb_app_rule.test3 my-org.my-org-vdc.my-edge-gw.script2
vcd_lb_app_rule.test3: Importing from ID "my-org.my-org-vdc.my-edge-gw.script2"...
vcd_lb_app_rule.test3: Import complete!
  Imported vcd_lb_app_rule
vcd_lb_app_rule.test3: Refreshing state... [id=applicationRule-42]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```

vcd/resource_vcd_external_network_test.go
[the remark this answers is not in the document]
vbauzysvmware: tx

vcd/resource_vcd_org_user.go
dataclouder: There is no user refresh

Didainius: Probably both ways are fine. `refresh` is being triggered automatically so if `id` is set `refresh` would fill in all the other fields anyway.

website/docs/r/org_user.html.markdown
dataclouder: @vbauzysvmware Please read the docs. Here's the executive summary:

Terraform import uses a combination of a **minimal definition** of a resource + a import command to retrieve data from the provider and add it to the state file.

The syntax is `terraform import resource_type.resource_name resource_path`
where `resource_path` is how we get to the resource from the top. In our case is `org_name.user_name`.
In places more complicated it could be `org_name.vdc_name.edge_name.service_name`

vcd/resource_vcd_org_user.go
dataclouder: If you know how to make such fieldDiffFunc, let me know. The ones I tried would suppress the password before it reaches the provider.

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,364 @@
...
+	userData.EmailAddress = d.Get("email_address").(string)
+	userData.Telephone = d.Get("telephone").(string)
+	userData.ProviderType = d.Get("provider_type").(string)
+	userData.IsEnabled = d.Get("is_enabled").(bool)
+	userData.IsLocked = d.Get("is_locked").(bool)
+	userData.DeployedVmQuota = d.Get("deployed_vm_quota").(int)
+	userData.StoredVmQuota = d.Get("stored_vm_quota").(int)
+	userData.IM = d.Get("instant_messaging").(string)
+
+	password := d.Get("password").(string)
```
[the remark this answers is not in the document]
dataclouder: I can add the check in code, although I can imagine that if someone removes the conflict definition, they can also remove the further check.

website/docs/r/org_user.html.markdown
vbauzysvmware: Thank you!

website/docs/r/org_user.html.markdown
```diff
@@ -0,0 +1,165 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to which the VDC belongs. Optional if defined at provider level.
+* `name` - (Required) A unique name for the user.
+* `password` - (Optional, but required if `password_file` was not given) The user password. This value is never returned 
+  on read. It is inspected on create and modify. On modify, the absence of this element indicates that the password 
+  should not be changed.
```
lvirbalas: > On modify, the absence of this element indicates that the password 
>  should not be changed.

This sentence leaves questions... Please rephrase to a more simple sentence describing workflow: how to change the password and how to not change it.

vcd/resource_vcd_org_user.go
vbauzysvmware: DiffSuppressFunc: func(k string, old string, new string, d *schema.ResourceData) bool {
return old == ""
}

vcd/resource_vcd_org_user.go
lvirbalas: That's a good catch. We did agree to use `enabled`. The only occurrence `is_enabled` is in the old Org resource. @dataclouder let's change to `enabled`.

lvirbalas: > In this case, the default is always false.
> I feel more comfortable mimicking vCD behavior.

We are defaulting all other enabled fields to true in Terraform. So, having this set to false is setting inconsistent expectations from the user. Let's make it true by default like in other places and people will have one less field to worry about.

vcd/resource_vcd_org_user.go
lvirbalas: No need to invert the logic of the field name. Suggestion could be this (please confirm however whether it is really the way it works):

"If the user account has been locked due to too many invalid login attempts, the value will change to true (only the system can set the value to true). To unlock the user re-set this flag to false."

website/docs/r/org_user.html.markdown
lvirbalas: Just posted one above.

vcd/resource_vcd_org_user.go
dataclouder: This one will prevent the password from reaching the provider, and fails on creation

vcd/resource_vcd_org_user.go
dataclouder: Confirmed. That's the way it works. Changed text accordingly

vcd/resource_vcd_org_user.go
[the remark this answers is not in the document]
dataclouder: check added

vcd/resource_vcd_org_user.go
dataclouder: done

vcd/resource_vcd_org_user.go
dataclouder: done

website/docs/r/org_user.html.markdown
dataclouder: applied

website/docs/r/org_user.html.markdown
dataclouder: done

dataclouder:
> LGTM. One ask to add: `Sensitive: true` for the password. And for the NOTE: I am not sure password handling is good enough

The Sensitive field is a good suggestion. It will prevent the password from being displayed during plan. 
It still will appear in `terraform.tfstate` and it will show a change when it becomes blank.
I believe this is all working as intended.

website/docs/r/org_user.html.markdown
```diff
@@ -0,0 +1,165 @@
...
+```hcl
+resource "vcd_org_user" "my-org-admin" {
+  org = "my-org"
+
+  name          = "my-org-admin"
+  description   = "a new org admin"
+  role          = "Organization Administrator"
+  provider_type = "INTEGRATED"
+  password      = "change-me"
+  enabled       = true
```
lvirbalas: Suggestion: remove enabled from this example, which purpose should be to show the simplest possible way of creating a user.

Also, would be good to have comments above two examples explaining the purpose (and what's the idea behind the second one).

website/docs/r/org_user.html.markdown
```diff
@@ -0,0 +1,165 @@
...
+* `is_locked` - (Optional)aIf the user account has been locked due to too many invalid login attempts, the value will 
+  change to true (only the system can lock the user). To unlock the user re-set this flag to false. 
+* `take_ownership` - (Optional) Take ownership of user's objects on deletion.
+* `deployed_vm_quota` - (Optional) Quota of vApps that this user can deploy. A value of 0 specifies an unlimited quota.
+  The default is 10.
+* `stored_vm_quota` - (Optional) Quota of vApps that this user can store. A value of 0 specifies an unlimited quota.
+  The default is 10.
+
+
+## Attribute Reference
```
lvirbalas: Extra empty line above?

website/docs/r/org_user.html.markdown
```diff
@@ -0,0 +1,165 @@
...
+
+The following attributes are exported on this resource:
+
+* `id` - The ID of the Organization user
+
+
+## Importing
```
lvirbalas: Extra empty line above?

vcd/resource_vcd_org_user.go
```diff
@@ -0,0 +1,372 @@
...
+	var userData govcd.OrgUserConfiguration
+	userData.RoleName = d.Get("role").(string)
+
+	userData.Name = d.Get("name").(string)
+	userData.Description = d.Get("description").(string)
+	userData.FullName = d.Get("full_name").(string)
+	userData.EmailAddress = d.Get("email_address").(string)
+	userData.Telephone = d.Get("telephone").(string)
+	userData.ProviderType = d.Get("provider_type").(string)
+	userData.IsEnabled = d.Get("enabled").(bool)
```
lvirbalas: It would probably be cleaner if we renamed `userData.IsEnabled` to `userData.Enabled` as well to match the adjusted resource field name?

dataclouder: you want me to open another PR in govcd for this?
No users of Terraform will see it

website/docs/r/org_user.html.markdown
dataclouder: removed

website/docs/r/org_user.html.markdown
dataclouder: removed

vcd/resource_vcd_org_user.go
lvirbalas: Forgot that this is in govcd, never mind.

PR #282, opened 2019-07-15
NAT rule support  IDs and external or Org networks

vbauzysvmware:
Ref: terraform-providers/terraform-provider-vcd#244
Ref: https://github.com/terraform-providers/terraform-provider-vcd/pull/224
Depends on: https://github.com/vmware/go-vcloud-director/pull/216

Uses nat rule Id's for handling rules. Also network_name and network_type used to handle which network to use. Old way left to be intact (allows only creating rule for first external network)


CHANGELOG.md
```diff
@@ -15,13 +15,19 @@ FEATURES:
...
 
 IMPROVEMENTS:
 * `resource/vcd_org_vdc`: Fix ignoring of resource guarantee values - [GH-265]
 * `resource/vcd_org_vdc`: Org VDC state ID changed from name to vCD ID - [GH-275]
 * Change resource handling to use locking mechanism when resource parallel handling is not supported by vCD. [GH-255] 
 * Fix issue when vApp is power cycled during member VM deletion. [GH-261]
 
+BUG FIXES:
+
+* `resource/vcd_dnat and resource/vcd_snat` - delete works 100% if you use network_name and type [GH-282]
```
Didainius: Maybe we need a bit more context here. For someone not knowing the problem it wouldn't make much sense.

vcd/resource_vcd_dnat.go
```diff
@@ -167,27 +168,53 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
...
 	vcdClient := meta.(*VCDClient)
 
-	e, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
+	edgeGateway, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
 
 	if err != nil {
 		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
 	}
 
+	// Terraform refresh won't work if Rule was edit in UI. vCD API uses tag elements to map edge gtw Id's
```
Didainius: If I understand correctly if this is lost we will actually mark the resource as deleted - `d.SetId("")`. I think it's important to note that (especially in website documentation)

vcd/resource_vcd_dnat.go
```diff
@@ -167,27 +168,53 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
...
+
+		orgVdcNetwork, _ := getOrgVdcNetwork(d, vcdClient, natRule.GatewayNatRule.Interface.Name)
+		if orgVdcNetwork != nil {
+			d.Set("network_type", "org")
+		} else {
+			externalNetwork, _ := govcd.GetExternalNetwork(vcdClient.VCDClient, natRule.GatewayNatRule.Interface.Name)
+			if externalNetwork != nil && externalNetwork != (&govcd.ExternalNetwork{}) {
+				d.Set("network_type", "ext")
+			} else {
+				return fmt.Errorf("issue to find network")
```
Didainius: A bit more context would help - network name, network type

vcd/resource_vcd_dnat.go
```diff
@@ -220,31 +247,96 @@ func resourceVcdDNATDelete(d *schema.ResourceData, meta interface{}) error {
...
-		portString,
-		d.Get("internal_ip").(string),
-		translatedPortString)
+
+	natRule, err := edgeGateway.FetchNatRule(d.Id())
 	if err != nil {
-		return fmt.Errorf("error setting DNAT rules: %#v", err)
+		log.Printf("Error: Nat rule isn't found")
+		d.SetId("")
+		return err
```
Didainius: Could this error give more context? I've checked `FetchNatRule` and it doesn't return the ID or more details as well.

vcd/resource_vcd_dnat_test.go
```diff
@@ -14,14 +15,14 @@ import (
 var baseDnatName string = "TestAccVcdDNAT"
 var orgVdcNetworkName = "TestAccVcdDNAT_BasicNetwork"
 
-func TestAccVcdDNAT_Basic(t *testing.T) {
+func TestAccVcdDNAT_WithOrgNetw(t *testing.T) {
```
Didainius: Would be nice to test "update" as well.

vcd/resource_vcd_dnat_test.go
```diff
@@ -47,45 +49,52 @@ func TestAccVcdDNAT_Basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr(
+						"vcd_dnat."+dnatName, "description", "test run1"),
 				),
 			},
 		},
 	})
 }
 
-func TestAccVcdDNAT_tlate(t *testing.T) {
+func TestAccVcdDNAT_WithExtNetw(t *testing.T) {
```
Didainius: Would be nice to test "update" as well.

vcd/resource_vcd_snat.go
```diff
@@ -119,23 +124,44 @@ func resourceVcdSNATCreate(d *schema.ResourceData, meta interface{}) error {
...
 	vcdClient := meta.(*VCDClient)
 
-	e, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
+	edgeGateway, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
 	if err != nil {
 		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
 	}
 
+	// Terraform refresh won't work if Rule was edit in UI. vCD API uses tag elements to map edge gtw Id's
+	// and UI will reset the tag element on update.
```
Didainius: Same thing on note about removed object as in DNAT.

vcd/resource_vcd_snat.go
```diff
@@ -119,23 +124,44 @@ func resourceVcdSNATCreate(d *schema.ResourceData, meta interface{}) error {
...
-		for _, r := range e.EdgeGateway.Configuration.EdgeGatewayServiceConfiguration.NatService.NatRule {
-			if r.RuleType == "SNAT" && r.GatewayNatRule.TranslatedIP == d.Get("internal_ip").(string) && r.GatewayNatRule.Interface.Name == d.Get("network_name").(string) {
-				found = true
-				d.Set("external_ip", r.GatewayNatRule.OriginalIP)
+	networkName := d.Get("network_name")
+	if nil != networkName && networkName.(string) != "" {
+		natRule, err := edgeGateway.FetchNatRule(d.Id())
+		if err != nil {
+			d.SetId("")
+			return err
```
Didainius: A bit more context would help the error as this is the one going for user.

vcd/resource_vcd_snat.go
```diff
@@ -119,23 +124,44 @@ func resourceVcdSNATCreate(d *schema.ResourceData, meta interface{}) error {
...
+
+		orgVdcNetwork, _ := getOrgVdcNetwork(d, vcdClient, natRule.GatewayNatRule.Interface.Name)
+		if orgVdcNetwork != nil {
+			d.Set("network_type", "org")
+		} else {
+			externalNetwork, _ := govcd.GetExternalNetwork(vcdClient.VCDClient, natRule.GatewayNatRule.Interface.Name)
+			if externalNetwork != nil && externalNetwork != (&govcd.ExternalNetwork{}) {
+				d.Set("network_type", "ext")
+			} else {
+				return fmt.Errorf("issue to find network")
```
Didainius: More context would help (network name, network type)

vcd/resource_vcd_snat.go
```diff
@@ -164,16 +187,77 @@ func resourceVcdSNATDelete(d *schema.ResourceData, meta interface{}) error {
...
-		return fmt.Errorf("error setting SNAT rules: %#v", err)
+		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
 	}
-	err = task.WaitTaskCompletion()
+
+	natRule, err := edgeGateway.FetchNatRule(d.Id())
 	if err != nil {
+		log.Printf("Error: Nat rule isn't found")
+		d.SetId("")
 		return err
```
Didainius: Would like to see more context rule ID

website/docs/r/snat.html.markdown
```diff
@@ -28,5 +30,12 @@ The following arguments are supported:
...
 * `internal_ip` - (Required) The IP or IP Range of the VM(s) to map from
+* `network_type` - (Optional; *v2.4+*) Type of the network on which to apply the NAT rule. *`network_type` will be a required field in the next major version.*
+* `network_name` - (Optional; *v2.4+*) The name of the network on which to apply the SNAT. *`network_name` will be a required field in the next major version.*
 * `org` - (Optional; *v2.0+*) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
 * `vdc` - (Optional; *v2.0+*) The name of VDC to use, optional if defined at provider level
+
+## NOTE
+
+When advanced edge gateway is used and rule is updated using UI, then Id mapping will be lost 
+and terraform won't find rule anymore.
```
Didainius: I'd find explicit note that it will be removed from statefile useful. Would also put it as the other notes we put in the top of the page as seldom someone scrolls down to the bottom :)

website/docs/r/dnat.html.markdown
```diff
@@ -49,5 +52,12 @@ The following arguments are supported:
...
 * `protocol` - (Optional; *v2.0+*) The protocol type. Possible values are TCP, UDP, TCPUDP, ICMP, ANY. TCP is default to be backward compatible with previous version
 * `icmp_sub_type` - (Optional; *v2.0+*) The name of ICMP type. Possible values are   address-mask-request, destination-unreachable, echo-request, echo-reply, parameter-problem, redirect, router-advertisement, router-solicitation, source-quench, time-exceeded, timestamp-request, timestamp-reply, any
+* `network_type` - (Optional; *v2.4+*) Type of the network on which to apply the NAT rule. *`network_type` will be a required field in the next major version.*
+* `network_name` - (Optional; *v2.4+*) The name of the network on which to apply the SNAT. *`network_name` will be a required field in the next major version.*
 * `org` - (Optional; *v2.0+*) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
 * `vdc` - (Optional; *v2.0+*) The name of VDC to use, optional if defined at provider level
+
+## NOTE
+
+When advanced edge gateway is used and rule is updated using UI, then Id mapping will be lost
```
Didainius: I'd find explicit note that it will be removed from statefile useful. Would also put it as the other notes we put in the top of the page as seldom someone scrolls down to the bottom :)

website/docs/r/dnat.html.markdown
```diff
@@ -49,5 +52,12 @@ The following arguments are supported:
 * `internal_ip` - (Required) The IP of the VM to map to
 * `protocol` - (Optional; *v2.0+*) The protocol type. Possible values are TCP, UDP, TCPUDP, ICMP, ANY. TCP is default to be backward compatible with previous version
 * `icmp_sub_type` - (Optional; *v2.0+*) The name of ICMP type. Possible values are   address-mask-request, destination-unreachable, echo-request, echo-reply, parameter-problem, redirect, router-advertisement, router-solicitation, source-quench, time-exceeded, timestamp-request, timestamp-reply, any
+* `network_type` - (Optional; *v2.4+*) Type of the network on which to apply the NAT rule. *`network_type` will be a required field in the next major version.*
```
Didainius: Should we mention options?

website/docs/r/snat.html.markdown
```diff
@@ -28,5 +30,12 @@ The following arguments are supported:
 * `edge_gateway` - (Required) The name of the edge gateway on which to apply the SNAT
 * `external_ip` - (Required) One of the external IPs available on your Edge Gateway
 * `internal_ip` - (Required) The IP or IP Range of the VM(s) to map from
+* `network_type` - (Optional; *v2.4+*) Type of the network on which to apply the NAT rule. *`network_type` will be a required field in the next major version.*
```
Didainius: Should we mention options?

vcd/resource_vcd_snat.go
```diff
@@ -164,16 +187,77 @@ func resourceVcdSNATDelete(d *schema.ResourceData, meta interface{}) error {
...
+}
+
+func resourceVcdSNATUpdate(d *schema.ResourceData, meta interface{}) error {
+
+	vcdClient := meta.(*VCDClient)
+
+	// Update supports only when network name and network type provided
+	networkName := d.Get("network_name")
+	if nil == networkName || networkName.(string) == "" {
+		return fmt.Errorf("update works only when network_name and network_type is provided and rule created using them \n")
```
Didainius: This is a wider team discussion but we'll have to think how to make it less hard to migrate for that major version 3.0. At least imports could help or could we handle somehow internally and populate the variables based on the data so that user has migration/update path.
Not intending to solve it here, just flagging.
CC @dataclouder, @lvirbalas

vcd/resource_vcd_dnat.go
vbauzysvmware: Sorry, didn't get question...

Didainius: I mean the `return err`. This error goes to user but it doesn't contain any context. Like nat rule ID or something more.

Didainius: Same applies to log.Printf("Error: Nat rule isn't found"). It doesn't give answer to "which nat rule isn't found"

vcd/resource_vcd_snat.go
vbauzysvmware: It's a little complicated. For user it will mislead. Cause ind standard edge gtw you won't see Id's in advanced edge gtw it won't match. vCD uses <tags> element to map internal mapping between NSX rule mapping. So it will mislead exposing ID cause user won't find anywhere to match it in UI.

CHANGELOG.md
vbauzysvmware: Changed

vcd/resource_vcd_dnat.go
vbauzysvmware: Changed

vcd/resource_vcd_dnat.go
vbauzysvmware: I explained in other Tread

website/docs/r/dnat.html.markdown
vbauzysvmware: Changed

website/docs/r/snat.html.markdown
vbauzysvmware: Changed

vcd/resource_vcd_dnat.go
vbauzysvmware: Well in reality this though we use d.SetId("") but return err, so resource won't be removed. This kind'a global issue in our codebase. We do it such behaviour randomly. We aren't consistent. @lvirbalas @dataclouder Seems we need to decided if we don't find something - do throw error and leave user to handle it or do we remove automatically and happy with it.

Didainius: Yes. I know we usually do this, but this is a bit different because user action can make the rule `invisible` to provider and then we will automatically remove it.

vbauzysvmware: So options are: throw err as we do know and leave state as it is.
Or Print message and remove resource.

vcd/resource_vcd_dnat_test.go
vbauzysvmware: Added

vcd/resource_vcd_snat.go
lvirbalas: Would it match what is in Terraform state file? If yes, it would still be useful to print.

vbauzysvmware: Yes, it would match state file ID

vcd/resource_vcd_dnat.go
lvirbalas: Need more info - pros/cons of one approach vs. the other.

Didainius: My initial point was to simply flag it for user (note at the top of resource page) that it can happen so. 


> I think it's important to note that (especially in website documentation)

vcd/resource_vcd_snat.go
lvirbalas: Then I'd print it. If you think it may be misleading, maybe say it's "internal NSX ID" or something like that.

vcd/resource_vcd_dnat.go
vbauzysvmware: So:
if leave it - we leave user handle state manually. If delete it - user can't do anything if he may fix issue in env he has to add resource back manually or with import when it exist.

vcd/resource_vcd_dnat.go
```diff
@@ -167,27 +168,53 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
...
 	vcdClient := meta.(*VCDClient)
 
-	e, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
+	edgeGateway, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
 
 	if err != nil {
 		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
 	}
 
+	// Terraform refresh won't work if Rule was edit in advanced edge gateway UI. vCD API uses tag elements to map edge gtw Id's
```
dataclouder: "Id's" -> "IDs"

vcd/resource_vcd_dnat.go
```diff
@@ -220,31 +247,96 @@ func resourceVcdDNATDelete(d *schema.ResourceData, meta interface{}) error {
...
+	return nil
+}
+
+func resourceVcdDNATUpdate(d *schema.ResourceData, meta interface{}) error {
 
+	vcdClient := meta.(*VCDClient)
+
+	// Update supports only when network name and network type provided
+	networkName := d.Get("network_name")
+	if nil == networkName || networkName.(string) == "" {
```
dataclouder: [Yoda notation](https://en.wikipedia.org/wiki/Yoda_conditions) is not encouraged in Go. Besides, it should be consistent : either always `"literal" == variable` or always `variable == "literal"`. The second one is recommended.
If you run staticcheck, it will show you all the occurrences that should be addressed.
```
resource_vcd_dnat.go:114:5: don't use Yoda conditions (ST1017)
resource_vcd_dnat.go:114:26: don't use Yoda conditions (ST1017)
resource_vcd_dnat.go:128:12: don't use Yoda conditions (ST1017)
resource_vcd_dnat.go:128:33: don't use Yoda conditions (ST1017)
resource_vcd_dnat.go:160:5: don't use Yoda conditions (ST1017)
resource_vcd_dnat.go:254:5: don't use Yoda conditions (ST1017)
resource_vcd_external_network.go:240:45: don't use Yoda conditions (ST1017)
resource_vcd_external_network.go:244:45: don't use Yoda conditions (ST1017)
resource_vcd_external_network.go:248:51: don't use Yoda conditions (ST1017)
resource_vcd_independent_disk.go:192:5: don't use Yoda conditions (ST1017)
resource_vcd_independent_disk.go:201:5: don't use Yoda conditions (ST1017)
resource_vcd_snat.go:78:5: don't use Yoda conditions (ST1017)
resource_vcd_snat.go:78:26: don't use Yoda conditions (ST1017)
resource_vcd_snat.go:89:12: don't use Yoda conditions (ST1017)
resource_vcd_snat.go:89:33: don't use Yoda conditions (ST1017)
resource_vcd_snat.go:115:5: don't use Yoda conditions (ST1017)
resource_vcd_snat.go:190:5: don't use Yoda conditions (ST1017)
resource_vcd_vapp_vm.go:460:33: don't use Yoda conditions (ST1017)
```

vcd/resource_vcd_dnat.go
dataclouder: What @Didainius is suggesting is return more context, such as `return fmt.Errorf(" rule %s not found: %s", d.Id(), err)`

vcd/resource_vcd_dnat.go
vbauzysvmware: Fixed

vcd/resource_vcd_dnat.go
vbauzysvmware: Fixed for this PR context

vcd/resource_vcd_dnat.go
vbauzysvmware: I understood what Dainius suggested, my concern was other thing. Overall I pushed change, what logging and resouce is removed from state if not found.

vcd/resource_vcd_dnat.go
vbauzysvmware: changed to remove from state

vcd/resource_vcd_snat.go
vbauzysvmware: fixed

vcd/resource_vcd_snat.go
vbauzysvmware: updated

vcd/resource_vcd_snat.go
vbauzysvmware: updated

vcd/resource_vcd_snat.go
vbauzysvmware: updated

website/docs/r/snat.html.markdown
vbauzysvmware: updated

website/docs/r/dnat.html.markdown
vbauzysvmware: updated

CHANGELOG.md
```diff
@@ -15,13 +15,19 @@ FEATURES:
...
 
 IMPROVEMENTS:
 * `resource/vcd_org_vdc`: Fix ignoring of resource guarantee values - [GH-265]
 * `resource/vcd_org_vdc`: Org VDC state ID changed from name to vCD ID - [GH-275]
 * Change resource handling to use locking mechanism when resource parallel handling is not supported by vCD. [GH-255] 
 * Fix issue when vApp is power cycled during member VM deletion. [GH-261]
 
+BUG FIXES:
+
+* `resource/vcd_dnat and resource/vcd_snat` - fix resource destroy as it would still leave nat rule in edge gateway. Fix works if network_name and network_type is used. [GH-282]
```
lvirbalas: Nit: "nat rule" -> "NAT rule"

vcd/resource_vcd_dnat.go
```diff
@@ -35,49 +38,46 @@ func resourceVcdDNAT() *schema.Resource {
...
 				Required: true,
-				ForceNew: true,
 			},
 
 			"port": &schema.Schema{
 				Type:     schema.TypeInt,
-				Required: true,
-				ForceNew: true,
+				Optional: true,
+				Computed: true,
```
lvirbalas: https://www.terraform.io/docs/extend/schemas/schema-behaviors.html states:

> Computed is often used to represent values that are **not user configurable** or can not be known at time of terraform plan or apply

Why are we changing "port", "translated_port" and "icmp_sub_type" to Computed when they seem to be casual user-settable parameters?

vbauzysvmware: Cause api still returns value of you haven't set. Computed allows for user to set and be provided by external means.

CHANGELOG.md

vcd/resource_vcd_dnat.go
lvirbalas: Interesting and I see more occurrences of this approach in the code base, but it looks like it conflicts with the quote from Terraform above. @Didainius @dataclouder what do you think?

Didainius: Hm. It's kind of mixed feeling from me. On one way if the user didn't set it - then he probably doesn't need it. On the other hand it wouldn't hurt him. I haven't yet played with installed binary. Does it complain about diff being different on `terraform plan`? If it doesn't then it probably shouldn't be bad.

vbauzysvmware: It is due the reason we read and set everything to state file - for user to see state uptodate. As so `terraform plan` will complain cause user haven't set, but state file has. In similar approach Org Vdc was configured and do things.

website/docs/r/snat.html.markdown
```diff
@@ -28,5 +32,7 @@ The following arguments are supported:
 * `edge_gateway` - (Required) The name of the edge gateway on which to apply the SNAT
 * `external_ip` - (Required) One of the external IPs available on your Edge Gateway
 * `internal_ip` - (Required) The IP or IP Range of the VM(s) to map from
+* `network_type` - (Optional; *v2.4+*) Type of the network on which to apply the NAT rule. Possible values org or ext. *`network_type` will be a required field in the next major version.*
```
Didainius: Can we briefly mention the default here.

website/docs/r/dnat.html.markdown
```diff
@@ -49,5 +54,7 @@ The following arguments are supported:
 * `internal_ip` - (Required) The IP of the VM to map to
 * `protocol` - (Optional; *v2.0+*) The protocol type. Possible values are TCP, UDP, TCPUDP, ICMP, ANY. TCP is default to be backward compatible with previous version
 * `icmp_sub_type` - (Optional; *v2.0+*) The name of ICMP type. Possible values are   address-mask-request, destination-unreachable, echo-request, echo-reply, parameter-problem, redirect, router-advertisement, router-solicitation, source-quench, time-exceeded, timestamp-request, timestamp-reply, any
+* `network_type` - (Optional; *v2.4+*) Type of the network on which to apply the NAT rule. Possible values org or ext. *`network_type` will be a required field in the next major version.*
```
Didainius: Can we briefly mention the default here.

website/docs/r/snat.html.markdown
```diff
@@ -11,11 +11,15 @@ description: |-
 Provides a vCloud Director SNAT resource. This can be used to create, modify,
 and delete source NATs to allow vApps to send external traffic.
 
+!> **Warning:** When advanced edge gateway is used and the rule is updated using UI, then Id mapping will be lost and terraform won't find the rule anymore and remove it from state.
```
Didainius: This catches attention nicely. 
image

vcd/resource_vcd_dnat.go
Didainius: Are you sure it does? I think acceptance tests would be failing (state not clean) if it was so. Need to validate it. The `terraform plan` after `terraform apply` should report 0 changes. Otherwise we'll get bugs reported, we got that for networks constantly updating IPs previously.

vcd/resource_vcd_snat_test.go
```diff
@@ -63,6 +67,20 @@ func TestAccVcdSNAT_Basic(t *testing.T) {
 						"vcd_snat."+snatName, "internal_ip", startIpAddress),
 				),
 			},
+			resource.TestStep{
+				Config: updateText,
+				// TODO remove with Terraform 0.12 support
```
Didainius: We are at Terraform 0.12. Maybe we can remove these comments?

vcd/resource_vcd_dnat.go
vbauzysvmware: Definition of computed:
// The fields below relate to diffs.
//
// If Computed is true, then the result of this value is computed
// (unless specified by config) on creation.

lvirbalas: Thank you. I got tricked by the doc page above. Your pasted description from Terraform's schema.go file is much more correct. "unless specified by config" is the key here.

website/docs/r/snat.html.markdown
vbauzysvmware: no default here

website/docs/r/dnat.html.markdown
vbauzysvmware: no default here

website/docs/r/snat.html.markdown
vbauzysvmware: nice

vcd/resource_vcd_snat_test.go
vbauzysvmware: removed comments

vbauzysvmware:
Bump to tagged version of govcd

Didainius:
I have `ext` and `org` networks with the same name.
This test fails in such case:
```
$ go test -tags functional -v -run "TestAccVcdDNAT_WithExtNetw" .
Connecting to https://192.168.1.109/api
as user administrator@System
Checking resources to create for test suite...
Skipping catalog creation - found preconfigured one: my-catalog
Catalog found successfully
Skipping catalog item creation - found preconfigured one: photon-os
=== RUN   TestAccVcdDNAT_WithExtNetw
--- FAIL: TestAccVcdDNAT_WithExtNetw (39.59s)
    testing.go:561: Step 1, expected error:

        Check failed: Check 3/8 error: vcd_dnat.TestAccVcdDNAT_WithExtNetw: Attribute 'network_type' expected "ext", got "org"

        To match:

        After applying this step and refreshing, the plan was not empty:


FAIL
```

vbauzysvmware:
> TestAccVcdDNAT_WithExtNetw

Why you have them? what is expectation here?

Didainius:
> > TestAccVcdDNAT_WithExtNetw
> 
> Well. Resource allows to change network type ext or org - so when I read/refresh I need to check which type it is. Now firstly search for org when for ext network if not found. I think it's ok - if anyone see value, so only one additional approach I see is throw error with complain that org and ext networks with same name exists.

Could we lookup the network based on it's type? If it was an `ext` then it should find and try to use `ext`?

vbauzysvmware:
> > > TestAccVcdDNAT_WithExtNetw
> > 
> > 
> > Well. Resource allows to change network type ext or org - so when I read/refresh I need to check which type it is. Now firstly search for org when for ext network if not found. I think it's ok - if anyone see value, so only one additional approach I see is throw error with complain that org and ext networks with same name exists.
> 
> Could we lookup the network based on it's type? If it was an `ext` then it should find and try to use `ext`?
e.g.
With terraform we created rule id 6565 `org` with `name1`. Using UI we changed rule 6565 tu use `ext` with `name2`. And in your case you have also created create network `org` with name `name2`. So type can be changed too.

vcd/resource_vcd_dnat.go
```diff
@@ -167,27 +168,56 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
...
+		if orgVdcNetwork != nil {
+			d.Set("network_type", "org")
+			found = true
+		}
+		externalNetwork, extNetwErr := govcd.GetExternalNetwork(vcdClient.VCDClient, natRule.GatewayNatRule.Interface.Name)
+		if extNetwErr != nil && strings.Contains(extNetwErr.Error(), "could not find external network named") {
+			d.Set("network_type", "ext")
+			found = true
+		}
+		if orgVdcNetwork != nil && extNetwErr == nil {
```
lvirbalas: A code readability suggestion:

`if orgVdcNetwork != nil && extNetwErr == nil {`
->
`if orgVdcNetwork != nil && externalNetwork != nil {`

vbauzysvmware: I can't if you check what returns GetExternalNet - it's a mess. So after spending a lot time, I found easiest way to match what  I need is err.

lvirbalas: I haven't realized that. Do I get correctly, that `GetExternalNetwork` may return some string that is **not** extnerla network in case of error?

vbauzysvmware: It can return empty ExternalNetwork{} or govcd.ExternalNetwork with empty types.ExternalNetwork or something else or error or error with message not found

lvirbalas: OK, marking this conversation as resolved. CC @dataclouder to have this in mind in the context of `Get` functions "harmonization".

vcd/resource_vcd_dnat.go
```diff
@@ -167,27 +168,56 @@ func resourceVcdDNATCreate(d *schema.ResourceData, meta interface{}) error {
...
+			return fmt.Errorf("find external network or org VCD network with same name: %s", natRule.GatewayNatRule.Interface.Name)
```
Didainius:
```suggestion
			return fmt.Errorf("found external network or org VCD network with same name: %s", natRule.GatewayNatRule.Interface.Name)
```

vbauzysvmware: changed

PR #283, opened 2019-07-18
Add Org data source

dataclouder:
* Change usage of GetOrg* methods to new ones
* Fix Org creation and update to handle task
* Fix Org defaults for deployed_vm_quota and stored_vm_quota.
* Add better tests for Org
* Add support for Org importing
* Add Org data source
* Add data source tests
* Improve Org docs
* Add docs for Org data source
* Add more tests for OrgUser (using Org data source)


vcd/config.go
```diff
@@ -195,68 +195,68 @@ func (cli *VCDClient) getVdcName(d *schema.ResourceData) string {
 // GetOrgAndVdc finds a pair of org and vdc using the names provided
 // in the args. If the names are empty, it will use the default
 // org and vdc names from the provider.
-func (cli *VCDClient) GetOrgAndVdc(orgName, vdcName string) (org govcd.Org, vdc govcd.Vdc, err error) {
+func (cli *VCDClient) GetOrgAndVdc(orgName, vdcName string) (org *govcd.Org, vdc govcd.Vdc, err error) {
```
vbauzysvmware: Why it returns only Org as pointer?

vcd/config.go
```diff
@@ -195,68 +195,68 @@ func (cli *VCDClient) getVdcName(d *schema.ResourceData) string {
...
 	if org.AdminOrg.Name == "" || org.AdminOrg.HREF == "" || org.AdminOrg.ID == "" {
-		return govcd.AdminOrg{}, fmt.Errorf("empty org %s found", orgName)
+		return nil, fmt.Errorf("empty org %s found", orgName)
 	}
 	return org, err
 }
 
 // Same as GetOrgAndVdc, but using data from the resource, if available.
-func (cli *VCDClient) GetOrgAndVdcFromResource(d *schema.ResourceData) (org govcd.Org, vdc govcd.Vdc, err error) {
+func (cli *VCDClient) GetOrgAndVdcFromResource(d *schema.ResourceData) (org *govcd.Org, vdc govcd.Vdc, err error) {
```
vbauzysvmware: same here: Why it returns only Org as pointer?

vcd/config.go
```diff
@@ -195,68 +195,68 @@ func (cli *VCDClient) getVdcName(d *schema.ResourceData) string {
...
 	orgName := d.Get("org").(string)
 	vdcName := d.Get("vdc").(string)
 	return cli.GetOrgAndVdc(orgName, vdcName)
 }
 
 // Same as GetOrgAndVdc, but using data from the resource, if available.
-func (cli *VCDClient) GetAdminOrgFromResource(d *schema.ResourceData) (org govcd.AdminOrg, err error) {
+func (cli *VCDClient) GetAdminOrgFromResource(d *schema.ResourceData) (org *govcd.AdminOrg, err error) {
```
vbauzysvmware: same here: Why it returns only Org as pointer?

vcd/datasource_vcd_org.go
```diff
@@ -0,0 +1,78 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Organization name for lookup",
+			},
+			"full_name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
vbauzysvmware: All other attributes should be Computed, not optional and forcenew. Cause read is done only by name

vcd/datasource_vcd_org.go
```diff
@@ -0,0 +1,78 @@
...
+			},
+		},
+	}
+}
+
+func datasourceVcdOrgRead(d *schema.ResourceData, meta interface{}) error {
+	vcdClient := meta.(*VCDClient)
+
+	identifier := d.Get("name").(string)
+	log.Printf("Reading Org with id %s", d.Get("name").(string))
```
vbauzysvmware: `identifier` can be used to print

vcd/datasource_vcd_org.go
```diff
@@ -0,0 +1,78 @@
...
+	adminOrg, err := vcdClient.VCDClient.GetAdminOrgByNameOrId(identifier)
+
+	if err != nil {
+		log.Printf("Org with id %s not found. Setting ID to nothing", identifier)
```
vbauzysvmware: Should return error, cause this datasource will fail silently and user won't see issue, only that referencing data is strange.

vcd/datasource_vcd_org.go
```diff
@@ -0,0 +1,78 @@
...
+			"delay_after_power_on_seconds": &schema.Schema{
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "Specifies this organization's default for virtual machine boot delay after power on.",
+			},
+		},
+	}
+}
+
+func datasourceVcdOrgRead(d *schema.ResourceData, meta interface{}) error {
```
vbauzysvmware: What about org Admin - not sys admin, this still work?

vcd/datasource_vcd_org_test.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+// Cloning an organization using an existing organization as data source
+func TestAccVcdDatasourceOrg(t *testing.T) {
+
+	if !usingSysAdmin() {
```
vbauzysvmware: I am not sure, but for me seems that should be valid case where admin org user still should be able access orgs within his rights.

vcd/resource_vcd_org.go
```diff
@@ -53,24 +54,25 @@ func resourceOrg() *schema.Resource {
 				Description: "True if this organization is enabled (allows login and all other operations).",
 			},
 			"deployed_vm_quota": &schema.Schema{
-				Type:        schema.TypeInt,
-				Optional:    true,
-				Default:     -1,
-				Description: "Maximum number of virtual machines that can be deployed simultaneously by a member of this organization.",
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      0,
```
vbauzysvmware: Doesn't we released that? and this change impacts back compatibility?

vcd/resource_vcd_org.go
```diff
@@ -148,8 +159,8 @@ func resourceOrgDelete(d *schema.ResourceData, m interface{}) error {
 
 	//fetches org
 	log.Printf("Reading Org with id %s", d.State().ID)
-	org, err := govcd.GetAdminOrgByName(vcdClient.VCDClient, d.Get("name").(string))
-	if err != nil || org == (govcd.AdminOrg{}) {
+	org, err := vcdClient.VCDClient.GetAdminOrgByName(d.Get("name").(string))
```
vbauzysvmware: I think we technical debt issue to cover this `vcdClient.VCDClient` ? what do you think? :)

vcd/resource_vcd_org.go
```diff
@@ -171,51 +182,120 @@ func resourceOrgUpdate(d *schema.ResourceData, m interface{}) error {
...
+
+	return err
+}
+
+// Retrieves an Org resource from vCD
 func resourceOrgRead(d *schema.ResourceData, m interface{}) error {
 	vcdClient := m.(*VCDClient)
 
+	identifier := d.State().ID
 	log.Printf("Reading Org with id %s", d.State().ID)
```
vbauzysvmware: `identifier` should be used

vcd/resource_vcd_org.go
```diff
@@ -171,51 +182,120 @@ func resourceOrgUpdate(d *schema.ResourceData, m interface{}) error {
...
-	return nil
+	log.Printf("Org with id %s found", identifier)
+	d.SetId(adminOrg.AdminOrg.ID)
+	return setOrgData(d, adminOrg)
+}
+
+// Imports an Org into Terraform state
+// This function task is to get the data from vCD and fill the resource data container
+// Expects the d.Id() to be a Org name
+//
```
vbauzysvmware: Example of import path missing

vcd/resource_vcd_vapp.go
```diff
@@ -416,7 +416,7 @@ func resourceVcdVAppRead(d *schema.ResourceData, meta interface{}) error {
 
 		if newIp != "allocated" {
 			log.Printf("[DEBUG] IP is assigned. Lets get it (%s)", d.Get("ip"))
-			ip, err = getVAppIPAddress(d, meta, vdc, org)
+			ip, err = getVAppIPAddress(d, meta, vdc, *org)
```
vbauzysvmware: Why only Org as pointer?

website/docs/d/org.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+}
+
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) Org name
+* `full_name` - (Optional) Org full name
```
vbauzysvmware: This is general question for LB documentation too. We saying attributes `optional`, but in reality if user would provide we don't care cause we fetch only by name, but still may don't match what we get from vCD ? ^^ @Didainius

CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
 ## 2.5.0 (Unreleased)
+
+FEATURES:
+
+* **New Data Source:** `vcd_org` Organization - ([#218](https://github.com/terraform-providers/terraform-provider-vcd/issues/218))
```
vbauzysvmware: Change log misses data on improvements (like `-1` -> `0`), new import ability, others.

vcd/config.go
dataclouder: As mentioned in the corresponding govcd PR, I am replacing structures with pointers as I implement the Get* methods. This function will be modified again when I implement the VDC `Get*`

vcd/config.go
dataclouder: see above

vcd/config.go
dataclouder: ditto

vcd/resource_vcd_vapp.go
dataclouder: see above

vcd/resource_vcd_org.go
dataclouder: Please explain.

vcd/resource_vcd_org.go
dataclouder: No. Using `-1` was wrong. It's a bug.

`-1` is an illegal value, which was silently converted to 0. It will impact compatibility, but in the way that a bug does. We are fixing what was wrong in the first place.
Also the GUI does not accept `-1`, and the API gives an error on such a value.

vcd/resource_vcd_org.go
vbauzysvmware: Does this vcdClient.VCDClient look nice for you? :)

dataclouder: It doesn't, but then there are many ugly things in this repository.

What do you suggest?
And should it be done in this PR?

vcd/datasource_vcd_org_test.go
dataclouder: The test creates another Org. For this, it requires Syestem Admin rights

vcd/datasource_vcd_org.go
dataclouder: It does. Try the test `TestAccVcdOrgUserWithDS` as org admin: it will work.

CHANGELOG.md
dataclouder: done

vcd/datasource_vcd_org.go
dataclouder: done

vcd/datasource_vcd_org.go
dataclouder: done

vcd/datasource_vcd_org.go
dataclouder: fixed

vcd/resource_vcd_org.go
dataclouder: done. The import path is just the org name

vcd/resource_vcd_org.go
vbauzysvmware: No don't do in this PR. Maybe we should or have Issue with such notes regarding what to fix?

vcd/resource_vcd_org.go
```diff
@@ -171,51 +182,119 @@ func resourceOrgUpdate(d *schema.ResourceData, m interface{}) error {
...
+	return err
+}
+
+// Retrieves an Org resource from vCD
 func resourceOrgRead(d *schema.ResourceData, m interface{}) error {
 	vcdClient := m.(*VCDClient)
 
-	log.Printf("Reading Org with id %s", d.State().ID)
-	org, err := govcd.GetAdminOrgByName(vcdClient.VCDClient, d.Get("name").(string))
+	identifier := d.State().ID
```
Didainius: Wondering if there is any difference between `d.Id()` we regularly use.

website/docs/d/org.html.markdown
Didainius: I left only the fields which are allowed to be input in `Argument` section. Everything else should probably in `Attribute reference` section. I have left it so for load balancer components.
image

Here's a better example how others do it - https://www.terraform.io/docs/providers/vsphere/d/compute_cluster.html

vcd/resource_vcd_org.go
```diff
@@ -171,51 +182,119 @@ func resourceOrgUpdate(d *schema.ResourceData, m interface{}) error {
...
+	log.Printf("Reading Org with id %s", identifier)
+	adminOrg, err := vcdClient.VCDClient.GetAdminOrgByNameOrId(identifier)
```
Didainius: Wouldn't there be problems after upgrade to this version. Here's the scenario.

* Before this PR the `Create` method used `d.SetId(task.Task.ID[15:])` and user has his resource created this way and stored in statefile
* User updates the provider to newer version with this PR.
* User does a `refresh` but the above function `vcdClient.VCDClient.GetAdminOrgByNameOrId(identifier)` does not find the resources because the `identifier` was  `task.Task.ID[15:]`, no the real `adminOrg.AdminOrg.ID`
* This read method gets an `err` and removes the resource from statefile `d.SetId("")`

dataclouder: Probably yes, but this was clearly a bug for many reasons. 
The best we can do is adding a note to the change log, saying that a bug was storing the wrong ID for orgs, and a refresh on existing resources will cause a reload.

However, the problem should be mitigated by the possibility of importing

Didainius: I agree this was bad. As for user friendliness -  not sure how safe it would be, but what if we tried to lookup by name if the ‘id’ is of old type (if we can distinguish between them) or we did not find using the existing ‘id’

dataclouder: I re-wrote several functions in Org resource to address this problem as you suggest and I added a warning and a practical guide to upgrade in the documentation.

vcd/resource_vcd_org.go
dataclouder: there isn't. I changed to using it. I actually cleaned up the whole function

vcd/provider.go
```diff
@@ -112,6 +112,7 @@ func Provider() terraform.ResourceProvider {
 		},
 
 		DataSourcesMap: map[string]*schema.Resource{
+			"vcd_org":                datasourceVcdOrg(),              // 2.4
```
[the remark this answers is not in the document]
dataclouder: fixed

dataclouder:
Updated data source docs

website/docs/r/org.html.markdown
```diff
@@ -3,16 +3,21 @@ layout: "vcd"
...
 # vcd\_org
 
-Provides a vCloud Director Org resource. This can be used to create and delete an organization.
+Provides a vCloud Director Org resource. This can be used to create, update, and delete an organization.
 Requires system administrator privileges.
 
 Supported in provider *v2.0+*
 
+!> **Warning:** Up to version 2.4, there were two bugs in the handling of this resource. If you have existing resources
+created in versions 2.0 to 2.4, you should re-create them using `terraform state rm` and `terraform import`
```
lvirbalas: 1.
"If you have existing resources created in versions 2.0 to 2.4," ->
"If you have existing vcd_org resource created with earlier versions,"

2.
"you should re-create them using `terraform state rm` and `terraform import`
(See *Upgrading Org resources to 2.5* below for the steps)."
->
"you should re-create them by following *Upgrading Org resources to 2.5* section below"

dataclouder: done

website/docs/r/org.html.markdown
```diff
@@ -43,13 +48,100 @@ The following arguments are supported:
...
+```hcl
+resource "vcd_org" "my-orgadmin" {
+  name             = "my-org"
+  full_name        = "guessing"
+  delete_recursive = "true"
+  delete_force     = "true"
+}
+```
+
+You can import such user into terraform state using this command
```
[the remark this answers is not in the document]
dataclouder: fixed

website/docs/r/org.html.markdown
```diff
@@ -43,13 +48,100 @@ The following arguments are supported:
...
 * `delay_after_power_on_seconds` - (Optional) - Specifies this organization's default for virtual machine boot delay after power on. Default is `0`.
 
+## Importing
+
+Supported in provider *v2.5+*
+
+~> **Note:** The current implementation of Terraform import can only import resources into the state. It does not generate
+configuration. [More information.][docs-import]
+
+An existing Org can be [imported][docs-import] into this resource via supplying the path for an Org.Since the Org is
```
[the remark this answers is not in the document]
dataclouder: fixed

vcd/resource_vcd_org.go
```diff
@@ -138,84 +150,199 @@ func getSettings(d *schema.ResourceData) *types.OrgSettings {
...
-	return nil
+	log.Printf("Org with id %s found", identifier)
+	d.SetId(adminOrg.AdminOrg.ID)
+	return setOrgData(d, adminOrg)
+}
+
+// Imports an Org into Terraform state
+// This function task is to get the data from vCD and fill the resource data container
+// Expects the d.Id() to be an Org name, which is the full path
+func resourceVcdOrgImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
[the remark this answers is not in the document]
dataclouder: done

CHANGELOG.md
```diff
@@ -1,4 +1,18 @@
...
+
+IMPROVEMENTS:
+
+* `vcd_org` Add import capability
+
+BUG FIXES:
+* Change default value for `vcd_org.deployed_vm_quota` and `vcd_org.stored_vm_quota`. It was incorrectly set at `-1` instead of `0`.
+* Change Org Id from partial task ID to real Org ID during creation.
```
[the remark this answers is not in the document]

website/docs/r/org.html.markdown
```diff
@@ -3,16 +3,20 @@ layout: "vcd"
...
 # vcd\_org
 
-Provides a vCloud Director Org resource. This can be used to create and delete an organization.
+Provides a vCloud Director Org resource. This can be used to create, update, and delete an organization.
 Requires system administrator privileges.
 
 Supported in provider *v2.0+*
 
+!> **Warning:** Up to version 2.4, there were two bugs in the handling of this resource. If you have existing resources
+created in versions 2.0 to 2.4, you should re-create them by following *Upgrading Org resources to 2.5* below.
```
lvirbalas: Better, but still too wordy and not specific enough. Suggestion to simplify and keep the details in that other section:

**Warning:** If you have existing vcd_org resources in state file created with versions earlier than 2.5, you should re-create them by following *Upgrading Org resources to 2.5* section below.

Didainius: It is not that you should. I have successfully tested creating it in 2.4.0 and updating provider with this PR. It did not even suggest to do any updates. Only in rare case it misbehaves should someone update. In my case it went smooth which is what I like in non major version update. The simple `read` fixed previously incorrect ID (used from task).

vcd/resource_vcd_org.go
Didainius: Works perfectly now. Thanks.

website/docs/r/org.html.markdown
dataclouder: I have changed the text as suggested, and toned down the certainty of failure with a more cautious wording

lvirbalas: Guys, we need to get to the bottom of this, because as a user after reading "may" I am left feeling uncertain:
- Should I upgrade the state file?
- How to know whether I should?
- What are consequences if I ignore this?

In other words:
- Do we need this warning at all?
- If yes, we need to make it clear **in which cases** the user should follow that state upgrade section.

Let's discuss it here until we reach a plausible formulation.

lvirbalas: Adding this from Dainius:

> My manual testing of migrations from 2.4.0 to 2.5.0 are smooth now with no need for import. The only case should someone need to update config is if they manually set stored_vm_quota=-1 and deployed_vm_quota=-1 previously and still it goes without triggering updates. If they used default values instead - the upgrade is hassle-free.

dataclouder: You know that you need to rebuild when you get a plan that suggests updating without any changes. 
I added the warning before I accepted Dainius suggestion to handle some possible inconsistencies in the code, and fixing existing bugs in the same sweep.
Perhaps we can remove the warning and leave just the section that says "in rare cases".
How rare, we don't know, but when I reproduced the inconsistencies manually, the procedure we suggest works fine.
With the latest code, the inconsistencies seem to have disappeared in all known cases.
So, the warning could be seen as overcautious, and could be removed.

Didainius: I would agree that it can be removed. I don't know how to get into this situation now.

PR #284, opened 2019-07-22
Add load balancer virtual server resource and data source

Didainius:
Work continues on https://github.com/terraform-providers/terraform-provider-vcd/issues/223

This PR introduces load balancer virtual server - the last component in the puzzle.


website/docs/r/lb_virtual_server.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful
+when connected as sysadmin working across different organisations
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+* `edge_gateway` - (Required) The name of the edge gateway on which the virtual server is to be
+created
+* `name` - (Required) Virtual Server name
+* `description` - (Optional) Virtual Server description
```
lvirbalas: Sometimes you use "Virtual Server" and sometimes "virtual server". Please chose one in user-facing messages.

website/docs/r/lb_virtual_server.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+  edge_gateway = "my-edge-gw"
+  
+  name       = "http-virtual-server"
+  ip_address = "1.1.1.1" # Edge Gateway uplink interface IP
+  protocol   = "http"    # Must be the same as specified in application profile
+  port       = 80
+  
+  app_profile_id = "${vcd_lb_app_profile.http.id}"
+  server_pool_id = "${vcd_lb_server_pool.web-servers.id}"
+  app_rule_ids   = ["${vcd_lb_app_rule.redirect.id}", "${vcd_lb_app_rule.language.id}"]
```
lvirbalas: In one of the first LB PRs we discussed a need for an "uber" example which connects all the LB resources together. Maybe having it here as second example would be a good place for it? Meaning, have one simple exampe for only virtual server and have the second one "extended" example where all the LB resources connect up?

vcd/resource_vcd_lb_virtual_server_test.go
```diff
@@ -0,0 +1,258 @@
...
+		"EdgeGatewayIp":     testConfig.Networking.ExternalIp,
+		"VirtualServerName": t.Name(),
+		"Tags":              "lb lbVirtualServer",
+	}
+
+	configText := templateFill(testAccVcdLbVirtualServer_step0, params)
+	debugPrintf("#[DEBUG] CONFIGURATION for step 0: %s", configText)
+
+	params["FuncName"] = t.Name() + "-step1"
+	configText2 := templateFill(testAccVcdLbVirtualServer_step2, params)
```
vbauzysvmware: where is step 1 ? :)

vcd/resource_vcd_lb_virtual_server_test.go
```diff
@@ -0,0 +1,258 @@
...
+			},
+
+			// Check that import works
+			resource.TestStep{
+				ResourceName:      "vcd_lb_virtual_server.virtual-server-import",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateIdFunc: importStateIdByOrgVdcEdge(testConfig, params["VirtualServerName"].(string)),
+			},
+			// Simple check - without app profile, rule ids or server pool
```
vbauzysvmware: Seems that validates update? or not?

vcd/datasource_vcd_lb_virtual_server.go
```diff
@@ -0,0 +1,113 @@
...
+			},
+			"edge_gateway": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Edge gateway name in which the Virtual Server is located",
+			},
+			"name": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
```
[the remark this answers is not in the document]
Didainius: Both have it. Fixed.

vcd/datasource_vcd_lb_virtual_server.go
```diff
@@ -0,0 +1,113 @@
...
+				Description: "Virtual Server name for lookup",
+			},
+			"description": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
```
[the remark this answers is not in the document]
Didainius: It's a mistake. Thanks

vcd/resource_vcd_lb_virtual_server.go
```diff
@@ -0,0 +1,285 @@
...
+	defer vcdClient.unLockParentEdgeGtw(d)
+
+	edgeGateway, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
+	if err != nil {
+		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	}
+
+	lBVirtualServer, err := getLBVirtualServerType(d)
+	if err != nil {
+		return fmt.Errorf("unable to expand load balancer virtual server: %s", err)
```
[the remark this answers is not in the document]
Didainius: changed

vcd/resource_vcd_lb_virtual_server.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	createdVirtualServer, err := edgeGateway.CreateLBVirtualServer(lBVirtualServer)
+	if err != nil {
+		return fmt.Errorf("error creating new load balancer virtual server: %s", err)
+	}
+
+	d.SetId(createdVirtualServer.Id)
+	return nil
```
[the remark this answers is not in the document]
Didainius: Added

vcd/resource_vcd_lb_virtual_server.go
```diff
@@ -0,0 +1,285 @@
...
+
+func resourceVcdLBVirtualServerRead(d *schema.ResourceData, meta interface{}) error {
+	vcdClient := meta.(*VCDClient)
+
+	edgeGateway, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
+	if err != nil {
+		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	}
+
+	readVirtualServer, err := edgeGateway.ReadLBVirtualServerById(d.Id())
```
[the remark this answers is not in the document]
Didainius: I will have a separate PR for all the syntax/naming changes as I'm noting what has to be changed. It has to be changed in many places but if I put it in this PR then there will be additional 10+ files changes and this PR will become huge.

vcd/resource_vcd_lb_virtual_server.go
```diff
@@ -0,0 +1,285 @@
...
+	defer vcdClient.unLockParentEdgeGtw(d)
+
+	edgeGateway, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
+	if err != nil {
+		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	}
+
+	updateVirtualServerConfig, err := getLBVirtualServerType(d)
+	if err != nil {
+		return fmt.Errorf("could not expand load balancer virtual server for update: %s", err)
```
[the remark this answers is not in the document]
Didainius: Done.

vcd/resource_vcd_lb_virtual_server.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	orgName, vdcName, edgeName, virtualServerName := resourceURI[0], resourceURI[1], resourceURI[2], resourceURI[3]
+
+	vcdClient := meta.(*VCDClient)
+	edgeGateway, err := vcdClient.GetEdgeGateway(orgName, vdcName, edgeName)
+	if err != nil {
+		return nil, fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	}
+
+	readVirtualServer, err := edgeGateway.ReadLBVirtualServerByName(virtualServerName)
```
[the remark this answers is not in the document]
Didainius: As I mentioned in another comment there is a separate PR coming in for name changes.

vcd/resource_vcd_lb_virtual_server.go
```diff
@@ -0,0 +1,285 @@
...
+	d.SetId(readVirtualServer.Id)
+	return []*schema.ResourceData{d}, nil
+}
+
+// getLBVirtualServerType converts schema.ResourceData to *types.LBVirtualServer and is useful
+// for creating API requests
+func getLBVirtualServerType(d *schema.ResourceData) (*types.LBVirtualServer, error) {
+	lbVirtualServer := &types.LBVirtualServer{
+		Name:        d.Get("name").(string),
+		Description: d.Get("description").(string),
```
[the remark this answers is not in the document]
Didainius: True, which probably means that maintaining this separation may become difficult. Removed it at all.

vcd/resource_vcd_lb_virtual_server.go
```diff
@@ -0,0 +1,285 @@
...
+	lbVirtualServer.ApplicationRuleIds = appRuleIds
+
+	return lbVirtualServer, nil
+}
+
+// setlBVirtualServerData sets object state from *types.LBVirtualServer
+func setlBVirtualServerData(d *schema.ResourceData, lBVirtualServer *types.LBVirtualServer) error {
+	d.Set("name", lBVirtualServer.Name)
+	d.Set("description", lBVirtualServer.Description)
+	d.Set("enabled", lBVirtualServer.Enabled)
```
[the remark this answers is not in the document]
Didainius: Just as above - removed it at all.

vcd/resource_vcd_lb_virtual_server_test.go
Didainius: Step 1 is the `import` test. I'm naming it so that step numbers are correct when error comes out

vcd/resource_vcd_lb_virtual_server_test.go
Didainius: Yes. Further steps are doing "updates" when resource names are left the same.

website/docs/r/lb_virtual_server.html.markdown
Didainius: Lowercase it will be then :)

website/docs/r/lb_virtual_server.html.markdown
Didainius: Added

vcd/resource_vcd_lb_virtual_server.go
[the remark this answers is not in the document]
Didainius: Before 2.4 of course :)

vcd/resource_vcd_lb_virtual_server.go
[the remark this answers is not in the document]
Didainius: Added a note in my list to fix other resources in that name/cleanup PR.

vcd/resource_vcd_lb_virtual_server_test.go
vbauzysvmware: Maybe we should state in comment

Didainius: Added comment for each step number

website/docs/r/lb_virtual_server.html.markdown
```diff
@@ -0,0 +1,206 @@
...
+
+Supported in provider *v2.4+*
+
+## Example Usage 1 (HTTP virtual server)
+
+```hcl
+provider "vcd" {
+  user     = "${var.admin_user}"
+  password = "${var.admin_password}"
+  org      = "System"
```
dataclouder: This should be "sysorg", to avoid clashes with the "org" below. 
Or the provider section could be removed altogether

vcd/resource_vcd_lb_virtual_server.go
```diff
@@ -0,0 +1,281 @@
...
+				Required:     true,
+				Description:  "Protocol that the virtual server accepts",
+				ValidateFunc: validateCase("lower"),
+			},
+			"port": &schema.Schema{
+				Type:        schema.TypeInt,
+				Required:    true,
+				Description: "Port number that the load balancer listens on",
+			},
+			"enable_acceleration": &schema.Schema{
```
dataclouder: Why is the default  for this property `false`, while we know that without acceleration the LB won't work?

Didainius: It's not necessarily not going to work. The idea behind it is such as per [vCD doc](https://docs.vmware.com/en/vCloud-Director/9.0/com.vmware.vcloud.tenantportal.doc/GUID-7E082E77-B459-4CE7-806D-2769F7CB5624.html). A asked for it in the test so that it is less steps. The UI has it disabled by default:
> Click the Acceleration Enabled icon to configure the load balancer to use the faster L4 engine rather than L7 engine. The L4 TCP VIP is processed before the edge gateway firewall so no Allow firewall rule is required. 

> L7 VIPs for HTTP and HTTPS are processed after the firewall, so when Acceleration Enabled is not selected, an edge gateway firewall rule must exist to allow access to the L7 VIP for those protocols. When Acceleration Enabled is selected and the server pool is in non-transparent mode, an SNAT rule is added, so you must ensure that the firewall is enabled on the edge gateway.

website/docs/r/lb_virtual_server.html.markdown
Didainius: Removed. I actually think it's not worth to set the `provider` everywhere. Noted to cleanup in other Lb examples. I think the index page shows how to use it and it's quite standard accros providers. https://www.terraform.io/docs/providers/vcd/index.html

PR #286, opened 2019-07-23
Improve tests

dataclouder:
Add variable to change testacc timeout on-the-fly
Add personalized file to skip tests locally
Add a skipping notice to some update tests
   (binary tests can only run init-apply-destroy, while
   these tests are part of a multi-script combo)

### DIFF
--- a/scripts/runtest.sh
+++ b/scripts/runtest.sh
@@ -11,7 +11,7 @@ then
 fi
 
 wanted=$1
-timeout=90m
+timeout=120m
 if [ -n "$VCD_TIMEOUT" ]
 then
     timeout=$VCD_TIMEOUT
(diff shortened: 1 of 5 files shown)


scripts/runtest.sh
```diff
@@ -92,13 +97,13 @@ function acceptance_test {
     if [ -n "$VERBOSE" ]
     then
         echo "# check for config file"
-        echo "TF_ACC=1 go test -tags '$tags' -v -timeout 120m ."
```
lvirbalas: This PR reduced the timeout from 120m to 90m. @vbauzysvmware please confirm this is OK as you increased it just recently.

dataclouder: fixed

PR #288, opened 2019-07-24
Add load balancing configuration for edge gateway

Didainius:
As part of load balancer introduction ( https://github.com/terraform-providers/terraform-provider-vcd/issues/223) this PR adds capability to set general load balancing settings (such as enable/disable load balancing)

### DIFF
--- a/vcd/resource_vcd_edgegateway.go
+++ b/vcd/resource_vcd_edgegateway.go
@@ -109,11 +109,6 @@ func resourceVcdEdgeGateway() *schema.Resource {
 func resourceVcdEdgeGatewayCreate(d *schema.ResourceData, meta interface{}) error {
 	log.Printf("[TRACE] edge gateway creation initiated")
 
-	// We use partial mode here because edge gateway creation and configuration consists of two
-	// parts (API calls) - creating edge gateway and configuring load balancer. If the second part
-	// fails we still want to persist state for the first part
-	d.Partial(true)
-
 	vcdClient := meta.(*VCDClient)
 
 	rawExternalNetworks := d.Get("external_networks").([]interface{})
@@ -173,11 +168,10 @@ func resourceVcdEdgeGatewayCreate(d *schema.ResourceData, meta interface{}) erro
 		log.Printf("[DEBUG] Error creating edge gateway: %#v", err)
 		return fmt.Errorf("error creating edge gateway: %#v", err)
 	}
-	// Edge gateway creation succeeded therefore we can flush related fields now. Edge
-	// gateway is already created even if further process fails.
-	log.Printf("[TRACE] flushing partial edge gateway creation fields")
+	// Edge gateway creation succeeded therefore we save related fields now to preserve Id.
+	// Edge gateway is already created even if further process fails
+	log.Printf("[TRACE] flushing edge gateway creation fields")
 	err = setEdgeGatewayValues(d, edge)
-	setPartialEdgeGatewayValues(d)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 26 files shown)


vcd/resource_vcd_edgegateway.go
```diff
@@ -145,12 +168,107 @@ func resourceVcdEdgeGatewayCreate(d *schema.ResourceData, meta interface{}) erro
...
+			return err
+		}
+	}
+
+	log.Printf("[TRACE] edge gateway read completed: %#v", edgeGateway.EdgeGateway)
+	return nil
+}
+
+// resourceVcdEdgeGatewayUpdate updates an edge gateway from a resource definition
+func resourceVcdEdgeGatewayUpdate(d *schema.ResourceData, meta interface{}) error {
```
lvirbalas: Does this function update only LB general params of EG currently? If yes, a fat comment about such behavior is needed.

vcd/resource_vcd_edgegateway.go
```diff
@@ -196,44 +314,45 @@ func setEdgeGatewayValues(d *schema.ResourceData, egw govcd.EdgeGateway) error {
...
+		return fmt.Errorf("unable to update general load balancer settings: %s", err)
 	}
 
-	err = edgeGateway.Delete(true, true)
+	return nil
+}
 
-	log.Printf("[TRACE] edge gateway deletion completed\n")
-	return err
+// isEdgeGatewayLbConfigured checks if any of load balancer related settings are set
```
lvirbalas: I suggest adding that this only applies to LB **general parameters**.

website/docs/r/edgegateway.html.markdown
```diff
@@ -60,5 +63,17 @@ The following arguments are supported:
 * `advanced` - (Optional) True if the gateway uses advanced networking. Default is `true`.
 * `ha_enabled` - (Optional) Enable high availability on this edge gateway. Default is `false`.
 * `distributed_routing` - (Optional) If advanced networking enabled, also enable distributed routing. Default is `false`.
+* `lb_enabled` - (Optional) Enable load balancing. Default is `false`.
```
lvirbalas: Currently these new optional LB parameters in resource definition don't have default values, but we are saying three of them are false by default. I suspect it becomes false if undefined? What about actually adding a default = false, so the behavior has less assumptions? Or would that introduce something unwanted?

Didainius: I experimented with that. One thing I didn't like:
If I set `default` I can't make the check if value `is set` in config or is it a `default`. Then it always tries to configure load balancer (with default values if not set).
The actual default configuration at least on 9.5 (but I think others as well) after creation is enabled=false, lb_acceleration_enabled=false, lb_logging_enabled=false, lb_loglevel='info'.
I could set such defaults, then after initial read it wouldn't complain as these values coincide with defaults, but if in some future version of vCD the defaults change then I would start overriding them.

vcd/resource_vcd_edgegateway.go
dataclouder: We have an issue about implementing Update for EGW. Please add a note saying that this Update is only dealing with LB items so far.

vbauzysvmware: Then maybe we need create Issue to tackle full edge gtw update?

dataclouder: we have one https://github.com/vmware/go-vcloud-director/issues/202

Didainius: I changed the comment to explicitly state that it is general load balancer params. Also resource schema hints it quite well with only `lb_*` fields not having `ForceNew: true`

vcd/resource_vcd_edgegateway.go
Didainius: Getting rid of it as we are going to set default load balancer settings always. This check doesn't help anymore when defaults are used.

website/docs/r/edgegateway.html.markdown
Didainius: Changed these fields to always have default values (same as edge gateway currently gets during provisioning):
enabled=false
lb_acceleration_enabled=false
lb_logging_enabled=false
lb_loglevel='info'

The only scenario when they are ignored is when edge gateway is _not_ advanced.
It also prints a warning if someone is trying to enable it on `simple` edge gateway during `updates procedures and the website docs have a note now.

website/docs/r/edgegateway.html.markdown
```diff
@@ -17,8 +17,10 @@ Supported in provider *v2.4+*
 You must use `System Adminstrator` account in `provider` configuration
 and then provide `org` and `vdc` arguments for edge gateway to work.
 
-~> **Note:** To enable load balancing capabilities the edge gateway must be `advanced`. Refer to
-[official vCloud Director documentation](https://docs.vmware.com/en/vCloud-Director/9.7/com.vmware.vcloud.tenantportal.doc/GUID-7E082E77-B459-4CE7-806D-2769F7CB5624.html) for more information.
+~> **Note:** Load balancing capabilities will work only when edge gateway is `advanced`. Load
```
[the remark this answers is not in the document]
Didainius: The `make website` handles it. Link is active and working.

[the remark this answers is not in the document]
dataclouder: OK, then. It's only my IDE preview that complains

PR #290, opened 2019-07-25
Load balancer syntax cleanup and bugfixes

Didainius:


vcd/resource_vcd_lb_service_monitor_test.go
```diff
@@ -103,7 +103,7 @@ func testAccCheckVcdLbServiceMonitorDestroy(serviceMonitorName string) resource.
 			return fmt.Errorf(errorUnableToFindEdgeGateway, err)
 		}
 
-		monitor, err := edgeGateway.ReadLBServiceMonitor(&types.LBMonitor{Name: serviceMonitorName})
+		monitor, err := edgeGateway.GetLbServiceMonitor(&types.LbMonitor{Name: serviceMonitorName})
```
vbauzysvmware: Doesn't we should have wappers here?

vbauzysvmware:
I am not sure I saw all changes/fixes from issues I mentioned in emails.

vcd/resource_vcd_lb_service_monitor_test.go
Didainius: Yes indeed. I was hoping `make install` to fail on such cases but I was wrong. Replaced them.

PR #32, opened 2017-08-04
adding hostname initscript fix to vm

bobbydeveaux:
Additional fix for #24 as I only added it to vApp before :(

Issue #22, opened 2017-06-27
Allow specifying protocol for SNAT/DNAT rules


pasikarkkainen:
Currently Terraform vcloud director provider allows specifying "protocol" (tcp/udp/icmp/any) for firewall rules, but not for snat/dnat rules. Instead currently snat/dnat rules always assume protocol to be "tcp".

To be able to create port forwarding rules for, say, DNS, NTP, or OpenVPN connections one needs to be able to configure protocol to be tcp/udp/icmp/any, just like one can for firewall rules.

For backwards compatibility we should default to "tcp" if protocol hasn't been specified.

bobbydeveaux:
This is currently hardcoded in the govcloudair library:

```
natRule := &types.NatRule{
		RuleType:  nattype,
		IsEnabled: true,
		GatewayNatRule: &types.GatewayNatRule{
			Interface: &types.Reference{
				HREF: uplink.HREF,
			},
			OriginalIP:     externalIP,
			OriginalPort:   externalPort,
			TranslatedIP:   internalIP,
			TranslatedPort: internalPort,
			Protocol:       "tcp",
		},
	}
```

We'll just need to parameterise this.

urajio:
please fix this

pasikarkkainen:
Let us know when it's possible to specify protocol for snat/dnat rules. I'm happy to test the functionality!

vbauzysvmware:
Hi Everyone, 

Changes are pushed and you are welcome to try.

P.s. Note that missing in documentation that "-1" for port means "any".

pasikarkkainen:
@vbauzysvmware : Great, I'll give it a go soon and let you know what happens :)

One small typo I noticed with a quick glance in go-vcloud-director changes: "provided protocol is not one of TCP, UDP, TCP_UDP, ICPM, ANY". That should obviously say "ICMP", not "ICPM".

vbauzysvmware:
Thank you, @pasikarkkainen. Mistype noted.
Also resource supports now ICMP sub types. Current implementations supports everything you can do it in VCD UI.

pasikarkkainen:
@vbauzysvmware : Seems to work fine for me! I was able to do for example:

resource "vcd_dnat" "dnat-test1" {
    edge_gateway  = "${var.edge_gateway}"
    external_ip   = "${var.edge_ext_ip}"
    port = 1234
    translated_port = 10234
    protocol = "udp"
[...]

Issue #12, opened 2017-06-13
VMware - Vcloud Air/ Vcloud Director: error: "* Something went wrong: error finding LoginUrl: error parsing error body for non-200 request: EOF"


hashibot:
_This issue was originally opened by @EllisFisher1994 as hashicorp/terraform#11455. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

getting the error * Something went wrong: error finding LoginUrl: error parsing error body for non-200 request: EOF

Have variables in a separate file.
`# Configure the VMware vCloud Director Provider
provider "vcd" {
    user                 = "${var.vcd_user}"
    password             = "${var.vcd_pass}"
    org                  = "${var.vcd_org}"
    url                  = "${var.vcd_url}"
    vdc                  = "${var.vcd_vdc}"
}

resource "vcd_vapp" "web" {
    name          = "web"
    catalog_name  = "Test-Catalog"
    template_name = "lampstack-1.10.1-ubuntu-10.04"
    memory        = 2048
    cpus          = 1

    network_name  = "testing_network"

    ip            = "10.10.104.160"

    metadata {
        role    = "web"
        env     = "staging"
        version = "v1"
    }

}
`

Can anyone help?

srinarayanant:
facing same issue 
2017/08/16 15:06:18 [DEBUG] plugin: waiting for all plugin processes to complete...
Error running plan: 1 error(s) occurred:

* provider.vcd: Something went wrong: error authorizing: error parsing error body for non-200 request: EOF
2017-08-1

srinarayanant:
Any updates on the issue , is it ok to use  api use =  administrator@SYSTEM

srinarayanant:
It works after configuring with org user with just the user part since appending is performed by code internally 

// Set Basic Authentication Header
	req.SetBasicAuth(user+"@"+org, pass)

bobbydeveaux:
hi @srinarayanant - sorry I was away yesterday. 

Yes, your user and org should be separate, as the code will concatenate them.

Obviously the error return is no good and we should fix. Could you just confirm how we can replicate?

Cheers

srinarayanant:
Hi @bobbydeveaux 

Just using a wrong user name produces the error.
user = administrator@vsphere.local    or user=user1@O2 
like below in vcd section
provider "vcd" {
  user                 = "user1@O2"

[...]

PR #293, opened 2019-07-26
Fix for protocol upper/lower case handling

vbauzysvmware:
Now lower case configured required and vCD returns lower case and they match


website/docs/r/dnat.html.markdown
```diff
@@ -52,9 +52,9 @@ The following arguments are supported:
 * `port` - (Required) The port number to map. -1 translates to "any"
 * `translated_port` - (Optional) The port number to map
 * `internal_ip` - (Required) The IP of the VM to map to
-* `protocol` - (Optional; *v2.0+*) The protocol type. Possible values are TCP, UDP, TCPUDP, ICMP, ANY. TCP is default to be backward compatible with previous version
+* `protocol` - (Optional; *v2.0+*) The protocol type. Possible values are tcp, udp, tcpupd, icmp, any. `tcp` is default to be backward compatible with previous version
```
lvirbalas: Please quote all these values. Now that they are lower case they need emphasis.

vbauzysvmware: done

website/docs/r/dnat.html.markdown
```diff
@@ -52,9 +52,9 @@ The following arguments are supported:
...
 * `translated_port` - (Optional) The port number to map
 * `internal_ip` - (Required) The IP of the VM to map to
-* `protocol` - (Optional; *v2.0+*) The protocol type. Possible values are TCP, UDP, TCPUDP, ICMP, ANY. TCP is default to be backward compatible with previous version
+* `protocol` - (Optional; *v2.0+*) The protocol type. Possible values are `tcp`, `udp`, `tcpupd`, `icmp`, `any`. `tcp` is default to be backward compatible with previous version
 * `icmp_sub_type` - (Optional; *v2.0+*) The name of ICMP type. Possible values are   address-mask-request, destination-unreachable, echo-request, echo-reply, parameter-problem, redirect, router-advertisement, router-solicitation, source-quench, time-exceeded, timestamp-request, timestamp-reply, any
-* `network_type` - (Optional; *v2.4+*) Type of the network on which to apply the NAT rule. Possible values org or ext. *`network_type` will be a required field in the next major version.*
+* `network_type` - (Optional; *v2.4+*) Type of the network on which to apply the NAT rule. Possible values `org` or `ext`. *`network_type` will be a required field in the next major version.*
```
lvirbalas: Please do the same quoting in SNAT doc.

lvirbalas:
BTW, in the warning of SNAT and DNAT pages:
```
Warning: When advanced edge gateway is used and the rule is updated using UI, then Id mapping will be lost and terraform won't find the rule anymore and remove it from state.
```
Please change:
1. "Id" -> "ID"
2. "terraform" -> "Terraform"

vbauzysvmware:
> BTW, in the warning of SNAT and DNAT pages:
> 
> ```
> Warning: When advanced edge gateway is used and the rule is updated using UI, then Id mapping will be lost and terraform won't find the rule anymore and remove it from state.
> ```
> 
> Please change:
> 
>     1. "Id" -> "ID"
> 
>     2. "terraform" -> "Terraform"

done

lvirbalas:
For reference: for those who have `protocol` not set from previous versions in the state file, this change will lead to a single occurrence of `protocol` field's value update from upper to lower case during the next `plan`?

vbauzysvmware:
> For reference: for those who have `protocol` not set from previous versions in the state file, this change will lead to a single occurrence of `protocol` field's value update from upper to lower case during the next `plan`?

this happens today, cause vCD api returns lowercase and everytime update happens

CHANGELOG.md
```diff
@@ -18,7 +18,7 @@ FEATURES:
 * `resource/vcd_org_vdc` added Org VDC update and full state read - [GH-275]
 * `resource/vcd_org_vdc` added Org VDC metadata support - [GH-276]
 * `resource/vcd_snat` added ability to choose network name and type. [GH-282] 
-* `resource/vcd_dnat` added ability to choose network name and type. [GH-282]
+* `resource/vcd_dnat` added ability to choose network name and type. [GH-282] [GH-292] [GH-293]
```
Didainius: Never tested before, but should be like that as per https://www.terraform.io/docs/extend/best-practices/versioning.html
> IMPROVEMENTS:
> * resource/load_balancer: Add `ATTRIBUTE` argument (support X new functionality) [GH-12]
> * resource/subnet: Now better [GH-22, GH-32]

```suggestion
* `resource/vcd_dnat` added ability to choose network name and type. [GH-282, GH-292, GH-293]
```

vbauzysvmware: changed

website/docs/r/dnat.html.markdown
```diff
@@ -52,9 +54,10 @@ The following arguments are supported:
...
```
lvirbalas: I used org user and hit this with `ext`:
```
vcd_dnat.dnat_ext_net: Creating...

Error: unable to find orgVdcNetwork: my-org-network, err: can't find VDC Network: my-org-network
```
Let's add a sentence:
```
* `network_type` - (Optional; *v2.4+*) Type of the network on which to apply the NAT rule. Possible values `org` or `ext`. `ext` requires system administrator privileges. *`network_type` will be a required field in the next major version.*
```

PR #310, opened 2019-08-12
Add update support for `network` block in `vcd_vapp_vm`, allow force guest customization

### LINKED ISSUE #299: Add update support for vcd_vapp_vm networking
At the moment `network` block in resource `vcd_vapp_vm` does not support `update` operations therefore one cannot change networking configuration without recreating `VM`.

This issue is for tracking for for introducing `update` support for VM networking.

### LINKED ISSUE #300: Add support for `force_customization` setting in resource vcd_vapp_vm
At the moment resource `vcd_vapp_vm`only performs customization on first boot, however this may be needed in some other scenarios. One of such scenarios is to alter network card configuration (https://github.com/terraform-providers/terraform-provider-vcd/issues/299) and then trigger guest OS configuration update.

Didainius:
Closed #299 Closes #300 
This PR goes on top of https://github.com/vmware/go-vcloud-director/pull/229/files and adds two features:
* Allows updating network configuration for `vcd_vapp_vm`
* Introduces `customization block with one field `force`which allows to force guest customization on next `terraform apply` (causing a VM to reboot).

Testing features:
* To shorten test suite length some tests will be run in parallel. A new `make` command `make seqtestacc` is introduced to force running tests sequentially

**Note** The `customization` block was introduced as there is already demand to make other customization features available as the GUI has in "Guest customization" tab. The idea is to group all customization settings under this block in future.


vcd/resource_vcd_vapp_vm.go
```diff
@@ -611,19 +658,62 @@ func resourceVcdVAppVmUpdateExecute(d *schema.ResourceData, meta interface{}) er
...
 		}
 
+		// When customization is requested VM must be un-deployed before starting it
+		if customizationNeeded {
+			log.Printf("[TRACE] forced customization for VM %s was requested. Current state %s",
+				vm.VM.Name, vmStatus)
+
+			if vmStatus != "POWERED_OFF" {
+				log.Printf("[TRACE] VM %s is in state %s. Un-deploying", vm.VM.Name, vmStatus)
+				task, err := vm.Undeploy()
```
lvirbalas: Above you already have code which does Undeploy if customization is needed. Why do you do it here again? Or from the other perspective, why do you Undeploy above, if you have this logic here?

Didainius: It is because this resource needs a major refactor at some point in time. Probably even with my next PR for "guest properties".

The main problem here is that `create` method _always_ calls `update` to fully customize the VM (this should not happen as `create` and `update` should be separate operations). The above "undeploy" is only called if some other resource settings have changed during update (like `memory`, `cpu`, `cpu_cores`, `power_on`, `disk`, `expose_hardware_virtualization`, `network`).

But if the user only forces customization `customization.force = true` without altering any other options (listed in my above comment) - then the above undeploy will not be triggered and I need to do it here.

website/docs/r/vapp_vm.html.markdown
```diff
@@ -175,3 +175,11 @@ its parameters do force recreation of VMs!
...
 
   * `ip_allocation_mode=NONE` - **`ip`** field can be omitted or set to an empty string "". Empty string may be useful when doing HCL variable interpolation.
+
+<a id="customization"></a>
+## Customization
+
+* `force` (Optional) This field works as a flag and triggers force customization when `true` during an update 
+(`terraform apply`) every time. It never complains about a change in statefile. It can be used when guest customization
+is needed after a NIC change and then set back to `false`. **Note** this setting will cause a VM reboot and will not
+have effect when `power_on` field is set to `false`.
```
lvirbalas: As this scenario is stretching the boundaries of who Terraform is used, it would be great to have a workflow example representing the use case from beginning to end (though with HCL snippets showing only the actual parts - no need to show everything). I suggest adding it in this section as opposed to the top. For example, show one snippet with old settings, then how you change the network settings, set force, run apply and set force back to false. Assuming I got everything right... :)

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1086,3 +1177,24 @@ func readNetworks(vm govcd.VM, vapp govcd.VApp) ([]map[string]interface{}, error
 	}
 	return nets, nil
 }
+
+// isForcedConfiguration checks "customization" block in resource and checks if the value of field "force"
+// is set to "true". It returns false if the value is not set or is set to false
+func isForcedConfiguration(customizationBlock interface{}) bool {
```
vbauzysvmware: func name is misleading I would say, block custimization but func name is configuration

website/docs/r/vapp_vm.html.markdown
```diff
@@ -175,3 +178,12 @@ its parameters do force recreation of VMs!
...
   * `ip_allocation_mode=NONE` - **`ip`** field can be omitted or set to an empty string "". Empty string may be useful when doing HCL variable interpolation.
+
+<a id="customization"></a>
+## Customization
+
+* `force` (Optional) **Warning.** `true` value will cause the VM to reboot on every `apply` operation.
+This field works as a flag and triggers force customization when `true` during an update 
+(`terraform apply`) every time. It never complains about a change in statefile. Can be used when guest customization
+is needed after VM configuration (e.g. NIC change, customization options change, etc.) and then set back to `false`.
+**Note.** It will not have effect when `power_on` field is set to `false`.
```
vbauzysvmware: Maybe we should throw warning from the code? @lvirbalas

lvirbalas: We discussed this with Dainius Yesterday. Yes, we need, and ideally during *plan*, if possible, because it will rebooting VM.

Didainius:
The tests have become quite lengthy, especially with this PR so I'm experimenting with parallelizing test runs using `resource.ParallelTest`. During initial testing it looks as if it saves ±30mins (takes 100 mins instead of 133mins) with current commit.

vcd/resource_vcd_vapp_vm.go
Didainius: renamed to `isForcedCustomization()`

website/docs/r/vapp_vm.html.markdown
Didainius: Added workflow example.

website/docs/r/vapp_vm.html.markdown
Didainius: I have added a Terraform validator which emits warning.
Here's how the `terraform plan` looks like (only when `customization.force=true`):
```bash
...
      + template_name                  = "LB"
      + vapp_name                      = "lb-test-web"
      + vdc                            = "tf_vdc"

      + customization {
          + force = true
        }
    }

Plan: 2 to add, 0 to change, 0 to destroy.

Warning: Using 'true' value for field 'vcd_vapp_vm.customization.force' will reboot VM on every 'terraform apply' operation



  on main.tf line 62, in resource "vcd_vapp_vm" "web":
  62: resource "vcd_vapp_vm" "web" {



------------------------------------------------------------------------

Note: You didn't specify an "-out" parameter to save this plan, so Terraform
can't guarantee that exactly these actions will be performed if
"terraform apply" is subsequently run.
```

When doing `terraform apply` however these warnings are not shown. I think I'm ok with that, but need other opinions.

The thing I like about using built in validator for warning is that it prints very nicely and points out location automatically while if I add a print to stdout manually I can't control it so nicely.

Didainius: Adding output for `terraform validate` as well:
```bash
$  terraform validate

Warning: Using 'true' value for field 'vcd_vapp_vm.customization.force' will reboot VM on every 'terraform apply' operation



  on main.tf line 62, in resource "vcd_vapp_vm" "web":
  62: resource "vcd_vapp_vm" "web" {


Success! The configuration is valid, but there were some validation warnings as shown above.

```

lvirbalas: Perfect. Does this get printed with `apply` without --auto-approve option too?

website/docs/r/vapp_vm.html.markdown
```diff
@@ -106,6 +106,65 @@ resource "vcd_vapp_vm" "web2" {
 
 ```
 
+
+## Example forced customization workflow
```
lvirbalas: What about moving this sample to the `## Customization` section below? This way it would appear contextually there where it's needed the most.

website/docs/r/vapp_vm.html.markdown
Didainius: No. That is the problem as I mentioned above.

> When doing `terraform apply` however these warnings are not shown. I think I'm ok with that, but need other opinions.

website/docs/r/vapp_vm.html.markdown
Didainius: True. I somehow forced myself into thinking that we put examples on top. Should be better now.

website/docs/r/vapp_vm.html.markdown
vbauzysvmware: You can write warning on apply phase using functionality from upload resource if we want that.

Didainius: Yes. I could, but the somewhat less nice thing is that it would be "somewhere in the middle" while the warning in validator prints it very clearly.

Issue #313, opened 2019-08-20
Add support for VM and vApp guest properties in `vcd_vapp_vm` and `vcd_vapp` resources


Didainius:
VM has a possibility to set guest properties (OVF properties) for images that support this. This provider does not yet support this feature and this issue is set to track it.

swerveshot:
Currently there seems to be some support for ovf attributes, but only at the vApp level. And when you try to use it you get the following error message:
```
PS H:\Documents\Code\Terraform-vCD-vSphereLab> terraform apply --auto-approve
vcd_vapp.vsphere67lab: Creating...

Error: error setting the ovf: &errors.errorString{s:"vApp doesn't contain any children, aborting customization"}

  on main.tf line 19, in resource "vcd_vapp" "vsphere67lab":
  19: resource "vcd_vapp" "vsphere67lab" {
```

My use case: creating vSphere lab environments based on @lamw's nested ESXi virtual appliances.

Didainius:
Hello @swerveshot,
The support for OVF properties is already in `master` branch and will be released in 2.5.0 (aimed for end of october).
If you have a chance and wish to test you could try building the provider from master and say what you think. We cannot guarantee that it will be 100% as it is in master when it is released (if we find some bugs/problems as final pre-release testing).

This is how it looks in both vApp and VM levels:
```hcl
guest_properties = {
    "guest.hostname" = "my-hostname"
    "guest.other"    = "another-setting"
  }
```
You could also explore the VM and vApp documentation in master for examples:
https://github.com/terraform-providers/terraform-provider-vcd/blob/master/website/docs/r/vapp_vm.html.markdown

https://github.com/terraform-providers/terraform-provider-vcd/blob/master/website/docs/r/vapp.html.markdown
```

Didainius:
The functionality is in `master` and is targeted for 2.5.0 release. Feel free to reopen if you see any issues.

swerveshot:
Thanks for the feedback @Didainius! And good to know where to find the most up-to-date documentation.

marwanhammad:
do you publish your code to create nested ESX in vcd for  vSphere lab?

marwanhammad:
> 
> 
> Currently there seems to be _some_ support for [ovf attributes](https://www.terraform.io/docs/providers/vcd/r/vapp.html#ovf), but only at the vApp level. And when you try to use it you get the following error message:
> 
> ```
[...]

PR #319, opened 2019-08-26
Add support for guest properties in `vcd_vapp` and `vcd_vapp_vm` resources

Didainius:
ref #313

guest properties allow to pass variables into guest os for access via VMware tools. Some appliances require these properties to initialize properly.


Didainius:
binary and acceptance tests passed for me

vcd/resource_vcd_vapp.go
```diff
@@ -249,6 +255,26 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	if _, ok := d.GetOk("properties"); ok {
+		// TODO - this is just for testing. Real issue must be solved
```
Didainius: I have put this in during development and spotted only now. I need to fix this yet.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -853,6 +883,17 @@ func resourceVcdVAppVmRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("href", vm.VM.HREF)
 	d.Set("expose_hardware_virtualization", vm.VM.NestedHypervisorEnabled)
 
+	// update guest properties
+	guestProperties, err := vm.GetGuestProperties()
+	if err != nil {
+		return fmt.Errorf("unable to read guest properties: %s", err)
+	}
+
+	err = setProductSectionListData(d, guestProperties)
```
lvirbalas: When looking at govcd PR, it looked like the idea was to hide this "product section" naming from the caller and expose "guest properties" instead?

vcd/resource_vcd_vapp.go
```diff
@@ -100,6 +101,11 @@ func resourceVcdVApp() *schema.Resource {
 				Optional: true,
 				Default:  true,
 			},
+			"properties": {
```
lvirbalas: We had this idea of keeping vApp thin... What do guest properties on a vApp (as opposed to VM) actually do? Thanks.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -197,6 +197,11 @@ func resourceVcdVAppVm() *schema.Resource {
 				Default:     false,
 				Description: "Expose hardware-assisted CPU virtualization to guest OS.",
 			},
+			"properties": {
```
lvirbalas: For me the "properties" is a too generic name, "reserved" for some lower-level stuff.
What about "guest_properties" instead?
@dataclouder @vbauzysvmware your thoughts?

vcd/resource_vcd_vapp.go
Didainius: Well. There is a concept of vApp template. In some cases it can be a bundle of VMs for an app. In we don't support much of "vApp deployments" so far so there is a question for its usability.

UI also has it:
image

vcd/resource_vcd_vapp_vm.go
Didainius: These functions worked in the "concept" of guest properties, but the API calls are ProductSection. I'll look if I can refine them somehow clearer

vcd/resource_vcd_vapp.go
vbauzysvmware: what is this ?why we need sleep?

website/docs/r/vapp.html.markdown
```diff
@@ -137,3 +142,6 @@ The following arguments are supported:
 * `org` - (Optional; *v2.0+*) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
 * `vdc` - (Optional; *v2.0+*) The name of VDC to use, optional if defined at provider level
 * `accept_all_eulas` - (Optional; *v2.0+*) Automatically accept EULA if OVA has it. Default is `true`
+* `properties` - (Optional; *v2.5+*) Key value map of vApp guest properties **Note** `ovf` attribute
```
vbauzysvmware: As I am not very enlighten with this functionality - I didn't follow this comment.

vcd/resource_vcd_vapp.go
Didainius: Just as per the comment above :) This is a `TODO` leftover

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: I think here we operate with State file, in state file we have guest properties.

vcd/resource_vcd_vapp.go
```diff
@@ -249,6 +255,26 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
...
 
+	if _, ok := d.GetOk("properties"); ok {
+		// TODO - this is just for testing. Real issue must be solved
+		time.Sleep(10 * time.Second)
+		vapp, err := vdc.FindVAppByName(d.Get("name").(string))
+		if err != nil {
+			return fmt.Errorf("unable to find vApp by name %s: %s", d.Get("name").(string), err)
+		}
+
+		vappProperties, err := getProductSectionListType(d)
```
vbauzysvmware: I think here we operate with State file, in state file we have guest properties.

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: I would also be more specific.

vcd/resource_vcd_vapp_vm.go
Didainius: We agreed to use `setXData` and `getXType` where X=struct name

lvirbalas: @dataclouder your thoughts?

vbauzysvmware: > We agreed to use `setXData` and `getXType` where X=struct name

Yes, but I think set/GetTerrraformStructure, IMHO. In govcd API struct.

Didainius: I have changed govcd method namings

vcd/resource_vcd_vapp.go
Didainius: This is no longer used and 
`err = vapp.BlockWhileStatus("UNRESOLVED", vcdClient.MaxRetryTimeout)` was introduced

vcd/resource_vcd_vapp_vm.go
Didainius: `guest_properties` it is

vcd/resource_vcd_vapp.go
Didainius: This is not removed and a solution of `err = vapp.BlockWhileStatus("UNRESOLVED", vcdClient.MaxRetryTimeout)` is put into place

website/docs/r/vapp.html.markdown
Didainius: The ugly part is that I don't find a link vCD oficial documentation about it (although this option is available in UI). Didn't want to put something like a blog link into docs https://blogs.vmware.com/vsphere/2012/06/leveraging-vapp-vm-custom-properties-in-vcloud-director.html

lvirbalas: Yes, this blog is a bit too old:
> one of the new features in the vCloud Director 1.5 release

:)

lvirbalas: The above blog has useful description how **vApp** level guest properties behave (that's the non-intuitive part):
> Custom Properties can be defined at the vApp level which will then be cascaded down to all Virtual Machines within the vApp

vcd/resource_vcd_vapp.go
Didainius: changed name

Didainius:
> My only concern that internal terraform functions `setProductSectionListData/getProductSectionListType` I would like to be called setGuestProperties/GetGuestProperties and leave production section to govcd level, as they are working with state file and in state file we have guest_properties.

Not happy about it but changed. This is killing the last drop of consistency attempt

PR #320, opened 2019-09-04
Unbind from Go 1.12 in go.mod

Didainius:
Currently `go.mod` has specified Go version as 1.12. As Go 1.13 we can remove this constraint to utilize some of the newer features (if needed).


lvirbalas:
Oops! Just noticed that Travis failed?

PR #324, opened 2019-09-06
Org VDC import, data source and read fix

vbauzysvmware:
Continuing changed for supporting data sources and imports.

* Org VDC import added
* Data source for Org VDC
* Fixed read issue with complex types


go.sum
```diff
@@ -12,6 +12,7 @@ github.com/Azure/go-autorest v10.15.4+incompatible/go.mod h1:r+4oMnoxhatjLLJ6zxS
 github.com/Azure/go-ntlmssp v0.0.0-20180810175552-4a21cbd618b4/go.mod h1:chxPXzSsl7ZWRAuOIE23GDNzjWuZquvFlgA8xmpunjU=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/ChrisTrenkamp/goxpath v0.0.0-20170922090931-c385f95c6022/go.mod h1:nuWgzSkT5PnyOd+272uUmV0dnAnAn42Mk7PiQC5VzN4=
+github.com/Didainius/go-vcloud-director/v2 v2.3.0-beta.4.0.20190725120514-394519871f9c/go.mod h1:+Hq7ryFfgZqsO6mXH29RQFnpIMSujCOMI57otHoXHhQ=
```
lvirbalas: Reminder on personal references

go.sum
```diff
@@ -310,6 +311,8 @@ github.com/tmc/grpc-websocket-proxy v0.0.0-20171017195756-830351dc03c6/go.mod h1
 github.com/ugorji/go v0.0.0-20180813092308-00b869d2f4a5/go.mod h1:hnLbHMwcvSihnDhEfx2/BzKp2xb0Y+ErdfYcrs9tkJQ=
 github.com/ulikunitz/xz v0.5.5 h1:pFrO0lVpTBXLpYw+pnLj6TbvHuyjXMfjGeCwSqCVwok=
 github.com/ulikunitz/xz v0.5.5/go.mod h1:2bypXElzHzzJZwzH67Y6wb67pO62Rzfn7BSiF4ABRW8=
+github.com/vbauzysvmware/go-vcloud-director/v2 v2.0.0-20190909085019-a4faa0b2c572 h1:dKauI+f7/xUe2Ky5O7cp9IiZep184vE3x9Cu1qz+zCU=
+github.com/vbauzysvmware/go-vcloud-director/v2 v2.0.0-20190909085019-a4faa0b2c572/go.mod h1:HonlGxbjJ1NAibWh99eE4/S2l6ZOZ5KJzKK1rh2a9vc=
```
lvirbalas: Reminder on personal references

vcd/datasource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+  metadata = {
+    vdc_metadata = "VDC Metadata"
+  }
+
+  enabled                  = true
+  enable_thin_provisioning = true
+  enable_fast_provisioning = true
+  delete_force             = true
+  delete_recursive         = true
```
lvirbalas: Do these fields get read out of vCD?

```
  delete_force             = true
  delete_recursive         = true
```

vcd/resource_vcd_org_vdc.go
```diff
@@ -297,10 +302,15 @@ func resourceVcdVdcRead(d *schema.ResourceData, meta interface{}) error {
 	adminVdc, err := adminOrg.GetAdminVDCByName(d.Get("name").(string), false)
 	if err != nil {
 		log.Printf("[DEBUG] Unable to find VDC")
-		return fmt.Errorf("unable to find VDC %#v", err)
+		return fmt.Errorf("unable to find VDC2 %s", err)
```
lvirbalas: What's "VDC2"?

vcd/resource_vcd_org_vdc.go
```diff
@@ -269,7 +274,7 @@ func resourceVcdVdcCreate(d *schema.ResourceData, meta interface{}) error {
 	adminVdc, err := adminOrg.GetAdminVDCByName(d.Get("name").(string), false)
 	if err != nil {
 		log.Printf("[DEBUG] Unable to find vdc.")
-		return fmt.Errorf("unable to find vdc.. %#v", err)
+		return fmt.Errorf("unable to find VDC.. %#v", err)
```
lvirbalas: Do we need the ".."?

vcd/resource_vcd_org_vdc.go
```diff
@@ -323,14 +333,23 @@ func resourceVcdVdcRead(d *schema.ResourceData, meta interface{}) error {
...
+
+	if err := d.Set("storage_profile", storageProfileStateData); err != nil {
+		return fmt.Errorf("error setting compute_capacity: %s", err)
+	}
 
 	vdc, err := adminOrg.GetVDCByName(d.Get("name").(string), false)
 	if err != nil {
 		log.Printf("[DEBUG] Unable to find VDC")
-		return fmt.Errorf("unable to find VDC %#v", err)
+		return fmt.Errorf("unable to find VDC3 %s", err)
```
lvirbalas: What's with the numbers? "VDC3"...

website/docs/d/org_vdc.html.markdown
```diff
@@ -0,0 +1,70 @@
+---
+layout: "vcd"
+page_title: "vCloudDirector: vcd_org_vdc"
+sidebar_current: "docs-vcd-data-source-org-vdc"
+description: |-
+  Provides a organization VDC data source.
```
lvirbalas: "a organization" -> "an organization"

website/docs/d/org_vdc.html.markdown
```diff
@@ -0,0 +1,70 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional, but required if not set at provider level) Org name 
+* `name` - (Required) Organization VDC name
+
+## Attribute Reference
```
lvirbalas: This section is somewhat misleading, because it states that some parameters are required, while some are optional. At the end, it just duplicates what resource has. Hence, what about not listing them at all, like we do in other data source docs? I mean just giving the link/reference:
> All the attributes defined in ...

If we decide to do this, we should do this in all data sources. @Didainius and @dataclouder your thoughts?

go.sum
vbauzysvmware: This will be removed after real tag happens

go.sum
vbauzysvmware: This will be removed after real tag happens

vcd/datasource_vcd_org_vdc_test.go
vbauzysvmware: no, this only from user input

vcd/resource_vcd_org_vdc.go
vbauzysvmware: fixed

vcd/resource_vcd_org_vdc.go
vbauzysvmware: fixed

vcd/resource_vcd_org_vdc.go
vbauzysvmware: fixed

website/docs/d/org_vdc.html.markdown
vbauzysvmware: fixed

vcd/datasource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+	datasourceVdc := "vcd_org_vdc.asDatasource"
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { preRunChecks(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckVdcDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config:      configText,
+				ExpectError: regexp.MustCompile(`After applying this step and refreshing, the plan was not empty`),
```
dataclouder: Please explain why we expect this error.

vcd/datasource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  enabled                  = true
+  enable_thin_provisioning = true
+  enable_fast_provisioning = true
+  delete_force             = true
+  delete_recursive         = true
+}
+
+data "vcd_org_vdc" "asDatasource"{
```
dataclouder: This is a hackish way of using a data source. We can use a data source for an existing resource, but we shouldn't.
The test should use a data source for an existing entity, without defining a resource for it.

CHANGELOG.md
```diff
@@ -23,12 +24,14 @@ IMPROVEMENTS:
 * `resource/vcd_vapp` supports guest properties [#319]
 * `resource/vcd_vapp_vm` supports guest properties [#319]
 * Upgrade Terraform SDK dependency to 0.12.6 [#302]
+* `vcd_org_vdc` Add import capability - ([#323])
```
Didainius: We need to agree on using or not `()` for enclosing GitHub issue IDs. Half of the items have it, have don't.

vcd/datasource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,239 @@
...
+		PreCheck:     func() { preRunChecks(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckVdcDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config:      configText,
+				ExpectError: regexp.MustCompile(`After applying this step and refreshing, the plan was not empty`),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckVcdVdcExists("vcd_org_vdc."+vdcName),
+					resource.ComposeTestCheckFunc(testAccDataSourceVcdOrgVdc("data."+datasourceVdc, vdcName)),
```
Didainius: Any reason why we are not using the regular `resource.TestCheckResourceAttr...` and similar functions here? if you need to compare values between two objects `resource.TestCheckResourceAttrPair` can be helpful

vcd/datasource_vcd_org_vdc_test.go
Didainius: But probably depends if we have it. If we don't it is cheaper to set it up such way (although in practice it wouldn't be useful in any way)

vcd/resource_vcd_org_vdc.go
```diff
@@ -269,7 +274,7 @@ func resourceVcdVdcCreate(d *schema.ResourceData, meta interface{}) error {
 	adminVdc, err := adminOrg.GetAdminVDCByName(d.Get("name").(string), false)
 	if err != nil {
 		log.Printf("[DEBUG] Unable to find vdc.")
-		return fmt.Errorf("unable to find vdc.. %#v", err)
+		return fmt.Errorf("unable to find VDC. %#v", err)
```
Didainius: No need for interface here, it doesn't show more information, just wraps it with additional structures which is less user friendly.
```suggestion
		return fmt.Errorf("unable to find VDC. %s", err)
```

website/docs/d/org_vdc.html.markdown
Didainius: Both ways could work for me although I am more of a fan for referencing to resource page. Our "read" functions force the coupling between resource and data source parameters and if we have attributes here and in resource it will make us maintain documentation in two places, which is easy to diverge if we need to change/update something.

vcd/datasource_vcd_org_vdc.go
```diff
@@ -0,0 +1,218 @@
...
+				Computed:    true,
+				Description: "Set to false to disallow creation of the VDC if the AllocationModel is AllocationPool or ReservationPool and the ComputeCapacity you specified is greater than what the backing Provider VDC can supply. Default is true.",
+			},
+			"enable_vm_discovery": &schema.Schema{
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "True if discovery of vCenter VMs is enabled for resource pools backing this VDC. If left unspecified, the actual behaviour depends on enablement at the organization level and at the system level.",
+			},
+
+			"delete_force": &schema.Schema{
```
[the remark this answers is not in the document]
vbauzysvmware: Right. Removed

website/docs/d/org_vdc.html.markdown
vbauzysvmware: Well, in this case I updated doc to be inline with others. Also there little difference between normal resource - `delete_recursive` and `delete_force` aren't part of data source as they are user input.

vcd/datasource_vcd_org_vdc.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	return &schema.Resource{
+		Read: datasourceVcdOrgVdcRead,
+
+		Schema: map[string]*schema.Schema{
+			"org": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
```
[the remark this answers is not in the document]
vbauzysvmware: Fixed

vcd/datasource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,239 @@
...
+		"MemoryGuaranteed": "1",
+		"CpuGuaranteed":    "1",
+	}
+
+	if !usingSysAdmin() {
+		t.Skip("TestAccVcdVdcDatasource requires system admin privileges")
+		return
+	}
+
+	params["FuncName"] = t.Name() + "-Datasource"
```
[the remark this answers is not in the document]
vbauzysvmware: I haven't dig deep, but somehow clashed with resource test name.

website/docs/d/org_vdc.html.markdown
```diff
@@ -0,0 +1,70 @@
...
+* `enable_fast_provisioning` - (Optional) Request fast provisioning. Request will be honored only if the underlying datastore supports it. Fast provisioning can reduce the time it takes to create virtual machines by using vSphere linked clones. If you disable fast provisioning, all provisioning operations will result in full clones.
+* `network_pool_name` - (Optional) Reference to a network pool in the Provider VDC. Required if this VDC will contain routed or isolated networks.
+* `allow_over_commit` - (Optional) Set to false to disallow creation of the VDC if the AllocationModel is AllocationPool or ReservationPool and the ComputeCapacity you specified is greater than what the backing Provider VDC can supply. Default is true.
+* `enable_vm_discovery` - (Optional) If true, discovery of vCenter VMs is enabled for resource pools backing this VDC. If false, discovery is disabled. If left unspecified, the actual behaviour depends on enablement at the organization level and at the system level.
+* `delete_force` - (Required) When destroying use `delete_force=True` to remove a VDC and any objects it contains, regardless of their state.
```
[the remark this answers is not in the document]
vbauzysvmware: removed. Tx for noticing.

vcd/datasource_vcd_org_vdc_test.go
vbauzysvmware: Yes it is problem `if we have it`, so I create to have it.

CHANGELOG.md

vcd/resource_vcd_org_vdc.go
vbauzysvmware: fixed

vcd/datasource_vcd_org_vdc_test.go
vbauzysvmware: Cause I getting this with testing datasource:
```

File: vcd/datasource_vcd_org_vdc_test.go (hunk 12 above)
Comment (vbauzysvmware) [reply vcd/datasource_vcd_org_vdc_test.go]: Moved what works with TestCheckResourceAttrPair

File: vcd/datasource_vcd_org_vdc_test.go (hunk 10 above)
Comment (dataclouder) [reply vcd/datasource_vcd_org_vdc_test.go]: Our test suite requires a VDC. Thus we have it, and we can make a data source for it. Similar for what we do for Org and catalog.

Comment (vbauzysvmware) [reply vcd/datasource_vcd_org_vdc_test.go]: Yes, that's true, but in that way I have to expect everywhere it will be same and hardcode values which I don't have control.

File: vcd/datasource_vcd_org_vdc_test.go (hunk 9 above)
Comment (dataclouder) [reply vcd/datasource_vcd_org_vdc_test.go]: This doesn't look OK. It seems that the read function doesn't do what it should.
Need to investigate.

File: vcd/datasource_vcd_org_vdc_test.go (hunk 10 above)
Comment (dataclouder) [reply vcd/datasource_vcd_org_vdc_test.go]: Please have a look at `TestAccVcdDatasourceOrg` or `TestAccVcdDNAT_WithExtNetw`.
There is no need for either hard coding or knowing all the values in advance.

File: CHANGELOG.md (hunk 11 above)
Comment (lvirbalas) [reply CHANGELOG.md]: My vote for no "()"

File: vcd/datasource_vcd_org_vdc_test.go (hunk 9 above)
Comment (vbauzysvmware) [reply vcd/datasource_vcd_org_vdc_test.go]: Removed expected error

File: vcd/datasource_vcd_org_vdc_test.go (hunk 10 above)
Comment (vbauzysvmware) [reply vcd/datasource_vcd_org_vdc_test.go]: Changed accordingly

**Comment by vbauzysvmware** [comment]
> Typo in "exixtingVdc" in this commit.

Fixed

**Comment by vbauzysvmware** [comment]
I changed test to extract Set hashmaps and use them to access fields. Overall now all fields are checked.
One minus is that test complain that plan isn't empty. I don't see any issue where - my guess don't like Sets or computed fields. 

```

vbauzysvmware:
Latest change by removing custom hash functionality and using internal. In test I haven't calculated hash values, but parsed as calculate hash in test would need create all structure of Schema and pass for hash function which is possible, but complicated.

vbauzysvmware:
Added convertion changes from typeSet to typeList

website/docs/d/org_vdc.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+layout: "vcd"
+page_title: "vCloudDirector: vcd_org_vdc"
+sidebar_current: "docs-vcd-data-source-org-vdc"
+description: |-
+  Provides a organization VDC data source.
+---
+
+# vcd\_org\_vcd
+
+Provides a vCloud Director Organization VDC data source. A Organization VDC can be used to reference a VCD and use its
```
Didainius: Just as above. "an organization VDC can be used"

vcd/resource_vcd_org_vdc.go
```diff
@@ -757,3 +835,43 @@ func getStorageProfileHREF(vcdClient *VCDClient, name string) (string, error) {
...
+		return nil, fmt.Errorf(errorRetrievingOrg, err)
+	}
+
+	adminVdc, err := adminOrg.GetAdminVDCByName(vdcName, false)
+	if err != nil {
+		log.Printf("[DEBUG] Unable to find VDC")
+		return nil, fmt.Errorf("unable to find VDC %s", err)
+	}
+
+	d.Set("org", orgName)
```
Didainius: No biggie, just linters complaining
```suggestion
	_ = d.Set("org", orgName)
```

vcd/resource_vcd_org_vdc.go
```diff
@@ -757,3 +835,43 @@ func getStorageProfileHREF(vcdClient *VCDClient, name string) (string, error) {
...
+	d.Set("name", vdcName)
```
Didainius:
```suggestion
	_ = d.Set("name", vdcName)
```

vcd/api_test.go
```diff
@@ -154,3 +154,20 @@ func getMajorVersion() string {
 	// Where the first element is the full text matched, and the second one is the first captured text
 	return versionList[0][1]
 }
+
+// getHashValuesFromKey returns the hash values of a TypeSet Key
+// Given "parent.1234.child.5678.something", it will return "1234" and "5678"
+// for Example: compute_capacity.315866465.memory.508945747.limit
+//              (returns 315866465 and 508945747)
+func getHashValuesFromKey(stateFileMap map[string]string, parentKey, childKey string) (string, string, error) {
```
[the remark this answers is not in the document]
Didainius: I guess now this gets to a question if this is used at all. Although it may come in handy in some other place in future.

vcd/datasource_vcd_org_vdc_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+package vcd
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccVcdVdcDatasource(t *testing.T) {
```
Didainius: I get the impression, that this test does not test the data source and its fields, but vice versa. That is - it uses a data source to generate some vdc config and checks that the resource is provisioned as specified from the data source. Wondering if this can mask some issues with data source itself (for example if it does not retrieve some fields correctly but resource provisioning still works)

vcd/resource_vcd_org_vdc.go
```diff
@@ -79,14 +82,15 @@ func resourceVcdOrgVdc() *schema.Resource {
...
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"cpu":    &capacityWithUsage,
 						"memory": &capacityWithUsage,
 					},
 				},
 				Description: "The compute capacity allocated to this VDC.",
+				//Set:         hashMapStringForCapacity,
```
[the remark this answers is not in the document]
vbauzysvmware: fixed

CHANGELOG.md
```diff
@@ -22,6 +23,8 @@ IMPROVEMENTS:
 * `resource/vcd_vapp_vm` allows to force guest customization [#310]
 * `resource/vcd_vapp` supports guest properties [#319]
 * `resource/vcd_vapp_vm` supports guest properties [#319]
+* Upgrade Terraform SDK dependency to 0.12.6 [#302]
+* `vcd_org_vdc` Add import capability - [#324]
```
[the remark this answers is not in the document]
vbauzysvmware: Add note

website/docs/d/org_vdc.html.markdown
vbauzysvmware: fixed

vcd/resource_vcd_org_vdc.go
vbauzysvmware: fixed

vcd/resource_vcd_org_vdc.go
vbauzysvmware: fixed

vcd/api_test.go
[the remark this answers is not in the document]
vbauzysvmware: removed

vcd/datasource_vcd_org_vdc_test.go
vbauzysvmware: This test according Giuseppes recommendations, his done previously his investigation. It's best approach though it has own issues.

vcd/datasource_vcd_org_vdc.go
```diff
@@ -0,0 +1,206 @@
...
+			"metadata": {
+				Type:        schema.TypeMap,
+				Computed:    true,
+				Description: "Key and value pairs for Org VDC metadata",
+			},
+		},
+	}
+}
+
+func datasourceVcdOrgVdcRead(d *schema.ResourceData, meta interface{}) error {
```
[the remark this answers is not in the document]
vbauzysvmware: cause they a little different due ID handling part, same issue everywhere - we figure out that with @Didainius  already, I asked him many times :-D

vcd/datasource_vcd_org_vdc_test.go
vbauzysvmware: Sorry don't get idea - what is difference if I would check : resource.TestCheckResourceAttr("vcd_org_vdc."+vdcName, "provider_vdc_name", testConfig.VCD.ProviderVdc.Name)
Why we would need output?

dataclouder: Your approach is equivalent. 
Using `output` shows explicitly that we are accessing the data source in the HCL script.

vcd/resource_vcd_org_vdc.go
```diff
@@ -757,3 +835,43 @@ func getStorageProfileHREF(vcdClient *VCDClient, name string) (string, error) {
...
+// (the resource must be already defined in .tf config otherwise `terraform import` will complain)
+// 5. `terraform refresh` is being implicitly launched. The Read method looks up all other fields
+// based on the known ID of object.
+//
+// Example resource name (_resource_name_): vcd_org_vdc.my_existing_vdc
+// Example import path (_the_id_string_): org.my_existing_vdc
+func resourceVcdOrgVdcImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	resourceURI := strings.Split(d.Id(), ".")
+	if len(resourceURI) != 2 {
+		return nil, fmt.Errorf("resource name must be specified as org.vdc.my_existing_vdc")
```
dataclouder: There is an extra element in the example path: 
`"resource name must be specified as org.vdc.my_existing_vdc"` should be
`"resource name must be specified as org.my_existing_vdc"`

website/docs/d/org_vdc.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+Supported in provider *v2.5+*
+
+## Example Usage
+
+```hcl
+data "vcd_org_vdc" "my-org-vdc" {
+  org     = "my-org"
+  name    = "my-vdc"
+}
+```
```
dataclouder: Please add a few `output` statements to show how to reference the data from this source

vcd/resource_vcd_org_vdc.go
vbauzysvmware: good catch, fixed

website/docs/d/org_vdc.html.markdown
vbauzysvmware: Added

PR #334, opened 2019-09-20
Computername fix

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/273

This is fix for an issue. Overall decision was made to introduce new parameter `computer_name` which explicitly sets computer name for VM. Existing implementation was left for back compatibility where computer name was set using `name`, but only when `initscript` was provided.
Also to help our customer warning message was added.
Note: Read haven't be added as Giuseppe is working on read, datasource and import capabilities.


vcd/resource_vcd_vapp_vm.go
```diff
@@ -33,6 +33,10 @@ func resourceVcdVAppVm() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+			"computer_name": &schema.Schema{
```
dataclouder: Is this the host name of the VM?

vcd/resource_vcd_vapp_vm.go
```diff
@@ -88,6 +92,8 @@ func resourceVcdVAppVm() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
+				ValidateFunc: anyValueWarningValidator(true,
+					"With next version `name` as computer name won't be set together with `initscript`. Please use `computer_name`."),
```
dataclouder: Which version are we talking about?
It is unclear what behavior we are promising now as opposed to later

vcd/resource_vcd_vapp_vm_test.go
```diff
@@ -54,6 +55,8 @@ func TestAccVcdVAppVm_Basic(t *testing.T) {
 					testAccCheckVcdVAppVmExists(vappName2, vmName, "vcd_vapp_vm."+vmName, &vapp, &vm),
 					resource.TestCheckResourceAttr(
 						"vcd_vapp_vm."+vmName, "name", vmName),
+					resource.TestCheckResourceAttr(
```
dataclouder: I think the read function should be updated just enough to deal with this field, or else the test is meaningless

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: yes
image

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: We need dedication, I could change to 2.6, but would need @lvirbalas confirmation

vcd/resource_vcd_vapp_vm_test.go
vbauzysvmware: I would, but decided not, cause I would need new function in govcd to fetch VM guest customization and this changes are in your future PR.

dataclouder: If this is urgent, I suggest doing it within this PR. I still have two projects to finish before working on vApp/VM.

vcd/resource_vcd_vapp_vm.go
dataclouder: Please define what is the current behavior and what is the promised one. From this description, it is unclear.

vcd/resource_vcd_vapp_vm_test.go
vbauzysvmware: https://github.com/vmware/go-vcloud-director/pull/244

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ IMPROVEMENTS:
 * `vcd_org_vdc` Add import capability - [#324]
 * `vcd_org_vdc` Add full read capability - [#324]
 * Upgrade Terraform SDK dependency to 0.12.8 [#320]
+* `resource/vcd_vapp_vm` have new field `computer_name` [#334]
 
 BUG FIXES:
```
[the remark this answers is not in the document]
lvirbalas: Haven't we decided to leave this old behaviour intact? I.e. set computer name _only_ if initscript is not empty (and use the new `computer_name` otherwise).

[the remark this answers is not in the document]
vbauzysvmware: I left. So won't add as bug fix comment. IMHO

website/docs/r/vapp_vm.html.markdown
```diff
@@ -124,6 +124,7 @@ The following arguments are supported:
 * `vdc` - (Optional; *v2.0+*) The name of VDC to use, optional if defined at provider level
 * `vapp_name` - (Required) The vApp this VM should belong to.
 * `name` - (Required) A unique name for the VM
+* `computer_name` - (Optional; *v2.5+*) A unique name
```
lvirbalas: For clarity, I'd like to see the long description of this field, copied either from API docs or UI.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -309,15 +315,32 @@ func resourceVcdVAppVmCreate(d *schema.ResourceData, meta interface{}) error {
...
 		}
+	} else {
+		task, err := vm.Customize(computerName, "", false)
+		if err != nil {
+			return fmt.Errorf("error with applying computer name: %#v", err)
+		}
+		err = task.WaitTaskCompletion()
+		if err != nil {
+			return fmt.Errorf(errorCompletingTask, err)
+		}
```
lvirbalas: This comment is about the whole 318:343 approach. As it is right now, we're mixing the old behaviour with the new one, but I believe we should be **prioritising new behaviour** and leaving the old logic intact. In other words, we can simplify as follows:

1. Leave old code intact, only add a warning message inside it:

```
	if initScript, ok := d.GetOk("initscript"); ok {
			task, err := vm.RunCustomizationScript(d.Get("name").(string), initScript.(string))
			if err != nil {
				return fmt.Errorf("error with init script setting: %#v", err)
			}
			err = task.WaitTaskCompletion()
			if err != nil {
				return fmt.Errorf(errorCompletingTask, err)
			}
```

2. Remove "else" and just do this part if the new `computer_name` is defined:

```
	} else { // Remove and check for computer_name instead
			task, err := vm.Customize(computerName, "", false)
			if err != nil {
				return fmt.Errorf("error with applying computer name: %#v", err)
			}
			err = task.WaitTaskCompletion()
			if err != nil {
				return fmt.Errorf(errorCompletingTask, err)
			}
```

So, if the user defines the new `computer_name` we don't care about old way of setting it. If they don't, we use the old code without modifications and print a warning.

CC: @Didainius @dataclouder

vcd/resource_vcd_vapp_vm.go
lvirbalas: I believe we should be printing the warning from within the function call in that very specific use case (I just posted a comment in that part), because printing it on initscript means spamming the users who are using initscript but who don't care about computer name.

website/docs/r/vapp_vm.html.markdown
vbauzysvmware: Copied from API description.

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: Don't forget we need to leave `initscript` intact! And with your proposal we would overwrite `initscript` to "".

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: In function warning is workaround/hack which we use when I found for upload functionality. Yes we can, but downside you will not see such error on `plan` phase - only on `execution`. P.s. in general if add warning in function - we only avoid showing warning when `computer_name` is set, but otherwise always also will be showed.

vcd/resource_vcd_vapp_vm.go
lvirbalas: No, because:
"} else { // Remove and check for computer_name instead"
Idea is that you set computer_name only if it's defined. If it's undefined, you don't "overwrite" the old logic.

vbauzysvmware: With current API, to change computer_name you have to provide initscript - what's why it's implemented as it is now. In general all logic is left - no previous behaviour is removed - it works as it was working.

CHANGELOG.md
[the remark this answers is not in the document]
Didainius: Oh my. Indeed.

vcd/resource_vcd_vapp_vm.go
Didainius: I think @lvirbalas is right. The `else` block will always be triggered now and VMname will always be set.

vbauzysvmware: Yep, for this case I added change

Issue #218, opened 2019-04-10
Add `data source` support


vbauzysvmware:
REF: https://github.com/terraform-providers/terraform-provider-vcd/pull/149
According to megalord:

> That's exactly what data sources are for. I'd encourage y'all to look into how other providers have used them (I can suggest AWS and DO from personal experience). Most resources that rely on other resources require the unfriendly id (specifically an arn in the AWS case) to be passed around.

> I'll go ahead and switch to the names, but you may keep this in mind. The more resources the provider defines, the worse this problem will get.

Links to documentation:
https://www.terraform.io/docs/configuration/data-sources.html
https://www.terraform.io/docs/configuration-0-11/data-sources.html

Such behaviour would add nice capabilities to provider to query resources which are not defined in terraform and link them together using IDs and not names. Names some times gives mentioned headaches and some resources like Independent disks allows disks with same name. This means some resources names are not unique.

lvirbalas:
Would it be practical to create separate issues for different resources?
- Add Datasource for vApps
- Add Datasource for Catalogs and Catalog Items
- Add Datasource for Org VDC Networks

vbauzysvmware:
What would be visible after POC for one of resources.

dataclouder:
Here's an example of a data source implementation for `catalog_item`

```go
package vcd

import (
	"fmt"

	"github.com/hashicorp/terraform/helper/schema"
	"github.com/vmware/go-vcloud-director/v2/govcd"
)

func dataSourceVcdCatalogItem() *schema.Resource {
	return &schema.Resource{
		Read: dataSourceVcdCatalogItemRead,

		Schema: map[string]*schema.Schema{
			"name": &schema.Schema{
				Type:     schema.TypeString,
				Required: true,
			},
			"org": {
				Type:     schema.TypeString,
				Required: true,
			},
			"catalog": &schema.Schema{
				Type:     schema.TypeString,
				Required: true,
			},
			"description": &schema.Schema{
				Type:     schema.TypeString,
				Optional: true,
			},
		},
	}
}

func dataSourceVcdCatalogItemRead(d *schema.ResourceData, meta interface{}) error {
	vcdClient := meta.(*VCDClient)

	adminOrg, err := vcdClient.GetAdminOrgFromResource(d)
	if err != nil {
		return fmt.Errorf(errorRetrievingOrg, err)
	}

	catalogName := d.Get("catalog").(string)
	catalogItemName := d.Get("name").(string)
[...]

PR #339, opened 2019-09-27
Add vcd_vapp datasource and importing

dataclouder:
Add datasource vApp
Add importing for vApp
Deprecate implicit VM creation withhin vApp definition

### DIFF
--- a/vcd/datasource_vcd_vapp.go
+++ b/vcd/datasource_vcd_vapp.go
@@ -10,46 +10,42 @@ func datasourceVcdVApp() *schema.Resource {
 			"name": {
 				Type:        schema.TypeString,
 				Required:    true,
-				ForceNew:    true,
 				Description: "A name for the vApp, unique within the VDC",
 			},
 			"org": {
 				Type:     schema.TypeString,
-				Required: false,
 				Optional: true,
-				ForceNew: true,
 				Description: "The name of organization to use, optional if defined at provider " +
 					"level. Useful when connected as sysadmin working across different organizations",
 			},
 			"vdc": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				ForceNew:    true,
 				Description: "The name of VDC to use, optional if defined at provider level",
 			},
 			"description": {
 				Type:        schema.TypeString,
-				Optional:    true,
+				Computed:    true,
 				Description: "Optional description of the vApp",
 			},
 			"metadata": {
 				Type:        schema.TypeMap,
-				Optional:    true,
+				Computed:    true,
 				Description: "Key value map of metadata to assign to this vApp. Key and value can be any string.",
 			},
 			"href": {
-				Type:     schema.TypeString,
-				Computed: true,
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "vApp Hyper Reference",
 			},
 			"power_on": {
 				Type:        schema.TypeBool,
-				Optional:    true,
-				Default:     true,
+				Computed:    true,
 				Description: "A boolean value stating if this vApp should be powered on",
 			},
 			"guest_properties": {
 				Type:        schema.TypeMap,
-				Optional:    true,
+				Computed:    true,
 				Description: "Key/value settings for guest properties",
 			},
 			"status": {
--- a/vcd/datasource_vcd_vapp.go
+++ b/vcd/datasource_vcd_vapp.go
@@ -11,7 +11,7 @@ func datasourceVcdVApp() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 				ForceNew:    true,
-				Description: "A unique name for the vApp",
+				Description: "A name for the vApp, unique within the VDC",
 			},
 			"org": {
 				Type:     schema.TypeString,
--- a/vcd/datasource_vcd_vapp.go
+++ b/vcd/datasource_vcd_vapp.go
@@ -1,0 +1,71 @@
+package vcd
+
+import "github.com/hashicorp/terraform/helper/schema"
+
+func datasourceVcdVApp() *schema.Resource {
+	return &schema.Resource{
+		Read: datasourceVcdVAppRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "A unique name for the vApp",
+			},
+			"org": {
+				Type:     schema.TypeString,
+				Required: false,
+				Optional: true,
+				ForceNew: true,
+				Description: "The name of organization to use, optional if defined at provider " +
+					"level. Useful when connected as sysadmin working across different organizations",
+			},
+			"vdc": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "The name of VDC to use, optional if defined at provider level",
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Optional description of the vApp",
+			},
+			"metadata": {
+				Type:        schema.TypeMap,
+				Optional:    true,
+				Description: "Key value map of metadata to assign to this vApp. Key and value can be any string.",
+			},
+			"href": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"power_on": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "A boolean value stating if this vApp should be powered on",
+			},
+			"guest_properties": {
+				Type:        schema.TypeMap,
+				Optional:    true,
+				Description: "Key/value settings for guest properties",
+			},
+			"status": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "Shows the status code of the vApp",
+			},
+			"status_text": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Shows the status of the vApp",
+			},
+		},
+	}
+}
+
+func datasourceVcdVAppRead(d *schema.ResourceData, meta interface{}) error {
+	return genericVcdVAppRead(d, meta, "datasource")
+}
--- a/vcd/datasource_vcd_vapp_test.go
+++ b/vcd/datasource_vcd_vapp_test.go
@@ -108,6 +108,10 @@ output "href" {
   value = data.vcd_vapp.{{.VappName}}.href
 }
 
+output "status" {
+  value = data.vcd_vapp.{{.VappName}}.status
+}
+
 output "status_text" {
   value = data.vcd_vapp.{{.VappName}}.status_text
 }
--- a/vcd/datasource_vcd_vapp_test.go
+++ b/vcd/datasource_vcd_vapp_test.go
@@ -1,0 +1,114 @@
+// +build vapp ALL functional
+
+package vcd
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/vmware/go-vcloud-director/v2/govcd"
+)
+
+// getAvailableVapp collects one available Vapp to use in data source tests
+func getAvailableVapp() (*govcd.VApp, error) {
+	// Get the data from configuration file. This client is still inactive at this point
+	vcdClient, err := getTestVCDFromJson(testConfig)
+	if err != nil {
+		return nil, fmt.Errorf("error getting client configuration: %s", err)
+	}
+	err = vcdClient.Authenticate(testConfig.Provider.User, testConfig.Provider.Password, testConfig.Provider.SysOrg)
+	if err != nil {
+		return nil, fmt.Errorf("authentication error: %s", err)
+	}
+	org, err := vcdClient.GetAdminOrgByName(testConfig.VCD.Org)
+	if err != nil {
+		return nil, fmt.Errorf("org not found : %s", err)
+	}
+	vdc, err := org.GetVDCByName(testConfig.VCD.Vdc, false)
+	if err != nil {
+		return nil, fmt.Errorf("vdc not found : %s", err)
+	}
+
+	for _, resourceEntities := range vdc.Vdc.ResourceEntities {
+		for _, resourceReference := range resourceEntities.ResourceEntity {
+			if resourceReference.Type == "application/vnd.vmware.vcloud.vApp+xml" {
+				return vdc.GetVAppByHref(resourceReference.HREF)
+			}
+		}
+	}
+
+	return nil, fmt.Errorf("no vApp found in VDC %s", testConfig.VCD.Vdc)
+}
+
+// TestAccVcdVappDS tests a vApp data source if a vApp is found in the VDC
+func TestAccVcdVappDS(t *testing.T) {
+	// This test requires access to the vCD before filling templates
+	// Thus it won't run in the short test
+	if vcdShortTest {
+		t.Skip(acceptanceTestsSkipped)
+		return
+	}
+
+	vapp, err := getAvailableVapp()
+	if err != nil {
+		t.Skip("No suitable vApp found for this test")
+		return
+	}
+
+	var params = StringMap{
+		"Org":      testConfig.VCD.Org,
+		"VDC":      testConfig.VCD.Vdc,
+		"VappName": vapp.VApp.Name,
+		"FuncName": "TestVappDS",
+		"Tags":     "vapp",
+	}
+	configText := templateFill(datasourceTestVapp, params)
+	debugPrintf("#[DEBUG] CONFIGURATION: %s", configText)
+
+	statusText, err := vapp.GetStatus()
+	if err != nil {
+		statusText = vAppUnknownStatus
+	}
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: configText,
+				Check: resource.ComposeTestCheckFunc(
+
+					resource.TestCheckOutput("name", vapp.VApp.Name),
+					resource.TestCheckOutput("description", vapp.VApp.Description),
+					resource.TestCheckOutput("href", vapp.VApp.HREF),
+					resource.TestCheckOutput("status_text", statusText),
+					resource.TestCheckResourceAttr("data.vcd_vapp."+vapp.VApp.Name, "status", fmt.Sprintf("%d", vapp.VApp.Status)),
+				),
+			},
+		},
+	})
+}
+
+const datasourceTestVapp = `
+data "vcd_vapp" "{{.VappName}}" {
+  name             = "{{.VappName}}"
+  org              = "{{.Org}}"
+  vdc              = "{{.VDC}}"
+}
+
+output "name" {
+  value = data.vcd_vapp.{{.VappName}}.name
+}
+
+output "description" {
+  value = data.vcd_vapp.{{.VappName}}.description
+}
+
+output "href" {
+  value = data.vcd_vapp.{{.VappName}}.href
+}
+
+output "status_text" {
+  value = data.vcd_vapp.{{.VappName}}.status_text
+}
+`
--- a/vcd/resource_vcd_vapp.go
+++ b/vcd/resource_vcd_vapp.go
@@ -552,6 +552,12 @@ func genericVcdVAppRead(d *schema.ResourceData, meta interface{}, origin string)
 		return fmt.Errorf("unable to set guest properties in state: %s", err)
 	}
 
+	statusText, err := vapp.GetStatus()
+	if err != nil {
+		statusText = vAppUnknownStatus
+	}
+	_ = d.Set("status", vapp.VApp.Status)
+	_ = d.Set("status_text", statusText)
 	// Power status is not easy to define.
 	// It should be set when status == 4, but even when we request it the status
 	// change may not happen until late. Returning the power status at an early
@@ -561,14 +567,9 @@ func genericVcdVAppRead(d *schema.ResourceData, meta interface{}, origin string)
 	// * status (numeric status code, such as "4")
 	// * status_text (status as a string, such as "POWERED_ON")
 	// _ = d.Set("power_on", vapp.VApp.Status == 4)
+
 	_ = d.Set("href", vapp.VApp.HREF)
 	_ = d.Set("description", vapp.VApp.Description)
-	statusText, err := vapp.GetStatus()
-	if err != nil {
-		statusText = vAppUnknownStatus
-	}
-	_ = d.Set("status", vapp.VApp.Status)
-	_ = d.Set("status_text", statusText)
 	metadata, err := vapp.GetMetadata()
 	if err != nil {
 		return fmt.Errorf("[vapp read] error retrieving metadata: %s", err)
--- a/vcd/resource_vcd_vapp.go
+++ b/vcd/resource_vcd_vapp.go
@@ -11,7 +11,7 @@ import (
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 )
 
-const implicitVmInVappDeprecation = "Creation of vApp and an implicit single VM in the same structure is deprecated"
+const implicitVmInVappDeprecation = "Creation of vApp and an implicit single VM in the same structure is deprecated."
 const vAppUnknownStatus = "-unknown-status-"
 
 func resourceVcdVApp() *schema.Resource {
@@ -29,7 +29,7 @@ func resourceVcdVApp() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 				ForceNew:    true,
-				Description: "A unique name for the vApp",
+				Description: "A name for the vApp, unique withing the VDC",
 			},
 			"org": {
 				Type:     schema.TypeString,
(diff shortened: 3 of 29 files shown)


vcd/datasource_vcd_vapp_test.go
```diff
@@ -0,0 +1,114 @@
...
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: configText,
+				Check: resource.ComposeTestCheckFunc(
+
+					resource.TestCheckOutput("name", vapp.VApp.Name),
+					resource.TestCheckOutput("description", vapp.VApp.Description),
+					resource.TestCheckOutput("href", vapp.VApp.HREF),
+					resource.TestCheckOutput("status_text", statusText),
+					resource.TestCheckResourceAttr("data.vcd_vapp."+vapp.VApp.Name, "status", fmt.Sprintf("%d", vapp.VApp.Status)),
```
lvirbalas: Could we test for all/more parameters? I'm especially concerned about `guest_properties`.

vcd/resource_vcd_vapp.go
```diff
@@ -4,24 +4,32 @@ import (
...
 
 		Schema: map[string]*schema.Schema{
 			"name": {
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "A unique name for the vApp",
```
lvirbalas: We should write unique at which scope, like Org VDC or whole Org?

vcd/resource_vcd_vapp.go
```diff
@@ -33,81 +41,112 @@ func resourceVcdVApp() *schema.Resource {
...
 			},
 			"network_name": {
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Name of the network this vApp should join",
+				Deprecated:  implicitVmInVappDeprecation,
```
lvirbalas: Why is network being deprecated? Please see in vCD API docs and/or UI if you can connect/"assign" an empty vApp to a network without a VM.

CC: @Didainius @vbauzysvmware

vcd/resource_vcd_vapp.go
```diff
@@ -33,81 +41,112 @@ func resourceVcdVApp() *schema.Resource {
...
 			},
 			"ip": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "The IP to assign to this vApp",
+				Deprecated:  implicitVmInVappDeprecation,
```
lvirbalas: Similar concern to the above.

vcd/resource_vcd_vapp.go
```diff
@@ -4,24 +4,32 @@ import (
...
 	"log"
 	"regexp"
+	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/vmware/go-vcloud-director/v2/govcd"
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 )
 
+const implicitVmInVappDeprecation = "Creation of vApp and an implicit single VM in the same structure is deprecated"
```
lvirbalas: There is an issue with providing a **generic** deprecation message to all the fields. In addition to the above message, we need to print "the new alternative" - i.e. what to use instead of the deprecated field. For example, if we really deprecate the `network_name` field (see concerns about it below), then we need to point the user to which parameter to use in the VM resource **specifically**.

All in all, all the deprecation messages should contain both the above **generic** statement + the **specific** statement on what to use instead.

CC: @Didainius @vbauzysvmware

vcd/resource_vcd_vapp.go
```diff
@@ -282,7 +323,7 @@ func resourceVcdVAppCreate(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
-	d.SetId(d.Get("name").(string))
+	d.SetId(vapp.VApp.ID)
```
lvirbalas: Looks like we're switching from saving name to ID. Won't these cause unexpected regressions? If not, we still need to tell this in the change log as we did with others.

vcd/resource_vcd_vapp.go
```diff
@@ -439,18 +480,32 @@ func resourceVcdVAppUpdate(d *schema.ResourceData, meta interface{}) error {
...
 
 	org, vdc, err := vcdClient.GetOrgAndVdcFromResource(d)
 	if err != nil {
 		return fmt.Errorf(errorRetrievingOrgAndVdc, err)
 	}
+	identifier := d.Id()
 
-	vapp, err := vdc.GetVAppByNameOrId(d.Id(), false)
+	if identifier == "" {
+		identifier = d.Get("name").(string)
```
lvirbalas: What is the purpose of this double-edged identifier? Why don't we know in the code here whether it is ID or name?

vcd/resource_vcd_vapp.go
```diff
@@ -577,3 +652,39 @@ func tryUndeploy(vapp govcd.VApp) error {
...
+// 1. The user supplies `terraform import _resource_name_ _the_id_string_` command
+// 2. `_the_id_string_` contains a dot formatted path to resource as in the example below
+// 3. The functions splits the dot-formatted path and tries to lookup the object
+// 4. If the lookup succeeds it sets the ID field for `_resource_name_` resource in statefile
+// (the resource must be already defined in .tf config otherwise `terraform import` will complain)
+// 5. `terraform refresh` is being implicitly launched. The Read method looks up all other fields
+// based on the known ID of object.
+//
+// Example resource name (_resource_name_): vcd_vapp.my-vapp
+// Example import path (_the_id_string_): org.vdc.my-vapp
```
lvirbalas: By looking at this, and the above changes to use ID, it's not clear whether this is expecting name or ID. I suggest we write `org-name.vdc-name.my-vapp-name` or something like that.

website/docs/d/vapp.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+
+## Attribute reference
+
+* `href` - The vApp Hyper Reference
+* `metadata` -  Key value map of metadata to assign to this vApp. Key and value can be any string. (Since *v2.2+* metadata is added directly to vApp instead of first VM in vApp)
+* `power_on` -  A boolean value stating if this vApp should be powered on. Default is `true`
+* `guest_properties` -  Key value map of vApp guest properties **Note** `ovf` attribute
+sets guest properties on the first VM using a legacy ability of this resource to spawn 1 VM. Please
+use resources `vcd_vapp_vm` to provision VMs.
```
lvirbalas: It's good that we have this explanation, but after reading it it's still not clear what this `guest_properties` represent. Is it guest properties of the first VM? Is it the deprecated code that is setting it? If so, why are we returning it through vApp as opposed to through VM data source (we agreed not to burden data source with deprecated fields)?

website/docs/r/vapp.html.markdown
```diff
@@ -68,6 +68,7 @@ resource "vcd_vapp_vm" "web2" {
 ## Example of vApp with single VM
 
 **Not recommended in v2.0+** : in the earlier version of the provider it was possible to define a vApp with a single VM in one resource, but it is not recommended as of *v2.0+* provider. Please define vApp and VM in separate resources instead.
+**Deprecated in 2.5**: The implicit inclusion of one VM in a vApp will be removed in the next major version.
```
lvirbalas: Please remove the last part of "will be removed in the next major version", it's enough to state it's deprecated.

website/docs/r/vapp.html.markdown
```diff
@@ -122,26 +123,58 @@ resource "vcd_vapp" "web" {
...
+  name              = "my-vapp"
+  org               = "my-org"
+  vdc               = "my-vdc"
+}
+```
+
+You can import such vapp into terraform state using this command
+
+```
+terraform import vcd_vapp.tf-vapp my-org.my-vdc.my-vapp
```
lvirbalas: For clarity, I propose adding "-name" for these references.

vcd/resource_vcd_vapp.go
Didainius: I look at this description and I think it has a bit incorrect comment. In fact this is the `network_name` to which the single VM inside vApp (old behavior) will be attached (the org network will be attached to vApp as a side effect)

Currently our `vcd_vapp_vm` resources automa**g**ically attach a network to vApp if it is not attached, but in future we should fix it with https://github.com/terraform-providers/terraform-provider-vcd/issues/264 (a breaking change once we remove that *automagic* part of attachment from vcd_vapp_vm code).
Field `network_name` wouldn't still suit it because it must accept a set of networks, not a single one. And for now I am leaning towards using IDs instead of names (this field being `org_network_ids` rather than `org_networks`). Especially that our networks will have proper IDs after Giuseppe's merge a few days ago.

lvirbalas: Not sure I agree with the use of IDs, just because it's harder for user, but otherwise good comment.

> I look at this description and I think it has a bit incorrect description. In fact this is the network_name to which the single VM inside vApp (old behavior) will be attached (the org network will be attached to vApp as a side effect)

I would like the current misleading description to be updated to the correct one. Your write-up above is almost a perfect fit for the new description. Even if it takes so many words to explain, we need to make it clear what the field does and that we are deprecating it for a reason. Also, description should be clarified in the doc page too. @dataclouder please look into this.

vcd/resource_vcd_vapp.go
dataclouder: It's unique among the vApps.

vcd/resource_vcd_vapp.go
dataclouder: We are deprecating everything that goes to the implicit VM. The network is one of such properties.
Currently, we don't assign networks to the vApp directly. We assign them to the VM, and from there it goes assigned to the vApp. We should change this behavior, but at the moment the vApp resource doesn't own the network.

vcd/resource_vcd_vapp.go
dataclouder: Ditto. The IP is for the implicit VM

vcd/resource_vcd_vapp.go
dataclouder: There is no replacement for the properties we are deprecating. The alternative is to define a VM explicitly.
Please suggest a wording that you think better conveys the meaning.

vcd/resource_vcd_vapp.go
dataclouder: I have used this method in other PRs. This allows us to use the same read function from both the resource and data source

vcd/resource_vcd_vapp.go
dataclouder: Nobody expects to use a resource ID explicitly. After 12 data sources in this release this concept should have been clear by now.

IDs are always generated and not provided manually in the scripts. I will add `-name` although I think it's unnecessary

website/docs/d/vapp.html.markdown
dataclouder: `guest properties` are assigned to the vApp, and any VM we add after that will inherit them.
This is not deprecated, and should not be

vcd/resource_vcd_vapp.go
dataclouder: For the record, I copied the descriptions from the vApp documentation page (docs/r/vapp.html.markdown). If that is not correct, we should change both.

website/docs/r/vapp.html.markdown
dataclouder: The example is using the names in the sample script. If I add `-name`, the example won't work.

vcd/resource_vcd_vapp.go
dataclouder: We have done this for all resources in which we have implemented import and data sources.
I will add to the change log

vcd/datasource_vcd_vapp_test.go
dataclouder: I've tested what is not deprecated.

Regarding `guest_properties`, the way it works now, requires a setup similar to what we do for load balancers. We need to set the guest properties, then add a VM, use a initscript to set a HTTP server, and then get the result of the guest properties from the server. Not a trivial task. This goes beyond the purpose of this PR.


cc: @Didainius

vcd/resource_vcd_vapp.go
dataclouder: Also, consider that deprecation works by showing a warning for every deprecated field. 
We are deprecating several fields, and so users that create a VM implicitly will see the same message half a dozen times. 
The alternative is to **deprecate the resource and make a new one with a different name.**

vcd/datasource_vcd_vapp_test.go
dataclouder: We have already, however, a simpler test for `guest properties` in  `resource_vcd_vapp_vm_properties_test.go` and `resource_vcd_vapp_properties_test.go`

vcd/resource_vcd_vapp.go
lvirbalas: > There is no replacement for the properties we are deprecating.

That's not true. Each of these fields have a counter part in the VM resources and we should point that out to the user.

> We are deprecating several fields, and so users that create a VM implicitly will see the same message half a dozen times.

That's one more reason for showing a customized (i.e. specific) message for each of the deprecated fields. This way the messages will be tailored.

Something like this:

```
network_name deprecated: In favor of vcd_vapp_vm.network. Creation of vApp and an implicit single VM in the same structure is deprecated.
```

This would also be in-line with what we have inside the VM resource itself:
```
			"network_name": &schema.Schema{
				ConflictsWith: []string{"network"},
				Deprecated:    "In favor of network",
				ForceNew:      true,
				Optional:      true,
			},
```

vcd/datasource_vcd_vapp_test.go
lvirbalas: > Regarding guest_properties, the way it works now, requires a setup similar to what we do for load balancers.

That's not needed here. All I'm asking is to set them and read them out to see that we have set them correctly, similar to the way we're doing for VM in vcd/resource_vcd_vapp_vm_properties_test.go but for vApp.

vcd/resource_vcd_vapp.go
lvirbalas: The question is whether the name needs to be unique across the whole Org or only inside VDC. This is useful information to figure out & put in Description and docs. Especially, given that our users query by names in our provider.

dataclouder: Will change the description to "A name for the vApp, unique withing the VDC"

vcd/datasource_vcd_vapp_test.go
dataclouder: I understood that you wanted to test the true purpose of `guest_properties` in vApp.

If we just need to check whether the properties are set, isn't what `resource_vcd_vapp_properties_test.go` does?

vcd/datasource_vcd_vapp.go
```diff
@@ -0,0 +1,71 @@
...
+				Description: "Key value map of metadata to assign to this vApp. Key and value can be any string.",
+			},
+			"href": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"power_on": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
```
Didainius: "Default" probably not needed

vcd/resource_vcd_vapp.go
```diff
@@ -485,6 +540,26 @@ func resourceVcdVAppRead(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return fmt.Errorf("unable to set guest properties in state: %s", err)
 	}
+	// _ = d.Set("power_on", vapp.VApp.Status == 4)
```
Didainius: Is this missing?

website/docs/r/vapp.html.markdown
dataclouder: done

vcd/resource_vcd_vapp.go
dataclouder: done

vcd/resource_vcd_vapp.go
dataclouder: done

website/docs/d/vapp.html.markdown
Didainius: `guest_properties` is the new one. `ovf` was the old one. This "propagation" to VMs is not done by us, this is how vCD behaves automatically.

website/docs/r/vapp.html.markdown
```diff
@@ -122,26 +123,58 @@ resource "vcd_vapp" "web" {
...
+* `network_name` - (Optional; **Deprecated**) Name of the network this vApp should join
+* `storage_profile` - (Optional)
+* `ip` - (Optional; **Deprecated**) The IP to assign to this vApp. Must be an IP address or
   one of dhcp, allocated or none. If given the address must be within the
   `static_ip_pool` set for the network. If left blank, and the network has
   `dhcp_pool` set with at least one available IP then this will be set with
   DHCP.
 * `metadata` - (Optional) Key value map of metadata to assign to this vApp. Key and value can be any string. (Since *v2.2+* metadata is added directly to vApp instead of first VM in vApp)
-* `ovf` - (Optional) Key value map of ovf parameters to assign to VM product section
+* `ovf` - (Optional; **Deprecated**) Key value map of ovf parameters to assign to VM product section
```
Didainius: May refer to `guest_properties` on vApp or VM so that user is not lost. Something like "use 'guest_properties' block on vApp or VM instead"

vcd/resource_vcd_vapp.go
dataclouder: This field works well as input, but not as output. The status of "POWERED_OFF" corresponds to VApp.Status == 4, but this is not always true when we check in our current tests. The vApp will be eventually set to that state, but not immediately. That's why I added `status` and `status_text`, so that users have some visibility on what the vApp is up to.

vcd/datasource_vcd_vapp.go
dataclouder: Yep, thanks. And also most of the `Optional` ones should be `Computed` instead. Fixing

website/docs/r/vapp.html.markdown
dataclouder: Well spotted. This was some leftover from copy-and-paste.  Cleaning up

dataclouder:
> Why did you remove the "Since _v2.2+_..." part? Granted it's an old release, but the message seemed valid still.

It's the data source. Everything is new in there

dataclouder:
> Description missing.

added

dataclouder:
> "explicit VM blocks" -> "explicit VM resources `vcd_vapp_vm`"

fixed

vcd/resource_vcd_vapp.go
lvirbalas: I saw you made changes to the deprecation message could you show how the latest variant looks when running `terraform plan`? Thanks.

vcd/resource_vcd_vapp.go
lvirbalas: @Didainius you have indicated a gravitation to use IDs, do you have any import where you require or will require ID soon?

vcd/resource_vcd_vapp.go
dataclouder: Here's an example:
```

  # vcd_network_routed.TestAccVcdVAppNet2 will be created
  + resource "vcd_network_routed" "TestAccVcdVAppNet2" {
      + dns1         = "8.8.8.8"
      + dns2         = "8.8.4.4"
      + edge_gateway = "gw-datacloud"
      + gateway      = "10.10.103.1"
      + href         = (known after apply)
      + id           = (known after apply)
      + name         = "TestAccVcdVAppNet2"
      + netmask      = "255.255.255.0"
      + org          = "datacloud"
      + shared       = false
      + vdc          = "vdc-datacloud"

      + dhcp_pool {
          + end_address    = "10.10.103.254"
          + max_lease_time = 7200
          + start_address  = "10.10.103.171"
        }

      + static_ip_pool {
          + end_address   = "10.10.103.170"
          + start_address = "10.10.103.2"
        }
    }

  # vcd_vapp.TestAccVcdVAppVappPowerOff will be created
  + resource "vcd_vapp" "TestAccVcdVAppVappPowerOff" {
      + accept_all_eulas = true
      + catalog_name     = "cat-datacloud"
      + cpus             = 1
      + href             = (known after apply)
      + id               = (known after apply)
      + ip               = "10.10.103.160"
      + memory           = 1024
      + metadata         = {
          + "vapp_metadata" = "vApp Metadata."
        }
      + name             = "TestAccVcdVAppVappPowerOff"
      + network_name     = "TestAccVcdVAppNet2"
      + org              = "datacloud"
      + power_on         = false
      + status           = (known after apply)
      + status_text      = (known after apply)
      + template_name    = "photon-hw11"
      + vdc              = "vdc-datacloud"
    }

Plan: 2 to add, 0 to change, 0 to destroy.

Warning: "memory": [DEPRECATED] Creation of vApp and an implicit single VM in the same structure is deprecated. Use vcd_vapp_vm.memory instead

  on config.tf line 39, in resource "vcd_vapp" "TestAccVcdVAppVappPowerOff":
  39: resource "vcd_vapp" "TestAccVcdVAppVappPowerOff" {



Warning: "ip": [DEPRECATED] Creation of vApp and an implicit single VM in the same structure is deprecated. Use vcd_vapp_vm.network instead

  on config.tf line 39, in resource "vcd_vapp" "TestAccVcdVAppVappPowerOff":
  39: resource "vcd_vapp" "TestAccVcdVAppVappPowerOff" {



Warning: "template_name": [DEPRECATED] Creation of vApp and an implicit single VM in the same structure is deprecated. Use vcd_vapp_vm.template_name instead

  on config.tf line 39, in resource "vcd_vapp" "TestAccVcdVAppVappPowerOff":
  39: resource "vcd_vapp" "TestAccVcdVAppVappPowerOff" {



Warning: "catalog_name": [DEPRECATED] Creation of vApp and an implicit single VM in the same structure is deprecated. Use vcd_vapp_vm.catalog_name instead

  on config.tf line 39, in resource "vcd_vapp" "TestAccVcdVAppVappPowerOff":
  39: resource "vcd_vapp" "TestAccVcdVAppVappPowerOff" {



Warning: "network_name": [DEPRECATED] Creation of vApp and an implicit single VM in the same structure is deprecated. Use vcd_vapp_vm.network instead

  on config.tf line 39, in resource "vcd_vapp" "TestAccVcdVAppVappPowerOff":
  39: resource "vcd_vapp" "TestAccVcdVAppVappPowerOff" {



Warning: "cpus": [DEPRECATED] Creation of vApp and an implicit single VM in the same structure is deprecated. Use vcd_vapp_vm.cpus instead

  on config.tf line 39, in resource "vcd_vapp" "TestAccVcdVAppVappPowerOff":
  39: resource "vcd_vapp" "TestAccVcdVAppVappPowerOff" {


```

vcd/datasource_vcd_vapp_test.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/vmware/go-vcloud-director/v2/govcd"
+)
+
+// getAvailableVapp collects one available Vapp to use in data source tests
+func getAvailableVapp() (*govcd.VApp, error) {
```
vbauzysvmware: What happens here? why we don't use setup prepared `firstVapp` ?

vcd/resource_vcd_vapp.go
```diff
@@ -486,6 +552,35 @@ func resourceVcdVAppRead(d *schema.ResourceData, meta interface{}) error {
...
 
+	// Power status is not easy to define.
+	// It should be set when status == 4, but even when we request it the status
+	// change may not happen until late. Returning the power status at an early
+	// stage may mislead both users and terraform, with the result of having
+	// unnecessary update requests.
+	// We have two fields that produce better information:
+	// * status (numeric status code, such as "4")
+	// * status_text (status as a string, such as "POWERED_ON")
+	// _ = d.Set("power_on", vapp.VApp.Status == 4)
```
vbauzysvmware: this line of comment isn't needed anymore?

vcd/resource_vcd_vapp.go
```diff
@@ -577,3 +672,39 @@ func tryUndeploy(vapp govcd.VApp) error {
...
+// (the resource must be already defined in .tf config otherwise `terraform import` will complain)
+// 5. `terraform refresh` is being implicitly launched. The Read method looks up all other fields
+// based on the known ID of object.
+//
+// Example resource name (_resource_name_): vcd_vapp.vapp_name
+// Example import path (_the_id_string_): org-name.vdc-name.vapp-name
+func resourceVcdVappImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	resourceURI := strings.Split(d.Id(), ".")
+	if len(resourceURI) != 3 {
+		return nil, fmt.Errorf("[network direct import] resource name must be specified as org-name.vdc-name.vapp-name")
```
vbauzysvmware: error message are wrong `network direct import`

vcd/resource_vcd_vapp.go
```diff
@@ -577,3 +672,39 @@ func tryUndeploy(vapp govcd.VApp) error {
...
+	}
+	orgName, vdcName, vappName := resourceURI[0], resourceURI[1], resourceURI[2]
+
+	vcdClient := meta.(*VCDClient)
+	_, vdc, err := vcdClient.GetOrgAndVdc(orgName, vdcName)
+	if err != nil {
+		return nil, fmt.Errorf("[network direct import] unable to find VDC %s: %s ", vdcName, err)
```
vbauzysvmware: error message are wrong `network direct import`

website/docs/d/vapp.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) A unique name for the vApp
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+
+## Attribute reference
```
vbauzysvmware: If everyone knows that `Attributes` are `Computed` ? - just thinking if should that mentioned

vcd/datasource_vcd_vapp_test.go
dataclouder: findFirstVapp is in go-vcloud-director.

vcd/resource_vcd_vapp.go
dataclouder: I left it on purpose, to point that we're talking about this property

website/docs/d/vapp.html.markdown
dataclouder: Everything except the mandatory items is computed in data source

vcd/datasource_vcd_vapp_test.go
vbauzysvmware: Does it means that this test expects that other tests create and left vApp ?

dataclouder: No. It means that if we find a vApp, we test the data source. If not, we skip the test

dataclouder:
> General asks in addition to the inline ones:
> 
>     * Please add references to the data source PRs to the covering issue for the users to know at #218
> 
>     * Need a change log entry for deprecations

Done

vcd/resource_vcd_vapp.go
lvirbalas: That's actually perfect.

website/docs/d/vapp.html.markdown
vbauzysvmware: For new users isn't obvious, I would say

vcd/resource_vcd_vapp.go
dataclouder: fixed

vcd/resource_vcd_vapp.go
dataclouder: fixed

vcd/datasource_vcd_vapp_test.go
```diff
@@ -0,0 +1,114 @@
...
+  value = data.vcd_vapp.{{.VappName}}.description
+}
+
+output "href" {
+  value = data.vcd_vapp.{{.VappName}}.href
+}
+
+output "status_text" {
+  value = data.vcd_vapp.{{.VappName}}.status_text
+}
```
lvirbalas: Please also add output of "status", given the caveat how it behaves.

PR #340, opened 2019-09-27
Add Media datasource/import and enhanced read

vbauzysvmware:
* Add Media datasource
* Add Media import
* Add a few `computed` fields to resource read
* Improved read - fixed issues
* Changed ID from constructed name to ID


vcd/catalogitem.go
```diff
@@ -55,18 +56,30 @@ func findCatalogItem(d *schema.ResourceData, vcdClient *VCDClient) (*govcd.Catal
...
+
+	catalogItem, err := catalog.GetCatalogItemByNameOrId(identifier, false)
 	if err != nil {
 		log.Printf("[DEBUG] Unable to find catalog item. Removing from tfstate")
-		return nil, fmt.Errorf("unable to find catalog item: %s", err)
+		d.SetId("")
+		return nil, nil
+	}
+
+	entityId, err := govcd.GetBareEntityUuid(catalogItem.CatalogItem.ID)
```
dataclouder: Why we can't use the simple catalogItem.ID instead of doing this transformation?

vbauzysvmware: Cause anyways we will need todo transformation when we use or refernce ID and you have to know, that you need to parse it. So this chage allow to save "384be749-2d58-41e6-b4a0-0ce407748cea" and not "urn:vcloud:media:84be749-2d58-41e6-b4a0-0ce407748cea"

dataclouder: It seems to me that we are adding complexity to terraform-provider-vcd instead of dealing with it in go-vcloud-director.

I would expect to have methods that retrieve a media item with all the elements that are needed here, and then we simply use it.
One thing I don't understand is why sometimes we look for MediaItem through a Query and sometimes we use GetCatalogItem.

vbauzysvmware: This I think is wider discussion regarding what type of ID we save and fetch from state file. 
Regarding MediaItem and CatalogItem as Media -> you can upload Media to catalog and to vdc. Also mediaitem resides as catalogitem in catalog. Its a mess already in vCD.

CHANGELOG.md
```diff
@@ -41,6 +42,8 @@ IMPROVEMENTS:
 * `vcd_org_vdc` Add import capability and full read support [#218]
 * Upgrade Terraform SDK dependency to 0.12.8 [#320]
 * `resource/vcd_vapp_vm` has new field `computer_name` [#334]
+* `resource/vcd_catalog_media` Add computed properties (is_iso, owner_name, is_published, creation_date, size, status, storage_profile_name) and full read support [#340]
```
lvirbalas: Please quote the property names.

CHANGELOG.md
```diff
@@ -41,6 +42,8 @@ IMPROVEMENTS:
...
+* `resource/vcd_catalog_media`: MediaItem state ID changed from colon separated list of catalog name and media name to vCD ID [#340]
```
lvirbalas: The colon ":" is extra - please remove it here and from a few entries above in this section to make it look consistent.

vcd/catalogitem.go
```diff
@@ -55,18 +56,30 @@ func findCatalogItem(d *schema.ResourceData, vcdClient *VCDClient) (*govcd.Catal
...
-		d.SetId("")
+		log.Printf("[DEBUG] Unable to find catalog.")
 		return nil, fmt.Errorf("unable to find catalog: %s", err)
 	}
 
-	catalogItem, err := catalog.GetCatalogItemByName(d.Get("name").(string), false)
+	identifier := d.Id()
+
+	// check if identifier still is in deprecated style `catalogName:mediaName`
+	if identifier == "" || strings.Contains(identifier, ":") {
```
lvirbalas: Is this needed for a smooth upgrade from 2.4.0 to 2.5.0?

vcd/resource_vcd_catalog_item.go
```diff
@@ -187,23 +193,10 @@ func createOrUpdateCatalogItemMetadata(d *schema.ResourceData, meta interface{})
...
-		return fmt.Errorf(errorRetrievingOrg, err)
-	}
-
-	catalog, err := adminOrg.GetCatalogByName(d.Get("catalog").(string), false)
-	if err != nil {
-		log.Printf("[DEBUG] Unable to find catalog: %s", err)
-		return nil
-	}
-
-	catalogItem, err := catalog.GetCatalogItemByName(d.Get("name").(string), false)
```
lvirbalas: Please tell more why this code became not needed after this PR?

website/docs/d/catalog_media.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+## Example Usage
+
+```hcl
+data "vcd_catalog_media" "existing-media" {
+  org     = "my-org"
+  catalog = "my-cat"
+  name    = "my-media"
+}
+
+output "gateway" {
```
lvirbalas: gateway?

website/docs/d/catalog_media.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+  value = data.vcd_catalog_media.size
+}
+
+output "dns1" {
```
lvirbalas: dns1?

website/docs/r/catalog_media.html.markdown
```diff
@@ -45,3 +45,45 @@ The following arguments are supported:
...
+
+## Importing
+
+Supported in provider *v2.5+*
+
+~> **Note:** The current implementation of Terraform import can only import resources into the state. It does not generate
+configuration. [More information.][docs-import]
+
+An existing media item can be [imported][docs-import] into this resource via supplying its path.
+The path for this resource is made of orgName.catalogName.mediaName.
```
lvirbalas: I suggest to use notion with dashes everywhere. I.e. "org-name.catalog-name.media-name" - it's just more consistent and easier to scan with eyes.

vcd/catalogitem.go
lvirbalas: > This I think is wider discussion regarding what type of ID we save and fetch from state file.

What's the question specifically?

> you can upload Media to catalog and to vdc

Please elaborate on how to upload to VDC only?

CHANGELOG.md
vbauzysvmware: done

CHANGELOG.md
vbauzysvmware: done

vcd/catalogitem.go
vbauzysvmware: Yes, cause in state file we could get old style id

vcd/resource_vcd_catalog_item.go
vbauzysvmware: Just removed repeatable code which was in already existing function

website/docs/d/catalog_media.html.markdown
vbauzysvmware: fixed

website/docs/d/catalog_media.html.markdown
vbauzysvmware: fixed

website/docs/r/catalog_media.html.markdown
vbauzysvmware: fixed

vcd/catalogitem.go
vbauzysvmware: Question what type of Id save in state file:
urn:vcloud:media:84be749-2d58-41e6-b4a0-0ce407748cea
vs
384be749-2d58-41e6-b4a0-0ce407748cea
----
Seems we have to save in extended ID style and parse UUID when we need, as NSX API requests extended ID style.

vbauzysvmware: After analysis - API allows to upload to vCD or catalog. Our current govcd supports only to catalog way.

vcd/catalogitem.go
lvirbalas: Got it. Please change/extend the comment with something like this:
```
// Check if identifier is still in deprecated style `catalogName:mediaName`
// Required for backwards compatibility as identifier has been changed to vCD ID in 2.5.0
```

website/docs/r/catalog_media.html.markdown
lvirbalas: Thanks, just one more ask: there's a dot at the end of the sentence, which may trip inexperienced user, please remove it as well.

vcd/catalogitem.go
lvirbalas: You mean VDC? Interesting! Would you share a screen shot how to do it through UI?

lvirbalas: > Seems we have to save in extended ID style and parse UUID when we need, as NSX API requests extended ID style.

And what ID styles other resources use?

vbauzysvmware: Atleast I need short one cause usually we construct api request URL with short ID.

website/docs/r/catalog_media.html.markdown
vbauzysvmware: done

vcd/catalogitem.go
vbauzysvmware: done

vcd/catalogitem.go
vbauzysvmware: > You mean VDC? Interesting! Would you share a screen shot how to do it through UI?

From UI perspective - in flash I don't have working plugin - so haven't seen myself. In HTML5 UI - found only for Tenant upload capabilities, but both media and OVF requires provide catalog.

lvirbalas: Oh, so it's API-only capability?

vbauzysvmware: I know it's in API and in UI I can't confirm due reasons of flash current compatibility and in new HTML5 I didn't found.

lvirbalas: @dataclouder @Didainius we discussed the following three points in-depth with Vaidotas and converged to the following conclusions:

1. Use the `urn:...` long ID format, because we want to be a thin layer above vCD API where possible. Granted, this adds additional `GetBareEntityUuid` calls inside the code.
2. Continue using `Query...` functions where it makes sense. `Query...` functions are by agreement those, which call upon the `/query` API endpoint. Most of them return array, but some may return a filtered single record. We already have both return types in code, but it's no issue. In general, we had an agreement that we have `Get` and `Query` function classes some time ago. In this case it's easier to use /query, because it's only a single call to get the media, as opposed to getting it through the catalog.
3. Review which media functions may need to be deprecated. There's one `Find...` function that certainly has a bug - it errors out if there are two media items with same name in two catalogs. We want to move away from it.

Next step is for Vaidotas to refactor where appropriate given the above.

dataclouder: This decision goes against our own Coding guidelines. (https://github.com/vmware/go-vcloud-director/blob/master/CODING_GUIDELINES.md)

vbauzysvmware: I already added all Get* functions by code guidelines, Query is additional one to get more data.

vcd/resource_vcd_catalog_media.go
```diff
@@ -152,19 +191,40 @@ func resourceVcdMediaRead(d *schema.ResourceData, meta interface{}) error {
...
+	if err != nil {
+		log.Printf("[DEBUG] Unable to find media item: %s", err)
+		return err
+	}
+	if catalogItem == nil {
+		log.Printf("[DEBUG] Unable to find media item: %s. Removing from tfstate", err)
+		return nil
+	}
+
+	mediaItem, err := vdc.QueryMediaImage(catalogItem.CatalogItem.Name, d.Get("catalog").(string))
```
dataclouder: This call will fail. 
There is no VDC in the media arguments, and there is no guarantee that there will be such parameter in the provider section, since we can use catalogs and media without mentioning a VDC.
Please use `Catalog.GetMedia*` to retrieve the image

vcd/datasource_vcd_catalog_media.go
```diff
@@ -0,0 +1,76 @@
...
+				Computed: true,
+			},
+			"metadata": {
+				Type:        schema.TypeMap,
+				Computed:    true,
+				Description: "Key and value pairs for catalog item metadata",
+				// For now underlying go-vcloud-director repo only supports
+				// a value of type String in this map.
+			},
+			"is_iso": &schema.Schema{
```
dataclouder: This question should have been asked when the resource was created, but let's see it here:
Why properties `is_iso` and `is_published` are strings instead of Boolean, and why is `size` a string instead of a number?

vcd/resource_vcd_catalog_media.go
vbauzysvmware: it's not an option cause Catalog.GetMedia don't return all attributes I need

vcd/resource_vcd_catalog_media.go
```diff
@@ -226,3 +286,62 @@ func createOrUpdateMediaItemMetadata(d *schema.ResourceData, meta interface{}) e
...
+	if err != nil {
+		return nil, fmt.Errorf(errorRetrievingOrg, orgName)
+	}
+
+	catalog, err := adminOrg.GetCatalogByName(catalogName, false)
+	if err != nil {
+		return nil, govcd.ErrorEntityNotFound
+	}
+
+	catalogItem, err := catalog.GetCatalogItemByName(catalogItemName, false)
```
dataclouder: Why not using `GetMedia*`?

vcd/datasource_vcd_catalog_media.go
vbauzysvmware: Cause they are computed - so no difference, but I look if there is other implications

vcd/resource_vcd_catalog_media.go
vbauzysvmware: cause I haven't updated this PR yet

vcd/catalogitem.go
lvirbalas: > This decision goes against our own Coding guidelines
It's not good that coding guidelines do not describe Query parameters. We had an agreement that for /query endpoints we keep `Query...` functions. I don't see how `Get...` functions can replace `Query...` functions, especially those which return an array. Please amend the coding guidelines or let's discuss.

dataclouder: There is no conflict here.
We agreed that the `Get*` functions are used to retrieve single entities, while `Query*` functions retrieve arrays.
If the function retrieve a single entity, let's call it `GetEntityBySomething` and use it consistently.

vbauzysvmware: Updated by new govcd changes

vcd/resource_vcd_catalog_media.go
vbauzysvmware: changed

vbauzysvmware:
Guys, updates are done. Should be now more cleaner.

vcd/datasource_vcd_catalog_media.go
```diff
@@ -0,0 +1,76 @@
...
+				Computed:    true,
+				Description: "True if this media file is in a published catalog",
+			},
+			"creation_date": &schema.Schema{
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Creation date",
+			},
+			"size": &schema.Schema{
+				Type:        schema.TypeInt,
```
lvirbalas: Curious, why here we use `schema.TypeInt`, but in indendent disk "size" is `schema.TypeFloat`? Where's the caveat?

vcd/resource_vcd_catalog_media_test.go
```diff
@@ -99,12 +125,12 @@ func testAccCheckVcdCatalogMediaExists(mediaName string, catalogItem *govcd.Cata
 			return fmt.Errorf("catalog %s does not exist (%s)", testSuiteCatalogName, err)
 		}
 
-		newCatalogItem, err := catalog.GetCatalogItemByName(catalogMediaRs.Primary.Attributes["name"], false)
+		newMedia, err := catalog.GetMediaByName(catalogMediaRs.Primary.Attributes["name"], false)
```
lvirbalas: "new" and "old" are the evil twins of naming variables. Please use precise word which describes what this is, "mediaRecord" maybe (IIRC the govcd PR right)?

vcd/datasource_vcd_catalog_media.go
vbauzysvmware: This isn't independent disk. type is ```	Size         int64            `xml:"size,attr,omitempty"` ```

vcd/resource_vcd_catalog_media_test.go
vbauzysvmware: changed

vcd/datasource_vcd_catalog_media_test.go
```diff
@@ -0,0 +1,157 @@
...
+    catalogMedia_metadata2 = "catalogMedia Metadata2"
+  }
+}
+
+data "vcd_catalog_media" "{{.NewCatalogMedia}}" {
+  org     = "{{.Org}}"
+  catalog = "{{.Catalog}}"
+  name    = "{{.CatalogMediaName}}"
+  depends_on = ["vcd_catalog_media.{{.CatalogMediaName}}"]
+}
```
dataclouder: I was initially confused by this setting, because the `NewCatalogMedia` suggests something being created, but the `data` means that we are using an existing resource. Then I noticed that the `name` element is the same as the resource above.

Since we use a media item for govcd tests, meaning that our servers already have one, why not checking whether any item exists and use it, like we do for vApps and networks?

vcd/datasource_vcd_catalog_media_test.go
```diff
@@ -0,0 +1,157 @@
...
+		Providers:    testAccProviders,
+		CheckDestroy: catalogMediaDestroyed(testSuiteCatalogName, TestCatalogMediaDS),
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config:             configText,
+				ExpectNonEmptyPlan: true,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckVcdCatalogMediaExists("vcd_catalog_media."+TestAccVcdDataSourceMedia, &media),
+					resource.TestMatchOutput("owner_name", regexp.MustCompile(`^\w*$`)),
+					resource.TestMatchOutput("creation_date", regexp.MustCompile(`^(2019|2020)-`)),
```
dataclouder: This regexp is going to fail in 2021.
If that the expected date is in the format 'YYYY-MM-DD' I suggest using `^\d{4}-\d{2}-\d{2}`

vcd/datasource_vcd_catalog_media_test.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		_, err = cat.GetMediaByName(mediaName, false)
+		if err == nil {
+			return fmt.Errorf("catalog media %s not deleted", mediaName)
+		}
+		return nil
+	}
+}
+
+func importStateIdByCatalogMedia(objectName string) resource.ImportStateIdFunc {
```
dataclouder: The function `importStateIdByCatalogItem` will return the same result. We can reuse it

vcd/resource_vcd_catalog_media.go
```diff
@@ -132,39 +171,74 @@ func resourceVcdMediaCreate(d *schema.ResourceData, meta interface{}) error {
...
-	d.SetId(catalogName + ":" + mediaName)
-
 	log.Printf("[TRACE] Catalog media created: %#v", mediaName)
 
 	err = createOrUpdateMediaItemMetadata(d, meta)
 	if err != nil {
 		return fmt.Errorf("error adding media item metadata: %s", err)
 	}
 
+	//sets Id in findCatalogItem func - there isn't ID with media type
```
dataclouder: Is this comment obsolete? There is an ID in media type

vcd/resource_vcd_catalog_media.go
```diff
@@ -132,39 +171,74 @@ func resourceVcdMediaCreate(d *schema.ResourceData, meta interface{}) error {
...
+	if err != nil {
+		log.Printf("[DEBUG] Unable to find media: %s", err)
+		return err
+	}
+
+	d.SetId(media.Media.ID)
+
+	mediaRecord, err := catalog.QueryMedia(d.Get("name").(string))
+	if err != nil {
+		log.Printf("[DEBUG] Unable to query media: %s", err)
```
dataclouder: This double request makes the code difficult to read.
I think we should change `GetMediaBy(Name|Id|Href)` to return a full media record, meaning that inside these functions we call `QueryMedia` and return a single structure containing everything we need.

vcd/resource_vcd_catalog_media.go
```diff
@@ -211,18 +291,72 @@ func createOrUpdateMediaItemMetadata(d *schema.ResourceData, meta interface{}) e
...
+	media, err := catalog.GetMediaByName(mediaName, false)
+	if err != nil {
+		return nil, govcd.ErrorEntityNotFound
+	}
+
+	_ = d.Set("org", orgName)
+	_ = d.Set("catalog", catalogName)
+	_ = d.Set("name", mediaName)
+	_ = d.Set("description", media.Media.Description)
+	d.SetId(media.Media.ID)
```
dataclouder: See my comment in the `Read` function. We are not setting several attributes because we lack the data.

website/docs/r/catalog_media.html.markdown
```diff
@@ -45,3 +45,45 @@ The following arguments are supported:
...
+  org     = "my-org"
+  catalog = "my-vdc"
+  name    = "my-media"
+}
+```
+
+You can import such catalog media into terraform state using this command
+
+```
+terraform import vcd_catalog_media.tf-mymedia my-org.my-catalog-name.my-media-name
```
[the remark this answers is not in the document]
vbauzysvmware: updated

website/docs/r/catalog_media.html.markdown
```diff
@@ -45,3 +45,45 @@ The following arguments are supported:
...
+configuration. [More information.][docs-import]
+
+An existing media item can be [imported][docs-import] into this resource via supplying its path.
+The path for this resource is made of org-name.catalog-name.media-name
+For example, using this structure, representing a media item that was **not** created using Terraform:
+
+```hcl
+resource "vcd_catalog_media" "tf-mymedia" {
+  org     = "my-org"
+  catalog = "my-vdc"
```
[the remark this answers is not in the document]
vbauzysvmware: yep, updated

vcd/datasource_vcd_catalog_media_test.go
vbauzysvmware: Tx. updated

vcd/datasource_vcd_catalog_media_test.go
vbauzysvmware: I liked to be separated-explicit, but updated

vcd/resource_vcd_catalog_media.go
vbauzysvmware: yes, removed

vcd/resource_vcd_catalog_media.go
vbauzysvmware: I would agree on having wrapper, cause not always you need `all the data` and thus create two request when you need one. And yet more votes needed for that, IMHO I am ok with this approach.

vcd/resource_vcd_catalog_media.go
vbauzysvmware: Comment in above

vcd/datasource_vcd_catalog_media_test.go
```diff
@@ -0,0 +1,147 @@
...
+		PreCheck:     func() { preRunChecks(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: catalogMediaDestroyed(testSuiteCatalogName, TestCatalogMediaDS),
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config:             configText,
+				ExpectNonEmptyPlan: true,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckVcdCatalogMediaExists("vcd_catalog_media."+TestAccVcdDataSourceMedia, &media),
+					resource.TestMatchOutput("owner_name", regexp.MustCompile(`^\w*$`)),
```
dataclouder: This regular expression means "match a string of  **zero** or more word characters". Thus it will match "abc", or "a_bc", but also the empty string "". However, it won't match "ab-c"

vcd/datasource_vcd_catalog_media_test.go
```diff
@@ -0,0 +1,147 @@
...
+					resource.TestMatchOutput("creation_date", regexp.MustCompile(`^^\d{4}-\d{2}-\d{2}.*`)),
+					resource.TestCheckOutput("status", "RESOLVED"),
+					resource.TestMatchOutput("storage_profile_name", regexp.MustCompile(`(.|\s)*\S(.|\s)*`)),
```
dataclouder: This regular expression will easily match "a" or "aaa" or " a ".
`(.|\s)*` == **zero* or more occurrence of either any character or a space. 
The only mandatory character is `\S` which means "a non-space" character. Is this what we are trying to catch?

vbauzysvmware: My aim is non empty string, I can't guess/verify anything else

vcd/datasource_vcd_catalog_media_test.go
vbauzysvmware: Do we now user input restrictions? or just check non empty string?

vcd/datasource_vcd_catalog_media_test.go
vbauzysvmware: But this create very strange dependency - you have to run govcd tests before terraform test. Also you have option ignore configuring media in govcd and test will be ignored.

vcd/datasource_vcd_catalog_media_test.go
vbauzysvmware: updated to `non-empty`

vcd/datasource_vcd_catalog_media_test.go
vbauzysvmware: updated to `non-empty`

vcd/datasource_vcd_catalog_item_test.go
```diff
@@ -96,8 +96,8 @@ func catalogItemDestroyed(catalog, itemName string) resource.TestCheckFunc {
 
 func importStateIdByCatalogItem(objectName string) resource.ImportStateIdFunc {
```
Didainius: I think existing more generic function `importStateIdOrgObject` could be used.

vcd/catalogitem.go
```diff
@@ -55,15 +56,23 @@ func findCatalogItem(d *schema.ResourceData, vcdClient *VCDClient) (*govcd.Catal
...
+		log.Printf("[DEBUG] Unable to find catalog.")
 		return nil, fmt.Errorf("unable to find catalog: %s", err)
 	}
 
-	catalogItem, err := catalog.GetCatalogItemByName(d.Get("name").(string), false)
+	identifier := d.Id()
+
+	// Check if identifier is still in deprecated style `catalogName:mediaName`
+	// Required for backwards compatibility as identifier has been changed to vCD ID in 2.5.0
+	if identifier == "" || strings.Contains(identifier, ":") {
```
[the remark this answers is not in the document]
vbauzysvmware: Adjusted, as we agreed to use full Id's

vcd/resource_vcd_catalog_media_test.go
```diff
@@ -54,12 +55,17 @@ func TestAccVcdCatalogMediaBasic(t *testing.T) {
 						"vcd_catalog_media."+TestAccVcdCatalogMedia, "metadata.mediaItem_metadata", "mediaItem Metadata"),
 					resource.TestCheckResourceAttr(
 						"vcd_catalog_media."+TestAccVcdCatalogMedia, "metadata.mediaItem_metadata2", "mediaItem Metadata2"),
+					resource.TestMatchOutput("owner_name", regexp.MustCompile(`^\S+`)),
+					resource.TestMatchOutput("creation_date", regexp.MustCompile(`^^\d{4}-\d{2}-\d{2}.*`)),
+					resource.TestCheckOutput("status", "RESOLVED"),
+					resource.TestMatchOutput("storage_profile_name", regexp.MustCompile(`^\S+`)),
+					testCheckMediaNonStringOutputs(),
```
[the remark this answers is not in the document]
vbauzysvmware: no, it isn't - i tried

vcd/resource_vcd_catalog_media_test.go
```diff
@@ -99,12 +125,12 @@ func testAccCheckVcdCatalogMediaExists(mediaName string, catalogItem *govcd.Cata
...
 
-		newCatalogItem, err := catalog.GetCatalogItemByName(catalogMediaRs.Primary.Attributes["name"], false)
+		foundMedia, err := catalog.GetMediaByName(catalogMediaRs.Primary.Attributes["name"], false)
 		if err != nil {
-			return fmt.Errorf("catalog media %s does not exist (%#v)", catalogMediaRs.Primary.ID, catalogItem.CatalogItem)
+			return fmt.Errorf("catalog media %s does not exist (%#v)", catalogMediaRs.Primary.ID, foundMedia.Media)
 		}
 
-		catalogItem = newCatalogItem
+		media = foundMedia
```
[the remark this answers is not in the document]
vbauzysvmware: removed

vcd/resource_vcd_catalog_media_test.go
```diff
@@ -153,7 +179,35 @@ const testAccCheckVcdCatalogMediaBasic = `
     mediaItem_metadata2 = "mediaItem Metadata2"
   }
 }
-`
+
+output "creation_date" {
```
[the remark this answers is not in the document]
vbauzysvmware: could you rephrase

vcd/datasource_vcd_catalog_media_test.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+data "vcd_catalog_media" "{{.NewCatalogMedia}}" {
+  org     = "{{.Org}}"
+  catalog = "{{.Catalog}}"
+  name    = "{{.CatalogMediaName}}"
+  depends_on = ["vcd_catalog_media.{{.CatalogMediaName}}"]
+}
+
+output "size" {
```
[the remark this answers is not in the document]
vbauzysvmware: This for testing to see if variable is accessible and readed

vcd/datasource_vcd_catalog_media_test.go
```diff
@@ -0,0 +1,147 @@
...
+			resource.TestStep{
+				Config:             configText,
+				ExpectNonEmptyPlan: true,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckVcdCatalogMediaExists("vcd_catalog_media."+TestAccVcdDataSourceMedia, &media),
+					resource.TestMatchOutput("owner_name", regexp.MustCompile(`^\S+`)),
+					resource.TestMatchOutput("creation_date", regexp.MustCompile(`^^\d{4}-\d{2}-\d{2}.*`)),
+					resource.TestCheckOutput("status", "RESOLVED"),
+					resource.TestMatchOutput("storage_profile_name", regexp.MustCompile(`^\S+`)),
+					testCheckMediaNonStringOutputs(),
```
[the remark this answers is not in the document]
vbauzysvmware: I needed cause Output doesn't work with other than strings

vcd/datasource_vcd_catalog_item_test.go
vbauzysvmware: No it can't as it uses org:catalogname:objectname

vbauzysvmware:
Code updated

vcd/datasource_vcd_catalog_media_test.go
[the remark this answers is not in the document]
Didainius: Ok. I guess some of the output fields are not necessary here, but I am closing this discussion now and will raise it in future as a wider topic.

vcd/datasource_vcd_catalog_media_test.go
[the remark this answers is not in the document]
Didainius: Closing this discussion now and will raise it in future as a wider topic.

vcd/resource_vcd_catalog_media_test.go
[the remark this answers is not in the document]
Didainius: Doesn't work for `output` fields. Closing it now.

vcd/resource_vcd_catalog_media_test.go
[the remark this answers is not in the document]
Didainius: I think we do not need "pass-through" `output` fields as this is introducing just more lines without any big use. Closing this discussion now and will raise it in future as a wider topic.

vcd/datasource_vcd_catalog_media_test.go
dataclouder: What I mean is that we have already a media item created and pre-installed for the purpose of using it with govcd tests. The check I suggest should look for existing media and use it for data source testing.
With our pre-configured servers, the media item is always there.

website/docs/d/catalog_media.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+
+```hcl
+data "vcd_catalog_media" "existing-media" {
+  org     = "my-org"
+  catalog = "my-cat"
+  name    = "my-media"
+}
+
+output "media_size" {
+  value = data.vcd_catalog_media.size
```
Didainius: Is this a valid data access? I think it is missing resource name `existing-media`. Also I do not think we need these `output` fields in examples. People can access them directly using `data.vcd_catalog_media.existing-media.size`

vbauzysvmware: updated

vcd/datasource_vcd_catalog_media_test.go
```diff
@@ -0,0 +1,144 @@
...
+	debugPrintf("#[DEBUG] CONFIGURATION: %s", configText)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { preRunChecks(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: catalogMediaDestroyed(testSuiteCatalogName, TestCatalogMediaDS),
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config:             configText,
+				ExpectNonEmptyPlan: true,
```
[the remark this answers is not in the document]
vbauzysvmware: updated

vcd/datasource_vcd_catalog_media_test.go
vbauzysvmware: In current configuration we don't have media reference.

PR #341, opened 2019-09-30
Add vcd_nsxv_firewall_rule resource and data source for advanced edge gateways

Didainius:
This PR adds `vcd_nsxv_firewall_rule` resource and data source. It also adds additional options to edge gateway resource to control firewall state and default rule.

A few additional notes as it is quite a big one.
* Due to UI reporting number order rather than real firewall IDs there is a command `list` command introduced into `terraform import vcd_nsxv_firewall_rule.import list@my-org.my-org-vdc.my-edge-gw` command. Investigation is in progress to find out if the rule numbering is reliable and then import command could also accept UI rules order
* Resource code has commented fields `ipsets` and `security_groups` in `source` and `destination` blocks. This is because govcd does not yet support ipsets and security groups (https://github.com/terraform-providers/terraform-provider-vcd/issues/144). A bit more investigation is needed and introducing at least `Get` functions so that these can be read. (There are some // TODO notes left to mark places where code needs to be updated to support it). The PR can probably be split into two so that it is more convenient to review.
* The import now supports only import by `id` because names can be duplicate.

Test suite passed on 9.5

### DIFF
--- a/vcd/datasource_vcd_nsxv_firewall.go
+++ b/vcd/datasource_vcd_nsxv_firewall.go
@@ -4,9 +4,9 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
-func datasourceVcdNsxvFirewall() *schema.Resource {
+func datasourceVcdNsxvFirewallRule() *schema.Resource {
 	return &schema.Resource{
-		Read: resourceVcdNsxvFirewallRead,
+		Read: resourceVcdNsxvFirewallRuleRead,
 		Schema: map[string]*schema.Schema{
 			"org": {
 				Type:     schema.TypeString,
--- a/vcd/datasource_vcd_nsxv_firewall.go
+++ b/vcd/datasource_vcd_nsxv_firewall.go
@@ -1,0 +1,217 @@
+package vcd
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func datasourceVcdNsxvFirewall() *schema.Resource {
+	return &schema.Resource{
+		Read: resourceVcdNsxvFirewallRead,
+		Schema: map[string]*schema.Schema{
+			"org": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Description: "The name of organization to use, optional if defined at provider " +
+					"level. Useful when connected as sysadmin working across different organizations",
+			},
+			"vdc": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "The name of VDC to use, optional if defined at provider level",
+			},
+			"edge_gateway": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Edge gateway name in which the firewall rule is located",
+			},
+			"rule_id": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Firewall rule ID for lookup",
+			},
+			"name": &schema.Schema{
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Firewall rule name",
+			},
+			"rule_type": &schema.Schema{
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Read only. Possible values 'user', 'internal_high'",
+			},
+			"rule_tag": &schema.Schema{
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Optional. Allows to set custom rule tag",
+			},
+			"action": &schema.Schema{
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "'accept' or 'deny'. Default 'accept'",
+			},
+			"enabled": &schema.Schema{
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Whether the rule should be enabled. Default 'true'",
+			},
+			"logging_enabled": &schema.Schema{
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Whether logging should be enabled for this rule. Default 'false'",
+			},
+			"source": {
+				MinItems: 1,
+				MaxItems: 1,
+				Computed: true,
+				Type:     schema.TypeList,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"exclude": {
+							Type:     schema.TypeBool,
+							Computed: true,
+							Description: "Rule is applied to traffic coming from all sources " +
+								"except for the excluded source. Default 'false'",
+						},
+						"ip_addresses": {
+							Type:        schema.TypeSet,
+							Computed:    true,
+							Description: "IP address, CIDR, an IP range, or the keyword 'any'",
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+						"gateway_interfaces": {
+							Type:        schema.TypeSet,
+							Computed:    true,
+							Description: "'vse', 'internal', 'external' or network name",
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+						"virtual_machine_ids": {
+							Type:        schema.TypeSet,
+							Computed:    true,
+							Description: "Set of VM IDs",
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+						"org_networks": {
+							Type:        schema.TypeSet,
+							Computed:    true,
+							Description: "Set of org network names",
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+						// "ipsets": {
+						// 	Type:        schema.TypeSet,
+						// 	Computed:    true,
+						// 	Description: "Set of IP set names",
+						// 	Elem: &schema.Schema{
+						// 		Type: schema.TypeString,
+						// 	},
+						// },
+						// "security_groups": {
+						// 	Type:        schema.TypeSet,
+						// 	Computed:    true,
+						// 	Description: "Set of security group names",
+						// 	Elem: &schema.Schema{
+						// 		Type: schema.TypeString,
+						// 	},
+						// },
+					},
+				},
+			},
+			"destination": {
+				MinItems: 1,
+				MaxItems: 1,
+				Computed: true,
+				Type:     schema.TypeList,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"exclude": {
+							Type:     schema.TypeBool,
+							Computed: true,
+							Description: "Rule is applied to traffic going to any destinations " +
+								"except for the excluded destination. Default 'false'",
+						},
+						"ip_addresses": {
+							Type:        schema.TypeSet,
+							Computed:    true,
+							Description: "IP address, CIDR, an IP range, or the keyword 'any'",
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+						"gateway_interfaces": {
+							Type:        schema.TypeSet,
+							Computed:    true,
+							Description: "'vse', 'internal', 'external' or network name",
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+						"virtual_machine_ids": {
+							Type:        schema.TypeSet,
+							Computed:    true,
+							Description: "Set of VM IDs",
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+						"org_networks": {
+							Type:        schema.TypeSet,
+							Computed:    true,
+							Description: "Set of org network names",
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+						// "ipsets": {
+						// 	Optional:    true,
+						// 	Type:        schema.TypeSet,
+						// 	Description: "Set of IP set names",
+						// 	Elem: &schema.Schema{
+						// 		Type: schema.TypeString,
+						// 	},
+						// },
+						// "security_groups": {
+						// 	Optional:    true,
+						// 	Type:        schema.TypeSet,
+						// 	Description: "Set of security group names",
+						// 	Elem: &schema.Schema{
+						// 		Type: schema.TypeString,
+						// 	},
+						// },
+					},
+				},
+			},
+			"service": {
+				Computed: true,
+				MinItems: 1,
+				Type:     schema.TypeSet,
+				Set:      resourceVcdNsxvFirewallRuleServiceHash,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"protocol": {
+							Computed: true,
+							Type:     schema.TypeString,
+						},
+						"port": {
+							Computed: true,
+							Type:     schema.TypeString,
+						},
+						"source_port": {
+							Computed: true,
+							Type:     schema.TypeString,
+						},
+					},
+				},
+			},
+		},
+	}
+}
--- a/vcd/datasource_vcd_nsxv_firewall.go
+++ b/vcd/datasource_vcd_nsxv_firewall.go
@@ -107,6 +107,8 @@ func datasourceVcdNsxvFirewall() *schema.Resource {
 								Type: schema.TypeString,
 							},
 						},
+						// TODO - ipsets and security groups need further investigation and at least
+						// "Get" capability in govcd
 						// "ipsets": {
 						// 	Type:        schema.TypeSet,
 						// 	Computed:    true,
@@ -171,6 +173,8 @@ func datasourceVcdNsxvFirewall() *schema.Resource {
 								Type: schema.TypeString,
 							},
 						},
+						// TODO - ipsets and security groups need further investigation and at least
+						// "Get" capability in govcd
 						// "ipsets": {
 						// 	Optional:    true,
 						// 	Type:        schema.TypeSet,
--- a/vcd/provider.go
+++ b/vcd/provider.go
@@ -111,7 +111,7 @@ func Provider() terraform.ResourceProvider {
 			"vcd_lb_virtual_server":  resourceVcdLBVirtualServer(),  // 2.4
 			"vcd_nsxv_dnat":          resourceVcdNsxvDnat(),         // 2.5
 			"vcd_nsxv_snat":          resourceVcdNsxvSnat(),         // 2.5
-			"vcd_nsxv_firewall_rule": resourceVcdNsxvFirewall(),     // 2.5
+			"vcd_nsxv_firewall":      resourceVcdNsxvFirewall(),     // 2.5
 		},
 
 		DataSourcesMap: map[string]*schema.Resource{
--- a/vcd/provider.go
+++ b/vcd/provider.go
@@ -111,6 +111,7 @@ func Provider() terraform.ResourceProvider {
 			"vcd_lb_virtual_server":  resourceVcdLBVirtualServer(),  // 2.4
 			"vcd_nsxv_dnat":          resourceVcdNsxvDnat(),         // 2.5
 			"vcd_nsxv_snat":          resourceVcdNsxvSnat(),         // 2.5
+			"vcd_nsxv_firewall_rule": resourceVcdNsxvFirewall(),     // 2.5
 		},
 
 		DataSourcesMap: map[string]*schema.Resource{
--- a/vcd/provider.go
+++ b/vcd/provider.go
@@ -119,7 +119,11 @@ func Provider() terraform.ResourceProvider {
 			"vcd_org_vdc":            datasourceVcdOrgVdc(),           // 2.5
 			"vcd_catalog":            datasourceVcdCatalog(),          // 2.5
 			"vcd_catalog_item":       datasourceVcdCatalogItem(),      // 2.5
+			"vcd_edgegateway":        datasourceVcdEdgeGateway(),      // 2.5
 			"vcd_external_network":   datasourceVcdExternalNetwork(),  // 2.5
+			"vcd_network_routed":     datasourceVcdNetworkRouted(),    // 2.5
+			"vcd_network_direct":     datasourceVcdNetworkDirect(),    // 2.5
+			"vcd_network_isolated":   datasourceVcdNetworkIsolated(),  // 2.5
 			"vcd_lb_service_monitor": datasourceVcdLbServiceMonitor(), // 2.4
 			"vcd_lb_server_pool":     datasourceVcdLbServerPool(),     // 2.4
 			"vcd_lb_app_profile":     datasourceVcdLBAppProfile(),     // 2.4
--- a/vcd/provider.go
+++ b/vcd/provider.go
@@ -131,6 +131,7 @@ func Provider() terraform.ResourceProvider {
 			"vcd_lb_virtual_server":  datasourceVcdLbVirtualServer(),  // 2.4
 			"vcd_nsxv_dnat":          datasourceVcdNsxvDnat(),         // 2.5
 			"vcd_nsxv_snat":          datasourceVcdNsxvSnat(),         // 2.5
+			"vcd_nsxv_firewall":      datasourceVcdNsxvFirewall(),     // 2.5
 		},
 
 		ConfigureFunc: providerConfigure,
--- a/vcd/provider.go
+++ b/vcd/provider.go
@@ -124,6 +124,7 @@ func Provider() terraform.ResourceProvider {
 			"vcd_network_routed":     datasourceVcdNetworkRouted(),    // 2.5
 			"vcd_network_direct":     datasourceVcdNetworkDirect(),    // 2.5
 			"vcd_network_isolated":   datasourceVcdNetworkIsolated(),  // 2.5
+			"vcd_vapp":               datasourceVcdVApp(),             // 2.5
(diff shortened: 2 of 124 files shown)


vcd/datasource_vcd_nsxv_firewall.go
```diff
@@ -0,0 +1,221 @@
...
+				Required:    true,
+				ForceNew:    true,
+				Description: "Edge gateway name in which the firewall rule is located",
+			},
+			"rule_id": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Firewall rule ID for lookup",
+			},
+			"name": &schema.Schema{
```
vbauzysvmware: We can't use name, cause it isn't unique?

vcd/resource_vcd_edgegateway.go
```diff
@@ -113,6 +114,25 @@ func resourceVcdEdgeGateway() *schema.Resource {
 				Description: "Log level. One of 'emergency', 'alert', 'critical', 'error', " +
 					"'warning', 'notice', 'info', 'debug'. ('info' by default)",
 			},
+			"fw_enabled": &schema.Schema{
```
vbauzysvmware: Did we have vote between this variation ?:)
```
firewall{
 enabled = true
 default_logging_enabled=true
 default_rule_action = accept
}
```
Also accept and deny could be boolean with meaningful property name.

vcd/resource_vcd_nsxv_firewall_rule.go
```diff
@@ -0,0 +1,850 @@
...
+	"text/tabwriter"
+
+	"github.com/hashicorp/terraform/helper/hashcode"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/vmware/go-vcloud-director/v2/govcd"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+)
+
+func resourceVcdNsxvFirewall() *schema.Resource {
```
vbauzysvmware: resourceVcdNsxvFirewall -> resourceVcdNsxvFirewallRule ?

vcd/resource_vcd_nsxv_firewall_rule.go
```diff
@@ -0,0 +1,850 @@
...
+	return &schema.Resource{
+		Create: resourceVcdNsxvFirewallCreate,
```
vbauzysvmware: Same remark regarding `FirewallRule`

vcd/resource_vcd_nsxv_firewall_rule.go
```diff
@@ -0,0 +1,850 @@
...
+			"above_rule_id": &schema.Schema{
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Optional:    true,
+				Description: "This firewall rule will be inserted above the referred one",
+			},
+			"rule_type": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    false,
```
vbauzysvmware: `ForceNew:    false,` isn't needed

vcd/resource_vcd_nsxv_firewall_rule.go
```diff
@@ -0,0 +1,850 @@
...
+				Computed:    true,
+				Description: "Read only. Possible values 'user', 'internal_high'",
+			},
+			"rule_tag": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
```
vbauzysvmware: `ForceNew:    true,` ?

vcd/resource_vcd_nsxv_firewall_rule.go
```diff
@@ -0,0 +1,850 @@
...
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "accept",
+				Description:  "'accept' or 'deny'. Default 'accept'",
+				ValidateFunc: validation.StringInSlice([]string{"accept", "deny"}, false),
+			},
+			"enabled": &schema.Schema{
+				Type:        schema.TypeBool,
+				Optional:    true,
+				ForceNew:    false,
```
vbauzysvmware: not needed

vcd/resource_vcd_nsxv_firewall_rule.go
```diff
@@ -0,0 +1,850 @@
...
+				Default:     true,
+				Description: "Whether the rule should be enabled. Default 'true'",
+			},
+			"logging_enabled": &schema.Schema{
+				Type:        schema.TypeBool,
+				Optional:    true,
+				ForceNew:    false,
```
vbauzysvmware: not needed

vcd/resource_vcd_nsxv_firewall_rule.go
```diff
@@ -0,0 +1,850 @@
...
+//
+// Example resource name (_resource_name_): vcd_lb_nsxv_firewall_rule.my-test-fw-rule
+// Example import path (_the_id_string_): org.vdc.edge-gw.existing-firewall-rule-id
+func resourceVcdNsxvFirewallImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	var commandOrgName, orgName, vdcName, edgeName, firewallRuleId string
+	var listRules, importRule bool
+
+	resourceURI := strings.Split(d.Id(), ".")
+	helpError := fmt.Errorf("resource name must be specified as " +
+		"'org.vdc.edge-gw.firewall-rule-id' or 'list@org.vdc.edge-gw' to get a list of rules")
```
vbauzysvmware:
```suggestion
		"'org.vdc.edge-gw.firewall-rule-id' or 'list@org.vdc.edge-gw-name' to get a list of rules")
```

website/docs/d/nsxv_firewall.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+page_title: "vCloudDirector: vcd_nsxv_firewall"
+sidebar_current: "docs-vcd-data-source-nsxv-firewall"
+description: |-
+  Provides a vCloud Director firewall data sourcefor advanced edge gateways (NSX-V). This can be
+  used to read existing rule by ID and use its attributes in other resources.
+---
+
+# vcd\_nsxv\_firewall
+
+Provides a vCloud Director firewall data sourcefor advanced edge gateways (NSX-V). This can be
```
vbauzysvmware:
```suggestion
Provides a vCloud Director firewall data source for advanced edge gateways (NSX-V). This can be
```

website/docs/r/nsxv_firewall.markdown
```diff
@@ -0,0 +1,255 @@
+---
+layout: "vcd"
+page_title: "vCloudDirector: vcd_nsxv_firewall"
+sidebar_current: "docs-vcd-resource-nsxv-firewall"
+description: |-
+  Provides a vCloud Director firewall resource for advanced edge gateways (NSX-V). This can be used
+  to create, modify, and delete firewall rules.
+---
+
+# vcd\_nsxv\_firewall
```
vbauzysvmware: I would recommend in documentation refer as `Advanced edge gateway (NSX-V)`

website/docs/r/nsxv_firewall.markdown
```diff
@@ -0,0 +1,255 @@
...
+* `edge_gateway` - (Required) The name of the edge gateway on which to apply the firewall rule.
+* `action` - (Optional) Defines if the rule is set to `accept` or `deny` traffic. Default `accept`
+* `enabled` - (Optional) Defines if the rule is enabaled. Default `true`.
+* `logging_enabled` - (Optional) Defines if the logging for this rule is enabaled. Default `false`.
+* `name` - (Optional) Free text name. Can be duplicate.
+* `rule_tag` - (Optional) This can be used to specify user-controlled rule tag. If not specified,
+it will report rule ID after creation. Must be between 65537-131072.
+* `above_rule_id` - (Optional) This can be used to alter default rule placement order. By default
+every rule is appended to the end of firewall rule list. When a value of another rule is set - this
+rule will be placed above the specified rule.
```
lvirbalas: You have good examples there, but we also need one with **more than one** rule where `above_rule_id` is used.

website/docs/r/nsxv_firewall.markdown
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  service {
+    protocol = "tcp"
+    port     = "443"
+  }
+}
+```
+
+## Example Usage 3 (Use exclusion in source)
```
lvirbalas: Given that exclusion is described in separate section and that I just requested one more example, I suggest moving exclusions examples to the mentioned exclusion section below.

website/docs/r/nsxv_firewall.markdown
```diff
@@ -0,0 +1,255 @@
...
+
+```
+terraform import vcd_nsxv_firewall.imported my-org.my-org-vdc.my-edge-gw.my-firewall-rule-id
+```
+
+The above would import the application rule named `my-firewall-rule-id` that is defined on edge
+gateway `my-edge-gw` which is configured in organization named `my-org` and vDC named `my-org-vdc`.
+
+
+!> **Warning:** The UI does not show real firewall rule IDs, but their order from 1 to n.
```
lvirbalas: I'm thinking that we need a second import variant, which would take UI ID **directly** as well. This would reduce the amount of custom work the user has to go through.

The list command below may be useful regardless, so we could leave it (also it's a good precedent for other ID-listing commands).

@vbauzysvmware @dataclouder your thoughts?

vcd/provider.go
```diff
@@ -111,6 +111,7 @@ func Provider() terraform.ResourceProvider {
 			"vcd_lb_virtual_server":  resourceVcdLBVirtualServer(),  // 2.4
 			"vcd_nsxv_dnat":          resourceVcdNsxvDnat(),         // 2.5
 			"vcd_nsxv_snat":          resourceVcdNsxvSnat(),         // 2.5
+			"vcd_nsxv_firewall":      resourceVcdNsxvFirewall(),     // 2.5
```
lvirbalas: Important. I strongly suggest renaming this resource and data source from `vcd_nsxv_firewall` -> `vcd_nsxv_firewall_rule`. Reasons:
- In order to create more than one rule, you need to create more than one resources.
- When you create more than one, current `vcd_nsxv_firewall` name loses rational, because it looks as if you're defining multiple firewalls (as in multiple gateways).
- It's just more intuitive to see "rule" in the resource name.

CC: @vbauzysvmware @dataclouder

vbauzysvmware: I agree, I mentioned similar thing in my review.

website/docs/r/nsxv_firewall.markdown
vbauzysvmware: I agree - but not sure if it is possible.

vcd/datasource_vcd_nsxv_firewall.go
dataclouder: `name` is computed

vcd/resource_vcd_nsxv_firewall_rule.go
dataclouder: @vbauzysvmware  either we use `-name` for each entity, or we leave it everywhere. Adding only one in the list is inconsistent.

vcd/resource_vcd_nsxv_firewall_rule.go
```diff
@@ -0,0 +1,850 @@
...
+	d.SetId("")
+	return nil
+}
+
+// resourceVcdNsxvFirewallImport  is responsible for importing the resource.
+// The following steps happen as part of import
+// 1. The user supplies `terraform import _resource_name_ _the_id_string_` command
+// 2a. If the `_the_id_string_` contains a dot formatted path to resource as in the example below
+// it will try to import it. If it is found - the ID is set
+// 2b. If the `_the_id_string_` starts with `list@` and contains path to edge gateway similar to
```
dataclouder: Very neat trick this `list@`! 
Is  this practice documented somewhere?

vcd/resource_vcd_nsxv_firewall_rule.go
Didainius: Yes, because the tag is so odd that it should alter the ID.

website/docs/d/nsxv_firewall.html.markdown
```diff
@@ -0,0 +1,43 @@
+---
+layout: "vcd"
+page_title: "vCloudDirector: vcd_nsxv_firewall"
+sidebar_current: "docs-vcd-data-source-nsxv-firewall"
+description: |-
+  Provides a vCloud Director firewall data sourcefor advanced edge gateways (NSX-V). This can be
```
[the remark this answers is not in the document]
Didainius: Thanks, fixed.

website/docs/r/nsxv_firewall.markdown
Didainius: The example 4 actually uses the `above_rule_id`. I updated its title to explicitly state that.

vcd/resource_vcd_edgegateway.go
vbauzysvmware: @lvirbalas did we already have guideline for such cases ?:)

vcd/resource_vcd_nsxv_firewall_rule.go
vbauzysvmware: Could you add some comment, as who haven't developed this resource wouldn't know why

vcd/resource_vcd_nsxv_firewall_rule.go
vbauzysvmware: We agreed everthere add `-name` to be explicit

website/docs/r/nsxv_firewall.markdown
vbauzysvmware: It's just my humble view point view:) @lvirbalas what is your view?

vcd/resource_vcd_nsxv_firewall_rule.go
```diff
@@ -0,0 +1,884 @@
...
+// resourceVcdNsxvFirewallRuleImport  is responsible for importing the resource.
+// The following steps happen as part of import
+// 1. The user supplies `terraform import _resource_name_ _the_id_string_` command
+// 2a. If the `_the_id_string_` contains a dot formatted path to resource as in the example below
+// it will try to import it. If it is found - the ID is set
+// 2b. If the `_the_id_string_` starts with `list@` and contains path to edge gateway similar to
+// `list@org.vdc.edge-gw` then the function lists all firewall rules and their IDs in that edge
+// gateway.
+// 2c. If the `_the_id_string_` contains a dot formatted path with the 4th element starting with
+// substring 'ui-id:3' and number after color - the import function will try to lookup real ID and
```
lvirbalas: Suggestion:
`// substring 'ui-id:3' and number after color - the import function will try to lookup real ID and`
->
`// substring 'ui-id:' and number after colon - the import function will try to lookup a real ID by the given UI ID and`

vcd/resource_vcd_nsxv_firewall_rule.go
```diff
@@ -0,0 +1,884 @@
...
+// import the rule
+// 2d. If the `_the_id_string_` does not match format described neither in '2a', '2b', '2c' a
+// usage error message is printed
+//
+// Example resource name (_resource_name_): vcd_nsxv_firewall_rule.my-test-fw-rule
+// Example import path (_the_id_string_): org.vdc.edge-gw.existing-firewall-rule-id
```
lvirbalas: As this is an example, it would be more valuable to see some real-life representative number instead of `.existing-firewall-rule-id`

website/docs/r/nsxv_firewall_rule.markdown
```diff
@@ -0,0 +1,268 @@
...
+gateway `my-edge-gw` which is configured in organization named `my-org` and vDC named `my-org-vdc`.
+
+
+### Import by firewall rule number as show in the UI (No. field)
+
+```
+terraform import vcd_nsxv_firewall_rule.imported my-org.my-org-vdc.my-edge-gw.ui-no:3
+```
+
+**Pay attention** to the specific format of firewall rule number `ui-no:3`. The `ui-no` flags that
```
lvirbalas: Nit:
```
The `ui-no`
```
->
```
The `ui-no:`
```

website/docs/r/nsxv_firewall_rule.markdown
```diff
@@ -0,0 +1,268 @@
...
+
+```
+terraform import vcd_nsxv_firewall_rule.imported my-org.my-org-vdc.my-edge-gw.my-firewall-rule-id
+```
+
+The above would import the application rule named `my-firewall-rule-id` that is defined on edge
+gateway `my-edge-gw` which is configured in organization named `my-org` and vDC named `my-org-vdc`.
+
+
+### Import by firewall rule number as show in the UI (No. field)
```
lvirbalas: Have you meant with quotes?
`("No." field)`

vcd/resource_vcd_nsxv_firewall_rule.go
Didainius: Changed.

vcd/resource_vcd_nsxv_firewall_rule.go
Didainius: Changed

website/docs/r/nsxv_firewall_rule.markdown
Didainius: Changed

website/docs/r/nsxv_firewall_rule.markdown
Didainius: Makes sense, changed

vcd/resource_vcd_edgegateway.go
lvirbalas: There was no guideline, but after we learned all the nasty things about `Set` you have hit with the VDC data source, I think we became "fine" with flattened parameter names like here.

vcd/resource_vcd_nsxv_firewall_rule.go
lvirbalas: +1 much more clearer for the user when scan reading and we have `-name` or `-id` explicitly.

Didainius: So do we make it `org-name.vdc-name.edge-gw-name.firewall-rule-id`?

website/docs/r/nsxv_firewall.markdown
lvirbalas: Two things here:
1. Either GitHub misbehaves, but we have agreed that it should "vcd_nsxv_firewall_**rule**" and I even saw commits changing to that.
2. I agree that we should note that this is for advanced EG, but it is already here a line below, no?

vcd/resource_vcd_nsxv_firewall_rule.go
Didainius: Agreed. Done.

vcd/resource_vcd_nsxv_firewall_rule.go
Didainius: Removed.

website/docs/r/nsxv_firewall.markdown
Didainius: 1. I think it misbehaves. If I open the PR - it is changed.
2. @vbauzysvmware if you can point it more explicitly. I think it is noted in https://github.com/terraform-providers/terraform-provider-vcd/pull/341/files#diff-215be64c11703733b065d8ed7b543c5fR12

website/docs/r/nsxv_firewall.markdown
Didainius: I remember I had a comment but this got lost. Given that the `exclude` is a single attribute in the `source`/`destination`blocks I think it would be too much to insert an example near it.  I have added a link in to this example in the `exclude`attributes note. Is it any better?

lvirbalas: Yes, neat!

Didainius:
Test suite passes on 9.5 and 10.

vcd/resource_vcd_edgegateway.go
```diff
@@ -198,8 +218,20 @@ func resourceVcdEdgeGatewayCreate(d *schema.ResourceData, meta interface{}) erro
...
+		err = updateFirewall(d, edge)
+		if err != nil {
+			return fmt.Errorf("unable to update firewall settings: %s", err)
+		}
+
+		log.Printf("[TRACE] edge gateway firewall configured")
+
 	}
 
+	// TODO double validate if we need to use partial state here
```
vbauzysvmware: Do we leave this?

vbauzysvmware: Do we leave this?

Didainius:
Additionally this PR closes #348 as this is a very much related issue.

vcd/resource_vcd_nsxv_firewall_rule_test.go
```diff
@@ -0,0 +1,879 @@
...
+	}
+  
+	destination {
+		ip_addresses = ["5.5.5.5"]
+	}
+  
+	service {
+		protocol = "any"
+	}
+	depends_on = ["vcd_nsxv_firewall_rule.rule0"]
```
vbauzysvmware: Why we need this here?

vcd/resource_vcd_nsxv_firewall_rule_test.go
```diff
@@ -0,0 +1,879 @@
...
+  
+	destination {
+		ip_addresses = ["40.10.10.0/24", "41.10.10.0/24"]
+	}
+
+	service {
+		protocol = "any"
+	}
+
+	depends_on = ["vcd_nsxv_firewall_rule.rule6"]
```
vbauzysvmware: Do we need depends_on if have `above_rule_id = "${vcd_nsxv_firewall_rule.rule6.id}"` ?

website/docs/d/nsxv_firewall_rule.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+
+## Example Usage
+
+```hcl
+data "vcd_nsxv_firewall_rule" "my-rule" {
+  org                 = "my-org"
+  vdc                 = "my-org-vdc"
+  edge_gateway        = "my-edge-gw"
+
+  rule_id = "133048"
```
vbauzysvmware: This is UI rule or real Id ?:)

website/docs/r/nsxv_firewall_rule.markdown
```diff
@@ -0,0 +1,271 @@
...
+* `rule_type` - Possible values - `user`, `internal_high`.
+
+## Importing
+
+~> **Note:** The current implementation of Terraform import can only import resources into the state.
+It does not generate configuration. [More information.](https://www.terraform.io/docs/import/)
+
+
+
+An existing firewall rule can be [imported][docs-import] into this resource
```
vbauzysvmware: Maybe we need to update documentation regardin `.` dot

vcd/resource_vcd_edgegateway.go
Didainius: Yes. I intend to leave it for next iteration we touch it, because this is not a potential issue introduced by this PR and may be not an issue at all - this is more for consideration and ensuring that we either need it, either not.

vcd/resource_vcd_nsxv_firewall_rule_test.go
Didainius: This is one way of ordering firewall rules (besides using the `above_rule_id`). Line 141 has a check that `rule0` is created first in the order before `rule0-2`

`checkfirewallRuleOrder("vcd_nsxv_firewall_rule.rule0", "vcd_nsxv_firewall_rule.rule0-2"),`

website/docs/d/nsxv_firewall_rule.html.markdown
Didainius: This is rule id

website/docs/r/nsxv_firewall_rule.markdown
vbauzysvmware: ty

vcd/resource_vcd_nsxv_firewall_rule_test.go
vbauzysvmware: ok.

website/docs/d/nsxv_firewall_rule.html.markdown
Didainius: So this is the "real" ID. I actually tried to make distinction that ID is the real ID, while UI number means that number. There is also reference about it.

vbauzysvmware: so maybe this would help:
```suggestion
 # real rule Id, not UI Id
  rule_id = "133048"
```

vcd/resource_vcd_nsxv_firewall_rule_test.go
vbauzysvmware:
```suggestion
        # this another way to control rules order if `above_id` isn't used
	depends_on = ["vcd_nsxv_firewall_rule.rule6"]
```

website/docs/d/nsxv_firewall_rule.html.markdown
Didainius: Added

vcd/resource_vcd_nsxv_firewall_rule_test.go
Didainius: Added.

Issue #311, opened 2019-08-14
Add support for VM disk management in vcd_vapp_vm resource


Didainius:
At the moment resource `vcd_vapp_vm`supports only adding independent disks. There are no options to handle VM disks (add them, resize them), but there may be such a need for altering VM disks and their sizes.

buec011:
I know this issue 311 is for the implementation in `vcd_vapp_vm`. But if a developer implements the it, it may be good to have bug 257 in mind to define functions and code structure:
We would like to see this in the 2 resource approach suggested here:
https://github.com/terraform-providers/terraform-provider-vcd/issues/257

buec011:
Our business requires vm snapshots now. Since this is not possible with independent/namend disks at the moment, we got a demand for this feature. Unlike named disks, an extra attachment resource does not make sence for vm disks, since you cannot create a vm disk without a vm.

But maybe it make sence to have the disk in its own resource anyway (maybe it is possible to move a vm disk to another vm in future and we want to avoid https://github.com/terraform-providers/terraform-provider-vcd/issues/206 for vm disks)

May I suggest the following example code?:

```
resource vcd_vm_disk {
   attachment = vcd_vapp_vm.web1.name
   size = 1024
   [...]
}
```
If VCD support a move of vm disks in future, we may change the attachment value and it will be translated in the "move vm disk" api call.

lvirbalas:
@buec011 please explain more about your use case. Esp., what do you want to do with snapshots and Terraform VM disks.

CC: @Didainius

buec011:
@lvirbalas 

With independent disks / named disks (we use VCD 10), it is not possible to create a snapshot

(in GUI: VM -> Actions -> Create Snapshot) 

![image](https://user-images.githubusercontent.com/38046041/68662047-53c7c880-053c-11ea-8f72-3a086100f301.png)

We use snapshots before we or a third party deploys a new application in the VM. I think this is a usual use case why you want to use this vSphere / VCD feature.

If you use VM disks, you can take your snapshot as usual.

We do not need create and delete snapshots via terraform yet. We use the GUI for this at the moment, because they will resist temporary only.

lvirbalas:
@buec011 I see. Are you still using the independent disks and do you still need the attachment resource for them #257 ?

buec011:
> @buec011 I see. Are you still using the independent disks and do you still need the attachment resource for them #257 ?

[...]

PR #345, opened 2019-10-03
Add support for changing root disk size

cheald:
This depends on https://github.com/vmware/go-vcloud-director/pull/251

This simply adds the ability to specify a `root_disk` parameter when defining a VM, which indicates the disk's size in megabytes. For example:

```javascript
// vm.tf.json
{
  "resource": {
    "vcd_vapp_vm": {
      "az-app1": {
        "name": "az-app1",
        "root_disk": 256000,
        // ...
      }
    }
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ FEATURES:
 * **New Data Source:** `vcd_network_routed` Routed Network - ([#218])
 * **New Data Source:** `vcd_network_isolated` Isolated Network - ([#218])
 * **New Data Source:** `vcd_network_direct` Direct Network - ([#218])
+* **New Data Source:** `vcd_vapp` vApp - ([#218])
 * **New Data Source:** `vcd_nsxv_dnat` DNAT for advanced edge gateways using proxied NSX-V API - [#328]
 * **New Data Source:** `vcd_nsxv_snat` SNAT for advanced edge gateways using proxied NSX-V API - [#328]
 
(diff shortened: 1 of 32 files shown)


cheald:
I've added the replace clause in go.mod, and it's working as expected locally, but I don't see Travis performing a `go get` or similar; do I need to `go mod vendor` and commit the vendored changes as well? I don't want to clutter up the diff with dependency changes unless that's desired.

dataclouder:
> I've added the replace clause in go.mod, and it's working as expected locally, but I don't see Travis performing a `go get` or similar; do I need to `go mod vendor` and commit the vendored changes as well? I don't want to clutter up the diff with dependency changes unless that's desired.

You need to run `go mod tidy` and `go mod vendor`. Once your change in go-vcloud-director is merged, you will remove the replace, run the above commands again, and all should be well.

cheald:
Thanks! And thanks for your patience - I haven't worked with Go modules before! Still learning the ropes.

lvirbalas:
Thank you @cheald for contributing! I have some usage-based questions:
- How does operating system of the OVA react when you change the size? Have you tried with different OVAs of both Linux and Windows variants? I'd think that there should be cases where only partition gets resized, but the extra space is not instantly accessible from the OS. Would be great if you could elaborate on this aspect with different OVA examples.
- What happens if you reduce the size as opposed to making it larger?

vbauzysvmware:
From perspective point of view. If we can change any disk size in OVA, maybe we should let it in terraform too, not only `root`/first one ?

cheald:
> How does operating system of the OVA react when you change the size?

Under Linux, the larger disk size is immediately reflected via `lslbk`. Filesystems have to be resized to larger extents via `resize2fs` or similar. This is functionally identical to EBS resizing at AWS. I haven't tested it under Windows.

> What happens if you reduce the size as opposed to making it larger?

vCD will not permit this; attempting to shrink a disk results in an API error.

> From perspective point of view. If we can change any disk size in OVA, maybe we should let it in terraform too, not only root/first one ?

That's certainly possible. This work could be expanded into full management of internal disks, to complement the independent disk attachment. The govcd APIs take bus/unit IDs and aren't hardcoded to the root disk, so the Terraform provider would just need additional work to consume it.

(As an aside: how do I kick off a Travis build for my latest push? It didn't seem to happen automatically.)

Didainius:
@cheald,
Sorry for this being silent for so long. We have been busy on rolling out 2.5.0 and got buried under other tasks.

Now we have yet another task #311 which we want to bring in. I am looking at them and thinking - as the root disk is same as any other disk (just mostly often having address `0:0`, but I believe not necessarily for some templates) we could design a structure to accomplish both goals.

What are your thoughts on such proposal? As this may become more time demanding should someone else pick #311 and cover the root disk size functionality as well in the solution of that issue or would you like to bring #311 forward?

cheald:
I'm happy to let you all take it if you'd like. My disk management needs
are pretty simple, and I'm still quite new to vCloud, so someone with a
broader view of the tool may be more suited to take that one over the
finish line. Hopefully my work provides a jump start!

vbauzysvmware:
Hi @cheald. I started work on different approach due that we need support various features, not only `root` internal disk. You can check and give a spin in https://github.com/terraform-providers/terraform-provider-vcd/pull/412 it is still draft.

cheald:
Looking good! I'll close this PR, since it looks like you're headed in a more robust direction. Thank you!

PR #354, opened 2019-10-23
Fix docs and rename import separator variable

dataclouder:
*    Update docs with vCD 10.0 support
*    Rename ImportSeparationToken to ImportSeparator

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ FEATURES:
 
 IMPROVEMENTS:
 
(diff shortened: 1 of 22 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ FEATURES:
 
 IMPROVEMENTS:
 
+* Add support for vCD 10.0 (API 33)
```
Didainius: I think I wouldn't mention the API 33 as it may lead to false conclusion that we use API v33.

PR #356, opened 2019-10-24
Fix filtering

vbauzysvmware:
Fix for terraform-providers/terraform-provider-vcd#287

* Parentheses isn't needed - everything works the same and we fix issue for some customers.


vbauzysvmware:
> General query - I see some assumptions are made based on `vcdClient.APIVCDMaxVersionIs(">= 33.0")` but the comments state "in vCD version 10 with NSXT". Do we know if this is related to NSX-T only, or to version 10 in general?

vbauzysvmware:
Issue in vCD NSX-T

vbauzysvmware:
> General query - I see some assumptions are made based on `vcdClient.APIVCDMaxVersionIs(">= 33.0")` but the comments state "in vCD version 10 with NSXT". Do we know if this is related to NSX-T only, or to version 10 in general?

Issue in vCD 10 - NSX-T only

vcd/resource_vcd_org_vdc.go
```diff
@@ -331,13 +332,16 @@ func setOrgVdcData(d *schema.ResourceData, vcdClient *VCDClient, adminOrg *govcd
 	}
 	_ = d.Set("name", adminVdc.AdminVdc.Name)
 
+	// in vCD version 10 with NXT - network pool reference isn't returned with AdminVdc
```
Didainius:
```suggestion
	// in vCD version 10 with NSX-T - network pool reference isn't returned with AdminVdc
```

vcd/resource_vcd_org_vdc_test.go
```diff
@@ -122,13 +122,30 @@ func runOrgVdcTest(t *testing.T, params StringMap, allocationModel string) {
 	}
 	debugPrintf("#[DEBUG] CONFIGURATION: %s", configText)
 
+	usedNetworkPool := testConfig.VCD.ProviderVdc.NetworkPool
+	expectNonEmptyPlan := false
+	// in vCD version 10 with NXT - network pool reference isn't returned with AdminVdc
```
Didainius:
```suggestion
	// in vCD version 10 with NSX-T - network pool reference isn't returned with AdminVdc
```

vcd/resource_vcd_org_vdc_test.go
```diff
@@ -122,13 +122,30 @@ func runOrgVdcTest(t *testing.T, params StringMap, allocationModel string) {
...
+
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
 		Providers:    testAccProviders,
 		CheckDestroy: testAccCheckVdcDestroy,
 		Steps: []resource.TestStep{
 			resource.TestStep{
-				Config: configText,
+				Config:             configText,
+				ExpectNonEmptyPlan: expectNonEmptyPlan,
```
Didainius: If you have a suppress in code - why is this needed? The non empty plan signals that it will cause an "update" on every `apply`, isn't it so?

vcd/resource_vcd_org_vdc.go
```diff
@@ -173,9 +173,10 @@ func resourceVcdOrgVdc() *schema.Resource {
 				Description: "Boolean to request thin provisioning. Request will be honored only if the underlying datastore supports it. Thin provisioning saves storage space by committing it on demand. This allows over-allocation of storage.",
 			},
 			"network_pool_name": &schema.Schema{
-				Type:        schema.TypeString,
-				Optional:    true,
-				Description: "The name of a network pool in the Provider VDC. Required if this VDC will contain routed or isolated networks.",
+				Type:             schema.TypeString,
+				Optional:         true,
+				DiffSuppressFunc: suppressEmptyString(),
```
Didainius: When does this issue occur? I would assume that if someone is using NSX-T based vDC - he would always get empty value, while if he is not - he would always get a real value. Isn't it so?

vcd/resource_vcd_org_vdc_test.go
vbauzysvmware: this testing framework issue, I tested in real life and supressFunc works, but in testing case it sees that network pool is empty and provided one isn't.

vcd/resource_vcd_org_vdc.go
vbauzysvmware: yes

vcd/resource_vcd_org_vdc.go
vbauzysvmware: updated

vcd/resource_vcd_org_vdc_test.go
vbauzysvmware: updated

vcd/resource_vcd_org_vdc.go
Didainius: Wouldn't it be safer to `d.Set` the field in `Create` function as it is. That way there is no need for such a "wide capturing" suppress function.

vbauzysvmware: On create we set the value, issue is when we re read it/refresh

PR #357, opened 2019-10-24
Comment out leftovers

vbauzysvmware:
Binary test complains regarding left size_in_bytes, so have to be commented out.


vcd/resource_vcd_independent_disk_test.go
```diff
@@ -34,14 +34,14 @@ func TestAccVcdIndependentDiskBasic(t *testing.T) {
 
 	params["FuncName"] = t.Name() + "-Compatibility"
 	configTextForCompatibility := templateFill(testAccCheckVcdIndependentDiskForCompatibility, params)
-	params["FuncName"] = t.Name()
-	configText := templateFill(testAccCheckVcdIndependentDiskBasic, params)
-	if vcdShortTest {
+	/*	params["FuncName"] = t.Name()
+		configText := templateFill(testAccCheckVcdIndependentDiskBasic, params)
+	*/if vcdShortTest {
```
lvirbalas:
```suggestion
	*/
	if vcdShortTest {
```

vcd/resource_vcd_independent_disk_test.go
```diff
@@ -175,3 +176,4 @@ resource "vcd_independent_disk" "{{.secondResourceName}}" {
   storage_profile = "{{.storageProfileName}}"
 }
 `
+*/
```
lvirbalas: I don't understand, why not leave this test, but only comment out the size_in_bytes field?

vbauzysvmware: cause it's duplicate

dataclouder: It was skipped in the acceptance test but was still exported to the binary tests

PR #358, opened 2019-10-24
Address outstanding issues before the release

Didainius:
This PR solves the following:
* Adds correct CHANGELOG github issue format as per https://www.terraform.io/docs/extend/best-practices/versioning.html#example-changelog
* Splits load balancer and firewall statefile update from setEdgeGatewayValues to separate function setEdgeGatewayComponentValues and runs it after firewall and load balancer is configured during creation. There is also one more test step to cover this problem. Previously the problem was that it is called before firewall and load balancer general config is applied. Later on, when then config is being applied the d.Get() already read values from what was set using setEdgeGatewayValues
* Adds note to avoid mixing old and new dnat/snat resources in documentation website.
* Adds note about migrating to new DNAT/SNAT and firewall resources into `TROUBLESHOOTING.md`

### DIFF
--- a/website/docs/r/nsxv_dnat.html.markdown
+++ b/website/docs/r/nsxv_dnat.html.markdown
@@ -16,6 +16,9 @@ modify, and delete destination NATs to map an external IP/port to an internal IP
 ~> **Note:** This resource requires advanced edge gateway. For non-advanced edge gateways please
 use the [`vcd_dnat`](/docs/providers/vcd/r/dnat.html) resource.
 
+!> **Warning:** Do not use older [`vcd_dnat`](/docs/providers/vcd/r/dnat.html) resource with this one
+because it will change IDs and this resource will not be able to lookup rules.
+
 ## Example Usage 1 (Minimal input)
 
 ```hcl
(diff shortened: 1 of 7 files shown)


Didainius:
Cannot update the description, but here it goes:

During create phase there is a `setEdgeGatewayValues` function executed to save fields for edge gateway. The problem is that it is called before firewall and load balancer general config is applied. Later on, when then config is being applied the d.Get() already read values from what was set using  `setEdgeGatewayValues`

This PR splits load balancer and firewall statefile update from `setEdgeGatewayValues` to separate function `setEdgeGatewayComponentValues` and runs it after firewall and load balancer is configured. There is also one more test step to cover this problem.

website/docs/r/nsxv_dnat.html.markdown
```diff
@@ -16,6 +16,9 @@ modify, and delete destination NATs to map an external IP/port to an internal IP
 ~> **Note:** This resource requires advanced edge gateway. For non-advanced edge gateways please
 use the [`vcd_dnat`](/docs/providers/vcd/r/dnat.html) resource.
 
+!> **Warning:** Do not use older [`vcd_dnat`](/docs/providers/vcd/r/dnat.html) resource with this one
```
dataclouder: Should we add such information to TROUBLESHOOTING.md ?

Didainius: Indeed - added a note

PR #361, opened 2019-10-25
Add extra step in binary tests to ensure plan is stable, fix `shared_secret` in vcd_edgegateway_vpn

### LINKED ISSUE #362: BUG: resource `vcd_edgegateway_vpn` proposes replacement of `shared_secret` on every apply
At the moment every time `apply`/`plan` is executed the `shared_secret` field hints for update and because the field is not updatable - it requests to recreate the the resource.

```hcl
resource "vcd_edgegateway_vpn" "vpn" {
  org                 = "my-org"
  vdc                 = "my-org-vdc"
  edge_gateway        = "my-edge-gw"
  name                = "asd"
  description         = "Description"
  encryption_protocol = "AES256"
  mtu                 = 1400
  peer_id             = "192.168.5.2"
  peer_ip_address     = "192.168.5.2"
  local_id            = "192.168.1.110"
  local_ip_address    = "192.168.1.110"
  shared_secret       = "my-secret-string"

  peer_subnets {
    peer_subnet_name    = "DMZ_WEST"
    peer_subnet_gateway = "192.168.5.1"
    peer_subnet_mask    = "255.255.255.0"
  }


  local_subnets {
    local_subnet_name    = "DMZ_EAST"
    local_subnet_gateway = "192.168.1.1"
    local_subnet_mask    = "255.255.255.0"
  }

}
```

image

Didainius:
This PR adds an extra step for our `make test-binary` infrastructure. The step is `plancheck` and is called after `apply`.  It runs `terraform plan -detailed-exitcode`. The flag `-detailed-exitcode` forces `plan` command to emit exit code 2 (instead of 0 without it) when the plan is not empty.
This allows us to validate any field issues just after the resource is created.

In addition to this it fixes outstanding issues so that `make test-binary` works fine with this new step:
* `resource/vcd_edgegateway_vpn` - force replacement due to blank value set for `shared_secret` (because it is not returned from API). Closes #362
* remove `ip` setting because `POOL` is used in `cust.misc2.tf` sample config. It used to cause IP overrides because POOL assignes IP from the pool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,7 +65,8 @@ BUG FIXES:
 * Wait for task completion on creation and update, where tasks were not handled at all.
 * `resource/vcd_firewall_rules` force recreation of the resource when attributes of the sub-element `rule` are changed (fixes a situation when it tried to update a rule).
 * `resource/vcd_network_isolated` Fix definition of DHCP, which was created automatically with leftovers from static IP pool even when not requested.
-* `resource/vcd_network_routed` Fix retrieval with early vCD versions [GH-344]
+* `resource/vcd_network_routed` Fix retrieval with early vCD versions. [GH-344]
+* `resource/vcd_edgegateway_vpn` Required replacement every time for `shared_secret` field.  [GH-361]
 
 DEPRECATIONS
 * The ability of deploying a VM implicitly within a vApp is deprecated. Users are encouraged to set an empty vApp and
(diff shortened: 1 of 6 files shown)


lvirbalas:
This is for testing the "apply-apply" combo, which

Didainius:
> If I understand right, with this addition our test suite will do the 2nd apply now by default? If yes, that’s really needed!

Yes (`plan` instead of `apply`) which is the same and throws error, unless the tests add directive `# skip-plan-check` to definition.

PR #363, opened 2019-10-28
Fix Org resource and datasource

dataclouder:
* Remove "ForceNew" from data source
* Change "name" from Optional to Required
* Set "delay_after_power_on_seconds" for both resource and data source
* Update Org test


dataclouder:
> One more changelog note? :)

If this change goes as part of 2.5.0. it should not be needed.

PR #380, opened 2019-10-30
Vdc datasource crash fix

vbauzysvmware:
Crash happen when user is org admin and try to use data source.
vCD API returns different set of data according user role and as result some of values are empty and that wasn't expected in code.
Fix checks various values before using them.
* changelog update
* In docs added NOTE

Crash reproducing HCL:
```hcl
provider "vcd" {
  user                 = "datacloud-admin"
  password             = "CHANGE-ME"
  url                  = "https://vcd-server.org/api"
  org                  = "datacloud"
  vdc                  = "vdc-datacloud"
  allow_unverified_ssl = "true"
  max_retry_timeout    = 600
  version              = "~> 2.5"
  logging              = true
  logging_file         = "go-vcloud-director-10.150.170.30.log"
}

data "vcd_org_vdc" "myvdc" {
    name = "vdc-datacloud"
}
```


dataclouder:
Given that reproducing the bug only needs a data source on the default VDC, we should add a test for exactly this scenario.
In a separate PR we are introducing the ability of testing easily with Org User. Until then, having the test ready will help check that the fix works.

vbauzysvmware:
datasource_vcd_org_vdc_test.go already have such test, data source test. And with test infrastructure upgrade you prepared it will covered by running with Org admin.

dataclouder:
> datasource_vcd_org_vdc_test.go already have such test, data source test. And with test infrastructure upgrade you prepared it will covered by running with Org admin.

That test is skipped when using Org User, because it creates a VDC.
We need a test that only reads a VDC, so it should not be skipped with an Org User.

vbauzysvmware:
Again it will be difficult to document, as we will have to handle changes through versions. So don't recommend. At the moment major missing parts using org admin are which were nil's. Other data mostly available. For e.g. logical that we don't see port groups as org admin.

vbauzysvmware:
> > datasource_vcd_org_vdc_test.go already have such test, data source test. And with test infrastructure upgrade you prepared it will covered by running with Org admin.
> 
> That test is skipped when using Org User, because it creates a VDC.
> We need a test that only reads a VDC, so it should not be skipped with an Org User.

Added test only for org admin

Didainius:
> Again it will be difficult to document, as we will have to handle changes through vCD versions. So don't recommend. At the moment major missing parts using org admin are which were nil's. Other data mostly available. For e.g. logical that we don't see port groups as org admin.

Alright, my assumption was that everything under `adminVdc`is no returned, just that other variables were'nt pointers and defaulted to empty values.

vcd/datasource_vcd_org_vdc_test.go
```diff
@@ -9,25 +9,24 @@ import (
...
-
 	var params = StringMap{
 		"ExistingVdcName": testConfig.VCD.Vdc,
 		"VdcName":         vdcName,
 		"OrgName":         testConfig.VCD.Org,
-		"FuncName":        "TestAccVcdVdcDatasource",
+		"FuncName":        vdcName,
 	}
 
+	configText := templateFill(testAccCheckVcdVdcDatasource_basic, params)
```
dataclouder: Please change the test in such a way that only one `templateFill` is called. 
This instruction creates the test file in `test-artifacts`. If we call it twice, it will create two files, one of which will contain fields visible to system admin only, and the subsequent binary tests will fail.

vcd/datasource_vcd_org_vdc_test.go
```diff
@@ -36,56 +35,93 @@ func TestAccVcdVdcDatasource(t *testing.T) {
...
-						"data."+datasourceVdc, "compute_capacity.0.memory.0.allocated", "vcd_org_vdc."+vdcName, "compute_capacity.0.memory.0.allocated"),
-					resource.TestCheckResourceAttrPair(
-						"data."+datasourceVdc, "compute_capacity.0.memory.0.limit", "vcd_org_vdc."+vdcName, "compute_capacity.0.memory.0.limit"),
-					resource.TestMatchResourceAttr(
-						"data."+datasourceVdc, "compute_capacity.0.memory.0.overhead", regexp.MustCompile(`^\d+$`)),
-					resource.TestMatchResourceAttr(
-						"data."+datasourceVdc, "compute_capacity.0.memory.0.reserved", regexp.MustCompile(`^\d+$`)),
-					resource.TestMatchResourceAttr(
-						"data."+datasourceVdc, "compute_capacity.0.memory.0.used", regexp.MustCompile(`^\d+$`))),
+	if !usingSysAdmin() {
```
dataclouder: (Following up from the comment about `fillTemplate`)
Wouldn't be more readable to have two functions instead of a giant `if`? If we split it, we can just skip one of them when the user is not sys-admin, and keep the other clean.

website/docs/d/org_vdc.html.markdown
```diff
@@ -36,6 +36,8 @@ The following arguments are supported:
 
 ## Attribute Reference
 
+~> **Note:** Only part of fields are read if user is Organization administrator. With System Admin user all fields are populated.
```
dataclouder: Could you list which fields are skipped with Org admin?

Didainius:
Here is what are differences when run as org admin (1st value) vs sys admin (2nd value)
I wonder if we can do anything smarter about the bool fields which are defaulted to false. Any chance of setting them to null? Showing `false` while actually it is `true` may fool someone into wrong conclusions.

```shell
$ diff terraform.tfstate terraform.tfstate_admin
4c4
<   "serial": 2,
---
>   "serial": 1,
41c41
<             "cpu_guaranteed": null,
---
>             "cpu_guaranteed": 1,
44,45c44,45
<             "enable_fast_provisioning": false,
<             "enable_thin_provisioning": false,
---
>             "enable_fast_provisioning": true,
>             "enable_thin_provisioning": true,
49c49
<             "memory_guaranteed": null,
---
>             "memory_guaranteed": 1,
56c56
<             "network_pool_name": null,
---
>             "network_pool_name": "vc1-VXLAN-NP",
60c60
<             "provider_vdc_name": null,
---
>             "provider_vdc_name": "vc1",
66c66
<                 "name": ""
---
>                 "name": "Development"
```

lvirbalas:
- Really need `null` instead of `false` booleans, it's dangerous to leave it like this; or figure out an alternative way to avoid misinterpretation.

- @Didainius Which vCD version was this list generated with? @vbauzysvmware can you generate the same diff but with different vCD version, so to see if there is any version-depended behavior.

Thanks.

website/docs/d/org_vdc.html.markdown
```diff
@@ -36,6 +36,8 @@ The following arguments are supported:
...
+
 * `description` - VDC friendly description
 * `provider_vdc_name` -A name of the Provider VDC from which this organization VDC is provisioned.
```
lvirbalas:
```suggestion
* `provider_vdc_name` - name of the Provider VDC from which this organization VDC is provisioned.
```

website/docs/d/org_vdc.html.markdown
lvirbalas: I would like to ask the same, but to list it not here, but below. I propose using a flag in the parenthesis to reflect which fields are sys-admin-only. And use this flag across all the data sources where org admin has less visibility.

For example:
```
field_a - some common field ...
field_x - (System Admin) some field visible only to sys admins ...
field_z - some another common field ...
```

@dataclouder  @vbauzysvmware @Didainius can we agree on such notion?

Didainius:
> * @Didainius Which vCD version was this list generated with? @vbauzysvmware can you generate the same diff but with different vCD version, so to see if there is any version-depended behavior.

9.7

vbauzysvmware:
> > Again it will be difficult to document, as we will have to handle changes through vCD versions. So don't recommend. At the moment major missing parts using org admin are which were nil's. Other data mostly available. For e.g. logical that we don't see port groups as org admin.
> 
> Alright, my assumption was that everything under `adminVdc`is no returned, just that other variables were'nt pointers and defaulted to empty values.

`adminVdc` vs `vDc` that returned data set is different. Non system admin data isn't available.

vcd/datasource_vcd_org_vdc_test.go
vbauzysvmware: updated

vcd/datasource_vcd_org_vdc_test.go
vbauzysvmware: updated

website/docs/d/org_vdc.html.markdown
vbauzysvmware: Updated

vbauzysvmware:
>     * Really need `null` instead of `false` booleans, it's dangerous to leave it like this; or figure out an alternative way to avoid misinterpretation.
> 
>     * @Didainius Which vCD version was this list generated with? @vbauzysvmware can you generate the same diff but with different vCD version, so to see if there is any version-depended behavior.
> 
> 
> Thanks.

Checked against my env - add missing data handling

website/docs/d/org_vdc.html.markdown
lvirbalas: Sorry for confusion, please also capitalize the first word:
 "(Required, System Admin) **N**ame of the Provider VDC ..."

website/docs/d/org_vdc.html.markdown
lvirbalas: The result looks really easy to read, thanks! I'll leave it to @dataclouder to resolve this conversation as he started it.

lvirbalas:
@vbauzysvmware did you try to generate the same diff as in https://github.com/terraform-providers/terraform-provider-vcd/pull/380#issuecomment-548363904 but with different (not 9.7) vCD version, so to see if there is any version-depended behavior?

vbauzysvmware:
These values fields provider_vdc_name, storage_profile, memory_guaranteed, cpu_guaranteed, cpu_speed, enable_thin_provisioning, enable_fast_provisioning, network_pool_name won't provided by API for org admin.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ IMPROVEMENTS:
 * Switch to Terraform terraform-plugin-sdk v1.0.0 as per recent [HashiCorp
   recommendation](https://www.terraform.io/docs/extend/plugin-sdk.html) - [GH-382]
 
+BUG FIXES:
+* Fix `vcd_org_vdc` datasource read. When user was Organization administrator datasource failed.
```
lvirbalas: Please list all the fields which will not be returned to Org Admin here as well. Let's make it very clear for the user what is the caveat with Org Admins.

vbauzysvmware:
Changes in state file between versions and sys vs org admin. Org admin sate file isn't made due crash.
```
sdiff -s ~/Desktop/beforechangeSysAdmin ~/Desktop/afterchangeOrgAdmin 
  "serial": 217,					      |	  "serial": 211,
            "cpu_guaranteed": 0.5,			      |	            "cpu_guaranteed": null,
            "cpu_speed": 1000,				      |	            "cpu_speed": null,
            "enable_fast_provisioning": true,		      |	            "enable_fast_provisioning": null,
            "enable_thin_provisioning": true,		      |	            "enable_thin_provisioning": null,
            "memory_guaranteed": 0.5,			      |	            "memory_guaranteed": null,
            "network_pool_name": "providerVbVdc-VXLAN-NP",    |	            "network_pool_name": null,
            "provider_vdc_name": "providerVbVdc",	      |	            "provider_vdc_name": null,
                "name": "*"				      |	                "name": ""
                "name": "Development"			      |	                "name": ""
 sdiff -s ~/Desktop/beforechangeSysAdmin ~/Desktop/afterchangeSysAdmin 
  "serial": 217,					      |	  "serial": 214,
sdiff -s ~/Desktop/afterchangeSysAdmin ~/Desktop/afterchangeOrgAdmin 
  "serial": 214,					      |	  "serial": 211,
            "cpu_guaranteed": 0.5,			      |	            "cpu_guaranteed": null,
            "cpu_speed": 1000,				      |	            "cpu_speed": null,
            "enable_fast_provisioning": true,		      |	            "enable_fast_provisioning": null,
            "enable_thin_provisioning": true,		      |	            "enable_thin_provisioning": null,
            "memory_guaranteed": 0.5,			      |	            "memory_guaranteed": null,
            "network_pool_name": "providerVbVdc-VXLAN-NP",    |	            "network_pool_name": null,
            "provider_vdc_name": "providerVbVdc",	      |	            "provider_vdc_name": null,
                "name": "*"				      |	                "name": ""
                "name": "Development"			      |	                "name": ""
```

vcd/resource_vcd_org_vdc.go
```diff
@@ -311,38 +311,54 @@ func resourceVcdVdcRead(d *schema.ResourceData, meta interface{}) error {
...
-	_ = d.Set("memory_guaranteed", *adminVdc.AdminVdc.ResourceGuaranteedMemory)
+	if adminVdc.AdminVdc.ResourceGuaranteedMemory != nil {
+		_ = d.Set("memory_guaranteed", *adminVdc.AdminVdc.ResourceGuaranteedMemory)
+	}
 	_ = d.Set("name", adminVdc.AdminVdc.Name)
 
-	networkPool, err := govcd.GetNetworkPoolByHREF(vcdClient.VCDClient, adminVdc.AdminVdc.NetworkPoolReference.HREF)
-	if err != nil {
-		return fmt.Errorf("error retrieving network pool: %s", err)
+	if adminVdc.AdminVdc.NetworkPoolReference != nil {
```
Didainius: Just a check - are we safe to assume that if reference is returned - then it is a fact we are admins and this will work.

My thought - if the reference was returned (even if we have no access to it) the call below would fail and "break" read.

CHANGELOG.md
vbauzysvmware: added

vcd/resource_vcd_org_vdc.go
vbauzysvmware: For org admin in we don't have NetworkPoolReference in XML.

Didainius: Alright, agreed. I was afraid of different vCD versions misbehaving, but our testing process should capture any errors.

PR #382, opened 2019-10-30
Start using terraform-plugin-sdk v1.0.0

Didainius:
HashiCorp recently [split the provider plugin from main Terraform repo](https://www.terraform.io/docs/extend/plugin-sdk.html).

This PR is to start using the new plugin early in 2.6 development cycle so that it is properly tested before release.

**Note.** Acceptance and binary tests passed successfully on at least one version of vCD which should be sufficient to prove it didn't break functionality.


dataclouder:
Is there any procedural change or deprecation we should be aware of?

Didainius:
> Is there any procedural change or deprecation we should be aware of?

I have not yet noticed anything. Here's the quote from official site:
> The first release of the standalone plugin SDK aims to keep nearly 100% backwards compatibility, aside from a handful of APIs, so only the imports within your provider need to be replaced.

Didainius:
> Keeping the kitchen clean, nice!
> 
> Just one question: is there a measurable difference after the change in how much the provider imports (quite hard to tell from PR itself)?

It is going down a bit:
```shell
# master
$ find vendor/ | wc -l
    2575
$ cat go.sum | wc -l
     492


# PR branch
$ find vendor/ | wc -l
    2510
$ cat go.sum | wc -l
     314
```

PR #386, opened 2019-11-06
Add vm datasource and import

dataclouder:
*Implement new VM resource and data source tests
* Update docs
* Add docs for VM data source
* Add import for VM resource
* deprecate `storage_profile` in vapp, now that VM can handle it
* Add new tests for VM resource and data source


vcd/resource_vcd_vapp_vm.go
```diff
@@ -816,129 +865,172 @@ func attachDetachDisks(d *schema.ResourceData, vm govcd.VM, vdc *govcd.Vdc) erro
...
+			return fmt.Errorf("[ VM read] failed reading network details: %s", err)
 		}
-		d.Set("ip", ip)
-		d.Set("mac", mac)
+		_ = d.Set("ip", ip)
+		_ = d.Set("mac", mac)
 		// TODO v3.0 EO remove this case block when we cleanup deprecated 'ip' and 'network_name' attributes
-	case len(d.Get("network").([]interface{})) > 0:
+	default:
+		//case len(d.Get("network").([]interface{})) > 0:
```
Didainius: Not needed

vcd/resource_vcd_vapp_vm.go
```diff
@@ -816,129 +865,172 @@ func attachDetachDisks(d *schema.ResourceData, vm govcd.VM, vdc *govcd.Vdc) erro
...
-	d.Set("expose_hardware_virtualization", vm.VM.NestedHypervisorEnabled)
+	_ = d.Set("href", vm.VM.HREF)
+	_ = d.Set("expose_hardware_virtualization", vm.VM.NestedHypervisorEnabled)
+
+	cpus := 0
+	coresPerSocket := 0
+	memory := 0
+	for _, item := range vm.VM.VirtualHardwareSection.Item {
+		if item.ResourceType == 3 {
+			cpus += item.VirtualQuantity
```
Didainius: Is this being summed up from multiple resources? (the `+=` assignment)

vcd/resource_vcd_vapp_vm.go
dataclouder: fixed

vcd/resource_vcd_vapp_vm.go
```diff
@@ -816,129 +865,172 @@ func attachDetachDisks(d *schema.ResourceData, vm govcd.VM, vdc *govcd.Vdc) erro
...
 	if err != nil {
 		d.Set("disk", nil)
-		return fmt.Errorf("error reading attached disks : %s", err)
+		return fmt.Errorf("[VM read] error updating attached disks : %s", err)
 	}
 
 	guestCustomizationSection, err := vm.GetGuestCustomizationSection()
 	if err != nil {
-		return fmt.Errorf("error reading guest customization : %s", err)
+		return fmt.Errorf("[VM read] error reading guest custimization : %s", err)
```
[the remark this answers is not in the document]
dataclouder: fixed

CHANGELOG.md
```diff
@@ -1,13 +1,31 @@
...
 
 * Switch to Terraform terraform-plugin-sdk v1.0.0 as per recent [HashiCorp
   recommendation](https://www.terraform.io/docs/extend/plugin-sdk.html) - [GH-382]
+* `resource/vcd_vapp_vm`: VM state ID changed from VM name to vCD ID
+* `resource/vcd_vapp_vm`: Add properties `description` and `storage_profile`
+* `resource/vcd_vapp_vm` Add import capability and full read support ([#218](https://github.com/terraform-providers/terraform-provider-vcd/issues/218))
+
+DEPRECATIONS:
+
+* Deprecated property `storage_profile` in resource `vcd_vapp`, as the corresponding field is noe enabled in `vcd_vapp_vm`
```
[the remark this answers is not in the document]

CHANGELOG.md
```diff
@@ -1,13 +1,30 @@
 ## 2.6.0 (Unreleased)
 
+FEATURES:
+
+* **New Data Source:** `vcd_vapp_vm` VM - ([#218](https://github.com/terraform-providers/terraform-provider-vcd/issues/218))
```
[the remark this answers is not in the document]

vcd/datasource_vcd_vapp_vm.go
```diff
@@ -0,0 +1,137 @@
...
+							Type:        schema.TypeString,
+							Description: "Mac address of network interface",
+						},
+					},
+				},
+			},
+			"disk": {
+				Type: schema.TypeSet,
+				Elem: &schema.Resource{Schema: map[string]*schema.Schema{
+					"name": {
```
lvirbalas: Why don't we expose `bus_number` and `unit_number`? These two fields would be potentially pretty useful.

website/docs/d/vapp_vm.html.markdown
```diff
@@ -0,0 +1,91 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+* `vapp_name` - (Required) The vApp this VM belongs to.
+* `name` - (Required) A name for the VM, unique within the vApp 
+
+## Attribute reference
```
lvirbalas: Please confirm with an org admin user whether there are any fields which values are *not* returned and which we should mark with the newly introduced `(System Admin)` flag in the docs. It doesn't look like there should be in this data source, but just in case.

CHANGELOG.md
```diff
@@ -1,16 +1,34 @@
 ## 2.6.0 (Unreleased)
 
+FEATURES:
+
+* **New Data Source:** `vcd_vapp_vm` VM - [GH-218]
+
+BUG FIXES:
+
+* Removed `power_on` property from data source `vcd_vapp`, as it is a directive used during vApp build.
```
vbauzysvmware: This statement doesn't match UI behavior 
image

Could you explain us what is happening here

vcd/datasource_vcd_vapp_vm.go
dataclouder: They are not available after VM creation

vcd/datasource_vcd_vapp_vm_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/vmware/go-vcloud-director/v2/govcd"
+)
+
+// TestAccVcdVappDS tests a vApp data source if a vApp is found in the VDC
+func TestAccVcdVappVmDS(t *testing.T) {
+	// This test requires access to the vCD before filling templates
+	// Thus it won't run in the short test
+	if vcdShortTest {
```
vbauzysvmware: shouldn't this check should after template fill for binary tests to work?

vcd/datasource_vcd_vapp_vm_test.go
```diff
@@ -0,0 +1,87 @@
...
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: configText,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckOutput("name", vm.VM.Name),
+					resource.TestCheckOutput("storage_profile", vm.VM.StorageProfile.Name),
+					resource.TestCheckOutput("description", vm.VM.Description),
+					resource.TestCheckOutput("href", vm.VM.HREF),
```
vbauzysvmware: shouldn't we check all values are there after read - to support future changes?

vcd/datasource_vcd_vapp_vm.go
lvirbalas: They are if you look at VM details in UI:

Screen Shot 2019-11-06 at 15 30 14

vcd/resource_vcd_vapp_vm.go
```diff
@@ -91,34 +110,42 @@ func resourceVcdVAppVm() *schema.Resource {
 				Type:          schema.TypeString,
 			},
 			"initscript": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "A script to be run only on initial boot",
```
vbauzysvmware:
```suggestion
				Description: "A script to be run only on initial boot. Will be inherited from template if found in it",
```

vcd/resource_vcd_vapp_vm.go
```diff
@@ -287,20 +324,32 @@ func resourceVcdVAppVmCreate(d *schema.ResourceData, meta interface{}) error {
...
+	storageProfileName := d.Get("storage_profile").(string)
+	if storageProfileName != "" {
+		storageProfile, err = vdc.FindStorageProfileReference(storageProfileName)
+		if err != nil {
+			return fmt.Errorf("[vm creation] error retrieving storage profile %s : %s", storageProfileName, err)
+		}
+	} else {
+		storageProfilePtr = nil
+	}
+	task, err := vapp.AddNewVMWithStorageProfile(d.Get("name").(string), vappTemplate, &networkConnectionSection, storageProfilePtr, acceptEulas)
```
vbauzysvmware: Doesn't this work by avoiding  storageProfilePtr and just having var and passing `*storageProfile` ?

vcd/datasource_vcd_vapp_vm.go
dataclouder: Screenshot 2019-11-06 14 35 47

I trusted the opinion of the previous implementer about this matter.
Such information is not in the VM structure. I need to find out where it is is stored.

website/docs/d/vapp_vm.html.markdown
dataclouder: Tests with Org Admin pass without errors

website/docs/d/vapp_vm.html.markdown
```diff
@@ -0,0 +1,91 @@
...
+
+* `computer_name` -  Computer name to assign to this virtual machine. 
+* `catalog_name` -  The catalog name in which to find the given vApp Template
+* `template_name` -  The name of the vApp Template to use
+* `memory` -  The amount of RAM (in MB) allocated to the VM
+* `cpus` -  The number of virtual CPUs allocated to the VM
+* `cpu_cores` -  The number of cores per socket
+* `metadata` -  Key value map of metadata assigned to this VM
+* `disk` -  Independent disk attachment configuration.
```
vbauzysvmware: Just @lvirbalas @Didainius any ideas how we approach with naming regarding independent disks and internal-normal disks? as today `disk` is independent disk. Just wanted to point here maybe we need some decisions what todo with this and as result could impact current situation.

website/docs/r/vapp_vm.html.markdown
```diff
@@ -132,14 +132,15 @@ The following arguments are supported:
 * `cpu_cores` - (Optional; *v2.1+*) The number of cores per socket
 * `metadata` - (Optional; *v2.2+*) Key value map of metadata to assign to this VM
 * `initscript` (Optional) A script to be run only on initial boot
```
vbauzysvmware:
```suggestion
* `initscript` (Optional) A script to be run only on initial boot. Will be inherited from template if found in it
```

CHANGELOG.md
dataclouder: This has been explained in the vApp PR.
`power_on` is not a reliable state. It is defined as such when state code = 4. However, this code can have several values, which may or may not be compatible with the state of being powered on. 
For this reason, the property `power_on` is used as a directive, but we retrieve the naked status code and status text as separate fields.

The UI gets such status and has some internal code to determine when a status should be considered ON. According to the API, only status n. 4 is "ON", but this field is overloaded with several more meanings that have nothing to do with being on or not.

vcd/datasource_vcd_vapp_vm_test.go
dataclouder: Just as the comment above says, this cannot happen. 
The short test runs without connection to the vCD, but this test requires connection to the vCD just to determine whether it can run or not.

vcd/datasource_vcd_vapp_vm_test.go
dataclouder: There is another test that does a thorough check of all properties

vcd/resource_vcd_vapp_vm.go
dataclouder: No. We need to pass a `nil` when there is no storage profile. 
If we pass `*storageProfile` it would be a pointer to an empty structure

vcd/resource_vcd_vapp_vm.go
dataclouder: Is this documented somewhere?

vbauzysvmware: https://github.com/terraform-providers/terraform-provider-vcd/pull/347 haven't done that. So see opportunity to add missing detail.

CHANGELOG.md
vbauzysvmware: My idea come from diff point of view  - didn't this field allowed to initiate power one or power off?

vcd/resource_vcd_vapp_vm.go
dataclouder: Does everyone agree on this change?
@lvirbalas @Didainius ?

CHANGELOG.md
dataclouder: It does. It is used in the resource, but not fetched when we read for the reason stated above. Hence, it is not needed in the **data source**

vbauzysvmware: oh yes, this is only datasource. thank you for explanation

vcd/datasource_vcd_vapp_vm.go
dataclouder: Found it. It is stored in a structure that is returned using API 32+.
([VmSpecSectionType](https://vdc-download.vmware.com/vmwb-repository/dcr-public/037ccaee-649a-417e-b365-1331034fb28d/1f0fd9eb-0238-4af6-89b5-7e6636f29c65/doc/doc/types/VmSpecSectionType.html))
We can't use it currently, as we are still bound to version 27

website/docs/d/vapp_vm.html.markdown
dataclouder: Can we discuss this matter outside of this PR?
Renaming `disk` would be a breaking change, so it isn't something we can fix here.

vcd/resource_vcd_vapp_vm.go
dataclouder: That's my understanding of the structure.

vcd/resource_vcd_vapp_vm.go
Didainius: I have just manually checked these two points:
* In fact, when one does `customization.force=true` it also runs the script so this is not fully correct to say "only on initial boot".
* I have added init script to VM, created a catalog template out of it and spawned new VM with .tf without adding init script and the results is - VM has the script set and it is being executed on first boot (even when my `initscript` parameter is not specified).

website/docs/d/vapp_vm.html.markdown
lvirbalas: Interesting point, @vbauzysvmware . Most important is consistency, if `disk` is currently used exclusively for independent disk, we can keep it as is and introduce a new name for internal-normal disks. What about this approach?

vcd/resource_vcd_vapp_vm.go
lvirbalas: What about this then:

"Script to run on initial boot or with `customization.force=true` set. Also, inherited from catalog template if found in it"

BTW, when template has it implicitly, do we **really** see the script in this `initscript` field after read?

vcd/resource_vcd_vapp_vm.go
```diff
@@ -53,26 +59,40 @@ func resourceVcdVAppVm() *schema.Resource {
...
+				// Note: description is read only, as we lack the needed fields to set it at creation.
+				// Currently, this field has the description of the OVA used to create the VM
 			},
 			"memory": &schema.Schema{
-				Type:     schema.TypeInt,
-				Optional: true,
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Description:  "The amount of RAM (in MB) to allocate to the VM",
+				ValidateFunc: validateMultipleOf4(),
```
lvirbalas: Why is this validate function added? Is this what vCD expects and if yes, why is it a multiple of four and not eight?

dataclouder: If we don't have this validation, and the memory is incorrect (for example `1` or `103`) we get an error saying "The virtual machine memory size must be a multiple of 4 MB".
However, the error happens **after the virtual machine has already been created**, leaving the resource in a tainted state.
Using this validation, the failure happens immediately, and users can change the offending value and try again.

vcd/datasource_vcd_vapp_vm_test.go
vbauzysvmware: you mean resource test?

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: I think API returns it, so we see it.

vcd/datasource_vcd_vapp_vm_test.go
dataclouder: yes

website/docs/d/vapp_vm.html.markdown
vbauzysvmware: Well it was one of idea also, using some additional word as "prefix". Another way we decide to remove attaching to separate resource as we were asked. And not support attaching in VM with current limited behaviour.

vcd/resource_vcd_vapp_vm.go
lvirbalas: Which vCD version did you observe this error in?

vcd/resource_vcd_vapp_vm.go
Didainius: > BTW, when template has it implicitly, do we **really** see the script in this `initscript` field after read?

> If I recall correctly - API returns it, so we see it.

No. The answer is we don't. At least here is what I see in the statefile `"initscript": null,`after creating from template and after additionally trigerring `plan/apply`

vcd/resource_vcd_vapp_vm.go
lvirbalas: Adding the original comment from community user where the need for this originated for reference:
https://github.com/terraform-providers/terraform-provider-vcd/issues/239#issuecomment-550375527

vcd/resource_vcd_vapp_vm.go
lvirbalas: Thank you. Then modified suggestion:
"Script to run on initial boot or with customization.force=true set"

vcd/resource_vcd_vapp_vm.go
dataclouder: > Which vCD version did you observe this error in?

9.5 and 10.0

website/docs/r/vapp_vm.html.markdown
dataclouder: updated

vcd/resource_vcd_vapp_vm.go
dataclouder: updated

website/docs/d/vapp_vm.html.markdown
dataclouder: Is there any actionable item for this PR in this conversation?

CHANGELOG.md
```diff
@@ -1,16 +1,34 @@
...
+  Its state is never updated and the fields `status` and `status_text` already provide the necessary information.
+  [GH-379]
+
 IMPROVEMENTS:
 
 * Switch to Terraform terraform-plugin-sdk v1.0.0 as per recent [HashiCorp
   recommendation](https://www.terraform.io/docs/extend/plugin-sdk.html) - [GH-382]
+* `resource/vcd_vapp_vm`: VM state ID changed from VM name to vCD ID
+* `resource/vcd_vapp_vm`: Add properties `description` and `storage_profile`
```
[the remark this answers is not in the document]

CHANGELOG.md
```diff
@@ -1,16 +1,34 @@
...
+
+* Removed `power_on` property from data source `vcd_vapp`, as it is a directive used during vApp build.
+  Its state is never updated and the fields `status` and `status_text` already provide the necessary information.
+  [GH-379]
+
 IMPROVEMENTS:
 
 * Switch to Terraform terraform-plugin-sdk v1.0.0 as per recent [HashiCorp
   recommendation](https://www.terraform.io/docs/extend/plugin-sdk.html) - [GH-382]
+* `resource/vcd_vapp_vm`: VM state ID changed from VM name to vCD ID
```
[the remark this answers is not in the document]

website/docs/d/vapp_vm.html.markdown
Didainius: I think this discussion is outside of this PR. The decision about it can go within disk PR.

lvirbalas: I agree that this is a separate of this PR discussion.

vcd/resource_vcd_vapp_vm.go
Didainius: I think looking at XML I saw only one structure, but this shouldn't make any harm anyway.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -364,15 +414,15 @@ func resourceVcdVAppVmCreate(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
-	d.SetId(d.Get("name").(string))
+	d.SetId(vm.VM.ID)
```
lvirbalas: How does the "upgrade" from 2.5.0 looks like for someone who already has VMs? Are there any challenges?

Didainius: I haven't observed any while I was trying this out. Tried 2.5 -> 2.6 multiple times and `apply` was clean.

dataclouder: I have seen no issues so far.

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: > > BTW, when template has it implicitly, do we **really** see the script in this `initscript` field after read?
> 
> > If I recall correctly - API returns it, so we see it.
> 
> No. The answer is we don't. At least here is what I see in the statefile `"initscript": null,`after creating from template and after additionally trigerring `plan/apply`

You would see if you create VM from template which had init script configured in it.

Issue #323, opened 2019-09-05
Additional edge gateway external network features: multiple external networks, manual IP assignment, IP pool sub-allocation


okozarva:
### Terraform Version
Terraform v0.12.5
+ provider.vcd v2.4.0

### Affected Resource(s)
Please list the resources as a list, for example:
- vcd_edgegateway

### Expected Behavior
- We have multiple subnets in the external network. But, there is no option to which IP and gateway will be use when creating edge gateway.

![1](https://user-images.githubusercontent.com/23477945/64338319-29acd400-cfea-11e9-9c8b-91a0d9dd7ac8.PNG)
![2](https://user-images.githubusercontent.com/23477945/64338465-7e504f00-cfea-11e9-8fe7-8f86d46d81a2.PNG)

- Secondly, there is no option for sub allocate ip.

![3](https://user-images.githubusercontent.com/23477945/64338558-c0799080-cfea-11e9-95eb-d2ec535e5b6b.PNG)

Didainius:
Hi @okozarva, thanks you for the issue.
Just want to clarify a few more details before we pull it in for the roadmap.

So first thing. You need ability

Didainius:
Partly related to https://github.com/terraform-providers/terraform-provider-vcd/issues/308 (could be done together)

Didainius:
Hi @okozarva, thanks for the issue. I need to clarify a bit with you so that this is clearer when someone starts to work.

You want to be able to manually specify 1 or more IP addresses in different subnets (and  external networks) to be attached to edge gateway?

<img width="902" alt="image" src="https://user-images.githubusercontent.com/15804230/64413021-65eb3d80-d099-11e9-9879-8c5cba25036a.png">
<img width="887" alt="image" src="https://user-images.githubusercontent.com/15804230/64413286-e4e07600-d099-11e9-8073-25e61a99dedf.png">


As well as use sub-allocated ranges in the external network subnets?

<img width="873" alt="image" src="https://user-images.githubusercontent.com/15804230/64413652-8f589900-d09a-11e9-9fb3-76ab95c2ae67.png">

okozarva:
Hi @Didainius,

Thanks for your quick response.

That's features will be perfect.

-  be able to manually specify 1 or more IP addresses in different subnets

- be able to use sub-allocated ranges in the external network subnets

carmineesposito73:
[...]

PR #389, opened 2019-11-08
Add `default_external_network_ip` computed field to resource and data source `vcd_edge_gateway`

### LINKED ISSUE #308: Feature Request: Return assigned IP of default edge gateway network when creating/updating
### Terraform Version
Terraform v0.12.6
+ provider.vcd v2.4.0

### Affected Resource(s)
- vcd_edgegateway

When creating an edge gateway, it would be useful to fetch the IP assigned to the edge gateway for further processing, i.e. creating snat/dnat rules or sending the IP to other resources. Not sure if this is an easy fix, or if it requires much modification.

Didainius:
This PR adds ~~`default_network_ip`~~ `default_external_network_ip` field to resource and data source `vcd_edge_gateway` which allows to read and use IP address of default gateway interface. It can be very useful in scenarios where this IP is immediately useful in other places (like opening firewalls, running scripts, etc)

Closes #308, also puts some groundwork for #323

### DIFF
--- a/vcd/datasource_vcd_edgegateway.go
+++ b/vcd/datasource_vcd_edgegateway.go
@@ -51,7 +51,7 @@ func datasourceVcdEdgeGateway() *schema.Resource {
 				Computed:    true,
 				Description: "External network to be used as default gateway. Its name must be included in 'external_networks'. An empty value will skip the default gateway",
 			},
-			"default_network_ip": &schema.Schema{
+			"default_external_network_ip": &schema.Schema{
 				Type:        schema.TypeString,
 				Computed:    true,
 				Description: "IP address of edge gateway interface which is used as default.",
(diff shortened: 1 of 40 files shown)


website/docs/r/edgegateway.html.markdown
```diff
@@ -84,6 +84,12 @@ order) logging. Default `false`.
 * `fw_default_rule_action` (Optional) Default firewall rule (last in the processing order) action.
 One of `accept` or `deny`. Default `deny`.
 
+## Attribute Reference
+
+The following attributes are exported on this resource:
+
+* `default_network_ip` (*v2.6+*) - IP address of edge gateway used for default network
```
lvirbalas: I'm thinking that it would be useful to show case this new field in one of our examples. It could be here or in some other resource page where EG IP is used.

Didainius: Added two examples in load balancer virtual server and firewall rule

vcd/datasource_vcd_edgegateway.go
```diff
@@ -51,6 +51,11 @@ func datasourceVcdEdgeGateway() *schema.Resource {
 				Computed:    true,
 				Description: "External network to be used as default gateway. Its name must be included in 'external_networks'. An empty value will skip the default gateway",
 			},
+			"default_network_ip": &schema.Schema{
```
lvirbalas: Just before merging, let's decide whether it would be helpful to rename `default_network_ip` to `default_external_network_ip`

CC: @dataclouder @vbauzysvmware

Issue #262, opened 2019-06-12
Create an edge gateway resource


dataclouder:
A resource `vcd_edgegateway` allows the creation and deletion of edge gateways. This task is t is the extension of [go-vcloud-director Issue 130](https://github.com/vmware/go-vcloud-director/issues/130) where the functionality was implemented.

The basic features of an edge gateway resources are shown in this example
```hcl
resource "vcd_edgegateway" "egw" {
  org = "my-org"
  vdc = "my-vdc"

  name              = "my-egw"
  description       = "new edge gateway"
  backing_config    = "compact"
  default_gateway   = "my-ext-net1"
  external_networks = [ "my-ext-net1", "my-ext-net2" ]
}
```
The full list of properties:

* `org` - (Optional) The name of organization to which the VDC belongs.
* `vdc` - (Optional) The name of VDC that owns the edge gateway.
* `name` - (Required) A unique name for the edge gateway.
* `external_networks` - (Required) An array of external network names.
* `backing_config` - (Required) Configuration of the vShield edge VM for this gateway. One of: `compact`, `full`.
* `default_gateway` - (Optional) Name of the external network to be used as default gateway. It must be included in the 
  list of `external_networks`. Providing an empty string or omitting the argument will create the edge gateway without a default gateway.
* `advanced` - (Required) True if the gateway uses advanced networking.
* `ha_enabled` - (Optional) Enable high availability on this edge gateway. Default is `false`.
* `distributed_routing` - (Optional) If advanced networking enabled, also enable distributed routing. Default is `false`.

Current limitations: 

* [There is no updating capability](https://github.com/vmware/go-vcloud-director/issues/202)
* [only two configuration values are allowed](https://github.com/vmware/go-vcloud-director/issues/201)

lvirbalas:
Does `advanced_networking` option represent "Create as Advanced Gateway" in the GUI? If yes, I suggest we name it just `advanced`, as we are aligning Terraform part to the GUI.

dataclouder:
Yes, it does. I will change it to "advanced" and I will need to also make it Required, because in 9.7 it becomes `true` by default, and vCD ignores input on this field. If we leave it optional, Terraform will fail the check when the property is not provided.
It's easier to make it mandatory and make sure that it is true on 9.7+

lvirbalas:
Why not just make it true by default? Then the user can even not use this field and only define it if someone ones the not advanced one.

dataclouder:
[...]

PR #393, opened 2019-11-13
Add ability of using authorization tokens

dataclouder:
* Provider can connect using property 'token'
* When token is used, username and password are ignored
* Add script get_token.sh
* Add command 'make token'

Implementation notes:

* As mentioned in [go-vcloud-director PR #262](https://github.com/vmware/go-vcloud-director/pull/262), the user/password information and the token are not correlated within the provider code, meaning that we may accidentally use a token that was not created using those credentials.

* Properties `user` and `password` are still mandatory when using tokens. They can be empty, but not missing from the HCL script. If we make them optional, we may break compatibility with current functionality, where a missing property is detected during plan.

* `make token` is not mentioned in the provider documentation, as users of the provider may not be using the code, but downloading the provider directly. Also, the example on how to get the token is Unix only. The alternative would be creating a cross-platform tool, but it may be overkill.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ IMPROVEMENTS:
 * Fix code warnings from `staticcheck` and add command `make static` to Travis tests
 * `resource/vcd_edge_gateway` and `datasource/vcd_edge_gateway` add `default_external_network_ip`
   field to export default edge gateway IP address - [GH-389]
+* The `vcd` provider now has the ability of connecting through an authorization token - [GH-280]
+* Add command `make token` to create an authorization token from testing credentials
 
 BUG FIXES:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ IMPROVEMENTS:
 * Fix code warnings from `staticcheck` and add command `make static` to Travis tests
 * `resource/vcd_edge_gateway` and `datasource/vcd_edge_gateway` add `default_external_network_ip`
   field to export default edge gateway IP address - [GH-389]
-* The `vcd` provider now has the ability of connecting through an authorization token - [GH-280]
+* Add `token` to the `vcd` provider for the ability of connecting with an authorization token - [GH-280]
 * Add command `make token` to create an authorization token from testing credentials
 
 BUG FIXES:
--- a/scripts/get_token.sh
+++ b/scripts/get_token.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+# This script will connect to the vCD using username and password,
+# and show the header that cointains an authorization token.
+#
+user=$1
+password=$2
+org=$3
+IP=$4
+
+if [ -z "$IP" ]
+then
+    echo "Syntax $0 user password organization IP_address"
+    exit 1
+fi
+
+auth=$(echo -n "$user@$org:$password" |base64)
+
+curl -I -k --header "Accept: application/*;version=27.0" \
+    --header "Authorization: Basic $auth" \
+    --request POST https://$IP/api/sessions
+
+# If successful, the output of this command will include a line like the following
+# x-vcloud-authorization: 08a321735de84f1d9ec80c3b3e18fa8b
+#
+# The string after `x-vcloud-authorization:` is the token.
--- a/scripts/get_token.sh
+++ b/scripts/get_token.sh
@@ -9,13 +9,13 @@ IP=$4
 
 if [ -z "$IP" ]
 then
-    echo "Syntax $0 user password organization IP_address"
+    echo "Syntax $0 user password organization hostname_or_IP_address"
     exit 1
 fi
 
 auth=$(echo -n "$user@$org:$password" |base64)
 
-curl -I -k --header "Accept: application/*;version=27.0" \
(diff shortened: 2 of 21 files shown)


scripts/get_token.sh
```diff
@@ -0,0 +1,25 @@
...
+# and show the header that cointains an authorization token.
+#
+user=$1
+password=$2
+org=$3
+IP=$4
+
+if [ -z "$IP" ]
+then
+    echo "Syntax $0 user password organization IP_address"
```
lvirbalas: Maybe there's an option to **not** duplicate this script which is already in go-vcloud-director? Maybe some way to pull it in into `vendor/`?

website/docs/index.html.markdown
```diff
@@ -122,6 +177,10 @@ The following arguments are used to configure the VMware vCloud Director Provide
 * `password` - (Required) This is the password for vCloud Director API operations. Can
   also be specified with the `VCD_PASSWORD` environment variable.
   
+* `token` - (Optional; *v2.6+*) This is the authorization token that can be used
+   instead of username and password. When this is set, username and password will not
+    be used and may be left empty, or may contain other remarks.
```
lvirbalas: What other remarks? Need to be more specific.

vcd/provider_test.go
```diff
@@ -25,10 +25,10 @@ func TestProvider_impl(t *testing.T) {
 // When this function is called, the initialization in config_test.go has already happened.
 // Therefore, we can safely require that testConfig fields have been filled.
 func testAccPreCheck(t *testing.T) {
-	if testConfig.Provider.User == "" {
+	if testConfig.Provider.User == "" && testConfig.Provider.Token == "" {
 		t.Fatal("provider.user must be set for acceptance tests")
```
vbauzysvmware:
```suggestion
		t.Fatal("provider.user or token must be set for acceptance tests")
```

vcd/provider_test.go
```diff
@@ -25,10 +25,10 @@ func TestProvider_impl(t *testing.T) {
...
 	}
-	if testConfig.Provider.Password == "" {
+	if testConfig.Provider.Password == "" && testConfig.Provider.Token == "" {
 		t.Fatal("provider.password must be set for acceptance tests")
```
vbauzysvmware:
```suggestion
		t.Fatal("provider.password or token must be set for acceptance tests")
```

scripts/get_token.sh
dataclouder: In vendor it would not go unless moved explicitly, and we might create conflicts during `go mod tidy` operations.
One way would be to link to its address on GitHub.
Or we can remove it altogether and only use `make token` in this repository.

website/docs/index.html.markdown
dataclouder: Something like 
```
  user     = "USING TOKEN"
  password = "USING TOKEN"
  token    ="d1e2a3d4b5e5e6f7"
```
This way, if there are tools that extract such fields, they will show the remarks instead of the real value.

vcd/provider_test.go
dataclouder: thanks. Updated

vcd/provider_test.go
dataclouder: Thanks. Updated

website/docs/index.html.markdown
lvirbalas: Then suggesting this rephrase:

```
When `token` is set, username and password will be ignored, but should be left in configuration either empty or with any custom values.
```

dataclouder: done

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ IMPROVEMENTS:
 * Fix code warnings from `staticcheck` and add command `make static` to Travis tests
 * `resource/vcd_edge_gateway` and `datasource/vcd_edge_gateway` add `default_external_network_ip`
   field to export default edge gateway IP address - [GH-389]
+* The `vcd` provider now has the ability of connecting through an authorization token - [GH-280]
```
lvirbalas: Thinking how to make this more noticeable, maybe like this:
```suggestion
* Add `token` to the `vcd` provider for an ability of connecting with an authorization token - [GH-280]
```

dataclouder: updated

buec011:
Works for me:
```
./get_token.sh "$VCD_USER" "$VCD_PASSWORD" "$VCD_SYS_ORG" $(echo "$VCD_URL" | sed -r 's#^https{0,1}://##' | sed -r 's#/api.*##')

> export VCD_TOKEN=123456789
> export VCD_USER=x
> export VCD_PASSWORD=x
> unset VCD_SYS_ORG

> env | grep VCD
VCD_URL=https://testvcd/api
VCD_ORG=testorg
VCD_TOKEN=123456789
VCD_USER=x
VCD_PASSWORD=x
VCD_VDC=testorg-testvdc
VCD_MAX_RETRY_TIMEOUT=300

> terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

vcd_vapp.vapp: Refreshing state...

```

PR #400, opened 2019-11-19
Upgrade tests

dataclouder:
* Add upgrade test environment
    * Add a `PREVIOUS_VERSION` file, containing the version from which  we want to test the upgrade
    * Add script `test-upgrade.sh` to drive upgrade operations
    * Modify `test-binary.sh` to support upgrade ops
    * Add command `make test-upgrade`

* Remove interpolation-only expressions from tests as they are causing a huge amount of warning text during binary tests.

How to run the upgrade tests:
1. Make sure you are using `terraform` binary 0.12.13. Using the latest version will blow your screen with lots of warning about interpolation-only expressions. Although these expressions are fixed in the current release, part of the test will run using code from the previous release.
2. Make sure the branch doesn't have any uncommitted files.
3. Run `make test-upgrade`


vcd/provider_unit_test.go
```diff
@@ -48,6 +50,57 @@ func TestProviderVersion(t *testing.T) {
...
+	previousVersion, err := semver.NewVersion(previousVersionText)
+	if err != nil {
+		t.Logf("error converting previous version to Hashicorp version: %s", err)
+		t.Fail()
+		return
+	}
+	result := currentVersion.Compare(previousVersion)
+	// result < 0 means current version is lower than previous version
+	// result == 0 means current version is the same as previous version
+	// result == 0 means current version is higher than previous version
```
lvirbalas: "==" looks to be an inertia :)

dataclouder: Thanks. Fixed

Issue #374, opened 2019-10-30
Improvements needed for network routed


vbauzysvmware:
* resource doesn't provide ability to provide description
* when pool address wrong, error isn't precise:
```
Error: error adding DHCP pool: error reconfiguring Edge Gateway: API Error: 400: [ 84284038-c2bd-41c0-86ff-3c608d003191 ] validation error on field 'configureGatewayServices.arg0.gatewayServiceConfigurationSpec.dhcpSpec.dhcpPools[0].lowIpAddress': String value has invalid format
```
* when Dns ip is wrong, error isn't precise:
```
Error: error: error creating the network: error instantiating a new OrgVDCNetwork: API Error: 400: [ afe4b21d-bee8-4e94-ac10-a228ca64cf32 ] validation error on field 'ipSettingList[0].nameServer1': String value has invalid format
```
* !!! when one of pool address is wrong - network is created, but don't exist in state file cause error is thrown.
* dhcp_pool field with details isn't refreshed when dhcp pool settings enabled/disabled or ip range is removed in UI. This can be reproduced: create route netw, goto actions "configure services" dhcp pool tab. !!! Overall another question risen here: should we allow to change edge gtw dhcp pool settings - as we would impact other networks which routed form same edge gtw?

datacharmer:
Please add an example that reproduces the issue. What's a "wrong pool address" or "wrong Dns"?

vbauzysvmware:
resource "vcd_network_routed" "TestAccVcdNetworkRoutedDhcp" {
  name         = "TestAccVcdNetworkRoutedDhcp"
  org          = "my1"
  vdc          = "vbVdc"
  edge_gateway = "vb_edge2"
  gateway      = "10.10.102.1111"
  dhcp_pool {
    start_address = "10.10.102.5111"
    end_address   = "10.10.102.100"
  }
}

dataclouder:
Fixed in #402

Issue #373, opened 2019-10-30
Improvements needed for network isolated


vbauzysvmware:
* resource doesn't provide ability to provide description
* when Dns ip is wrong, error isn't precise:
```
Error: error: error creating the network: error instantiating a new OrgVDCNetwork: API Error: 400: [ afe4b21d-bee8-4e94-ac10-a228ca64cf32 ] validation error on field 'ipSettingList[0].nameServer1': String value has invalid format
```
* when one of pool address  is wrong then error isn't precise too:
```
Error: error: error creating the network: error instantiating a new OrgVDCNetwork: API Error: 400: [ 6b72cf9a-e95a-464f-bc76-0adbeff8e783 ] validation error on field 'ipSettingList[0].ipAddressRanges[0].startAddress': String value has invalid format
```
* dhcp_pool field with details isn't refreshed when dhcp pool settings enabled/disabled or ip range is removed in UI. This can be reproduced: create isolated netw, goto actions "configure services" :
![image](https://user-images.githubusercontent.com/41415459/67848932-4b1fcd00-fb0e-11e9-8686-1491d78998e5.png)

dataclouder:
Please provide sample HCL scripts that reproduces the issues.

vbauzysvmware:
`resource "vcd_network_isolated" "TestAccVcdNetworkIsoMixed" {
  name       = "TestAccVcdNetworkIsoMixed"
  org        = "my1"
  vdc        = "vbVdc"
  gateway    = "192.168.2.1111"
  dns1       = "192.168.2.1"
  static_ip_pool {
    start_address = "192.168.2.2222"
    end_address   = "192.168.2.50"
  }
  dhcp_pool {
    start_address = "192.168.2.51"
    end_address   = "192.168.2.100"
  }
}`

dataclouder:
Fixed in #402

Issue #321, opened 2019-09-05
Add subinterface as interface type when creating a routed Org VDC Network


swerveshot:
### Terraform Version
Terraform v0.12.7
\+ provider.vcd v2.4.0

### Affected Resource(s)
- vcd_network_routed
- vcd_edgegateway

### Expected Behavior
As a vCD system user I would like to have the ability to create an Organization VDC routed network that is created as a subinterface on an Edge Gateway.

![image](https://user-images.githubusercontent.com/10358044/64327753-85686480-cfcc-11e9-8a33-b2b917ec32f5.png)

### Actual Behavior
Currently when creating a routed network with the vCloud Director Provider this network is attached to the edge gateway as a regular interface. This means that only a maximum of 9 networks can be attached to that edge. As you can attach a maximum of 10 NICs to a vSphere VM. If you were to use the option to have the network use a subinterface a maximum of 200 networks can be used.

### Steps to Reproduce
- `terraform apply`
- observe the network created in the vCD UI and check the interface type

### References
- [vCloud Director Administrator's Guide](https://docs.vmware.com/en/vCloud-Director/9.7/com.vmware.vcloud.admin.doc/GUID-6E69AF88-31E0-4DD8-A79E-E8E4B6F68878.html)

lvirbalas:
Thank you @swerveshot for defining the needed feature so nicely!

dataclouder:
IMplemented with #402

PR #402, opened 2019-11-25
Add property "interface_type" and improve networks

dataclouder:
* Add property "interface_type" to routed network, allowing the choice
  among "internal" (default), "subinterface", and "distributed" [#321]
* Add property "description" in all three network resources.
* Add validation for IP fields in routed and isolated networks. [#373] [#374]
* Update documentation

Note for testing:
To test distributed networking, you need to have an edge gateway with distributed routing enabled, and then you need to enable such tests by setting the variable `VCD_TEST_DISTRIBUTED_NETWORK` to a non-empty value.


vcd/resource_vcd_network_routed.go
```diff
@@ -444,26 +511,42 @@ func resourceVcdNetworkIPAddressHash(v interface{}) int {
 // On success, returns the name of the edge gateway
 func findEdgeGatewayConnection(client *VCDClient, vdc *govcd.Vdc, network *govcd.OrgVDCNetwork) (string, error) {
```
lvirbalas: The name of this function is very confusing. I was reading the big comment at:
https://github.com/terraform-providers/terraform-provider-vcd/pull/402/files#diff-e99542a876fc291b70cf88bd368423b5R41-R48

It just didn't make any sense, then I looked at this function and realized it's not returning a "Connection". Better name needed.

vcd/datasource_vcd_network_test.go
```diff
@@ -38,6 +38,14 @@ func detectNetwork(vcdClient *VCDClient, vdc *govcd.Vdc, network *govcd.OrgVDCNe
...
 	}
 	edgeGatewayName, err := findEdgeGatewayConnection(vcdClient, vdc, network)
+	// There is a potential false positive returning from findEdgeGatewayConnection:
+	// When a non-routed network has the same name as the edge gateway,
+	// this function will return as if it had found an edge gateway.
+	// This fact won't affect regular operations, as this function is called
+	// from functions that assume a routed network. The only place where
+	// it could cause a problem is in the data source tests (this file), where this function
+	// is used to find the network type. If we avoid calling both the edge gateway and the
+	// isolated network "foo", all will be well.
```
lvirbalas: Why not to add this comment to the function itself?
But first, why is this happening, how come network is even considered together with a totally different type of EG?

vcd/resource_vcd_network_routed.go
dataclouder: The name says what the function is doing: finding a connection between a routed network and an edge gateway. Please suggest a better name if you have one.

vcd/datasource_vcd_network_test.go
dataclouder: The comment is here because it may only affect this kind of test. Putting it in the function would raise unnecessary questions about risks that don't exist for regular usage.

vcd/resource_vcd_network_routed.go
dataclouder: More background info:
when we create a routed network, we know the name of the edge gateway. When we retrieve a network for importing into a state file, we don't know the edge gateway name, as it is not needed to retrieve the network. Thus, once we have the network, we need to find to which edge gateway is connected. This function exists since the creation of the network data source. It was using a different algorithm then, by looking at the existing edge gateways for a connected network. However, for distributed networks the edge gateway doesn't store a connection, and we had to change the method of connection retrieval using a query.

lvirbalas: Makes it clearer. Assuming I got the purpose right, then the name suggestion is:
`findEdgeGatewayNameByNetwork`
This would be clearer and consistent with our format of `GetEdgeGatewayByName`

Finally, I just realized that this function is in provider, we should really move it to govcd.

lvirbalas: One more variant:
`findEdgeGatewayNameOfNetwork`

dataclouder:
`findGatewayConnection` moved to go-vcloud-director as `vdc.FindEdgeGatewayNameByNetwork`

vcd/datasource_vcd_network_test.go
```diff
@@ -285,6 +296,10 @@ output "network_name" {
   value = data.vcd_network_direct.{{.NetworkName}}.name
 }
 
+output "network_description" {
```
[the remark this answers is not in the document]
dataclouder: I did that for a specific reason: using output fields will allow me to use the generated HCL script to see the result when running it with `terraform` tool. If I don't put them, I will have to edit the file to see the result.  Since I had to do this operation several times when I was developing, I thought that I get the best of two worlds with this approach: result in the integrated tests, and visible output when running manually.

CHANGELOG.md
```diff
@@ -31,6 +31,12 @@ IMPROVEMENTS:
 * Add command `make token` to create an authorization token from testing credentials
 * Clean up interpolation-only expressions from tests (as allowed in terraform v0.12.11+)
 * Increment vCD API version used from 27.0 to 29.0 [GH-396]
+* `resource/vcd_network_routed` Add properties `description` and `interface` [GH-321] [GH-342] [GH-374]
```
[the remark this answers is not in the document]

vcd/resource_vcd_network_routed.go
dataclouder: Moved to govcd

vcd/datasource_vcd_network_test.go
dataclouder: removed comment and refactored as the interested function was moved to govcd

vcd/resource_vcd_network_routed.go
```diff
@@ -214,6 +246,27 @@ func resourceVcdNetworkRoutedCreate(d *schema.ResourceData, meta interface{}) er
...
+	case "internal":
+		// default: no configuration is needed
+		orgVDCNetwork.Configuration.SubInterface = nil
+	case "subinterface":
+		orgVDCNetwork.Configuration.SubInterface = &trueValue
+	case "distributed":
+		if distributedAllowed {
+			orgVDCNetwork.Configuration.DistributedInterface = &trueValue
+		} else {
+			return fmt.Errorf("interface 'distributed' requested, but edge gateway '%s' not enabled", edgeGateway.EdgeGateway.Name)
```
Didainius: I'd think this could be rephrased to better. Currently I read the error - "edge gateway is not enabled" while in fact "distributed routing is not enabled on edge gateway"

dataclouder: Thanks. Changed

vcd/resource_vcd_network_routed.go
```diff
@@ -50,35 +50,56 @@ func resourceVcdNetworkRouted() *schema.Resource {
...
+			},
+
+			"interface_type": &schema.Schema{
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          "internal",
+				ForceNew:         true,
+				Description:      "Which interface to use (one of `internal`, `subinterface`, `distributed`)",
+				ValidateFunc:     validation.StringInSlice([]string{"internal", "subinterface", "distributed"}, true),
+				DiffSuppressFunc: suppressNetworkUpgradedInterface(),
```
vbauzysvmware: As far as I got it it's support back compatibility. So comment would be good regarding that here

PR #406, opened 2019-11-27
Add vcd_ipset resource and data source

Didainius:
This PR introduces IP set resource and data source (`vcd_ipset`) for handling IP sets.
At the moment there is no much use of them, but soon PRs will come in to enable support in `vcd_nsxv_firewall_rule` resource and later on DHCP pool handling in edge gateway (will need additional new resources)

**Note** It bumps `terraform-provider-sdk` version 1.0.0 -> 1.3.0 therefore there are quite a lot of changes in `vendor`


website/docs/d/ipset.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+* `name` - (Required) IP set name for identifying the exact IP set
+
+## Attribute Reference
+
+All the attributes defined in `vcd_ipset` resource are available.
```
dataclouder: Please add a link to the resource

website/docs/r/ipset.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+
+# vcd\_ipset
+
+Provides a vCloud Director IP set resource. An IP set is a group of IP addresses that you can add as
+  the source or destination in a firewall rule or in DHCP relay configuration.
+
+
+Supported in provider *v2.6+*
+
+## Example Usage 1
```
dataclouder: Is there an example 2 coming?
I'd like to see several examples of different combinations allowed in `ip_addresses` and at least one example of how to use this in other resources, even if they don't exist yet.

website/docs/r/ipset.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+* `name` - (Required) Unique IP set name.
+* `description` - (Optional) An optional description for IP set.
+* `ip_addresses` - (Required) A set of IP addresses, CIDRs and ranges as strings.
+* `is_inheritance_allowed` (Optional) Toggle to enable inheritance to allow visibility at underlying scopes. (Default `true`)
```
lvirbalas: For consistency:
```suggestion
* `is_inheritance_allowed` (Optional) Toggle to enable inheritance to allow visibility at underlying scopes. Default `true`
```

website/docs/r/ipset.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+The following attributes are exported on this resource:
+
+* `id` - ID of IP set
+
+## Importing
+
+~> **Note:** The current implementation of Terraform import can only import resources into the state.
+It does not generate configuration. [More information.](https://www.terraform.io/docs/import/)
+
+An existing load balancer application rule can be [imported][docs-import] into this resource
```
lvirbalas: Load balancer? :)

website/docs/d/ipset.html.markdown
Didainius: Added

website/docs/r/ipset.html.markdown
Didainius: it will come either in IP set or in vcd_nsxv_firewall_rule when there is a PR to start supporting it.

website/docs/r/ipset.html.markdown
Didainius: Fixed.

website/docs/r/ipset.html.markdown
Didainius: Fixed.

Didainius:
> LGTM. Though I would like SDK upgrade as separate PR

Message noted for future. Although now all the tests were done in this version so I don't want to roll it back and redo all the tests as we willl will bump it anyway.

Issue #294, opened 2019-07-29
Dnat rule doesn't support ext network with org user


vbauzysvmware:
Current dnat resource has limitation using external network. User has to be sys admin, as it uses getExternalNetwork function which requires sys admin rights. According advanced gtw UI, seems there is possibility to acquire ext networks with org admin User.

pasikarkkainen:
Also the legacy Standard Edges did support configuring DNAT/SNAT for the External Network, using normal tenant admin account. I've used that a lot. Also with terraform vcd provider :)

pasikarkkainen:
Simple example which configures DNAT rule which is applied on the External Network of the EdgeGw:

resource "vcd_dnat" "jumpbox-ssh" {
    edge_gateway  = "${var.edge_gateway}"
    external_ip   = "${var.edge_ext_ip}"
    port          = 22
    internal_ip   = "${var.jumpbox_int_ip}"
}

This works OK with terraform vcd provider (at least up to v2.3.0, which I'm using atm). It works with both Standard and Advanced Edges. It requires the external IP is sub-allocated properly by the admin though, which allows the tenant to use it for NAT rules on the EdgeGw, otherwise it cannot be used.

PR #408, opened 2019-11-29
Enable org users to see direct network data

dataclouder:
Until now, direct network data was only visible to system administrators, because we were fetching the external network data directly, and such operation requires system administrator privileges.
We are now using a call to `GetNetworkList`, which is available to Org users, and they are now able to read the data.

Note that Org users are still unable to create a direct network, as such operation is still restricted to system users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ BUG FIXES:
   [GH-379]
 * Fix `vcd_independent_disk` reapply issue, which was seen when optional `bus_sub_type` and `bus_type` wasn't used - [GH-394]
 * Fix `vcd_vapp_network` apply issue, where the property `guest_vlan_allowed` was applied only to the last of multiple networks.
+* Fix `vcd_network_direct` data reading from Org Users. Such users can now read the data from existing direct networks.
 
 DEPRECATIONS:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,6 @@ IMPROVEMENTS:
 * `resource/vcd_network_direct` Add property `description`
 * `resource/vcd_network_routed` Add check for valid IPs [GH-374]
 * `resource/vcd_network_isolated` Add check for valid IPs [GH-373]
-* `resource/vcd_network_direct` Org User can now read this resource (previously it was only Sys Admin) - this change makes it possible to get the details of External Network as Org User
 
 BUG FIXES:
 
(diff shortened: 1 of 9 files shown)


dataclouder:
> LGMT. changelog update missing. And no test changes or new tests needed?
Change log updated.
The test is already there. We need to run it as org user
```
VCD_TEST_ORG_USER=1 go test -tags functional -v -timeout 0 -run TestAccVcdNetworkDirectDS .
```

dataclouder:
> BTW, please try whether we have the same issue still in the new NSX-V DNAT rule and, if yes, whether it could be fixed in the same way? We had such in the old DNAT rule:
> #294

The DNAT rule is a different issue. In this PR, we can get the relationship between a network and an external network through the list of networks.  For DNAT rules, the same workaround doesn't apply.

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ BUG FIXES:
   [GH-379]
 * Fix `vcd_independent_disk` reapply issue, which was seen when optional `bus_sub_type` and `bus_type` wasn't used - [GH-394]
 * Fix `vcd_vapp_network` apply issue, where the property `guest_vlan_allowed` was applied only to the last of multiple networks.
+* Fix `vcd_network_direct` data reading from Org Users. Such users can now read the data from existing direct networks.
```
lvirbalas: I suggest rephrasing the message to make it clearer and move it to improvements:

```suggestion
* `resource/vcd_network_direct` Org User can now use this resource (previously it was only Sys Admin) - this change makes it possible to get the details of External Network as Org User
```

vcd/resource_vcd_network_direct.go
```diff
@@ -95,6 +95,9 @@ func resourceVcdNetworkDirect() *schema.Resource {
 func resourceVcdNetworkDirectCreate(d *schema.ResourceData, meta interface{}) error {
 	vcdClient := meta.(*VCDClient)
 
+	if !vcdClient.Client.IsSysAdmin {
+		return fmt.Errorf("creation of a vcd_network_routed requires system administrator privileges")
```
Didainius:
```suggestion
		return fmt.Errorf("creation of a vcd_network_direct requires system administrator privileges")
```

CHANGELOG.md
dataclouder: changed

vcd/resource_vcd_network_direct.go
dataclouder: Thanks. Fixed

CHANGELOG.md
dataclouder: About the documentation page, for the resource it is still valid. Org users can't create a direct network. But they can read it.
We have the warning banner in the resource, but not in the data source

CHANGELOG.md
```diff
@@ -38,7 +38,7 @@ IMPROVEMENTS:
 * `resource/vcd_network_direct` Add property `description`
 * `resource/vcd_network_routed` Add check for valid IPs [GH-374]
 * `resource/vcd_network_isolated` Add check for valid IPs [GH-373]
-
+* `resource/vcd_network_direct` Org User can now read this resource (previously it was only Sys Admin) - this change makes it possible to get the details of External Network as Org User
```
lvirbalas:
```suggestion
* `datasource/vcd_network_direct` and `resource/vcd_network_direct` are now readable by Org User (previously it was only by Sys Admin), as this change made it possible to get the details of External Network as Org User [GH-408]
```

PR #414, opened 2019-12-05
Do not send fips_mode_enabled when it is not set (needed for vCD 9.0)

Didainius:
vCD 9.0 edge gateway does not support FIPS mode and API throws error if the XML field `FipsModeEnabled` is sent. This modifies edge gateway creation so that if the field `fips_mode_enabled` is not set - it is not sent.

The problem was identified here - https://github.com/terraform-providers/terraform-provider-vcd/issues/323#issuecomment-561575052

### DIFF
--- a/vcd/resource_vcd_edgegateway.go
+++ b/vcd/resource_vcd_edgegateway.go
@@ -244,11 +244,12 @@ func resourceVcdEdgeGateway() *schema.Resource {
 				ValidateFunc: validation.StringInSlice([]string{"accept", "deny"}, false),
 			},
 			"fips_mode_enabled": &schema.Schema{
-				Type:        schema.TypeBool,
-				ForceNew:    true,
-				Optional:    true,
-				Default:     false,
-				Description: "Enable FIPS mode. FIPS mode turns on the cipher suites that comply with FIPS. (False by default)",
+				Type:     schema.TypeBool,
+				ForceNew: true,
+				Optional: true,
+				// Not setting default here because vCD 9.0 does not support FIPS mode and there
+				// must be a clear indicator to distinguish when the field was not set by user
+				Description: "Enable FIPS mode. Only for vCD 9.1+. FIPS mode turns on the cipher suites that comply with FIPS. (False by default)",
 			},
 			"use_default_route_for_dns_relay": &schema.Schema{
 				Type:        schema.TypeBool,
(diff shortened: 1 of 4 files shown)


Didainius:
> Looks good overall, only a few asks:
> 
>     1. If we have such, please reference the issue number or comment, where this was found in, in the description of the PR.
Added.
> 
>     2. Would it still be possible to share a test output which failed on 9.0 before the fix and succeeded after the fix?
Actually the same reference I added has the example output as well - https://github.com/terraform-providers/terraform-provider-vcd/issues/323#issuecomment-561575052

Didainius:
> LGTM, only asking to confirm that `resource_vcd_edgegateway_test.go` (or other relevant test) is succeeding after this fix on your vCD 9.0 before merge too.

Sure- here is the output without the fix:
```shell
--- FAIL: TestAccVcdEdgeGatewayComplex (22.21s)
    testing.go:635: Step 0 error: errors during apply:

        Error: error creating edge gateway: error instantiating a new Edge Gateway: API Error: 400: [ 10acdb60-04fc-4287-a0bc-65d012ca63ae ] HTTP 400 Bad Request
         - cvc-complex-type.2.4.d: Invalid content was found starting with element 'FipsModeEnabled'. No child element is expected at this point.

          on /var/folders/h2/c60z324s6219c3hc_w28hbx80000gp/T/tf-test771078234/main.tf line 41:
          (source code not available)

```

Here is the output with fix:
```
=== RUN   TestAccVcdEdgeGatewayComplex
--- PASS: TestAccVcdEdgeGatewayComplex (108.14s)
```

However some other test `TestAccVcdEdgeGatewayExternalNetworks`has problems on vCD 9.0 which are not related to FIPS mode. Will link to a separate PR when it is raised...

vcd/resource_vcd_edgegateway.go
```diff
@@ -349,6 +349,18 @@ func resourceVcdEdgeGatewayCreate(d *schema.ResourceData, meta interface{}) erro
 		},
 	}
 
+	// vCD 9.0 does not support FIPS Mode and fails if XML tag <FipsModeEnabled> is sent therefore
+	// field value must be sent only if user specified its value
+	if fipsModeEnabled, ok := d.GetOkExists("fips_mode_enabled"); ok {
+		if vcdClient.APIVCDMaxVersionIs("<= 29.0") { // vCD 9.0 or less
+			_, _ = fmt.Fprint(getTerraformStdout(), "WARNING! FIPS mode is only supported starting"+
```
vbauzysvmware: what happens here if I set? - API error? maybe is worth it here send error and fail early?

Didainius: I had a doubt about it. Here is how it looks - https://github.com/terraform-providers/terraform-provider-vcd/issues/323#issuecomment-562136931


Probably yes - I could fail early.

vbauzysvmware: If you talking about `Invalid content was found starting with element 'FipsModeEnabled'. No child element is expected at this point.` Then my answer is that 100% we have fail with good explanation. @lvirbalas and @dataclouder can include they opinions too if it is needed

Didainius: Changed

PR #416, opened 2019-12-06
Add Edge Gateway DHCP relay configuration support

Didainius:
This PR adds support for NSX-V Edge Gateway DHCP relay configuration which represents relay DHCP relay configuration tab in edge gateway.

Note. This resource is a bit "special" because it actually does not create any objects and there is no real ID for it therefore there is a fake ID created which includes Edge Gateway ID and additional hardcoded string part.


website/docs/r/nsxv_dhcp_relay.markdown
```diff
@@ -0,0 +1,97 @@
...
+---
+
+# vcd\_nsxv\_dhcp\_relay
+
+Provides a vCloud Director Edge Gateway DHCP relay configuration resource. The DHCP relay capability
+provided by NSX in vCloud Director environment allows to leverage existing DHCP infrastructure from
+within vCloud Director environment without any interruption to the IP address management in existing
+DHCP infrastructure. DHCP messages are relayed from virtual machines to the designated DHCP servers
+in your physical DHCP infrastructure, which allows IP addresses controlled by the NSX software to
+continue to be in synch with IP addresses in the rest of your DHCP-controlled environments.
```
lvirbalas:
```suggestion
continue to be in sync with IP addresses in the rest of your DHCP-controlled environments. 
```

website/docs/r/nsxv_dhcp_relay.markdown
```diff
@@ -0,0 +1,97 @@
...
+
+```hcl
+resource "vcd_nsxv_dhcp_relay" "relay_config" {
+  org          = "my-org"
+  vdc          = "my-org-vdc"
+  edge_gateway = "my-edge-gw"
+
+  ip_addresses = ["1.1.1.1", "2.2.2.2"]
+  domain_names = ["servergroups.domainname.com", "other.domain.com"]
+  ip_sets      = [vcd_ipset.myset1.name, vcd_ipset.myset2.name]
```
lvirbalas: Thinking that it may be clearer for new users if we defined these IP set resources in this example as well.

website/docs/r/nsxv_dhcp_relay.markdown
```diff
@@ -0,0 +1,97 @@
...
+It does not generate configuration. [More information.](https://www.terraform.io/docs/import/)
+
+An existing DHCP relay configuration can be [imported][docs-import] into this resource
+via supplying the full dot separated path for your edge gateway. An example is
+below:
+
+[docs-import]: https://www.terraform.io/docs/import/
+
+```
+terraform import vcd_nsxv_dhcp_relay.imported my-org.my-org-vdc.my-edge-gw
```
lvirbalas: This raises thoughts. So, we have 1:1 mapping between DHCP relay resource and an EG.
- Would be useful to note in docs that only a single DHCP relay can be set for a single EG.
- What would happen if user defined two relay blocks for a single EG?

website/docs/d/nsxv_dhcp_relay.markdown
```diff
@@ -0,0 +1,44 @@
...
+---
+
+# vcd\_nsxv\_dhcp\_relay
+
+Provides a vCloud Director Edge Gateway DHCP relay configuration data source. The DHCP relay
+capability provided by NSX in vCloud Director environment allows to leverage existing DHCP
+infrastructure from within vCloud Director environment without any interruption to the IP address
+management in existing DHCP infrastructure. DHCP messages are relayed from virtual machines to the
+designated DHCP servers in your physical DHCP infrastructure, which allows IP addresses controlled
+by the NSX software to continue to be in synch with IP addresses in the rest of your DHCP-controlled
```
lvirbalas:
```suggestion
by the NSX software to continue to be in sync with IP addresses in the rest of your DHCP-controlled
```

vcd/resource_vcd_nsxv_dhcp_relay_test.go
```diff
@@ -0,0 +1,206 @@
...
+}
+`
+
+const testAccVcdNsxvDhcpRelayUpdate = testAccRoutedNet + `
+resource "vcd_nsxv_dhcp_relay" "relay_config" {
+	org          = "{{.Org}}"
+	vdc          = "{{.Vdc}}"
+	edge_gateway = "{{.EdgeGateway}}"
+	
+    ip_sets      = [vcd_ipset.myset1.name, vcd_ipset.myset2.name]
```
lvirbalas: Maybe it's just how GitHub is showing these testing const strings, but indent is off in some places - maybe go fmt would help?

vcd/datasource_vcd_nsxv_dhcp_relay.go
```diff
@@ -0,0 +1,74 @@
...
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"relay_agent": {
+				Computed: true,
+				Type:     schema.TypeSet,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"org_network": {
```
lvirbalas: Two comments:

1.) It seems, in H5 UI in addition to Org networks you can select external network:

Screen Shot 2019-12-06 at 16 30 20

Are you sure it's only org networks?

2.) I'm also thinking whether we shouldn't rename `org_network` -> `org_network_name` for clarity. It would be similar to `network_name` in `vcd_nsxv_snat` resource.

3.) And continuing on the train of thought of (1) and (2), if it turns out that we can select external network, then we could just name the field `network_name` as in NAT rules.

vcd/resource_vcd_nsxv_dhcp_relay.go
```diff
@@ -0,0 +1,434 @@
...
+	relayAgentSet := schema.NewSet(schema.HashResource(relayAgentResource), relayAgentSlice)
+	err := d.Set("relay_agent", relayAgentSet)
+	if err != nil {
+		return fmt.Errorf("could not save relay_agent to schema: %s", err)
+	}
+
+	return nil
+}
+
+// getDhclRelaySettingsId constructs a fake DHCP relay configuration ID which is needed for
```
lvirbalas: Typo in comment function name "Dhcl".

website/docs/r/nsxv_dhcp_relay.markdown
Didainius: > This raises thoughts. So, we have 1:1 mapping between DHCP relay resource and an EG.
> 
>     * Would be useful to note in docs that only a single DHCP relay can be set for a single EG.
I had thoughts to note it is a "singleton". Probably worth it if as now we both have this idea.
> 
>     * What would happen if user defined two relay blocks for a single EG?
They would "fight" each other. Latest would win. Next apply would propose to "update"...

lvirbalas: > I had thoughts to note it is a "singleton". Probably worth it if as now we both have this idea.
Sounds good, of course you'll need to define scope to be on EG name.

vcd/datasource_vcd_nsxv_dhcp_relay.go
Didainius: > Two comments:
> 
> 1.) It seems, in H5 UI in addition to Org networks you can select external network:
> 
> Screen Shot 2019-12-06 at 16 30 20
> 
> Are you sure it's only org networks?
Yes. It does allow to choose, but here is what happens:
image

> 
> 2.) I'm also thinking whether we shouldn't rename `org_network` -> `org_network_name` for clarity. It would be similar to `network_name` in `vcd_nsxv_snat` resource.
> 
Can do. I kind of thought between this and that. In some places we don't ask for "name", in some we do. (like `edge_gateway` also asks for name, but we don't mention it).

> 3.) And continuing on the train of thought of (1) and (2), if it turns out that we can select external network, then we could just name the field `network_name` as in NAT rules.

I can make it broader - maybe in future it works although for my limited understanding this wouldn't make sense. The general point of relay is to "forward" DHCP request messages to defined servers outside of NSX. One shouldn't be able to forward DHCP request messages from external network. I might be wrong though.

website/docs/r/nsxv_dhcp_relay.markdown
Didainius: Fixed

vcd/resource_vcd_nsxv_dhcp_relay_test.go
Didainius: I tried to apply `terraform fmt` on each of these examples. Should be better now.

vcd/resource_vcd_nsxv_dhcp_relay.go
Didainius: Changed

website/docs/r/nsxv_dhcp_relay.markdown
Didainius: Added a note

website/docs/d/nsxv_dhcp_relay.markdown
Didainius: fixed

website/docs/r/nsxv_dhcp_relay.markdown
Didainius: Added

vcd/resource_vcd_nsxv_dhcp_relay.go
```diff
@@ -0,0 +1,434 @@
...
+
+	edgeGateway, err := vcdClient.GetEdgeGatewayFromResource(d, "edge_gateway")
+	if err != nil {
+		return fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	}
+
+	_, vdc, err := vcdClient.GetOrgAndVdcFromResource(d)
+	if err != nil {
+		return fmt.Errorf(errorRetrievingOrgAndVdc, err)
+	}
```
vbauzysvmware: this should be part of `getDhcpRelayType`

vcd/resource_vcd_nsxv_dhcp_relay.go
```diff
@@ -0,0 +1,434 @@
...
+
+	_, err = edgeGateway.UpdateDhcpRelay(dhcpRelayConfig)
+	if err != nil {
+		return fmt.Errorf("unable to update DHCP relay settings for Edge Gateway %s: %s", edgeGateway.EdgeGateway.Name, err)
+	}
+
+	// This is not a real object but a settings property on Edge gateway - creating a fake composite
+	// ID
+	fakeId, err := getDhcpRelaySettingsId(edgeGateway)
```
vbauzysvmware: I don't like using `fake` as it says something bad. `generatedId`, `constructedId` or something similar would be suitable

vcd/resource_vcd_nsxv_dhcp_relay.go
```diff
@@ -0,0 +1,434 @@
...
+	}
+
+	err = setDhcpRelayAgentData(d, dhcpRelaySettings, edgeGateway, vdc)
+	if err != nil {
+		return fmt.Errorf("could not set DHCP relay agent settings: %s", err)
+	}
+
+	// This is not a real object but a settings property on Edge gateway - creating a fake composite
+	// ID
+	fakeId, err := getDhcpRelaySettingsId(edgeGateway)
```
vbauzysvmware: same comment as above

vcd/resource_vcd_nsxv_dhcp_relay.go
```diff
@@ -0,0 +1,434 @@
...
+	return nil
+}
+
+// resourceVcdNsxvDhcpRelayImport imports DHCP relay configuration. Because DHCP relay is just a
+// settings on edge gateway and not a separate object - the ID actually does not represent any
+// object
+func resourceVcdNsxvDhcpRelayImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	resourceURI := strings.Split(d.Id(), ImportSeparator)
+	if len(resourceURI) != 3 {
+		return nil, fmt.Errorf("resource name must be specified in such way org.vdc.edge-gw")
```
vbauzysvmware:
```suggestion
		return nil, fmt.Errorf("resource name must be specified in such way org.vdc.edge-gw-name")
```

vcd/resource_vcd_nsxv_dhcp_relay.go
```diff
@@ -0,0 +1,434 @@
...
+	}
+	orgName, vdcName, edgeName := resourceURI[0], resourceURI[1], resourceURI[2]
+
+	vcdClient := meta.(*VCDClient)
+	edgeGateway, err := vcdClient.GetEdgeGateway(orgName, vdcName, edgeName)
+	if err != nil {
+		return nil, fmt.Errorf(errorUnableToFindEdgeGateway, err)
+	}
+
+	fakeId, err := getDhcpRelaySettingsId(edgeGateway)
```
vbauzysvmware: same comment regarding fakeId

vcd/resource_vcd_nsxv_dhcp_relay.go
```diff
@@ -0,0 +1,434 @@
...
+
+	return dhcpRelayConfig, nil
+
+}
+
+// getDhcpRelayAgentsType converts relay_agent configuration blocks to []types.EdgeDhcpRelayAgent
+func getDhcpRelayAgentsType(relayAgentsSet *schema.Set, edge *govcd.EdgeGateway) ([]types.EdgeDhcpRelayAgent, error) {
+	relayAgentsSlice := relayAgentsSet.List()
+	relayAgentsStruct := make([]types.EdgeDhcpRelayAgent, len(relayAgentsSlice))
+	if len(relayAgentsSlice) > 0 {
```
vbauzysvmware: not needed condition

Didainius: Yes. removed.

vcd/resource_vcd_nsxv_dhcp_relay.go
Didainius: changed to compositeId

vcd/resource_vcd_nsxv_dhcp_relay.go
Didainius: changed

vcd/resource_vcd_nsxv_dhcp_relay.go
Didainius: compositeId

vcd/resource_vcd_nsxv_dhcp_relay.go
Didainius: compositeId

vcd/resource_vcd_nsxv_dhcp_relay.go
Didainius: But I guess this saves nothing. I would have to pass `vcdClient` instead of `vdc` then.

Didainius: Let it be. Changed.

vcd/datasource_vcd_nsxv_dhcp_relay.go
lvirbalas: OK, so we have three choices:
- `org_network`
- `org_network_name`
- `network_name`

@Didainius @dataclouder @vbauzysvmware your votes?

To me `network_name` looks most consistent with the same field from NSX-V NAT rules. However, it doesn't reflect the point that these are only Org networks.

dataclouder: `network_name` is more generic and will work even if we end up supporting different kinds of network

vbauzysvmware: I am fine with `network_name`

PR #422, opened 2019-12-16
Fix documentation formatting

dataclouder:
Fix documentation issues for 2.6 release

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -12,9 +12,9 @@ The VMware vCloud Director provider is used to interact with the resources suppo
 
 Use the navigation to the left to read about the available resources.
 
-~> **NOTE:** The VMware vCloud Director Provider went through a refresh at the beginning of 2019 and some semantic
-changes were made compared to the previously available initial version. Please check docs for *v2.x+* labels and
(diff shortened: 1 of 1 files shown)


website/docs/index.html.markdown
```diff
@@ -12,7 +12,9 @@ The VMware vCloud Director provider is used to interact with the resources suppo
 
 Use the navigation to the left to read about the available resources.
 
-~> **NOTE:** The VMware vCloud Director Provider went through a refresh at the beginning of 2019 and some semantic changes were made compared to the previously available initial version. Please check docs for *v2.0+*, *v2.1+*, *v2.2+*, *v2.4+* labels and your existing .tf configuration files carefully when shifting to this new version. 
+~> **NOTE:** The VMware vCloud Director Provider went through a refresh at the beginning of 2019 and some semantic
```
lvirbalas: What about updating entirely to something like this:

> ~> **NOTE:** The VMware vCloud Director Provider documentation pages include *v2.x+* labels in resource and/or field descriptions. These labels are designed to show at which provider version a certain feature was introduced. When upgrading the provider please check for such labels for the resources you are using.

dataclouder: updated

Issue #385, opened 2019-11-05
vCloud Director Provider - No organization policy lease time options


sysadmintutorials:
Hi,

I'm writing to ask for a feature request which is to add org policy lease options within the vcloud provider. 

### Terraform Version
Terraform v0.12.13
+ provider.vcd v2.5.0

### Affected Resource(s)
Resource: vcd_org

### Expected Behavior
Creating a new vcd_org, we should have an argument that allows us to specify the following:
- Maximum runtime lease
- Maximum storage lease
- Storage Cleanup

### Actual Behavior
Creating a new org sets the leases to default:
- Maximum runtime lease: 7 days
- Maximum storage lease: 30 days
- Storage cleanup: move to expired items

### Steps to Reproduce
1. terraform plan or apply does not give any options to specific lease times in Org creation or updates:

Terraform will perform the following actions:

  vcd_org.org-name will be updated in-place
  ~ resource "vcd_org" "org-name" {
        can_publish_catalogs         = false
        delay_after_power_on_seconds = 0
        delete_force                 = true
        delete_recursive             = true
        deployed_vm_quota            = 0
      ~ description                  = "Terraform Create Org" -> "Terraform1 Create Org"
        full_name                    = "My Terraform Organization"
        id                           = "urn:vcloud:org:ff1974f7-d893-4fd6-b129-531432ded60b"
        is_enabled                   = true
        name                         = "Terraform1"
        stored_vm_quota              = 0
    }

vbauzysvmware:
Thank you for feature proposal.

nilic:
I hate to leave a basically +1 post, but this is a very important feature for our use case. As a SP, we use Terraform for provisioning new tenants. Current provider covers 95% of our needs, but this is a setting which still requires manual workarounds for each and every tenant, since the default Org lease values are useless for everything except trial tenants (and even for them it would be useful to be able to define lease times, because the default runtime value of 7 days is too short).

dataclouder:
The feature is under review in PR #432
Please have a look, try it, and comment on the implementation.

dataclouder:
Implemented in PR #432, now merged to master

PR #432, opened 2020-01-14
Add section vapp_lease and vapp_template_lease

dataclouder:
* Enhance Org resource and datasource with vapp_lease section
  (parameters related to vApp lease management)
* Enhance Org resource and datasource with vapp_template_lease section
  (parameters related to vApp template lease management)
* Update documentation

Defined in issue #385

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ FEATURES:
 
 IMPROVEMENTS:
 
-* `resource/vcd_org` and `datasource/vcd_org` include a section `vapp_lease` and a section `vapp_template_lease` to define lease related parameters of depending entities - [GH-385]
+* `resource/vcd_org` and `datasource/vcd_org` include a section `vapp_lease` and a section `vapp_template_lease` to define lease related parameters of depending entities - [GH-432]
 
 NOTES:
 * Drop support for vCD 9.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ FEATURES:
 * **New Resource:** `vcd_vm_internal_disk` VM internal disk configuration
 * `vcd_vapp_vm` Internal disks in VM template allows to be edited by `override_template_disk` field 
 
+IMPROVEMENTS:
+
+* `resource/vcd_org` and `datasource/vcd_org` include a section `vapp_lease` and a section `vapp_template_lease` to define lease related parameters of depending entities - [GH-385]
+
 NOTES:
 * Drop support for vCD 9.0
 
--- a/vcd/datasource_vcd_org.go
+++ b/vcd/datasource_vcd_org.go
@@ -58,7 +58,7 @@ func datasourceVcdOrg() *schema.Resource {
 							Type:     schema.TypeBool,
 							Computed: true,
 							Description: "When true, vApps are powered off when the runtime lease expires. " +
-								"When false or missing, vApps are suspended when the runtime lease expires",
+								"When false, vApps are suspended when the runtime lease expires",
 						},
 						"maximum_storage_lease_in_sec": &schema.Schema{
 							Type:        schema.TypeInt,
--- a/vcd/datasource_vcd_org.go
+++ b/vcd/datasource_vcd_org.go
@@ -44,6 +44,55 @@ func datasourceVcdOrg() *schema.Resource {
 				Computed:    true,
 				Description: "True if this organization is allowed to share catalogs.",
 			},
+			"vapp_lease": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"maximum_runtime_lease_in_sec": &schema.Schema{
+							Type:        schema.TypeInt,
+							Computed:    true,
+							Description: "How long vApps can run before they are automatically stopped (in seconds)",
+						},
+						"power_off_on_runtime_lease_expiration": &schema.Schema{
+							Type:     schema.TypeBool,
+							Computed: true,
+							Description: "When true, vApps are powered off when the runtime lease expires. " +
+								"When false or missing, vApps are suspended when the runtime lease expires",
+						},
+						"maximum_storage_lease_in_sec": &schema.Schema{
+							Type:        schema.TypeInt,
+							Computed:    true,
+							Description: "How long stopped vApps are available before being automatically cleaned up (in seconds)",
+						},
+						"delete_on_storage_lease_expiration": &schema.Schema{
+							Type:     schema.TypeBool,
+							Computed: true,
+							Description: "If true, storage for a vApp is deleted when the vApp's lease expires. " +
+								"If false, the storage is flagged for deletion, but not deleted.",
+						},
+					},
+				},
+			},
+			"vapp_template_lease": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"maximum_storage_lease_in_sec": &schema.Schema{
+							Type:        schema.TypeInt,
+							Computed:    true,
+							Description: "How long vApp templates are available before being automatically cleaned up (in seconds)",
+						},
+						"delete_on_storage_lease_expiration": &schema.Schema{
+							Type:     schema.TypeBool,
+							Computed: true,
+							Description: "If true, storage for a vAppTemplate is deleted when the vAppTemplate lease expires. " +
+								"If false, the storage is flagged for deletion, but not deleted",
+						},
+					},
+				},
+			},
 			"delay_after_power_on_seconds": &schema.Schema{
 				Type:        schema.TypeInt,
 				Computed:    true,
--- a/vcd/resource_vcd_org.go
+++ b/vcd/resource_vcd_org.go
@@ -83,11 +83,10 @@ func resourceOrg() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"maximum_runtime_lease_in_sec": &schema.Schema{
-							Type:     schema.TypeInt,
-							Optional: true,
-							// If not set, the default is 604800 (7 days)
+							Type:         schema.TypeInt,
+							Optional:     true,
 							Description:  "How long vApps can run before they are automatically stopped (in seconds). 0 means never expires",
-							ValidateFunc: IntLeaseSeconds(), // Lease can be either 0 or 3600+
+							ValidateFunc: validateIntLeaseSeconds(), // Lease can be either 0 or 3600+
 						},
 						"power_off_on_runtime_lease_expiration": &schema.Schema{
 							Type:     schema.TypeBool,
@@ -97,11 +96,10 @@ func resourceOrg() *schema.Resource {
 								"When false or missing, vApps are suspended when the runtime lease expires",
 						},
 						"maximum_storage_lease_in_sec": &schema.Schema{
-							Type:     schema.TypeInt,
-							Optional: true,
-							// If not set, the default is 1209600 (14 days)
+							Type:         schema.TypeInt,
+							Optional:     true,
 							Description:  "How long stopped vApps are available before being automatically cleaned up (in seconds). 0 means never expires",
-							ValidateFunc: IntLeaseSeconds(), // Lease can be either 0 or 3600+
+							ValidateFunc: validateIntLeaseSeconds(), // Lease can be either 0 or 3600+
 						},
 						"delete_on_storage_lease_expiration": &schema.Schema{
 							Type:     schema.TypeBool,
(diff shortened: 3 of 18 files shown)


website/docs/r/org.html.markdown
```diff
@@ -47,6 +58,26 @@ The following arguments are supported:
...
 * `delay_after_power_on_seconds` - (Optional) - Specifies this organization's default for virtual machine boot delay after power on. Default is `0`.
+* `vapp_lease` - (Optional; *v2.7*) - Defines lease parameters for vApps created in this organization. See [vApp Lease](#vapp-lease) below for details. 
+* `vapp_template_lease` - (Optional; *v2.7*) - Defines lease parameters for vApp templates created in this organization. See [vApp Template Lease](#vapp-template-lease) below for details.
+
+<a id="vapp-lease"></a>
+## vApp Lease
+
+The `vapp_lease` section contains lease parameters for vApps created in the current organization, as defined below:
+
+* `maximum_runtime_lease_in_sec` - (Optional) - How long vApps can run before they are automatically stopped (in seconds). Minimum 3600. - Default is 604800 (7 days)
```
nilic: The value for 'Never Expires' should be documented

dataclouder: Thanks for looking at it. Good call.
I need to add a call for this item, as the current API can't set the `never expires` value.

dataclouder: Support for 'Never Expired' added

vcd/resource_vcd_org.go
```diff
@@ -74,6 +74,70 @@ func resourceOrg() *schema.Resource {
...
+				Optional:    true,
+				Computed:    true,
+				MaxItems:    1,
+				Description: "Defines lease parameters for vApps created in this organization",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"maximum_runtime_lease_in_sec": &schema.Schema{
+							Type:     schema.TypeInt,
+							Optional: true,
+							//Default:      604800, // (7 days)
```
Didainius: Just checking - are these commented out values are what vCD sets if nothing is set?
If so - I could say this could be phrased differently as it gives doubt whether that was commented out code block ar a comment which ment some system defaults

vcd/validate_funcs.go
```diff
@@ -112,3 +112,21 @@ func validateMultipleOf4() schema.SchemaValidateFunc {
 		return
 	}
 }
+
+func IntLeaseSeconds() schema.SchemaValidateFunc {
```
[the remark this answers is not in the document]
dataclouder: added

vcd/resource_vcd_org.go
dataclouder: Changed comments

website/docs/r/org.html.markdown
```diff
@@ -47,6 +58,26 @@ The following arguments are supported:
...
+
+<a id="vapp-lease"></a>
+## vApp Lease
+
+The `vapp_lease` section contains lease parameters for vApps created in the current organization, as defined below:
+
+* `maximum_runtime_lease_in_sec` - (Optional) - How long vApps can run before they are automatically stopped (in seconds). 0 means never expires. Values accepted from 3600+. - Default is 604800 (7 days)
+* `power_off_on_runtime_lease_expiration` - Optional - When true, vApps are powered off when the runtime lease expires. When false or missing, vApps are suspended when the runtime lease expires.
+* `maximum_storage_lease_in_sec` - (Optional) - How long stopped vApps are available before being automatically cleaned up (in seconds). 0 means never expires. Regular values accepted from 3600+. Default is 1209600 (14 days)
+* `delete_on_storage_lease_expiration` - (Optional) - If true, storage for a vApp is deleted when the vApp's lease expires. If false, the storage is flagged for deletion, but not deleted.
```
nilic: Documentation should not that false is the default. Same for `vapp_template_lease` section

vcd/validate_funcs.go
```diff
@@ -112,3 +112,24 @@ func validateMultipleOf4() schema.SchemaValidateFunc {
 		return
 	}
 }
+
+// IntLeaseSeconds validates amount of seconds for lease
+// A value of 0 is accepted, as it means "never expires"
+// Regular values must be > 3600
+func IntLeaseSeconds() schema.SchemaValidateFunc {
```
vbauzysvmware: currently our validation functions starts with checkSomething, validateSomething. Maybe there is better name for this function.

vcd/resource_vcd_org.go
```diff
@@ -134,19 +198,74 @@ func resourceOrgCreate(d *schema.ResourceData, m interface{}) error {
...
 	settings := &types.OrgSettings{}
-	General := &types.OrgGeneralSettings{}
-
-	General.DeployedVMQuota = d.Get("deployed_vm_quota").(int)
-	General.StoredVMQuota = d.Get("stored_vm_quota").(int)
+	vappLeaseSettings := &types.VAppLeaseSettings{}
+	vappTemplateLeaseSettings := &types.VAppTemplateLeaseSettings{}
 
-	delay, ok := d.GetOk("delay_after_power_on_seconds")
+	vappInputProvided := false
```
vbauzysvmware:
```suggestion
	vappLeaseInputProvided := false
```

vcd/resource_vcd_org.go
```diff
@@ -134,19 +198,74 @@ func resourceOrgCreate(d *schema.ResourceData, m interface{}) error {
...
+		}
+		deleteOnLeaseExpiration, isSet := itemMap["delete_on_storage_lease_expiration"]
+		if isSet {
+			tmpBool := deleteOnLeaseExpiration.(bool)
+			vappLeaseSettings.DeleteOnStorageLeaseExpiration = &tmpBool
+		}
+	}
+	item, ok = d.GetOk("vapp_template_lease")
+	if ok {
+		vappTemplateInputProvided = true
```
vbauzysvmware:
```suggestion
		vappTemplateLeaseInputProvided = true
```

vcd/validate_funcs.go
```diff
@@ -112,3 +112,24 @@ func validateMultipleOf4() schema.SchemaValidateFunc {
...
+	return func(i interface{}, k string) (s []string, es []error) {
+		v, ok := i.(int)
+		if !ok {
+			es = append(es, fmt.Errorf("expected type of %s to be int", k))
+			return
+		}
+
+		valid := i == 0 || v >= 3600
```
vbauzysvmware: I am not use `valid` here is needed. I think it's ok to check directly in condition

vcd/validate_funcs.go
dataclouder: Thanks. Renaming. I used the current name modeling on `validation.IntAtLeast` and similar. But consistency within our library is important.

vcd/validate_funcs.go
dataclouder: The current condition check seems to me more readable than

```
if !(i == 0 || v >= 3600) {
```

vcd/resource_vcd_org.go
```diff
@@ -74,6 +74,67 @@ func resourceOrg() *schema.Resource {
...
+						"maximum_storage_lease_in_sec": &schema.Schema{
+							Type:         schema.TypeInt,
+							Optional:     true,
+							Description:  "How long stopped vApps are available before being automatically cleaned up (in seconds). 0 means never expires",
+							ValidateFunc: validateIntLeaseSeconds(), // Lease can be either 0 or 3600+
+						},
+						"delete_on_storage_lease_expiration": &schema.Schema{
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
```
lvirbalas: Is it on purpose that `power_off_on_runtime_lease_expiration` and `delete_on_storage_lease_expiration` are optional and have defaults, but `maximum_runtime_lease_in_sec` and `maximum_storage_lease_in_sec` don't have defaults, but are also optional?

What would be the meaning of such definition (which seem to be allowed):

```
vapp_lease {
power_off_on_runtime_lease_expiration = true
delete_on_storage_lease_expiration = true
}
```

Or such:


```
vapp_lease {
}
```

website/docs/r/org.html.markdown
```diff
@@ -30,6 +30,17 @@ resource "vcd_org" "my-org" {
   is_enabled       = "true"
   delete_recursive = "true"
   delete_force     = "true"
+
+  vapp_lease {
+    maximum_runtime_lease_in_sec          = 3600  // 1 hour
```
lvirbalas: Syntax highlighting is off when using "//" for comments, but maybe it's only when running locally? If not, "#" seems to work fine.

Screenshot 2020-01-17 at 16 49 24

vcd/resource_vcd_org.go
dataclouder: The lease amounts can change depending on which vCD version you use.

If you set any values within `vapp_lease` block (or if you put an empty block), the result for all the lease amounts will be 0 (= never expires).

If you don't set the block at all, the result will be whatever default the vCD version uses.

vcd/resource_vcd_org.go
dataclouder: done

vcd/resource_vcd_org.go
dataclouder: done

website/docs/r/org.html.markdown
dataclouder: Changed

vcd/resource_vcd_org.go
lvirbalas: > If you set any values within vapp_lease block (or if you put an empty block), the result for **all the lease amounts will be 0 (= never expires)**.

Do you mean that in my first example where I don't set `maximum_runtime_lease_in_sec` it would be forced to 0? That's not what the user would expect, but also not what the docs are saying:

> maximum_runtime_lease_in_sec - (Optional) - How long vApps can run before they are automatically stopped (in seconds). 0 means never expires. Values accepted from 3600+. - **Default is 604800 (7 days) but may vary depending on vCD version**

vcd/validate_funcs.go
vbauzysvmware: For it isn't, but we can live with that.

vcd/resource_vcd_org.go
dataclouder: Made fields in lease sections mandatory. Docs updated

website/docs/r/org.html.markdown
dataclouder: Documentation updated, following some changes in the lease sections: the sections are still optional, but all the fields inside are mandatory. The docs were updated to reflect the new requirements.
The change was necessary because the API return different results depending on whether we put in the whole lease block or not. If we don't put the block, we'll get the same defaults that we get from the GUI (e.g. maximum runtime lease = 7 days) but if we put  other values in the block and omit `maximum_runtime_lease`, the default is zero (never expires).
By making all fields mandatory we avoid the ambiguity.

vcd/datasource_vcd_org.go
```diff
@@ -44,6 +44,55 @@ func datasourceVcdOrg() *schema.Resource {
...
+						"maximum_runtime_lease_in_sec": &schema.Schema{
+							Type:        schema.TypeInt,
+							Computed:    true,
+							Description: "How long vApps can run before they are automatically stopped (in seconds)",
+						},
+						"power_off_on_runtime_lease_expiration": &schema.Schema{
+							Type:     schema.TypeBool,
+							Computed: true,
+							Description: "When true, vApps are powered off when the runtime lease expires. " +
+								"When false or missing, vApps are suspended when the runtime lease expires",
```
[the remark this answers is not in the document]
dataclouder: fixed

vcd/resource_vcd_org.go
```diff
@@ -74,6 +74,65 @@ func resourceOrg() *schema.Resource {
...
+						"maximum_storage_lease_in_sec": &schema.Schema{
+							Type:         schema.TypeInt,
+							Required:     true,
+							Description:  "How long vApp templates are available before being automatically cleaned up (in seconds). 0 means never expires",
+							ValidateFunc: validateIntLeaseSeconds(), // Lease can be either 0 or 3600+
+						},
+						"delete_on_storage_lease_expiration": &schema.Schema{
+							Type:     schema.TypeBool,
+							Required: true,
+							//Default:  false,
```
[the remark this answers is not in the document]
dataclouder: removed

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ FEATURES:
 * **New Resource:** `vcd_vm_internal_disk` VM internal disk configuration
 * `vcd_vapp_vm` Internal disks in VM template allows to be edited by `override_template_disk` field 
 
+IMPROVEMENTS:
+
+* `resource/vcd_org` and `datasource/vcd_org` include a section `vapp_lease` and a section `vapp_template_lease` to define lease related parameters of depending entities - [GH-385]
```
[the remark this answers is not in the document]

vbauzysvmware:
Overall looks good. I have last question - in UI we have option to choose action. Is this another PR scope?
image

sysadmintutorials:
thanks for adding this support, will this be rolled into provider.vcd 2.7 ?

Didainius:
@sysadmintutorials - yes

sysadmintutorials:
awesome any ETA on 2.7 release ?

Didainius:
No exact date yet. We usually take ~2-3 months from the last release. We had a release 1 month ago so this makes  ~1-2 months until the release.

sysadmintutorials:
Ok great, thank you

Issue #316, opened 2019-08-21
vcd_vapp_vm - investigate options to ensure DHCP ip is reported in resource after provisioning


Didainius:
The network block does not guarantee to report `ip` when `ip_allocation_mode=DHCP` (https://www.terraform.io/docs/providers/vcd/r/vapp_vm.html#dhcp).
This happens because we cannot know such things in advance:
* If guest machine has guest tools installed
* If DHCP setting is enabled to be used in guest machine

However the `ip`field (when correct) is useful in referring the `ip` in other resources (like load balancer pool membership) so it would be nice to have a way for ensuring to get a proper IP address from DHCP.

This needs further investigation however one option could be to introduce a "wait_for_dhcp" or similar attribute which would wait for some time before finally reporting IP address. User would need to understand what he is doing and manage the timeout, but still would have an option to use the IP address further down the .tf configuration.

PR #436, opened 2020-01-17
Add support for `network_dhcp_wait_seconds` in`vcd_vapp_vm`

Didainius:
Closes https://github.com/terraform-providers/terraform-provider-vcd/issues/316

This PR adds a parameter `network_dhcp_wait_seconds` for `vcd_vapp_vm` resource and data source. The parameter makes it possible to retrieve `ip` field in `network` block on the first `apply` run when `ip_allocation_mode=DHCP`. The drawback of it is that READ for the functions takes longer. For this function to work VM must be powered on and at least one of the following must be true:
 * VM has guest tools. It waits for IP address to be reported in vCD UI. This is a slower option, but does not require for the VM to use Edge Gateways DHCP service.
 * VM DHCP interface is connected to routed Org network and is using Edge Gateways DHCP service (not relayed). It works by querying DHCP leases on edge gateway. In general it is quicker than waiting until UI reports IP addresses, but is more constrained. However this is the only option if guest tools are not present on the VM.

The feature got the current state after ruling out a few other options. I'm leaving it here to either avoid discussion of what is not possible or to spark some better ideas:

_Problem 1:_ Changing the value of field `network_dhcp_wait_seconds` will actually show a change in `plan` and an update although it will result in _no op `update`_. The reason for this is that in READ function `d.Get("network_dhcp_wait_seconds")` actually gets the value not from `.tf config`, but from `statefile`. Apparently in the READ function there is no way to access value of `.tf config` (open issue for that https://github.com/hashicorp/terraform-plugin-sdk/issues/133). If `DiffSuppressFunc` is used - then `d.Get("network_dhcp_wait_seconds")` always receives 0 (default TypeInt value).

_Problem 2:_ Initially I had idea to add this field to each of `network` block however there are a few issues about it:
* this wouldn't be suitable for data source because they don't have `network` block defined
* change of `network_dhcp_wait_seconds` value would cause whole `network` block to be reported as changed. It could confuse the user of updating network adapter itself (which is not the case here). Also it does not make sense to have different timings for each DHCP NIC as the lookup checks all of them at the same time (to save on API calls) and would still take as the longest time defined.

Note. Test suite passed on 9.5.


vcd/resource_vcd_vapp_vm.go
```diff
@@ -668,7 +679,7 @@ func resourceVcdVAppVmUpdate(d *schema.ResourceData, meta interface{}) error {
 }
 
 func resourceVcdVAppVmUpdateExecute(d *schema.ResourceData, meta interface{}) error {
```
vbauzysvmware: Update should avoid any code handling if only network_dhcp_wait_seconds is changed. Or may call read instead ^^

Didainius: And it does. There is nothing happening when only this value is changed.

vbauzysvmware:
> Closes #316
> 
> This PR adds a parameter `network_dhcp_wait_seconds` for `vcd_vapp_vm` resource and data source. The parameter makes it possible to retrieve `ip` field in `network` block on the first `apply` run when `ip_allocation_mode=DHCP`. The drawback of it is that READ for the functions takes longer. For this function to work VM must be powered on and at least one of the following must be true:
> 
>     * VM has guest tools. It waits for IP address to be reported in vCD UI. This is a slower option, but does not require for the VM to use Edge Gateways DHCP service.
> 
>     * VM DHCP interface is connected to routed Org network and is using Edge Gateways DHCP service (not relayed). It works by querying DHCP leases on edge gateway. In general it is quicker than waiting until UI reports IP addresses, but is more constrained. However this is the only option if guest tools are not present on the VM.
> 
> 
> The feature got the current state after ruling out a few other options. I'm leaving it here to either avoid discussion of what is not possible or to spark some better ideas:
> 
> _Problem 1:_ Changing the value of field `network_dhcp_wait_seconds` will actually show a change in `plan` and an update although it will result in _no op `update`_. The reason for this is that in READ function `d.Get("network_dhcp_wait_seconds")` actually gets the value not from `.tf config`, but from `statefile`. Apparently in the READ function there is no way to access value of `.tf config` (open issue for that [hashicorp/terraform-plugin-sdk#133](https://github.com/hashicorp/terraform-plugin-sdk/issues/133)). If `DiffSuppressFunc` is used - then `d.Get("network_dhcp_wait_seconds")` always receives 0 (default TypeInt value).
> 
> _Problem 2:_ Initially I had idea to add this field to each of `network` block however there are a few issues about it:
> 
>     * this wouldn't be suitable for data source because they don't have `network` block defined
> 
>     * change of `network_dhcp_wait_seconds` value would cause whole `network` block to be reported as changed. It could confuse the user of updating network adapter itself (which is not the case here). Also it does not make sense to have different timings for each DHCP NIC as the lookup checks all of them at the same time (to save on API calls) and would still take as the longest time defined.
> 
> 
> Note. Test suite passed on 9.5.

Regarding 1:
You can avoid 0 issue in suppress function if you use other type for e.g. string. Though then additional validation needed. Also I am kind'a think were will be situation where users use one timeout, sees isn't enough, increases and apply. So maybe reread in that case should happen ^^

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: You mean update function isn't called?

Didainius:
> Regarding 1:
> You can avoid 0 issue in suppress function if you use other type for e.g. string. Though then additional validation needed. Also I am kind'a think were will be situation where users use one timeout, sees isn't enough, increases and apply. So maybe reread in that case should happen ^^

If I suppress - the value does not get to `statefile`. If it does not get into `statefile` - then during READ I cannot access this variable (because `d.Get` in read function reads from `statefile`).

vcd/resource_vcd_vapp_vm.go
Didainius: The function is called, because I can't do "DiffSuppressFunc" (for the reason that the value doesn't get to state and Read only gets values from read), but all actual changes in Update are guarded by `d.HasChange`

If I explicitly list all of the fields we have, it will have to be "maintained" everytime and I didn't find a function that would say something like `d.IfOnlyFieldChanged("wait_for_dhcp_seconds")` although that would be good to possibly exit early.

Didainius:
> Also I am kind'a think were will be situation where users use one timeout, sees isn't enough, increases and apply. So maybe reread in that case should happen ^^

Yes, this is exactly what I thought, but sadly enough I can't make this happen due to current limitations. When a user changes values this is what will happen:
```
terraform apply
# Terraform will at first READ with old value
# Terraform will propose to UPDATE (if it doesn't update - the value does not get to statefile)
# Terraform will do "empty update" and then perform a READ with new DHCP wait value

File: vcd/resource_vcd_vapp_vm.go (hunk 1 above)
Comment (vbauzysvmware) [reply vcd/resource_vcd_vapp_vm.go]: It's risk one against another. I mean early(though have to maintain) exist save time(as now it makes at least 3 request before ends function), but makes it secure that nothing will called when someone will add new features and it won't be guarded for some reason.

**Comment by vbauzysvmware** [comment]
> > Regarding 1:
> > You can avoid 0 issue in suppress function if you use other type for e.g. string. Though then additional validation needed. Also I am kind'a think were will be situation where users use one timeout, sees isn't enough, increases and apply. So maybe reread in that case should happen ^^
> 
> If I suppress - the value does not get to `statefile`. If it does not get into `statefile` - then during READ I cannot access this variable (because `d.Get` in read function reads from `statefile`).

Good point. Forgot that :)

File: vcd/resource_vcd_vapp_vm.go (hunk 1 above)
Comment (Didainius) [reply vcd/resource_vcd_vapp_vm.go]: I will check if I can make some function having behavior as `d.IfOnlyFieldChanged("wait_for_dhcp_seconds")` with what is available.

**Comment by vbauzysvmware** [comment]
> > Also I am kind'a think were will be situation where users use one timeout, sees isn't enough, increases and apply. So maybe reread in that case should happen ^^
> 
> Yes, this is exactly what I thought, but sadly enough I can't make this happen due to current limitations. When a user changes values this is what will happen:
> 
> ```
> terraform apply
> # Terraform will at first READ with old value timeout value
> # Terraform will propose to UPDATE (if it doesn't update - the value does not get to statefile)
> # Terraform will do "empty update" and then perform a READ with new DHCP wait value from statefile
> ```

Not nice. Yet we can only improve update if possible or agree todo that.

**Comment by Didainius** [comment]
> > > Also I am kind'a think were will be situation where users use one timeout, sees isn't enough, increases and apply. So maybe reread in that case should happen ^^
> > 
> > 
> > Yes, this is exactly what I thought, but sadly enough I can't make this happen due to current limitations. When a user changes values this is what will happen:
> > ```
> > terraform apply
> > # Terraform will at first READ with old value timeout value
> > # Terraform will propose to UPDATE (if it doesn't update - the value does not get to statefile)
> > # Terraform will do "empty update" and then perform a READ with new DHCP wait value from statefile
> > ```
> 
> Not nice. Yet we can only improve update if possible or agree todo that.


fb837a5  has the changes to support that

File: vcd/resource_vcd_vapp_vm.go (hunk 1 above)
Comment (Didainius) [reply vcd/resource_vcd_vapp_vm.go]: I have implemented a custom fb837a5 function to check that based on the schema of resource (didn't find anything useful in `d.XXX` functions)

File: website/docs/d/vapp_vm.html.markdown [hunk 2]
```diff
@@ -71,6 +71,16 @@ The following arguments are supported:
 * `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
```
lvirbalas: **Not** introduced by this PR, but maybe you could fix this sample output formatting on the way:

Screenshot 2020-02-17 at 13 21 03

CC: @dataclouder (in case you planned/fixed this too)

website/docs/d/vapp_vm.html.markdown
```diff
@@ -71,6 +71,16 @@ The following arguments are supported:
...
 * `vapp_name` - (Required) The vApp this VM belongs to.
 * `name` - (Required) A name for the VM, unique within the vApp 
+* `network_dhcp_wait_seconds` - (Optional; *v2.7+*) Allows to wait for up to a defined amount of
+  seconds before IP address is reported for NICs with `ip_allocation_mode=DHCP` setting. It
+  constantly checks if IP is reported so the time given is a maximum. VM must be powered on and 
+  __at least one__ of the folowing __must be true__:
+ * VM has guest tools. It waits for IP address to be reported in vCD UI. This is a slower option, but
```
lvirbalas: > waits for IP address to be reported in vCD UI

The mention of the UI here and below are a bit unexpected. I'm pretty sure you are not waiting for vCD UI - you're checking the vCD API. We should change the formulation in such a way, that user is not left thinking that we are querying the UI (which we are not). I mean, we can mention that "at that point, UI reports IP address as well" or something like that, but we should not make it seem like provider is depended on UI.

website/docs/d/vapp_vm.html.markdown
```diff
@@ -71,6 +71,16 @@ The following arguments are supported:
 * `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
 * `vapp_name` - (Required) The vApp this VM belongs to.
 * `name` - (Required) A name for the VM, unique within the vApp 
+* `network_dhcp_wait_seconds` - (Optional; *v2.7+*) Allows to wait for up to a defined amount of
+  seconds before IP address is reported for NICs with `ip_allocation_mode=DHCP` setting. It
```
lvirbalas: > Allows to wait for up to a defined amount of seconds before IP address is reported for NICs with `ip_allocation_mode=DHCP` setting

One question that is left after reading this statement is whether `network_dhcp_wait_seconds` is used with other `ip_allocation_mode` modes at all. If not, we should make it clear, for example:

> This field is ignored with non-DHCP `ip_allocation_mode`.

lvirbalas: Just noticed that you have it formulated even more nicely here, so could just copy/paste your variant:
https://github.com/terraform-providers/terraform-provider-vcd/pull/436/files#diff-73ac9414e0dd641233165392db1d4329R79-R80

lvirbalas:
Tried with `bitnami-wordpress-5.2.2-3-linux-centos-7-x86_64` and vCD 10.0. Worked quite nicely! VM got created, IP retrieved and all the NAT+FW rules created automatically with the IP assigned to VM by DHCP.
```vcd_vapp_vm.demo_vm_wordpress: Creation complete after 5m2s [id=urn:vcloud:vm:e5a82105-6c72-4dbd-b8b3-086ad1db0fea]```

File: website/docs/r/vapp_vm.html.markdown [hunk 5]
```diff
@@ -140,6 +140,35 @@ resource "vcd_vapp_vm" "internalDiskOverride" {
...
+  vapp_name     = vcd_vapp.TestAccVcdVAppVmDhcpWait.name
+  name          = "brr"
+  computer_name = "dhcp-vm"
+  catalog_name  = "cat-dserplis"
+  template_name = "photon-rev2"
+  memory        = 512
+  cpus          = 2
+  cpu_cores     = 1
+
+  network_dhcp_wait_seconds = 220
```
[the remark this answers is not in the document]
Didainius: No. this is something like a generic thing. Maybe the example deserves a "conservatime" time of 300s (5 minutes)? It may take that time if the image is a beefy one.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1596,6 +1640,55 @@ func readNetworks(vm govcd.VM, vapp govcd.VApp) ([]map[string]interface{}, error
...
+			nicIps, timeout, err := vm.WaitForDhcpIpByNicIndexes(dhcpNicIndexes, maxDhcpWaitSecondsInt, true)
+			if err != nil {
+				return nil, fmt.Errorf("unable to to lookup DHCP IPs for VM NICs '%v': %s", dhcpNicIndexes, err)
+			}
+
+			if timeout {
+				log.Printf("[DEBUG] [VM read] VM %s timed out waiting %d seconds "+
+					"to report DHCP IPs. You may want to increase 'network_dhcp_wait_seconds' or ensure "+
+					"your DHCP settings are correct.\n", vm.VM.Name, maxDhcpWaitSeconds)
+				_, _ = fmt.Fprintf(getTerraformStdout(), "WARNING: VM %s timed out waiting %d seconds "+
```
lvirbalas: I tried to hit the timeout. It's nice that we get the WARNING inside the console. However, the further plan ended up running with a bunch of errors, because NAT+FW rules tried to use empty IP. Do we have any way to protect against that or workaround in the `.tf` config itself?

```
vcd_vapp_vm.demo_vm_wordpress: Still creating... [4m50s elapsed]
WARNING: VM demo-vm-wordpress timed out waiting 60 seconds to report DHCP IPs. You may want to increase 'network_dhcp_wait_seconds' or ensure your DHCP settings are correct.
vcd_vapp_vm.demo_vm_wordpress: Creation complete after 4m52s [id=urn:vcloud:vm:5322ba15-9fc6-4e56-a193-6072f6c66129]
vcd_nsxv_dnat.demo_dnat_wordpress_ssh: Modifying... [id=196613]
vcd_nsxv_snat.web: Modifying... [id=196612]
vcd_nsxv_dnat.demo_dnat_wordpress_https: Modifying... [id=196614]
vcd_nsxv_firewall_rule.demo-nsxv-fw-rule-outbound: Modifying... [id=131079]
vcd_nsxv_dnat.demo_dnat_wordpress_https: Still modifying... [id=196614, 10s elapsed]
vcd_nsxv_firewall_rule.demo-nsxv-fw-rule-outbound: Still modifying... [id=131079, 10s elapsed]
vcd_nsxv_firewall_rule.demo-nsxv-fw-rule-outbound: Still modifying... [id=131079, 20s elapsed]
vcd_nsxv_firewall_rule.demo-nsxv-fw-rule-outbound: Modifications complete after 25s [id=131079]

Error: unable to update NAT rule with ID 196614: error while updating NAT rule : vShield Edge Invalid IP Address input /32 for field translatedAddress. (API error: 15012)

  on xendi.tf line 393, in resource "vcd_nsxv_dnat" "demo_dnat_wordpress_https":
 393: resource "vcd_nsxv_dnat" "demo_dnat_wordpress_https" {



Error: unable to update NAT rule with ID 196613: error while updating NAT rule : vShield Edge Invalid IP Address input /32 for field translatedAddress. (API error: 15012)

  on xendi.tf line 411, in resource "vcd_nsxv_dnat" "demo_dnat_wordpress_ssh":
 411: resource "vcd_nsxv_dnat" "demo_dnat_wordpress_ssh" {



Error: unable to update NAT rule with ID 196612: error while updating NAT rule : vShield Edge Invalid IP Address input /32 for field originalAddress. (API error: 15012)

  on xendi.tf line 430, in resource "vcd_nsxv_snat" "web":
 430: resource "vcd_nsxv_snat" "web" {
```

lvirbalas:
Tried one more time with Org VDC network with DHCP pool:
```
resource "vcd_network_routed" "demo_routed_net_web" {
  name         = "demo-net-web"
  edge_gateway = "${vcd_edgegateway.demo_gw.name}"

  gateway = "192.168.0.1"

  static_ip_pool {
    start_address = "192.168.0.2"
    end_address   = "192.168.0.100"
  }

  dhcp_pool {
    start_address = "192.168.0.101"
    end_address   = "192.168.0.132"
  }
}
```
Now the IP assigned to VM was from this pool and the log said it got it after lease check (as opposed to guest tools):
```
$ tail -f go-vcloud-director.log | grep IPs
<...>
2020/02/17 16:14:34 [DEBUG] VM 'demo-vm-wordpress' NICs with indexes [0] did not all report their IPs using guest tools
2020/02/17 16:14:34 [DEBUG] VM 'demo-vm-wordpress' NICs with indexes [0] did not all report their IPs using DHCP leases
<...>
2020/02/17 16:16:07 [DEBUG] VM 'demo-vm-wordpress' NICs with indexes [0] did not all report their IPs using guest tools
2020/02/17 16:16:07 [TRACE] VM 'demo-vm-wordpress' NICs with indexes [0] all reported their IPs after lease check
```
And IP from the pool inside Terraform:
```
Apply complete! Resources: 18 added, 0 changed, 0 destroyed.

Outputs:

demo_vm_wordpress_ip = 192.168.0.101
```

website/docs/r/vapp_vm.html.markdown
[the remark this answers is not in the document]
lvirbalas: I concur - better longer timeout here (300s).

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1596,6 +1640,55 @@ func readNetworks(vm govcd.VM, vapp govcd.VApp) ([]map[string]interface{}, error
...
+					"to report DHCP IPs. You may want to increase 'network_dhcp_wait_seconds' or ensure "+
+					"your DHCP settings are correct.\n", vm.VM.Name, maxDhcpWaitSeconds)
+			}
+
+			log.Printf("[DEBUG] [VM read] '%s' waiting for DHCP IPs took '%s' (of '%ds')",
+				vm.VM.Name, time.Since(start), maxDhcpWaitSeconds)
+
+			for sliceIndex, nicIndex := range dhcpNicIndexes {
+				log.Printf("[DEBUG] [VM read] '%s' NIC %d reported IP %s",
```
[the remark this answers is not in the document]
Didainius: Both govcd and terraform log messages should include "DHCP IP Lookup" string now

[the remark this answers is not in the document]
lvirbalas: Thanks! Total nit, but here's one idea, maybe let's use the existing bracket scheme for it:

>[DEBUG] [VM read] NIC '%d' **DHCP IP Lookup -** is not using DHCP in 'ip_allocation_mode'. Skipping IP wait
->
>[DEBUG] [VM read] **[DHCP IP Lookup]** NIC '%d' is not using DHCP in 'ip_allocation_mode'. Skipping IP wait

website/docs/d/vapp_vm.html.markdown
Didainius: Fixed. 
image

website/docs/r/vapp_vm.html.markdown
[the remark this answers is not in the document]
Didainius: Adjusted with extra comment to show time value in minutes.

vcd/resource_vcd_vapp_vm.go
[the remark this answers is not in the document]
Didainius: Changed.

website/docs/d/vapp_vm.html.markdown
Didainius: I have rephrased website doc - do have a check if it is better.

Also - when `network_dhcp_wait_seconds` is specified and none of the adapters use DHCP - user would see something like below during READ:

```
...
vcd_vapp_vm.TestAccVcdVAppVmDhcpWaitVM: Creating...
vcd_vapp_vm.TestAccVcdVAppVmDhcpWaitVM: Still creating... [10s elapsed]
vcd_vapp_vm.TestAccVcdVAppVmDhcpWaitVM: Still creating... [20s elapsed]
vcd_vapp_vm.TestAccVcdVAppVmDhcpWaitVM: Still creating... [30s elapsed]
vcd_vapp_vm.TestAccVcdVAppVmDhcpWaitVM: Still creating... [40s elapsed]
vcd_vapp_vm.TestAccVcdVAppVmDhcpWaitVM: Still creating... [50s elapsed]
vcd_vapp_vm.TestAccVcdVAppVmDhcpWaitVM: Still creating... [1m0s elapsed]
INFO: Using 'network_dhcp_wait_seconds' only makes sense if at least one NIC is using 'ip_allocation_mode=DHCP'
...
```

vcd/resource_vcd_vapp_vm.go
Didainius: The first thing that comes to my mind is actually returning an **error** during READ instead of **warning**. But there are a few problems and I believe this isn't good enough:
* READ is part of CREATE/UPDATE so creation could fail because of too small value.
* Because READ accesses the configuration from statefile, even changing the value to a higher one (say from 1 second to 300 seconds) would still at first try to READ with 1 second (because plan/apply triggers READ before UPDATE). If it fails immediately then the user never gets to 300 seconds.

vcd/resource_vcd_vapp_vm.go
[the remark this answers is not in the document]
lvirbalas: It's really clean now, thanks!
```
$ tail -f go-vcloud-director.log | grep "DHCP IP Lookup"
2020/02/18 15:38:27 [TRACE] [DHCP IP Lookup] VM 'demo-vm-wordpress' attempting to lookup IP addresses for DHCP NICs [0]
2020/02/18 15:38:31 [TRACE] [DHCP IP Lookup] VM 'demo-vm-wordpress' NIC with index 0 is attached to edge gateway routed network
2020/02/18 15:38:35 [DEBUG] [DHCP IP Lookup] VM 'demo-vm-wordpress' NICs with indexes [0] did not all report their IPs using guest tools
2020/02/18 15:38:35 [DEBUG] [DHCP IP Lookup] VM 'demo-vm-wordpress' NICs with indexes [0] did not all report their IPs using DHCP leases
<...>
2020/02/18 15:40:08 [DEBUG] [DHCP IP Lookup] VM 'demo-vm-wordpress' NICs with indexes [0] all reported their IPs after lease check
```

website/docs/r/vapp_vm.html.markdown
```diff
@@ -170,13 +199,24 @@ DHCP.
...
+* `network` - (Optional; *v2.2+*) A block to define network interface. Multiple can be used. See [Network](#network-block) and 
 example for usage details. **Deprecates**: `network_name`, `ip`, `vapp_network_name`.
 * `customization` - (Optional; *v2.5+*) A block to define for guest customization options. See [Customization](#customization)
 * `guest_properties` - (Optional; *v2.5+*) Key value map of guest properties
 * `description`  - (Computed; *v2.6+*) The VM description. Note: description is read only. Currently, this field has
   the description of the OVA used to create the VM
 * `override_template_disk` - (Optional; *v2.7+*) Allows to update internal disk in template before first VM boot. Disk is matched by `bus_type`, `bus_number` and `unit_number`. See [Override template Disk](#override-template-disk) below for details.
+* `network_dhcp_wait_seconds` - (Optional; *v2.7+*) Optional number of seconds to try and wait for DHCP IP (only valid
+  for adapters in `network` block with ip_allocation_mode=DHCP). It constantly checks if IP is present so the time given
```
lvirbalas: Small suggestion:
```suggestion
  for adapters in `network` block with `ip_allocation_mode=DHCP`). It constantly checks if IP is present so the time given
```

vcd/resource_vcd_vapp_vm.go
lvirbalas: Yes, looks like current behavior is actually correct. Given that VM _does_ get created even though without IP address, we _shouldn't_ report an error. In a sense, we're missing a special state in the Terraform state machine to model such incomplete executions. All in all, let's leave current implementation intact.

PR #441, opened 2020-01-24
Add network adapter type support in vcd_vapp_vm resource and data source.

### LINKED ISSUE #327: NIC adapter type for VM
Placeholder for ability to specify NIC adapter type (e.g. E1000, VMXNET3, PCNet32) in VM's network section.

Didainius:
Implements and closes #327

This PR adds support of reading and setting `adapter_type` in `vcd_vapp_vm`.

It has loose validation on the field, because there may be more adapters which are not in the current list as future vSphere version may introduce new ones. Case is also ignored as the API accepts in any case, but returns specifically cased. `DiffSuppressFunc` covers this.

Note. The `TestAccVcdVAppVmMultiNIC` specific test was tested on 9.1, 9.5 and 10. Whole test suite was tested on 10.


vcd/datasource_vcd_vapp_vm.go
```diff
@@ -107,6 +107,11 @@ func datasourceVcdVAppVm() *schema.Resource {
 							Type:        schema.TypeString,
 							Description: "Mac address of network interface",
 						},
+						"adapter_type": {
+							Computed:    true,
```
vbauzysvmware: Please swap lines, we use `Type` everywhere as first

vcd/resource_vcd_vapp_vm.go
```diff
@@ -203,6 +203,13 @@ func resourceVcdVAppVm() *schema.Resource {
 							Type:        schema.TypeString,
 							Description: "Mac address of network interface",
 						},
+						"adapter_type": {
+							Computed:         true,
+							Optional:         true,
+							Type:             schema.TypeString,
```
vbauzysvmware: Please swap lines, we use Type everywhere as first

vcd/datasource_vcd_vapp_vm.go
Didainius: Done.

vcd/resource_vcd_vapp_vm.go
Didainius: Done.

lvirbalas:
p.s. @dataclouder when reviewing this PR could you try on vCD 9.x? Thanks.

Didainius:
> Thinking more about [my last question](https://github.com/terraform-providers/terraform-provider-vcd/pull/441#pullrequestreview-352832978) of inability to **update** adapter type on existing VM.
> 
> In UI this field is read-only, hence the users are not allowed to change it. On the other hand, we allow to change it and get the error message. If there is a case where such update would work, it makes sense. If not, then we should do one of the following:
> 
> a. Warn the user in documentation that updating the value in adapter_type is usually unsupported by the underlying infrastructure and would cause an error.
> 
> b. Force recreation of the whole corresponding `network` block (the connection itself), but if there's no standard (non-hack) way of implementing (and representing in the _plan_) such recreation of only a block in a resource, let's _not_ overdo it. A sentence in documentation would be enough.
> 
> LGTM, otherwise :)

Yes. There is a problem - if I add the native Terraform `ForceNew: true` it will propose to re-provision whole resource on `adapter_type` change. It does not do so if I remove an existing adapter, but will always proposes to recreate whole resource if I add a new adapter.
Consider this workflow:
1. Provision VM with some specific adapter type
2. Remove definition of that block and apply (removing the block does not propose to recreate VM, just removes the adapter). I do this with intention to insert an adapter with different type later.
3. Add an adapter back with different type. And then the **problem occurs** (it proposes to recreate resource):
```
 # vcd_vapp_vm.TestAccVcdVAppVmMultiNICVM must be replaced
-/+ resource "vcd_vapp_vm" "TestAccVcdVAppVmMultiNICVM" {
        accept_all_eulas               = true
        catalog_name                   = "my-catalog"
      ~ computer_name                  = "PhotonOS-001" -> (known after apply)
        cpu_cores                      = 1
        cpus                           = 2
      ~ description                    = "photon-os-description" -> (known after apply)
        expose_hardware_virtualization = false
      - guest_properties               = {} -> null
      ~ href                           = "https://192.168.1.109/api/vApp/vm-99da32de-0a9e-40fe-bd1a-1124b805044a" -> (known after apply)
      ~ id                             = "urn:vcloud:vm:99da32de-0a9e-40fe-bd1a-1124b805044a" -> (known after apply)
      ~ internal_disk                  = [
          - {
              - bus_number       = 0
              - bus_type         = "paravirtual"
              - disk_id          = "2000"
              - iops             = 0
              - size_in_mb       = 16384
              - storage_profile  = "Development"
              - thin_provisioned = true
              - unit_number      = 0
            },
        ] -> (known after apply)
      + ip                             = (known after apply)
      + mac                            = (known after apply)
        memory                         = 512
      - metadata                       = {} -> null
        name                           = "TestAccVcdVAppVmMultiNICVM"
        org                            = "my-org"
        power_on                       = true
      ~ storage_profile                = "Development" -> (known after apply)
        template_name                  = "photon-os"
        vapp_name                      = "TestAccVcdVAppVmMultiNIC"
        vdc                            = "my-org-vdc"

      ~ network {
          ~ adapter_type       = "VMXNET3" -> (known after apply)
          ~ ip                 = "11.10.0.152" -> (known after apply)
            ip_allocation_mode = "POOL"
          - is_primary         = true -> null
          ~ mac                = "00:50:56:01:31:7b" -> (known after apply)
            name               = "multinic-net"
            type               = "org"
        }
      + network {
          + adapter_type       = "E1000E" # forces replacement
          + ip                 = (known after apply)
          + ip_allocation_mode = "DHCP"
          + is_primary         = true
          + mac                = (known after apply)
          + name               = "multinic-net"
          + type               = "org"
        }
    }

Plan: 1 to add, 0 to change, 1 to destroy.
```

In general, it would break the functionality of simply adding a new network adapter so this won't work.

As a result of this experiment I am going with your option **a** and added a note into website docs.

lvirbalas:
> * Provision VM with some specific adapter type
> 
> * Remove definition of that block and apply (removing the block does not propose to recreate VM, just removes the adapter). I do this with intention to insert an adapter with different type later.
> 
> * Add an adapter back with different type. And then the **problem occurs** (it proposes to recreate resource):

Maybe I'm missing some important detail here, but it shouldn't propose recreation in this workflow, because from Terraform's perspective it's just a new adapter. It shouldn't be saving any state of the previous adapter, hence how can it know that the type has changed?

I tried the following:

1. Added 2nd adapter to VM and applied:

 ```
 network {
    type               = "org"
    name               = vcd_network_routed.tf_org_net.name
    ip_allocation_mode = "DHCP"
    is_primary         = false
    adapter_type = "E1000E"
  }
```

2. Removed the above adapter and applied.

3. Added 2nd adapter again, but with different type this time:

```
 network {
    type               = "org"
    name               = vcd_network_routed.tf_org_net.name
    ip_allocation_mode = "DHCP"
    is_primary         = false
    adapter_type = "VMXNET3"
  }
```

Result is an **in-place** update of the VM:

```
# vcd_vapp_vm.tf_org_vm will be updated in-place
  ~ resource "vcd_vapp_vm" "tf_org_vm" {
        accept_all_eulas               = true
        catalog_name                   = "shared-catalog"
        computer_name                  = "PhotonOS-001"
        <...>
        internal_disk                  = [
            {
                bus_number       = 0
                bus_type         = "paravirtual"
                disk_id          = "2000"
                iops             = 0
                size_in_mb       = 16384
                storage_profile  = "*"
                thin_provisioned = true
                unit_number      = 0
            },
        ]
        memory                         = 384
        metadata                       = {}
        name                           = "tf-org-vm"
        power_on                       = true
        storage_profile                = "*"
        template_name                  = "photon-hw11-3.0-9355405"
        vapp_name                      = "tf-org-vapp"

        network {
            adapter_type       = "E1000"
            ip                 = "192.168.1.2"
            ip_allocation_mode = "POOL"
            is_primary         = true
            mac                = "00:50:56:29:00:53"
            name               = "tf-org-net"
            type               = "org"
        }
      + network {
          + adapter_type       = "VMXNET3"
          + ip_allocation_mode = "DHCP"
          + name               = "tf-org-net"
          + type               = "org"
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

vcd_vapp_vm.tf_org_vm: Modifying... [id=urn:vcloud:vm:7b11ebc6-d3bb-4d5f-b72c-bd7bdbb3d01d]
vcd_vapp_vm.tf_org_vm: Still modifying... [id=urn:vcloud:vm:7b11ebc6-d3bb-4d5f-b72c-bd7bdbb3d01d, 10s elapsed]
<...>
vcd_vapp_vm.tf_org_vm: Modifications complete after 1m4s [id=urn:vcloud:vm:7b11ebc6-d3bb-4d5f-b72c-bd7bdbb3d01d]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

Didainius:
> Maybe I'm missing some important detail here, but it shouldn't propose recreation in this workflow, because from Terraform's perspective it's just a new adapter. It shouldn't be saving any state of the previous adapter, hence how can it know that the type has changed?

My above example was if I used `ForceNew: true` for `adapter_type` field. I had compiled such version for testing the hypothesis.

jpbuecken:
> Thinking more about [my last question](https://github.com/terraform-providers/terraform-provider-vcd/pull/441#pullrequestreview-352832978) of inability to **update** adapter type on existing VM.
> 
> In UI this field is read-only, hence the users are not allowed to change it. On the other hand, we allow to change it and get the error message. If there is a case where such update would work, it makes sense. If not, then we should do one of the following:
> 
> a. Warn the user in documentation that updating the value in adapter_type is usually unsupported by the underlying infrastructure and would cause an error.
> 
> b. Force recreation of the whole corresponding `network` block (the connection itself), but if there's no standard (non-hack) way of implementing (and representing in the _plan_) such recreation of only a block in a resource, let's _not_ overdo it. A sentence in documentation would be enough.
> 
> LGTM, otherwise :)

As a long term task, I may suggest a solution c.
If you add a resource similar to `vm_internal_disk` , e.g. `vm_internal_network` or `vm_network_nic` or `vm_internal_nic` ...
then you can implement the behavior as it should be (force new if network adapter type changed).

If you look in the H5 this approach will be consistent. You have to sections (Hard Disks and NICs) in "Hardware" and you can click "ADD". It makes sence to me to implement similar GUI items as similar terraform configuration?

image

This idea should not delay this PR, but an idea to consider to improve this in future.

lvirbalas:
That's an interesting thought @jpbuecken . Realistically, we are hitting the limits of Terraform itself in these kind of cases, where on one side it's more convenient to have it in sub-blocks, but Terraform doesn't provide enough flexibility to manage life-cycle of the sub-blocks, while separate resources have different draw backs. At the end of the day, it's a balancing act...

PR #446, opened 2020-01-31
Fix/Unify data source behavior on Read errors

Didainius:
As per other major providers and general Terraform practice - a data source must return an error when it cannot find referenced object. It was not all consistent in our case and this PR introduces a generic Test `TestAccDataSourceNotFound` which should cover all current and future data sources.

**Note**. Resources also must undergo similar cleanup, but there looks to be no way to make a generic test therefore I am splitting this PR so that it is not that large.

As part of this testing I discovered a bug where `vcd_nsxv_dhcp_relay` data source would crash when DHCP relay is not configured on edge gateway. 86dc583  fixes it.

Acceptance tests passed on 9.7 and 10.


vcd/resource_vcd_nsxv_dhcp_relay.go
```diff
@@ -303,6 +303,10 @@ func getDhcpRelayAgentsType(relayAgentsSet *schema.Set, edge *govcd.EdgeGateway)
 // setDhcpRelayServerData sets DHCP relay server related fields into statefile
 func setDhcpRelayServerData(d *schema.ResourceData, edgeRelay *types.EdgeDhcpRelay, edge *govcd.EdgeGateway, vdc *govcd.Vdc) error {
 	relayServer := edgeRelay.RelayServer
+	// If relay server has no config - just return it empty
```
vbauzysvmware: Can we add tests to cover this cases?

Didainius: It is there. That is how I detected it.

PR #449, opened 2020-01-31
Improve import for vapp and vapp_vm if environment is used for ORG and VDC

jpbuecken:
Try to fix
https://github.com/terraform-providers/terraform-provider-vcd/issues/444
similar to
https://github.com/terraform-providers/terraform-provider-vcd/commit/94703ff8b400b6fb288ffc71f143669f8f517bbf

for resources vapp and vapp_vm

### DIFF
--- a/vcd/resource_vcd_vapp.go
+++ b/vcd/resource_vcd_vapp.go
@@ -706,8 +706,12 @@ func resourceVcdVappImport(d *schema.ResourceData, meta interface{}) ([]*schema.
 		return nil, fmt.Errorf("[vapp import] error retrieving vapp %s: %s", vappName, err)
 	}
 	_ = d.Set("name", vappName)
-	_ = d.Set("org", orgName)
-	_ = d.Set("vdc", vdcName)
+	if vcdClient.Org != orgName {
+		_ = d.Set("org", orgName)
+	}
+	if vcdClient.Vdc != vdcName {
+		_ = d.Set("vdc", vdcName)
+	}
 	d.SetId(vapp.VApp.ID)
 	return []*schema.ResourceData{d}, nil
 }
--- a/vcd/resource_vcd_vapp_vm.go
+++ b/vcd/resource_vcd_vapp_vm.go
@@ -1679,8 +1679,12 @@ func resourceVcdVappVmImport(d *schema.ResourceData, meta interface{}) ([]*schem
 		return nil, fmt.Errorf("[VM import] error retrieving VM %s: %s", vmName, err)
 	}
 	_ = d.Set("name", vmName)
-	_ = d.Set("org", orgName)
-	_ = d.Set("vdc", vdcName)
+	if vcdClient.Org != orgName {
+		_ = d.Set("org", orgName)
+	}
+	if vcdClient.Vdc != vdcName {
+		_ = d.Set("vdc", vdcName)
+	}
 	_ = d.Set("vapp_name", vappName)
 	d.SetId(vm.VM.ID)
 	return []*schema.ResourceData{d}, nil


Didainius:
Hi @jpbuecken,
Thanks for PR!

This change implies that one cannot have `org` and `vdc` defined at both `provider` and _resource_ level during import. Current import tests fail `TestAccVcdVApp_PowerOff` and `TestAccVcdVAppVm_Basic`:
```
=== RUN   TestAccVcdVApp_PowerOff
--- FAIL: TestAccVcdVApp_PowerOff (217.03s)
    testing.go:635: Step 3 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        (map[string]string) {
        }

        (map[string]string) (len=2) {
         (string) (len=3) "org": (string) (len=6) "my-org",
         (string) (len=3) "vdc": (string) (len=10) "my-org-vdc"
        }

FAIL
```

I am actually fine with this as it still is better than the current behavior, but you'd need to remove `org` and `vdc` defintions from the tests so that they pass:
https://github.com/jpbuecken/terraform-provider-vcd/blob/improve_import_vapp_vm/vcd/resource_vcd_vapp_test.go#L291-L292
and
https://github.com/jpbuecken/terraform-provider-vcd/blob/improve_import_vapp_vm/vcd/resource_vcd_vapp_vm_test.go#L191-L192

@vbauzysvmware ,
I think you introduced this way of import handling in  internal disk recently. Do you see any issues with such approach?

vbauzysvmware:
> Hi @jpbuecken,
> Thanks for PR!
> 
> This change implies that one cannot have `org` and `vdc` defined at both `provider` and _resource_ level during import. Current import tests fail `TestAccVcdVApp_PowerOff` and `TestAccVcdVAppVm_Basic`:
> 
> ```
> === RUN   TestAccVcdVApp_PowerOff
> --- FAIL: TestAccVcdVApp_PowerOff (217.03s)
>     testing.go:635: Step 3 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
>         (map[string]string) {
>         }
> 
>         (map[string]string) (len=2) {
>          (string) (len=3) "org": (string) (len=6) "my-org",
>          (string) (len=3) "vdc": (string) (len=10) "my-org-vdc"
>         }
> 
> FAIL
> ```
> 
> I am actually fine with this as it still is better than the current behavior, but you'd need to remove `org` and `vdc` defintions from the tests so that they pass:
> https://github.com/jpbuecken/terraform-provider-vcd/blob/improve_import_vapp_vm/vcd/resource_vcd_vapp_test.go#L291-L292
> and
> https://github.com/jpbuecken/terraform-provider-vcd/blob/improve_import_vapp_vm/vcd/resource_vcd_vapp_vm_test.go#L191-L192
> 
> @vbauzysvmware ,
> I think you introduced this way of import handling in internal disk recently. Do you see any issues with such approach?

I think if it works for most cases, we should use same approach everywhere.

lvirbalas:
> I think if it works for most cases, we should use same approach everywhere.

@vbauzysvmware After your fix, upon which this PR is based on too, does the following configuration still work? It has a default Org and VDC set for provider, but then specific Org and VDC set in the few exclusive resources?

1. A default Org and VDC in provider:

```
provider "vcd" {
  password             = "..."
  org                  = "Solpan"                                               # Default for resources
  vdc                  = "SolpanVDC"                                            # Default for resources
  url                  = "..."
  max_retry_timeout    = "300"
  allow_unverified_ssl = "true"
}
```

2. An overriding Org and VDC in the resources:

```
# vApp in the routed network
resource "vcd_vapp" "TerraformVAppCat" {
  org = "Cats"
  vdc = "CatsVDC"

  name = "TerraformVAppCat"
}

# vApp directly in the external network
resource "vcd_vapp" "TerraformVAppExternal" {
  vdc = "SolpanExtVDC"

  name = "TerraformVAppExternal"
}
```

vbauzysvmware:
> > I think if it works for most cases, we should use same approach everywhere.
> 
> @vbauzysvmware After your fix, upon which this PR is based on too, does the following configuration still work? It has a default Org and VDC set for provider, but then specific Org and VDC set in the few exclusive resources?
> 
>     1. A default Org and VDC in provider:
> 
> 
> ```
> provider "vcd" {
>   user                 = "administrator"
>   password             = "..."
>   sysorg               = "System"
>   org                  = "Solpan"                                               # Default for resources
>   vdc                  = "SolpanVDC"                                            # Default for resources
>   url                  = "..."
>   max_retry_timeout    = "300"
>   allow_unverified_ssl = "true"
> }
> ```
> 
>     1. An overriding Org and VDC in the resources:
> 
> 
> ```
> # vApp in the routed network
> resource "vcd_vapp" "TerraformVAppCat" {
>   org = "Cats"
>   vdc = "CatsVDC"
> 
>   name = "TerraformVAppCat"
> }
> 
> # vApp directly in the external network
> resource "vcd_vapp" "TerraformVAppExternal" {
>   vdc = "SolpanExtVDC"
> 
>   name = "TerraformVAppExternal"
> }
> ```

Similar change was introduced in VM internal disks. Import needs to check if provider and import org/vdc matches if not it has to write down to state file. If not, on apply resource won't be found and state will be delete or error, depends on implementation.

Didainius:
@jpbuecken ,
Just an ask not to invest too much time as we were thinking a lot about this problem and how to make it even more friendly. I have one idea I need to validate so If you could bare with me for another day.
I will give an update in a day, max two.

Thanks!

jpbuecken:
> @jpbuecken ,
> Just an ask not to invest too much time as we were thinking a lot about this problem and how to make it even more friendly. I have one idea I need to validate so If you could bare with me for another day.
> I will give an update in a day, max two.
> 
> Thanks!

OK, I was ill last week anyway. We will import a bunch of VCD VMs to terraform in the near future (next week). 
Maybe we just delete the ORG/VDC information from state file after import. We need to modify the state file anyway to set power_on to a value unequal null ( because of https://github.com/terraform-providers/terraform-provider-vcd/issues/448 )

Didainius:
@jpbuecken,
I have created a draft PR (https://github.com/terraform-providers/terraform-provider-vcd/pull/454) which has vapp and VM patched.

I think it is a better approch, but as you are hitting this problem directly I'd like your review/test if it does what you expect.

The main point here is to shift handling of org and vdc from _import_ to _DiffSuppressFunc_.

That way it wouldn't matter if `org` and/org `vdc` are set in resource definition or not, provided they are the same `provider` section. You would even have an option to either set them, or not set them if you like.

It should also even work for existing resources if someone wants to improve readability by adding/removing these fields without having to reprovision them.

jpbuecken:
> I think it is a better approch, but as you are hitting this problem directly I'd like your review/test if it does what you expect.

Commented in the other PR. Looks good. We can close this one and discuss further in #454

Didainius:
Closing this PR at the moment as part of moving discussion to https://github.com/terraform-providers/terraform-provider-vcd/pull/454

PR #451, opened 2020-02-05
Align READ/refresh behavior for all resources

Didainius:
As per other major providers and general Terraform practice - a resource _must not return an error_ when it _cannot find referenced object_. It simply should remove it from the state. It is not all consistent in our codebase and this PR is to clean this up.

However a _more important_ part of this PR is to create a test that ensures consistency in current and future resources.

That is not a trivial task and it will lengthen running test time because the general workflow has to be:
1. Create a resource in Terraform (there is no way to inject fake state into testing framework)
2. Delete the created resource using govcd SDK
3. Run acceptance test step and expect a non empty plan (because when READ does not return an error Terraform should simply propose to create the missing resource on plan/apply

This test should also capture all resources defined (now and in future) in provider and fail if there is no READ test defined for it.

# Draft state
This is still a work in progress to find the best solution with minimal extra code for performing tests for current and future resources to align the recommended "READ" standard


Didainius:
At this point in time it does not look as a 'nice to read" solution therefore will give it another think.

PR #453, opened 2020-02-11
Fix data race in client connection caching and add -race flag to tests

Didainius:
As part of terraform-providers/terraform-provider-vcd#445 there is a small issue in how connection cache is being accessed and this causes some data races during terraform acceptance 
testing (`go test -race -tags ALL -v -timeout=45m .`). 

The essence of this PR is to add thread-safety to connection cache related variables by adding mutexes to them. 

This PR also  improves testing by adding the following:
* Uses `-race` flag when `make test` or `make seqtestacc` are triggered. **Note** - `-race` is not enabled when `make testacc` is run because there is outstanding bug in Terraform SDK (https://github.com/hashicorp/terraform-plugin-sdk/issues/319) which causes data races when `resource.ParallelTest` is used

* Use a binary built with `-race` flag for `make test-binary`


CHANGELOG.md
```diff
@@ -17,11 +17,13 @@ BUG FIXES:
   values after refresh. [GH-433]
 * `resource/vcd_vapp_vm` `network` block changes caused MAC address changes in existing NICs
   [GH-436,GH-407]
+* Fix a potential data race in client connection caching [GH-453]
```
lvirbalas: Did this happen only when the cache was enabled? If yes, we could add a mention about that here, so it's clearer that the scope of effect was limited.

CHANGELOG.md
```diff
@@ -17,11 +17,13 @@ BUG FIXES:
...
 
 
 NOTES:
 * Drop support for vCD 9.0
 * Bump terraform-plugin-sdk to v1.5.0 [GH-442]
+* Testing uses `-race` flags for go test to check if there are no data races [GH-453]
```
lvirbalas: That's really cool.

CHANGELOG.md
Didainius: Yes, indeed. Added.

Didainius:
> Thanks for this. I'd like to see this merged sooner than later, while we still have time in release cycle to use both the fix to caching and, more importantly, the added `-race` check for the tests.

We will not be able to use these flags `-race` flags until https://github.com/hashicorp/terraform-plugin-sdk/issues/319 is fixed.
I am about to remove them now.

CHANGELOG.md
```diff
@@ -25,10 +25,14 @@ BUG FIXES:
   `vapp_template_lease` to define lease related parameters of depending entities - [GH-432]
 * `resource/vcd_vapp_vm` `network` block changes caused MAC address changes in existing NICs
   [GH-436,GH-407]
+* Fix a potential data race in client connection caching when VCD_CACHE is enabled [GH-453]
 
 NOTES:
+
 * Drop support for vCD 9.0
 * Bump terraform-plugin-sdk to v1.5.0 [GH-442]
+* `make seqtestacc` and `make test-binary` use `-race` flags for `go test` to check if there are no data races.
```
lvirbalas: The target name`seqtestacc` sends a bit different message - it looks like the test is a "downgrade" from the usual `testacc` as you don't see from the name that it uses `-race`. Maybe better:

- `testaccrace`
- `testaccraceseq`
- `testaccseqrace`
- `testacc-seq-race`
- `testacc-race-seq` (for me this one is clearest)

To me it looks like it and related functions should start with `test...` to be consistent:

Screenshot 2020-02-17 at 09 28 38

Didainius: Ok, we can change it. The `make seqtestacc` has actually been here for a while now. Before none of these had `-race` flag, but now I added it to `make seqtestacc`. This may be a good opportunity to rename it after all then.

`testacc-race-seq` suits me.

vcd/config.go
```diff
@@ -66,15 +67,20 @@ type cachedConnection struct {
 	connection *VCDClient
 }
 
+type cacheStorage struct {
+	// conMap holds  Cached VDC authenticated connection
```
vbauzysvmware:
```suggestion
	// conMap holds Cached VDC authenticated connection
```

Didainius: It is approved, but for reference - fixed.

Didainius:
Any small issues are now eliminated in go-vcloud-director and terraform-provider-vcd. There is a wider issue in terraform SDK acceptance tests, but this is to be handled by Terraform team.

Issue #427, opened 2020-01-08
VM guest customization force=false crashes with interface {} is nil 


rhysjtevans:
Hi folks, trying to deploy a VM to a vapp and terraform crashes on this line with the error below.

https://github.com/terraform-providers/terraform-provider-vcd/blob/759be139ecfadaf78198876d0392530d50f40391/vcd/resource_vcd_vapp_vm.go#L1343

Here's config

> resource "vcd_vapp_vm" "k8s-masters" {
  count         = var.k8s_master_count
  vapp_name     = var.vapp_name
  name          = "centos-7-test"
  computer_name = "centos-7-test"
  catalog_name  = var.catalog_name
  template_name = var.template_name
  memory        = 4096
  cpus          = 8
  cpu_cores     = 2
customization {
    force = false
  }
network {
    type               = "org"
    name               = var.network_name
    ip_allocation_mode = "POOL"
    is_primary         = true
  }
}
terraform versions: 
Terraform v0.12.18
+ provider.random v2.2.1
+ provider.vcd v2.6.0

> 2020-01-08T22:04:59.617Z [DEBUG] plugin.terraform-provider-vcd_v2.6.0_x4: github.com/terraform-providers/terraform-provider-vcd/vcd.isForcedCustomization(...)
2020-01-08T22:04:59.617Z [DEBUG] plugin.terraform-provider-vcd_v2.6.0_x4: 	/opt/teamcity-agent/work/5d79fe75d4460a2f/src/github.com/terraform-providers/terraform-provider-vcd/vcd/resource_vcd_vapp_vm.go:1343

Any thoughts?

Didainius:
Hi @rhysjtevans, 
Thanks for the issue. Could you try to remove the `customization` block?
```
customization {
  force = false
}
```

Skipping it will be the same as setting to `false`.

rhysjtevans:
Hi @Didainius 

customisation force = True - I get an issue with the vcloud web API (Looking into that separately)
customisation force = False - throws the error
No customization  - applies successfully but "Guest Customisation" is not enabled so the IP address from the IP pool in vcloud isn't applied to the guest machine.

Didainius:
Thanks for the info - curious what errors happens when you try to enable customization.
Also - what is the guest OS you are using? Is it some public image or a private one?

It may be that the guest OS itself doesn't support getting the IPs. For example I tested with Photon OS - even when guest customization is not set - it still gets the POOL allocated IP.

[...]

Issue #303, opened 2019-08-06
Add support for `Change SID`


vtodorovv:
Add support for `Change SID`

Didainius:
Hello, @vtodorovv 
Can you please share more details which resources do you want this to work for and any suggestion for how the HCL configuration should look like?

vtodorovv:
Hey, apologies, would be great if this was available for vcd_vapp_vm. Can probably go with https://github.com/terraform-providers/terraform-provider-vcd/issues/300 ? No suggestion for HCL i'm afraid.

buec011:
Hello,
we would like to see this as well. This is needed for Windows VMs that you want to join to an active directory.

The customization block mentioned in https://github.com/terraform-providers/terraform-provider-vcd/pull/310 looks like a good place for this:

> The customization block was introduced as there is already demand to make other customization features available as the GUI has in "Guest customization" tab.

Didainius:
This should go into the [customization](https://www.terraform.io/docs/providers/vcd/r/vapp_vm.html#customization) block for VM.

Didainius:
@vtodorovv,
I would appreciate if you can test out the code in pull request https://github.com/terraform-providers/terraform-provider-vcd/pull/462

vtodorovv:
Thank you @Didainius

Issue #52, opened 2017-08-28
Guest customization - cannot reset root password


ewypych:
### Terraform Version
v0.10.2

### Provider vCD Version
version = "~> 1.0"

### Affected Resource(s)
Please list the resources as a list, for example:
- vcd_vapp

### Terraform Configuration Files
```hcl
resource "vcd_vapp" "server" {
  name              = "TERRA-TEST"
  catalog_name      = "Linux"
  template_name     = "TEMP-CentOS-7.3"
  memory            = 1024
  cpus              = 1
  network_name      = "${var.vcd_number}_NETWORK_01"
  storage_profile   = "SILVER"
  ip                = "${cidrhost(var.internal_net, 110)}"
}
```

### Expected Behavior
vApp should be deployed and VM should be customized. The template has checked "Enable guest customization" and also "Allow local administrator password", "Auto generate password" and "Require administrator to change the password on first login", so the "Guest OS Customization" tab in the new VM should have the same configuration. Root password should be generated and listed on the tab. 

![image](https://user-images.githubusercontent.com/25039172/29762131-1b510a2e-8bcd-11e7-9c2e-6fd5dc1eb085.png)

### Actual Behavior
The new deployed VM has unchecked mentioned checkboxes (excluding "Enable guest customization") and the root password is not changed.

![image](https://user-images.githubusercontent.com/25039172/29762367-7c322ae8-8bce-11e7-97a1-ee5bf15e7284.png)

### Steps to Reproduction
1. Run terraform apply
2. Check "Guest OS Customization" tab in the newly deployed VM.
3. Check whether "Allow local administrator password", "Auto generate password" and "Require administrator to change the password on first login" check boxes are checked.

### Important Factoids
Running on vCloud Director 8.10 - Linux OS.

carmineesposito73:
any news on that?

lvirbalas:
This needs some investigation @vbauzysvmware 

Few remotely related updates in the meantime:

1. Ability to create VM through the vApp resource will be deprecated - only the proper way of creating vApp + n VM resources will be left.
2. Recently (still in master for 2.5.0 release) the following have been added:
2.b. `guest_properties` for both vApp and VM resources.
2.c.  `force` guest customization flag for VM resource.

vfuyong:
@vbauzysvmware @lvirbalas I can confirm that with 2.5 the issue still exists. In 2.6 release notes, there is no fix yet. 

[...]

PR #456, opened 2020-02-17
FIX VM guest customization force=false crashes with interface {} is nil 

Didainius:
As uncovered in issue  #427 `vcd_vapp_vm` crashes the provider when the `customization` block has only one field and it is being explicitly set to default value `false` as per below code.

```hcl
resource "vcd_vapp_vm" "first" {
...
customization {
    force = false
}
...
}
```

It was possible to use the provider without explicitly specifying this field before, but this needs a fix.

This PR adds a test to cover it and a fix.

Test suite passed on 10.


Didainius:
Closing this pull request because it is not fully approved and there is a coming PR which implements all other Guest customization features - such as #303, #52 and more. It also refactors this code part once again.

PR #469, opened 2020-03-06
Make datasource vcd_vapp_vm test predictable

Didainius:
The test `TestAccVcdVappVmDS` depended on existing VM. This makes some checks hard because VM may be already altered in some way which does not work with defined checks.

This PR introduces VM creation into the test itself.


Didainius:
It takes its time. I added power_on=false on purpose to shave off at least some of this time. Other option could be to move datasource test together with VM test where we spin up the vm anyway.

PR #475, opened 2020-03-12
Fix issue were vCD 9.1 version doesn't have property description in vapp network

vbauzysvmware:
Fix issue were vCD 9.1 version doesn't have property description in vapp network
* Improved documentation
* Improved test


vcd/resource_vcd_vapp_network_test.go
```diff
@@ -68,12 +73,23 @@ func TestAccVcdVappNetwork_Isolated(t *testing.T) {
 		"retainIpMacEnabledForUpdate": "false",
 	}
 
+	// In version 9.1 the property description are not supported
```
Didainius:
```suggestion
	// In version 9.1 the property description is not supported
```

vbauzysvmware: fixed

PR #478, opened 2020-03-18
Fix deprecated 'ip' and 'mac' read for legacy `vcd_vapp_vm` resource

Didainius:
In very rare cases vCD API returns unordered list of network cards. This impacts which IP is reported in deprecated `resource_vcd_vapp_vm.ip` and `resource_vcd_vapp_vm.mac` fields when both `resource_vcd_vapp_vm.network_name` and `resource_vcd_vapp_vm.vapp_network_name` networks are used.

By default the IP should always be returned for `network_name` (org network), but it happens to report IP for `vapp_network_name` when vCD API does not give correct order. It would also occasionally fail the test `TestAccVcdVAppVmSingleNIC`

This PR adds a unit test for a function and puts a fix for it. `make testvm` passed on 9.1, 9.5 and 10. In addition to this, test `TestAccVcdVAppVmSingleNIC` worked fine on my env when run 60+ times.


CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 ## 2.8.0 (Unreleased)
+
+BUG FIXES:
+* `resource/vcd_vapp_vm` sometimes reports incorrect `ip` in legacy network configuration [GH-478]
```
lvirbalas: Please be more specific by "legacy network configuration", for a casual user reading the changelog it raises questions :)

Didainius: Changed

lvirbalas: Thank you. Another small suggestion: consider changing the word "legacy" to "deprecated" for even greater clarity.

PR #491, opened 2020-04-08
Change member port to optional

simdevmat:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/490

This allows users to create members in LB pools without specifying port.

- Made LB pool member port optional

### DIFF
--- a/vcd/resource_vcd_lb_server_pool.go
+++ b/vcd/resource_vcd_lb_server_pool.go
@@ -102,7 +102,7 @@ func resourceVcdLBServerPool() *schema.Resource {
 							Description: "IP address of member in server pool",
 						},
 						"port": {
-							Required:    true,
+							Optional:    true,
 							ForceNew:    false,
 							Type:        schema.TypeInt,
 							Description: "Port at which the member is to receive traffic from the load balancer",
--- a/vcd/resource_vcd_lb_server_pool_test.go
+++ b/vcd/resource_vcd_lb_server_pool_test.go
@@ -99,6 +99,29 @@ func TestAccVcdLbServerPool(t *testing.T) {
 					resource.TestCheckResourceAttr("vcd_lb_server_pool.server-pool", "member.3.weight", "6"),
 					resource.TestCheckResourceAttr("vcd_lb_server_pool.server-pool", "member.3.min_connections", "0"),
 					resource.TestCheckResourceAttr("vcd_lb_server_pool.server-pool", "member.3.max_connections", "0"),
+					
+					// Member 5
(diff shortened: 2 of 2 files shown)


Didainius:
Hello @simdevmat ,
Thanks for the PR. I wonder what is your use case here? In general I see that the `port` field is shown as mandatory (having red asterisk in the screenshot below) while adding a member to server pool although I can see it allows to "apply" without specifying it.

We'd need to add a TestStep to cover such functionality into `TestAccVcdLbServerPool`. That way we are capable of always ensuring it works in all supported versions. Would you be able to do that or do you need some help?


image

simdevmat:
Thanks @Didainius 

We need this as we will be sending traffic with a range of ports and dont want to lock the member to a specific port. 

Its strange, if you don't add a port when you create the member it doesn't complain, and works as we expect. 

Here is the dialog when you edit and existing member. No asterisk
image


Will add some tests for this. I'll try figure out how, otherwise will ask for help.

simdevmat:
@Didainius 

May need some help with the schema validation. It looks like we either need monitor port and/or port. How can we validate at least one of the parameters are present?

Didainius:
> @Didainius
> 
> May need some help with the schema validation. It looks like we either need monitor port and/or port. How can we validate at least one of the parameters are present?

Terraform SDK only allows to validate single field (no combos). Otherwise it should be validated at run-time. `terraform validate` wont work, but there is no other way at this point in time. That is inside `resourceVcdLBServerPoolCreate` and `resourceVcdLBServerPoolUpdate` functions. On the other hand - if the API error returned is nicely formated when an error occurs - you could leave the validation error to happen at run time. It is happening at run-time anyway.

Didainius:
@simdevmat ,
I are you going to put some more commits?
I see the test is failing:

```
go test -tags ALL -v -run TestAccVcdLbServerPool
Connecting to https://192.168.1.160/api
as user administrator@System (using password)
Checking resources to create for test suite...
Skipping catalog creation - found preconfigured one: my-catalog
Catalog found successfully
Skipping catalog item creation - found preconfigured one: photon-os
=== RUN   TestAccVcdLbServerPool
=== PAUSE TestAccVcdLbServerPool
=== CONT  TestAccVcdLbServerPool
    TestAccVcdLbServerPool: testing.go:654: Step 0 error: config is invalid: Missing required argument: The argument "monitor_port" is required, but no definition was found.
--- FAIL: TestAccVcdLbServerPool (0.03s)
FAIL
exit status 1
FAIL    github.com/terraform-providers/terraform-provider-vcd/v2/vcd    2.618s
```

Didainius:
Closing this for time being. Feel free to re-open if there is a need.

PR #492, opened 2020-04-10
Increase API version

vbauzysvmware:
Depends on https://github.com/vmware/go-vcloud-director/pull/300

* Fix test after API changes in govcd


vcd/resource_vcd_edgegateway_test.go
```diff
@@ -729,9 +729,8 @@ resource "vcd_edgegateway" "egw" {
 	configuration           = "compact"
 	advanced                = true
   
-	# FIPS mode is not set because otherwise provider would try to send this field
-	# and vCD 9.0 complains about it, because it was introduced in vCD 9.1 only
-	# fips_mode_enabled               = false
+    # can be only true when system setting Allow FIPS Mode is enabled
+	fips_mode_enabled               = false
```
Didainius, line 733: Great fix - at last :)
Can I ask to also do a few cleanups based on this before we run into the testing parade.
* Remove that comment and description https://github.com/terraform-providers/terraform-provider-vcd/blob/v2.7.0/vcd/resource_vcd_edgegateway.go#L250-L252
* The extra "note" for 9.0 is not needed anymore in https://www.terraform.io/docs/providers/vcd/r/edgegateway.html#fips_mode_enabled
* Remove the warning https://github.com/terraform-providers/terraform-provider-vcd/blob/v2.7.0/vcd/resource_vcd_edgegateway.go#L355-L358

vbauzysvmware: Added changelog

vbauzysvmware: Well, regarding cleanup. My idea was at this moment just increase version if everything is fine - we make general cleanup, I created Explicit task for next version.

Didainius: Ok, let it be then

lvirbalas: @vbauzysvmware Is there a good reason not to do the cleanup now in these PRs? As we decided to release a focused 2.8.0 just to support 10.1 we have time and I believe we should do the cleanup now, so it's not lost in the technical debt bucket.

vbauzysvmware: Yes main reason is to split testing between API versioning changes and technical debt. We decide when todo cleanup in same version or next. I created task to track this.

PR #505, opened 2020-05-10
Add command line flags for tests

dataclouder:
The command line options are an alternative to environment variables.

For example:

```
$ go test -tags functional   -vcd-help
vcd flags:

  -vcd-add-provider                         add provider to test scripts (false)
  -vcd-debug                                enables debug output (false)
  -vcd-help                                 Show vcd flags (true)
  -vcd-remove-org-vdc-from-template         Remove org and VDC from template (false)
  -vcd-short                                runs short test (false)
  -vcd-skip-template-write                  Skip writing templates to file (false)
  -vcd-test-distributed                     enables testing of distributed network (false)
  -vcd-test-org-user                        Run tests with org user (false)
  -vcd-trace                                enables function calls tracing (false)
  -vcd-verbose                              enables verbose output (false)

ok  	github.com/terraform-providers/terraform-provider-vcd/v2/vcd	0.021s
```

The environment variables are still usable.

Using 
`go test -tags functional  -v -timeout 0 -vcd-debug -vcd-test-org-user`
is equivalent to using 
`VCD_TEST_ORG_USER=1 GOVCD_DEBUG=1 go test -tags functional  -v -timeout 0 `

**IMPORTANT**: these flags apply to all testing tags, except `unit`.


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ IMPROVEMENTS:
 
 * `resource/vcd_vapp_vm` allows creating empty VM. New fields added `boot_image`, `os_type` and `hardware_version`. Also, supports `description` updates. [GH-484]
 * Removed code that handled specific cases for API 29.0 and 30.0. This library now supports VCD versions from 9.5 to 10.1 included [GH-499]
+* Added command line flags to test suite, corresponding to environment variables listed in TESTING.md
```
vbauzysvmware, line 7: PR ref  is missing.

TESTING.md
```diff
@@ -468,26 +468,31 @@ To run these tests, you go inside `test-artifacts` and execute:
...
+* `VCD_TEST_ORG_USER=1` (`-vcd-test-org-user`) will enable tests with Org User, using the credentials from the configuration file
   (`testEnvBuild.OrgUser` and `testEnvBuild.OrgUserPassword`)
 * `VCD_TOKEN=string` : specifies the authentication token to use instead of username/password
    (Use `./scripts/get_token.sh` to retrieve one)
-* `VCD_TEST_DISTRIBUTED_NETWORK=1` runs testing of distributed networks (requires the edge gateway to have distributed
+* `VCD_TEST_DISTRIBUTED_NETWORK=1` (`-vcd-test-distributed`) runs testing of distributed networks (requires the edge gateway to have distributed
   routing enabled)
+
+When both the environment variable and the command line option are possible, the environment variable gets evaluated first: is it is set,
+also the flag becomes true.
```
lvirbalas, line 498: Distorted sentence.

vcd/config_test.go
```diff
@@ -832,3 +880,27 @@ func importStateIdVmObject(orgName, vdcName, vappName, vmName, objectIdentifier
 			objectIdentifier, nil
 	}
 }
+
+func setBoolFlag(variable *bool, name, envVar, help string) {
```
lvirbalas, line 884: You know, same as in govcd PR, for me personally this function is confusing, mostly because flag.BoolVar definition in its own is not user friendly. I'd appreciate a comment in the file.

vcd/config_test.go
```diff
@@ -832,3 +880,27 @@ func importStateIdVmObject(orgName, vdcName, vappName, vmName, objectIdentifier
...
+	if envVar != "" && os.Getenv(envVar) != "" {
+		*variable = true
+	}
```
lvirbalas, line 887: In govcd you have a `return` in this place. Here - no return.

dataclouder: The return was wrong in both places. I found out after you suggested checking what happens when we use both env variable and command line flag.
It was removed from govcd as well

vcd/config_test.go
dataclouder: Added comment. Flag operations are not easy to grasp at first.

TESTING.md
dataclouder: Reworked the text

lvirbalas: > When both the environment variable and the command line option are **possible**, the environment variable gets evaluated first.

Probably you wanted to say **used** instead.

dataclouder: No. I meant "possible", because there are environment variable without a corresponding command line flag

dataclouder: When they are both **used**, the environment variable will enable the internal variable, and after that it doesn't matter if the flag is used or not. This is a corner case that should not really matter, because the purpose of this change is to remove the awkwardness of using environment variables on the command line; thus if someone uses the environment variable, there is no point of increasing the amount of written text by also supplying the flag.
If it happens, nothing bad occurs, as the variable is initialized as the user wants

PR #514, opened 2020-05-31
Affinity rules

dataclouder:
Implement resource and data source `vcd_vm_affinity_rule`

Test with 
```
go test -tags functional -timeout 0 -v \
  -run TestAccVcdVmAffinityRule \
  -vcd-verbose \
  -vcd-add-provider
```
The test creates files `test-artifacts/vcd.Test_VmAffinityRule*` that you can edit and run stand-alone as binary tests.


vcd/resource_vcd_vm_affinity_rule.go
```diff
@@ -0,0 +1,387 @@
...
+				Optional:    true,
+				ForceNew:    true,
+				Description: "The name of VDC to use, optional if defined at provider level",
+			},
+			"name": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "VM affinity rule name",
+			},
+			"rule_id": &schema.Schema{
```
vbauzysvmware, line 43: why we need this? as this store Id in `setId`

dataclouder: It's computed. I make it explicit because it's visible in the data source, where it becomes the preferred way of retrieving the entity, although search by name is also supported.

vcd/resource_vcd_vm_affinity_rule.go
```diff
@@ -0,0 +1,387 @@
...
+		vmAffinityRule = ruleList[0]
+		method = "name"
+	} else {
+		vmAffinityRule, err = vdc.GetVmAffinityRuleById(identifier)
+		if method == "[undetected]" {
+			method = "id"
+		}
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving Rule by ID %s (%s)", identifier, method)
+		}
```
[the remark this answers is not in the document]
dataclouder: Good point. Changed to use that function.

vcd/resource_vcd_vm_affinity_rule.go
vbauzysvmware: I leave this to @lvirbalas to decide :)

lvirbalas: @dataclouder Correct me if I'm wrong, but we don't have this approach anywhere else. I don't understand why we need explicit `rule_id` if it's already accessible via the `id` field. From the first glance, looks a duplicate.

dataclouder: We use it in other resources and data sources (e.g. nsxv_firewall_rule, nsxv_dnat), and it helps conveying to the user the idea that the name of the entity could be duplicated and the entity ID is needed. By making it explicit, we reinforce the concept that the

vbauzysvmware: just for info: nsxv_dnat has `rule_id` in datasource, but not in resource.

lvirbalas: > We use it in other resources and data sources (e.g. nsxv_firewall_rule, nsxv_dnat)

@dataclouder Yours is the only resource that has it. Only data sources have it so far:

Screenshot 2020-06-19 at 08 56 58

> and it helps conveying to the user the idea that the name of the entity could be duplicated and the entity ID is needed

Please paste an example use case.

> By making it explicit, we reinforce the concept that the

Sentence ending lost :)

dataclouder: Removed `rule_id` from resource

website/docs/r/vm_affinity_rule.html.markdown
```diff
@@ -0,0 +1,125 @@
...
+  ]
+}
+```
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+* `name` - (Required) The name of VM affinity rule. Duplicates are allowed, although the name can be used to retrieve
```
lvirbalas, line 61: Retrieve? Please rephrase/add a remark what retrieve means in this context (importing?).

website/docs/r/vm_affinity_rule.html.markdown
```diff
@@ -0,0 +1,125 @@
...
+  used to create, modify, and delete VM affinity and anti-affinity rules.
+---
+
+# vcd\_vm\_affinity\_rule
+
+Provides a vCloud Director VM affinity rule resource. This can be
+used to create, modify, and delete VM affinity and anti-affinity rules.
+
+Supported in provider *v2.9+*
+
```
lvirbalas, line 16: We need a note here about what prerequisites VCD should meet for Affinity Rules to work at all. For example, I'm getting this when trying:
```
vcd_vm_affinity_rule.demo_affinity_rule: Creating...

Error: task did not complete successfully:  [500:INTERNAL_SERVER_ERROR] - [ 20415e53-4f4d-4b8b-bf85-44dfab0a380b ] Internal Server Error

  on yorktown.tf line 561, in resource "vcd_vm_affinity_rule" "demo_affinity_rule":
 561: resource "vcd_vm_affinity_rule" "demo_affinity_rule" {
```

vcd/resource_vcd_vm_affinity_rule_test.go
```diff
@@ -0,0 +1,607 @@
...
+		err = vapp.BlockWhileStatus(initialVappStatus, testConfig.Provider.MaxRetryTimeout)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return vapp, nil
+}
+
+// makeEmptyVm creates an empty VM inside a given vApp
+func makeEmptyVm(vapp *govcd.VApp, name string) (*govcd.VM, error) {
```
lvirbalas, line 429: Didn't we have a function to create empty VMs already? `addEmptyVm`?

dataclouder: That is what `makeEmptyVm` is using

website/docs/r/vm_affinity_rule.html.markdown
dataclouder: I doubt that an internal server error is a result of unmet prerequisites.
The only prerequisites are:
* the VMs exist
* they are not used in another affinity rule

website/docs/r/vm_affinity_rule.html.markdown
dataclouder: We are talking about the resource here, but eventually that resource becomes a data source, and we need to `retrieve` it. We have a similar predicament when we import.
The caveat must be here, because the server doesn't enforce uniqueness, and thus users must take care of the names if they want to use them later to find the entity.

dataclouder: Added a note about the above

website/docs/r/vm_affinity_rule.html.markdown
```diff
@@ -0,0 +1,125 @@
...
+}
+```
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+* `name` - (Required) The name of VM affinity rule. Duplicates are allowed, although the name can be used to retrieve
+  the rule only if it was unique.
```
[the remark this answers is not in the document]
dataclouder: fixed

website/docs/r/vm_affinity_rule.html.markdown
```diff
@@ -0,0 +1,125 @@
...
+    data.vcd_vapp_vm.Test_EmptyVm1b.id,
+    data.vcd_vapp_vm.Test_EmptyVm1c.id
+  ]
+}
+```
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
```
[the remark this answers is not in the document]
dataclouder: fixed

website/docs/r/vm_affinity_rule.html.markdown
lvirbalas: Yes, for some reason VMs from that vApp get "refused". Same error in UI. Works with VMs from a different vApp.

website/docs/d/vm_affinity_rule.html.markdown
```diff
@@ -0,0 +1,49 @@
...
+
+* `org` - The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `vdc` - The name of VDC to use, optional if defined at provider level
+* `name` - The name of VM affinity rule. Needed if we don't provide `rule_id`
+* `rule_id` - Is the ID of the affinity rule. It's the preferred way to retrieve the affinity
+rule, especially if the rule name could have duplicates
+ 
+## Attribute reference
+
+* `polarity` - One of `Affinity` or `Anti-Affinity`. This property cannot be changed. Once created, if we
```
lvirbalas, line 43: After more use, I believe we should change this to just bool

`affinity=true|false`

Right now you have to go to the docs to see what values are supported. Also, it's easy to miss the letter case:
```
$ terraform apply
Error: expected polarity to be one of [Affinity Anti-Affinity], got anti-affinity
```

Let's make it simpler to use :)

lvirbalas: I take this back after realizing it's coming from API.

PR #518, opened 2020-06-10
Deprecate resources dnat, snat, firewall_rules

dataclouder:
* Deprecate `vcd_snat` in favor of `vcd_nsxv_snat`
* Deprecate `vcd_dnat` in favor of `vcd_nsxv_dnat`
* Deprecate `vcd_firewall_rules` in favor of `vcd_nsxv_firewall_rule`

### DIFF
--- a/website/docs/r/snat.html.markdown
+++ b/website/docs/r/snat.html.markdown
@@ -11,7 +11,7 @@ description: |-
 Provides a vCloud Director SNAT resource. This can be used to create, modify,
 and delete source NATs to allow vApps to send external traffic.
 
-~> **Note:** This resource may corrupt UI edited NAT rules when used with advanced
+~> **Note:** DEPRECATED: This resource may corrupt UI edited NAT rules when used with advanced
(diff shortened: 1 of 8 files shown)


dataclouder:
> One ask - please add the note about non-advanced EG to the comments in changelog and `provider.go`, so it's clear where old resources are to be used.

Done. Also updated the documentation

website/docs/r/snat.html.markdown
```diff
@@ -11,7 +11,7 @@ description: |-
 Provides a vCloud Director SNAT resource. This can be used to create, modify,
 and delete source NATs to allow vApps to send external traffic.
 
-~> **Note:** This resource may corrupt UI edited NAT rules when used with advanced
+~> **Note:** DEPRECATED: This resource may corrupt UI edited NAT rules when used with advanced
```
vbauzysvmware, line 14: Not sure is rigth place:) now it seems that deprecated behavior currupting UI:)

dataclouder: I think we have piled up the right amount of warnings to users. It's deprecated and we are telling them that it can be dangerous.
We need to point out the danger and we need to inform about deprecation. I believe it is OK as it is

PR #533, opened 2020-07-07
Proof of concept for CHANGELOG assembler

dataclouder:
**NOT FOR MERGING.** 

This is a proof of concept of a method defined in issue #528.

It's the CHANGELOG for 2.9.0, converted to individual files as if they were produced one per PR.

The script `.changes/make-changelog.sh` shows the result of assembling the files into one single CHANGELOG entry.

The final script would also update the CHANGELOG, which we would only do prior to the release.

Feel free to comment.


Didainius:
This actually sounds like it could work. 

Just thinking how to make it more transparent in the project as people tend to peek at CHANGELOG.md (and at least I feel disappointed when I can't find one). One thought - maybe we could put some banner/reference at the top of CHANGELOG.md during development phase.

dataclouder:
Real implementation in PR #534

PR #541, opened 2020-07-30
change allows create VM with vApp org network with user limited rights

vbauzysvmware:
Ref: https://github.com/terraform-providers/terraform-provider-vcd/issues/539

Change allows to create VM with vApp org network with user limited rights. It removes need of fetching Org VDC network.

This change doesn't brake anything, but improves current implementation by removing need of fetching org VDC network.


Tensibai:
Just to record I've tested this version and I'm no longer blocked on the network part.

vbauzysvmware:
This PR will be cancel, when deprecation of old behaviour will be cleaned and one methods will moved into changes.

vbauzysvmware:
Solved in another PR #563

PR #544, opened 2020-08-18
Drop support for VCD 9.5

Didainius:
Drop support for VCD 9.5 and remove conditional code for API v31.0~32.0

### DIFF
--- a/.changes/v2.10.0/544-notes.md
+++ b/.changes/v2.10.0/544-notes.md
@@ -1,0 +1,1 @@
+* Dropped support for vCD 9.5 [GH-544]
--- a/.changes/v2.10.0/544-notes.md
+++ b/.changes/v2.10.0/544-notes.md
@@ -1,1 +1,1 @@
-* Dropped support for vCD 9.5 [GH-544]
+* Dropped support for VCD 9.5 [GH-544]
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/hashicorp/hcl/v2 v2.3.0 // indirect
 	github.com/hashicorp/terraform-config-inspect v0.0.0-20191212124732-c6ae6269b9d7 // indirect
 	github.com/hashicorp/terraform-plugin-sdk v1.8.0
-	github.com/vmware/go-vcloud-director/v2 v2.9.0-alpha.1
+	github.com/vmware/go-vcloud-director/v2 v2.9.0-alpha.2
 	golang.org/x/crypto v0.0.0-20200510223506-06a226fb4e37 // indirect
 )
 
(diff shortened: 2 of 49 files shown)


.changes/v2.10.0/544-notes.md
```diff
@@ -0,0 +1 @@
+* Dropped support for vCD 9.5 [GH-544]
```
lvirbalas, line 1: Let's use the new capitalization:

```suggestion
* Dropped support for VCD 9.5 [GH-544]
```

dataclouder (approved):

vbauzysvmware (approved):
Seems ok. 
Jus two concerns:

```
		// TODO adjust when we have option to switch between API versions or upgrade the default version
		// API does not return GuestVlanAllowed if API client version is 27.0 (default at the moment) therefore we rely
		// on updating statefile only if the field was returned. In API v31.0 - the field is returned.
		if config.GuestVlanAllowed != nil {
```

also 
```
get_token.sh
runtest.sh
index.html
```
has `version=31.0`

Didainius:
> Seems ok.
> Jus two concerns:
> 
> ```
> 		// TODO adjust when we have option to switch between API versions or upgrade the default version
> 		// API does not return GuestVlanAllowed if API client version is 27.0 (default at the moment) therefore we rely
> 		// on updating statefile only if the field was returned. In API v31.0 - the field is returned.
> 		if config.GuestVlanAllowed != nil {
> ```
> 

Fair point. I have removed conditional for this case and doing some testing now. 
> also
> 
> ```
> get_token.sh
> runtest.sh
> index.html
> ```
> 
> has `version=31.0`

Changed

vbauzysvmware:
> > Seems ok.
> > Jus two concerns:
> > ```
> > 		// TODO adjust when we have option to switch between API versions or upgrade the default version
> > 		// API does not return GuestVlanAllowed if API client version is 27.0 (default at the moment) therefore we rely
> > 		// on updating statefile only if the field was returned. In API v31.0 - the field is returned.
> > 		if config.GuestVlanAllowed != nil {
> > ```
> 
> Fair point. I have removed conditional for this case and doing some testing now.
> 
> > also
> > ```
> > get_token.sh
> > runtest.sh
> > index.html
> > ```
> > 
> > 
> > has `version=31.0`
> 
> Changed

Don't see commit regarding GuestVlanAllowed

Didainius:
> Don't see commit regarding GuestVlanAllowed

Still waiting for `make testnetwork` and `make testvapp` to finish before pushing.

Didainius:
> > Don't see commit regarding GuestVlanAllowed
> 
> Still waiting for `make testnetwork` and `make testvapp` to finish before pushing.

They both passed on 9.7 and 10

PR #550, opened 2020-08-27
Add NSX-T support for Org Vdc

Didainius:
This PR plays on top of recent default API version bump to V32.0 and tests out that NSX-T backed Org Vdc can be created.
It also adds a test for NSX-T backed Org Vdc when NSX-T related configuration is specified in config.

**Note** Additional NSX-T related configuration is mandatory to run the test `TestAccVcdOrgVdcNsxt`


dataclouder (approved):
LGTM.
Tests were successful in 10.1 and 10.2

vcd/config_test.go
```diff
@@ -1006,3 +1011,18 @@ func (env *envHelper) restoreVcdVars() {
 		os.Setenv(keyName, valueName)
 	}
 }
+
+// skipNoNsxtConfiguration allows to skip a test if NSX-T configuration is missing
+func skipNoNsxtConfiguration(t *testing.T) {
+	generalMessage := "Missing NSX-T config: "
+	if testConfig.VCD.NsxtProviderVdc.Name == "" {
+		t.Skip(generalMessage + "No provider vdc specified")
```
lvirbalas, line 1019:
```suggestion
		t.Skip(generalMessage + "No provider VDC specified")
```

.changes/v2.10.0/550-features.md
```diff
@@ -0,0 +1 @@
+* Add support for NSX-T Org Vdc [GH-550]
```
lvirbalas, line 1: I think it's better to use capital "VDC" in human-facing messages.
```suggestion
* Add support for NSX-T Org VDC [GH-550]
```

lvirbalas (approved):
Few small comments, LGTM otherwise!

vcd/org_vdc_shared_test.go
```diff
@@ -0,0 +1,335 @@
+// +build vdc nsxt ALL functional
```
vbauzysvmware, line 1: maybe is better to rename file to something like `vdc_util_test.go` `vdc_common_test.go` to avoid opening/read this as normal/running test as others are with same naming convention.

vbauzysvmware (approved):
LGTM
Also if possible in govcd, then here also would be good that not only configuration decides to run test or not.

vcd/config_test.go
Didainius: This was fixed. Didn't comment.

.changes/v2.10.0/550-features.md
Didainius: This was fixed as well. Comment was missing.

vcd/org_vdc_shared_test.go
Didainius: Renamed to `vdc_common_test.go`

PR #567, opened 2020-10-06
Upgrade to terraform 013 and HCL validation 

dataclouder:
Change `make install` and `make test-binary-*` to use terraform binary 0.13.

The updated procedure injects a `terraform` block to every HCL file, and deploys the plugin to path 
`$HOME/.terraform.d/plugins/registry.terraform.io/vmware/vcd/${VERSION}/${OS}_amd64/terraform-provider-vcd_v${VERSION}`

Also, we introduce the command `make test-binary-validate`, which prepares the binary tests, and then validates all HCL files, without executing them. The validation is performed also for files that would normally be skipped during binary tests.

### POSSIBLE BREAKING CHANGE

<s>The binary tests will not work between 2.9.0 and 3.0.0 using terraform 0.13
The reasons are: 

1. the plugin is installed in the wrong directory using 2.9.0
2. the HCL files have no `terraform` block</s>

Added a convoluted workaround that seems to be working


scripts/install-plugin.sh
```diff
@@ -16,6 +16,59 @@ then
...
+check_empty "$terraform_major" "terraform_version major not detected"
+terraform_minor=$(echo $terraform_version | tr '.' ' '| awk '{print $2}')
+check_empty "$terraform_minor" "terraform_version minor not detected"
+os=$(uname -s | tr '[A-Z]' '[a-z]')
+check_empty "$os" "operating system not detected"
+arch=${os}_amd64
+
+
+# if terraform executable is 0.13+, we use the new path
+if [[ $terraform_major -gt 0 || $terraform_major -eq 0 && $terraform_minor > 12 ]]
```
lvirbalas, line 67: In PR you wrote:

> **Change** make install and make test-binary-* to use terraform binary 0.13.

But it looks like it's addition of 0.13 support and 0.12 should still work?

dataclouder: For `make install`, yes. This will allow users to deploy the plugin in the correct directory with both 0.12 and 0.13.

For the test suite (binary tests), this change will force us to use terraform 0.13+

Didainius (approved):
Thanks for fix. It works with 0.12 and 0.13 versions for me (apply and validation).

My only probably concern is that hardcoded `amd64` suffix, but I don't know how to extract such format from OS tools (using shell) and probably there aren't many people using 32 bit systems (especially the ones who try to build it on their own) anyway...

vbauzysvmware (approved):
LGTM

dataclouder:
> My only probably concern is that hardcoded `amd64` suffix, but I don't know how to extract such format from OS tools (using shell) and probably there aren't many people using 32 bit systems (especially the ones who try to build it on their own) anyway...

I suspect this may cause problems once ARM-based Macs become commonplace. I think this arrangement should be enough for a while.

scripts/install-plugin.sh
dataclouder: UPDATE: after discussing with @Didainius, we saw that we can have the test suite work on 0.12 for the time being, at the price of a warning, which would be an incentive to upgrade.

scripts/install-plugin.sh
```diff
@@ -16,6 +16,59 @@ then
...
+then
+    target_dir=$HOME/.terraform.d/plugins/registry.terraform.io/vmware/vcd/$bare_version/$arch
```
lvirbalas, line 69: I'm trying this workflow:
1. Old (0.12) simple configuration with some catalog and catalog item.
2. Replace old terraform 0.12 binary with terraform 0.13 binary in the $PATH
3. Run `terraform init` and upgrade the config from 0.12 to 0.13
```
$ terraform init

Initializing the backend...

Initializing provider plugins...
- Using previously-installed -/vcd v2.9.0
- Finding latest version of hashicorp/vcd...

Error: Failed to install providers

Could not find required providers, but found possible alternatives:

  hashicorp/vcd -> terraform-providers/vcd

If these suggestions look correct, upgrade your configuration with the
following command:
    terraform 0.13upgrade .

$ terraform 0.13upgrade .

This command will update the configuration files in the given directory to use
the new provider source features from Terraform v0.13. It will also highlight
any providers for which the source cannot be detected, and advise how to
proceed.

We recommend using this command in a clean version control work tree, so that
you can easily see the proposed changes as a diff against the latest commit.
If you have uncommited changes already present, we recommend aborting this
command and dealing with them before running this command again.

Would you like to upgrade the module in the current directory?
  Only 'yes' will be accepted to confirm.

  Enter a value: yes

-----------------------------------------------------------------------------

Upgrade complete!

Use your version control system to review the proposed changes, make any
necessary adjustments, and then commit.
```
4. Run `make build` and `make install` from this PR:
```
$ make install
==> Checking that code complies with gofmt requirements...
go install -ldflags="-X 'github.com/vmware/terraform-provider-vcd/v3/vcd.BuildVersion=v1.0.0-438-gbbd4922'"
/Users/l/go/bin/terraform-provider-vcd -> /Users/l/.terraform.d/plugins/registry.terraform.io/vmware/vcd/3.0.0/darwin_amd64/terraform-provider-vcd_v3.0.0
Installed
total 66288
-rwxr-xr-x  1 l  33935556 Spa  7 18:19 terraform-provider-vcd_v3.0.0
```
5. Run `terraform init` with the hope that 3.0.0 will be used:
```
$ terraform init

Initializing the backend...

Initializing provider plugins...
- Using previously-installed terraform-providers/vcd v2.9.0
- Using previously-installed -/vcd v2.9.0

The following providers do not have any version constraints in configuration,
so the latest version was installed.

To prevent automatic upgrades to new major versions that may contain breaking
changes, we recommend adding version constraints in a required_providers block
in your configuration, with the constraint strings suggested below.

* -/vcd: version = "~> 2.9.0"
* terraform-providers/vcd: version = "~> 2.9.0"

Terraform has been successfully initialized!
```

What am I doing wrong? Why is it using published 2.9.0 instead of the built 3.0.0 and how can I fix this?
```

Comment (dataclouder) [reply scripts/install-plugin.sh:69]: The problem is that you need to run `make install` with terraform 0.13 to get the files in the right path. When you run that command with 0.12, the files go into the old path, where 0.13 doesn't find them.

The new `make install` looks at the terraform version and puts the files in the appropriate path.

Moreover, your HCL script needs to have a `terraform` block like this:

```
terraform {
  required_providers {
    vcd = {
      source = "vmware/vcd"
    }
  }
  # optional
  # required_version = ">= 0.13"
}
```

The files generated by our test scripts all have such block. But if you want to run it with your own scripts, you need to add it manually.

Lastly, you may need to add to the provider the property:

```
  version = "~> 3.0"
```

It should not be needed, but it might, depending on which other versions are already cached in your `$PWD/.terraform` directory.

Comment (dataclouder) [reply scripts/install-plugin.sh:69]: As a guideline, here's where `make install` produces when you have terraform 0.13.

```
.terraform.d/
├── checkpoint_signature
└── plugins
    ├── registry.terraform.io
    │   └── vmware
    │       └── vcd
    │           ├── 2.9.0
    │           │   └── darwin_amd64
    │           │       └── terraform-provider-vcd_v2.9.0
    │           └── 3.0.0
    │               └── darwin_amd64
    │                   └── terraform-provider-vcd_v3.0.0
```

If you are using Linux, you'll get a directory named `linux_amd64` instead

lvirbalas: > The problem is that you need to run make install with terraform 0.13 to get the files in the right path. When you run that command with 0.12, the files go into the old path, where 0.13 doesn't find them.

Yes, I actually was doing that.

> Moreover, your HCL script needs to have a terraform block like this:

It was really interesting. As I had run the `terraform 0.13upgrade` command, it did generate the `versions.tf` file with this block, but it contained the **old** 2.9 repo address (**not** in vmware namespace).

I believe we need a section in the docs explaining how to upgrade correctly.

lvirbalas (approved):
Thanks for clarifications in README.md!

PR #569, opened 2020-10-09
Remove redundant terraform block from inline tests

dataclouder:
The `terraform {}` block is needed to run HCL scripts with terraform 0.13.

However, should block was also added to the inline tests, and was failing tests when running with `vcd-add-provider` flag.

The binary tests already added such block to the files that missed it, so the block was redundant in the first place.

`test-binary.sh` was updated to add the block to validation tests as well as binary tests.

### DIFF
--- a/scripts/test-binary.sh
+++ b/scripts/test-binary.sh
@@ -245,6 +245,32 @@ function echo_verbose {
     fi
 }
 
+function check_terraform_version {
+
+    fname=$1
+    dir=$2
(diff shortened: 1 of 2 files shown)


lvirbalas (approved):
LGTM, only nit maybe commit message would be clearer like this:
"Moved redundant terraform block from inline tests to versions.tf"

vbauzysvmware (approved):

Didainius (approved):

PR #576, opened 2020-10-21
Upgrade to Terraform SDK 2.1

dataclouder:
* Change provider version to 3.1.0
* Update imports to SDK 2.1 (https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html)
* Remove `./vendor` directory
* [Import function hashcode.String from Hashicorp](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-helper-hashcode-package)
* [Import object MutexKv from Hashicorp](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-helper-mutexkv-package)
* Change `terraform.ResourceProvider` to `schema.Provider`
* Remove `MaxItems`/`MinItems` from datasource definitions
* Add `depends_on` to data source definitions where the entity is created in the same configuration.

**Prerequisite for testing: having `terraform 0.13+` in `$PATH`**

SDK v2 has many breaking changes.
On the plus side, it exposed several hidden bugs that were fixed while preparing this PR. On the minus side, it has changed some behaviors in a way that is difficult to fix, and sometimes required rewriting the tests.
The most interesting thing is that the inline tests now use the `terraform` binary: this may make part of the binary tests redundant (pending investigation). However, the tests may give different results depending on which version of terraform you are using.

Open issues:
1. CRUD + import functions were deprecated. We keep the existing ones in place, and use the new format in new entities.
2. Validation functions were deprecated, with no ready-made replacement. We are keeping them in place.
3. `GetOkExist` is deprecated without replacement. We should be able to keep them for the time being.


Didainius (approved):
having had the opportunity to work on it myself I am confident this is a step forward although it is not too easy to spot if we could have made any mechanical mistake

vcd/datasource_vcd_vapp_vm.go
```diff
@@ -311,6 +320,12 @@ func datasourceVcdVAppVm() *schema.Resource {
 				Computed:    true,
 				Description: "True if the virtual machine supports addition of memory while powered on.",
 			},
+			"sizing_policy_id": {
```
lvirbalas, line 323: SDK upgrade adds new fields to data source?

vcd/mutexkv.go
```diff
@@ -43,9 +45,9 @@ func (m *MutexKV) get(key string) *sync.Mutex {
 	return mutex
 }
 
-// Returns a properly initalized MutexKV
-func NewMutexKV() *MutexKV {
-	return &MutexKV{
+// Returns a properly initalized mutexKV
+func newMutexKV() *mutexKV {
```
lvirbalas, line 49: Curious, why are we making these functions private?

vcd/org_vdc_common_test.go
```diff
@@ -31,9 +31,9 @@ func runOrgVdcTest(t *testing.T, params StringMap, allocationModel string) {
 	debugPrintf("#[DEBUG] CONFIGURATION: %s", updateText)
 
 	resource.Test(t, resource.TestCase{
-		PreCheck:     func() { testAccPreCheck(t) },
-		Providers:    testAccProviders,
-		CheckDestroy: testAccCheckVdcDestroy,
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviders,
```
lvirbalas, line 35: Java :)

vcd/resource_vcd_lb_app_rule_test.go
```diff
@@ -152,6 +152,7 @@ data "vcd_lb_app_rule" "test" {
   vdc          = "{{.Vdc}}"
   edge_gateway = "{{.EdgeGateway}}"
   name         = vcd_lb_app_rule.test.name
+  depends_on   = [vcd_lb_app_rule.test]
```
lvirbalas, line 155: What is the reason for adding all these `depends_on` which were previously not there?

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1365,8 +1364,12 @@ func genericVcdVAppVmRead(d *schema.ResourceData, meta interface{}, origin strin
...
 	}
 
-	_ = d.Set("hardware_version", vm.VM.VmSpecSection.HardwareVersion.Value)
-	_ = d.Set("os_type", vm.VM.VmSpecSection.OsType)
+	if vm.VM.VmSpecSection != nil && vm.VM.VmSpecSection.HardwareVersion != nil && vm.VM.VmSpecSection.HardwareVersion.Value != "" {
+		_ = d.Set("hardware_version", vm.VM.VmSpecSection.HardwareVersion.Value)
+	}
+	if vm.VM.VmSpecSection != nil && vm.VM.VmSpecSection.OsType != "" {
+		_ = d.Set("os_type", vm.VM.VmSpecSection.OsType)
+	}
```
lvirbalas, line 1372: Not sure I understand why these checks are now needed.

CC: @vbauzysvmware

lvirbalas (commented):
Cool! Just a few questions.

vcd/datasource_vcd_vapp_vm.go
dataclouder: No. This was a leftover. In its original implementation, this field was forgotten, and the enhanced error detection of SDK 2.x created a panic because this one was missing. 
I have added it to make the data source complete.

vcd/mutexkv.go
dataclouder: These functions were in SDK 1.x, but were removed in 2.0.
Hashicorp has released the code in the public domain and I added it to our code base. But since we don't need to export them, I've made them private

vcd/org_vdc_common_test.go
dataclouder: it smells like it, yes

vcd/resource_vcd_lb_app_rule_test.go
dataclouder: All the tests where we have a resource and a data source for the same entity are now failing. We have seen this behavior when we ran the binary tests in 3.0 release. 
Since the tests now run with the `terraform` binary, also the inline tests with the same construct are failing.
We can either re-write these tests to use separate steps, but that would be quite a lot of code, with no immediate benefit.
Since this paradigm (resource + data source in the same configuration) only makes sense in our tests (there is no need for it in a production environment), this is the quickest fix

vcd/resource_vcd_vapp_vm.go
dataclouder: They were needed also before. This particular code was failing when I first implemented the upgrade, and I added the check for safeguard.

vcd/resource_vcd_lb_app_rule_test.go
lvirbalas: It would be helpful to have a comment somewhere in the code explaining this, as it's not very intuitive.

lvirbalas (approved):
Just one ask to add a comment about `depends_on` additions in tests. Overall, looks great!

vcd/resource_vcd_ipset_test.go
```diff
@@ -51,8 +51,8 @@ func TestAccVcdIpSet(t *testing.T) {
 					resource.TestCheckResourceAttr("vcd_nsxv_ip_set.test-ipset", "description", "test-ip-set-description"),
 					resource.TestCheckResourceAttr("vcd_nsxv_ip_set.test-ipset", "is_inheritance_allowed", "true"),
 					resource.TestCheckResourceAttr("vcd_nsxv_ip_set.test-ipset", "ip_addresses.#", "2"),
-					resource.TestCheckResourceAttr("vcd_nsxv_ip_set.test-ipset", "ip_addresses.2555711295", "192.168.1.1"),
-					resource.TestCheckResourceAttr("vcd_nsxv_ip_set.test-ipset", "ip_addresses.2329977041", "192.168.2.1"),
+					resource.TestCheckTypeSetElemAttr("vcd_nsxv_ip_set.test-ipset", "ip_addresses.*", "192.168.1.1"),
+					resource.TestCheckTypeSetElemAttr("vcd_nsxv_ip_set.test-ipset", "ip_addresses.*", "192.168.2.1"),
```
vbauzysvmware, line 55:
```
					resource.TestCheckTypeSetElemAttr("vcd_nsxv_ip_set.test-ipset", "ip_addresses.*", "192.168.1.1"),
					resource.TestCheckTypeSetElemAttr("vcd_nsxv_ip_set.test-ipset", "ip_addresses.*", "192.168.2.1"),
```
 Do we here match any element in Set?

dataclouder:
> Just one ask to add a comment about `depends_on` additions in tests
There are too many occurrences of this issue in the code. I have created issue #586 to take care of this at a later time

vcd/resource_vcd_vapp_network.go
```diff
@@ -92,18 +91,6 @@ func resourceVcdVappNetwork() *schema.Resource {
 				Optional:    true,
 				Description: "org network name to which vapp network is connected",
 			},
-			"firewall_enabled": {
```
vbauzysvmware, line 95: please remind me, why these are removed

vbauzysvmware (commented):
Just few questions. Will make a spin on my home env.

vcd/resource_vcd_vapp_network.go
dataclouder: These properties were marked as "Removed" and should have been deleted in 3.0.
Now the `Removed` field has been itself removed, and we have no choice but to eliminate the property

.travis.yml
```diff
@@ -13,18 +13,10 @@ matrix:
...
-        - make static
-        - make website-test
     - go: "1.15.x"
       env: GO111MODULE=on
       script:
         - make test
         - make vet
         - make static
-        - make vendor-check
-        - make website-test
```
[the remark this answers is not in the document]
dataclouder: Issue #587

vcd/resource_vcd_ipset_test.go
Didainius: The above two tests check that it contains these both addresses (note the "ip_addresses.*" notation)

vcd/resource_vcd_vapp_network.go
vbauzysvmware: ahf. ok - Great ;-)

vcd/resource_vcd_ipset_test.go
vbauzysvmware: ok. So that I guess correctly.

.travis.yml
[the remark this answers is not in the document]
vbauzysvmware: Great!

vbauzysvmware (approved):

khayama:
👍

Issue #546, opened 2020-08-18
vcd_vapp_vm: Update for storage_profile is missing


NilsBusche:
### Terraform Version
Terraform v0.12.29
terraform-provider-vcd_v2.9.0_x4

### Affected Resource(s)
- vcd_vapp_vm

### Terraform Configuration Files
```hcl
resource "vcd_vapp_vm" "instance" {
  vapp_name       = vcd_vapp.vapp.id == "always-not-equal" ? null : var.hostname
  name            = var.hostname
  computer_name   = var.hostname
  
  [...]

  storage_profile = "profile-1"

  [...]

}
```

### Debug Output

```shell
2020-08-18T12:08:22.999+0200 [DEBUG] plugin.terraform-provider-vcd_v2.9.0_x4: 2020/08/18 12:08:22 [DEBUG] [VM update] started with lock
2020-08-18T12:08:22.999+0200 [DEBUG] plugin.terraform-provider-vcd_v2.9.0_x4: 2020/08/18 12:08:22 [DEBUG] Locking "***"
2020-08-18T12:08:22.999+0200 [DEBUG] plugin.terraform-provider-vcd_v2.9.0_x4: 2020/08/18 12:08:22 [DEBUG] Locked "***"
2020-08-18T12:08:22.999+0200 [DEBUG] plugin.terraform-provider-vcd_v2.9.0_x4: 2020/08/18 12:08:22 [DEBUG] [VM update] started without lock
2020-08-18T12:08:22.999+0200 [DEBUG] plugin.terraform-provider-vcd_v2.9.0_x4: 2020/08/18 12:08:22 [DEBUG] [VM update] checking if only field 'network_dhcp_wait_seconds' has change during update
2020-08-18T12:08:22.999+0200 [DEBUG] plugin.terraform-provider-vcd_v2.9.0_x4: 2020/08/18 12:08:22 [DEBUG] [VM update] field 'storage_profile' has change
2020-08-18T12:08:26.317+0200 [DEBUG] plugin.terraform-provider-vcd_v2.9.0_x4: 2020/08/18 12:08:26 [DEBUG] [VM update] finished

```

### Expected Behavior
If you change `storage_profile` this should update the storage profile of your VM in vCloud Director

### Actual Behavior
`terraform apply` succeeds with `Apply complete` but nothing has changed in the real infrastructure. If you run `terraform apply` again, you will see the same behaviour again and again.

[...]

PR #580, opened 2020-10-22
Vm update storage profile

vbauzysvmware:
Ref:https://github.com/vmware/terraform-provider-vcd/issues/546

change allows to update VM storage profile. Before storage profile was used only on create.


vcd/resource_vcd_vapp_vm.go
```diff
@@ -812,6 +812,24 @@ func resourceVmHotUpdate(d *schema.ResourceData, meta interface{}) error {
...
 	}
 
+	if d.HasChange("storage_profile") {
+		var storageProfile types.Reference
+		storageProfilePtr := &storageProfile
+		storageProfileName := d.Get("storage_profile").(string)
+		if storageProfileName != "" {
+			storageProfile, err = vdc.FindStorageProfileReference(storageProfileName)
+			if err != nil {
+				return fmt.Errorf("[vm creation] error retrieving storage profile %s : %s", storageProfileName, err)
```
Didainius, line 822: ?
```suggestion
				return fmt.Errorf("[vm update] error retrieving storage profile %s : %s", storageProfileName, err)
```

vcd/resource_vcd_vapp_vm_hot_updates_test.go
```diff
@@ -152,6 +156,7 @@ func TestAccVcdVAppHotUpdateVm(t *testing.T) {
 
 					resource.TestCheckResourceAttr("vcd_vapp_vm."+hotVmName1, `guest_properties.guest.hostname`, "test-host2"),
 
+					resource.TestCheckResourceAttr("vcd_vapp_vm."+hotVmName1, `storage_profile`, params["StorageProfile2"].(string)),
```
Didainius, line 159: Can we check that VM `ID` is the same after update. Otherwise we are not even sure if VM supports update or it was re-created as part of this step.

Didainius (commented):
No biggie. A few small comments.

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: ty. Fixed

vcd/resource_vcd_vapp_vm_hot_updates_test.go
vbauzysvmware: in 160 line there is function which checks `testAccCheckVcdVmNotRestarted("vcd_vapp_vm."+hotVmName1, hotVappName, hotVmName1),`

vcd/resource_vcd_vapp_vm.go
```diff
@@ -812,6 +812,24 @@ func resourceVmHotUpdate(d *schema.ResourceData, meta interface{}) error {
...
+			storageProfile, err = vdc.FindStorageProfileReference(storageProfileName)
+			if err != nil {
+				return fmt.Errorf("[vm update] error retrieving storage profile %s : %s", storageProfileName, err)
+			}
+		} else {
+			storageProfilePtr = nil
+		}
+		_, err := vm.UpdateStorageProfile(storageProfilePtr)
+		if err != nil {
+			return fmt.Errorf("error updating sizing policy %s: %s", storageProfileName, err)
```
Didainius, line 829:
```suggestion
			return fmt.Errorf("error changing storage profile to %s: %s", storageProfileName, err)
```

vcd/resource_vcd_vapp_vm.go
```diff
@@ -812,6 +812,24 @@ func resourceVmHotUpdate(d *schema.ResourceData, meta interface{}) error {
...
+	if d.HasChange("storage_profile") {
+		var storageProfile types.Reference
+		storageProfilePtr := &storageProfile
+		storageProfileName := d.Get("storage_profile").(string)
+		if storageProfileName != "" {
+			storageProfile, err = vdc.FindStorageProfileReference(storageProfileName)
+			if err != nil {
+				return fmt.Errorf("[vm update] error retrieving storage profile %s : %s", storageProfileName, err)
+			}
+		} else {
```
Didainius, line 824: If this `else` clause needed? 
What happens when `UpdateStorageProfile` receives `nil`? Does it reset to something or leave the same (previous) value.
Could we not skip the `UpdateStorageProfile` at all if new value is empty? Probably by checking in some similar fashion as :
```
_,  newStorageProfile := d.GetChange("storage_profile")
if d.HasChange("storage_profile") && newStorageProfile != "" {
    ..      
} 
```

vcd/resource_vcd_vapp_vm_hot_updates_test.go
```diff
@@ -25,15 +25,17 @@ func TestAccVcdVAppHotUpdateVm(t *testing.T) {
...
+		"Org":             testConfig.VCD.Org,
+		"Vdc":             testConfig.VCD.Vdc,
+		"EdgeGateway":     testConfig.Networking.EdgeGateway,
+		"Catalog":         testSuiteCatalogName,
+		"CatalogItem":     testSuiteCatalogOVAItem,
+		"VAppName":        hotVappName,
+		"VMName":          hotVmName1,
+		"Tags":            "vapp vm",
+		"Media":           testConfig.Media.MediaName,
+		"StorageProfile1": testConfig.VCD.ProviderVdc.StorageProfile1,
```
Didainius, line 37: A bit hard to see in the test, but is there a case where One step has storage profile defined, but the next step removes value at all?

vcd/config_test.go
```diff
@@ -82,9 +82,10 @@ type TestConfig struct {
 		Org         string `json:"org"`
 		Vdc         string `json:"vdc"`
 		ProviderVdc struct {
-			Name           string `json:"name"`
-			NetworkPool    string `json:"networkPool"`
-			StorageProfile string `json:"storageProfile"`
+			Name            string `json:"name"`
+			NetworkPool     string `json:"networkPool"`
+			StorageProfile1 string `json:"storageProfile1"`
```
Didainius, line 87: At least the `sample_govcd_test_config.yaml` needs update. Could we possibly leave names `StorageProfile` and `StorageProfile2` for backwards compatibility of these configs?

Didainius (commented):
A few queries and checks

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: thank you, fixed

vcd/resource_vcd_vapp_vm.go
vbauzysvmware: change a little differently, but yes that explicit else wasn't needed

vcd/resource_vcd_vapp_vm_hot_updates_test.go
vbauzysvmware: no, there isn't such step.

vcd/config_test.go
vbauzysvmware: sample_govcd_test_config.yaml is already updated. I could leave `StorageProfile` but wanted to more explicitly readable with StorageProfile1 and StorageProfile2. @dataclouder what is your choice?

dataclouder: Are they going to be both mandatory? What happens when we have a server with only one?

dataclouder: in the configuration file we have already a `StorageProfile2` under `testEnvBuild`. I would like to avoid confusion

vbauzysvmware: In my test I made mandatory if one is missing, test will be skipped. Previous test working the same.

vbauzysvmware: > in the configuration file we have already a `StorageProfile2` under `testEnvBuild`. I would like to avoid confusion

so what is proposal?

dataclouder: Removing the field from `testEnvBuild` and update the code that uses it to get `StorageProfile2` instead

vbauzysvmware: changed that

vcd/resource_vcd_vapp_vm_hot_updates_test.go
```diff
@@ -24,16 +24,22 @@ func TestAccVcdVAppHotUpdateVm(t *testing.T) {
 		fmt.Println("Warning: `MediaName` is not configured: boot image won't be tested.")
 	}
 
+	if testConfig.VCD.ProviderVdc.StorageProfile1 == "" || testConfig.VCD.ProviderVdc.StorageProfile2 == "" {
+		t.Skip("Variable estConfig.VCD.ProviderVdc.StorageProfile1 and estConfig.VCD.ProviderVdc.StorageProfile2 must be configured")
```
dataclouder, line 28:
```suggestion
		t.Skip("Both variables testConfig.VCD.ProviderVdc.StorageProfile1 and testConfig.VCD.ProviderVdc.StorageProfile2 must be set")
```

vbauzysvmware: oh. Thank you. Fixed

vcd/resource_vcd_vapp_vm_hot_updates_test.go
Didainius: The question is - should we have such step then?

vbauzysvmware: you cant' remove storage profile for VM

vbauzysvmware: also, this field is computed and optional - as if user didn't provide value then API uses default one. As result if remove storage profile, no changes will be done.

vcd/config_test.go
dataclouder: after some missteps in my tests, I have to agree with Dainius. The field `StorageProfile1` should go back to `StorageProfile`

vbauzysvmware: renamed back

lvirbalas (approved):
LGTM.

Didainius (approved):
Thanks. `make testvm` passed

dataclouder (approved):

PR #585, opened 2020-10-27
Adopt github actions

dataclouder:
Intro to Github actions

IMPORTANT notes to reviewers:
1. the result of the checks will not show in the PR page (yet?) but in the *Actions* tab.
2. Before this PR is merged, the actions will only run in my repo. Thus, you won't see any activity here.
3. The main check will fail, due to a bug introduced in master in mid December, and fixed in PR #628


Didainius (approved):

lvirbalas (approved):
Looking forward to seeing "Actions" tab becoming useful :)

PR #608, opened 2020-12-16
3.1 pre-release fixes and cleanups

Didainius:
*  Fixes problems in `make test-upgrade`
Upgrade tests stopped working after Terraform changes (0.13 and 0.14). 
This PR makes it working again by removing dependency on _fake_ home environments and just using users home dir because new Terraform versions are able to hold multiple versions.
It also adds `-upgrade` flag to `terraform init` so that Terraform 0.14+ .lock files are respected.

* Adds missing documentation field for `dedicate_external_network` in `vcd_nsxt_edgegateway`
* Adds missing newlines for changelog
* New Terraform binary versions (0.13.5, 0.14.2)  include `output` variables in a plan diff and test `TestAccVcdEdgeGatewaySettingsFull` intentionally showed different values. Because `output` variables do not support `ignore_changes` - these output values are removed from test as it caused failures in binary test.
```hcl
Changes to Outputs:
  ~ egw = {
        ...
      ~ fw_default_rule_logging_enabled = false -> true
....
      ~ lb_acceleration_enabled         = false -> true
      ~ lb_enabled                      = false -> true
...
    }
```
* Add a bunch of tests to `skip-upgrade-tests.txt`. Reason is because upgrade tests are generated using previous version tag (3.0) and in version 3.0 we had quite a few tests where datasource would depend on a resource created in the same file. This is not supported as of Terraform 0.13+ (and really it makes sense only for testing)
* Fix `vcd_edgegateway_vpn`. As part of [bump to Terraform SDK 2.1](https://github.com/vmware/terraform-provider-vcd/commit/0b67643fd23ff64c6d1daab1ec29b9be99b61cee#) we tried to fix read for local_subnets and peer_subnets blocks. Testing caught an issue and apparently we cannot read these blocks because not all values are returned as they are sent. Now this part is left commented out with [clear example why it can't work](https://github.com/vmware/terraform-provider-vcd/pull/608/files#diff-ab1db943ccb373ff98ebe707ac31cd7a7be7a81c6753005fce0ace391481f205R316). **The resource will continue to work as it worked in 3.0 and older versions.**


lvirbalas (approved):

vcd/resource_vcd_edgegateway_vpn.go
```diff
@@ -10,6 +11,44 @@ import (
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 )
 
+var edgeVpnlocalSubnetResource = &schema.Resource{
```
lvirbalas, line 14:
```suggestion
var edgeVpnLocalSubnetResource = &schema.Resource{
```

vcd/resource_vcd_edgegateway_vpn.go
```diff
@@ -10,6 +11,44 @@ import (
...
+		},
+
+		"local_subnet_mask": &schema.Schema{
+			Type:     schema.TypeString,
+			Required: true,
+		},
+	},
+}
+
+var edgeVpnpeerSubnetResource = &schema.Resource{
```
lvirbalas, line 33:
```suggestion
var edgeVpnPeerSubnetResource = &schema.Resource{
```

lvirbalas (approved):

vcd/resource_vcd_edgegateway_vpn.go
Didainius: Fixed.

vcd/resource_vcd_edgegateway_vpn.go
Didainius: Fixed.

PR #644, opened 2021-02-23
Fix binary test version

dataclouder:
In binary tests, the version within the `provider` section is deprecated and causes a warning.
This PR removes the provider version from the main HCL file and use it in the generated `versions.tf` file.

### DIFF
--- a/vcd/config_test.go
+++ b/vcd/config_test.go
@@ -228,7 +228,7 @@ provider "vcd" {
   vdc                  = "{{.Vdc}}"
   allow_unverified_ssl = "{{.AllowInsecure}}"
   max_retry_timeout    = {{.MaxRetryTimeout}}
-  version              = "~> {{.VersionRequired}}"
+  #version             = "~> {{.VersionRequired}}"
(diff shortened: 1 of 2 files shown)


vcd/config_test.go
```diff
@@ -228,7 +228,7 @@ provider "vcd" {
   vdc                  = "{{.Vdc}}"
   allow_unverified_ssl = "{{.AllowInsecure}}"
   max_retry_timeout    = {{.MaxRetryTimeout}}
-  version              = "~> {{.VersionRequired}}"
+  #version             = "~> {{.VersionRequired}}"
```
lvirbalas, line 231: Do we need to comment the comment?

lvirbalas (approved):
One comment in-line, LGTM otherwise.

vcd/config_test.go
dataclouder: I disabled the line as a directive, but I've left there as a comment to help debugging.

Didainius (approved):

PR #645, opened 2021-02-23
Introduce NSX-T Imported Org VDC network resource and data source.

### LINKED ISSUE #621: Introduce NSX-T  imported Org Vdc network
NSX-V did not have imported network type, but NSX-T has it. This issue is to track implementation work for new resource relating to NSX-T imported Org Vdc network

Didainius:
Closes #621

This Org VDC network is NSX-T specific and allows to use existing NSX-T segment which is already defined (outside of VCD) . More details about - https://docs.vmware.com/en/VMware-Cloud-Director/10.2/VMware-Cloud-Director-Tenant-Portal-Guide/GUID-FB303D62-67EA-4209-BE4D-C3746481BCC8.html

Excerpt from https://fojta.wordpress.com/2020/04/14/vmware-cloud-director-10-1-nsx-t-integration/

> Existing NSX-T overlay logical segment can be imported (same as in VCD 9.7 or 9.5). Its routing/external connectivity must be managed outside of vCloud Director.

It adds resource and datasource with relevant tests and docs.

Additionally - fixes vcd_resource_list datasource to work properly with properties `vcd_network_isolated_v2` and vcd_nsxt_network_imported` and document it.

The following new tests pass. Also running full test suite.
```
--- PASS: TestAccVcdNsxtNetworkImportedDS (134.40s)
--- PASS: TestAccVcdNsxtNetworkImported (86.22s)
--- PASS: TestAccDataSourceNotFound/vcd_nsxt_network_imported (10.40s)
```


dataclouder (approved):

go.sum
```diff
@@ -36,6 +36,8 @@ dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7
 github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
+github.com/Didainius/go-vcloud-director/v2 v2.10.1-0.20210205074351-69c21764466c h1:LDfZUIeCOVpZgMf+mPIgydkfscRwBPWVmeANWlEd/Hc=
+github.com/Didainius/go-vcloud-director/v2 v2.10.1-0.20210205074351-69c21764466c/go.mod h1:czvTQZlB4/WsOsL7rMVCb+SYAPJhx/dYoS/Sk7rc/O0=
```
lvirbalas, line 40: Just in case, it's pointing to Didainius/

website/docs/r/nsxt_network_imported.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+page_title: "VMware Cloud Director: vcd_nsxt_network_imported"
+sidebar_current: "docs-vcd-resource-nsxt-network-imported"
+description: |-
+  Provides a VMware Cloud Director Org VDC imported Network. This can be used to create, modify, and
+  delete imported VDC networks (backed by NSX-T).
+---
+
+# vcd\_nsxt\_network\_imported
+
+Provides a VMware Cloud Director Org VDC imported Network. This can be used to create, modify, and
```
lvirbalas, line 12: I have an ask. I'm looking at this page and would like to ask for a clearer distinction between NSX-T imported type vs. Terraform imported resource. I think it's very easy for new users to get confused when they see "imported" all over the place.

I think in this paragraph we should explain that it is "NSX-T imported type network" and what exactly this means (that it is backed by a segment).

Also, where we say "imported network", maybe we could say "NSX-T imported type network" to be explicit?

website/docs/r/nsxt_network_imported.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+
+~> **Note:** After import the field `nsxt_logical_switch_name` will remain empty because it is
+impossible to read it in API once it is consumed by network.
+
+~> **Note:** The current implementation of Terraform import can only import resources into the state. It does not generate
+configuration. [More information.][docs-import]
+
+
+
+An existing imported network can be [imported][docs-import] into this resource via supplying its path.
```
lvirbalas, line 88: "existing imported network can be [imported]" - I think this maye be too much for a new user :)

Maybe this:
```suggestion
An existing NSX-T imported type network can be [imported][docs-import] into this resource via supplying its path.
```

lvirbalas (changes requested):
Mainly asking to make a distinction in the docs between NSX-T imported network type and Terraform imported resource.

go.sum
Didainius: `go mod tidy` cleaned it up, but no biggie - these only hold checksum, not necessarily actual imports.

website/docs/r/nsxt_network_imported.html.markdown
Didainius: Noted. I have made an attempt to improve this and datasource docs. Tried to differentiate these concepts and popped an additional note with link to official VCD docs. Looks like that in preview now:
image

website/docs/r/nsxt_network_imported.html.markdown
Didainius: Hopefully made it even clearer with:
> An existing NSX-T VDC Imported network can be [imported][docs-import] into this Terraform resource via supplying its path.
The path for this resource is made of orgName.vdcName.networkName.

website/docs/r/nsxt_network_imported.html.markdown
lvirbalas: It's great! Only one ask: please make the added notes blue instead of yellow - blue is enough in this case.

Didainius: Changed in resource and data source
image

lvirbalas (approved):

PR #652, opened 2021-03-07
Conditional running of test suite

dataclouder:
The whole test suite takes several hours to run. If some errors happen during the run, we need to clean up and try again
from the beginning, which is not always convenient.
There are a few tags that help us gain some control on the flow:

* `-vcd-pre-post-checks`    Perform checks before and after tests (false)
* `-vcd-re-run-failed`      Run only tests that failed in a previous run (false)
* `-vcd-remove-test-list`   Remove list of test runs (false)
* `-vcd-show-count`         Show number of pass/fail tests (false)
* `-vcd-show-elapsed-time`  Show elapsed time since the start of the suite in pre and post checks (false)
* `-vcd-show-timestamp`     Show timestamp in pre and post checks (false)
* `-vcd-skip-pattern`       Skip tests that match the pattern (implies vcd-pre-post-checks ()

When `-vcd-pre-post-checks` is used, we have several advantages:

1) After each successful test, the test name gets recorded in a file `vcd_test_pass_list_{VCD_IP}.txt`, and each failed
   test goes to `vcd_test_fail_list_{VCD_IP}.txt`. When running the suite on the same VCD a second time, all tests in
   the `pass` list are skipped. If the test run was interrupted (see #2 below), we can only run the tests that did not
   run in the previous attempt.
2) We can **gracefully** interrupt the tests by creating a file `skip_vcd_tests` in the `./vcd` directory. 
    When this file is found by the pre-run routine, all the tests are skipped. The file `skip_vcd_tests` will be removed 
    automatically at the next run.
3) We can skip one or more tests conditionally, using `-vcd-skip-pattern="{REGEXP}"`. All the test with a name that
   matches the pattern are skipped.
4) We can re-run only the tests that failed in the previous run, using `-vcd-re-run-failed`.
5) We can add monitoring information with `-vcd-show-count`, `-vcd-show-elapsed-time`, `-vcd-show-timestamp`.

If we use `-vcd-pre-post-checks` and the run was successful, the next run will skip all tests, because the test names
would be all found in `vcd_test_pass_list_{VCD_IP}.txt`. To run again the test from scratch, we could either remove
the file manually, or use the tag `-vcd-remove-test-list`.

**VERY IMPORTANT**: for the conditional running to work, each test must have a call to `preTestChecks(t)`  at the beginning
and to `postTestChecks(t)` right before the end.

**Default behavior**: by default, pre and post checks are disabled. The checks are activated only when we use one or more of the flags listed above.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -470,6 +470,39 @@ To run these tests, you go inside `test-artifacts` and execute:
 
 The execution then proceeds as explained in [Binary testing](#Binary-testing).
 
+## Conditional running of tests
+
+The whole test suite takes several hours to run. If some errors happen during the run, we need to clean up and try again
+from the beginning, which is not always convenient.
+There are a few tags that help us gain some control on the flow:
+
+* `-vcd-pre-post-checks`    Perform checks before and after tests (false)
+* `-vcd-re-run-failed`      Run only tests that failed in a previous run (false)
+* `-vcd-remove-test-list`   Remove list of test runs (false)
+* `-vcd-show-count`         Show number of pass/fail tests (false)
+* `-vcd-show-elapsed-time`  Show elapsed time since the start of the suite in pre and post checks (false)
+* `-vcd-show-timestamp`     Show timestamp in pre and post checks (false)
+* `-vcd-skip-pattern`       Skip tests that match the pattern (implies vcd-pre-post-checks ()
+
+When `-vcd-pre-post-checks` is used, we have several advantages:
+
+1) After each successful test, the test name gets recorded in a file `vcd_test_pass_list_{VCD_IP}.txt`, and each failed
+   test goes to `vcd_test_fail_list_{VCD_IP}.txt`. When running the suite on the same VCD a second time, all tests in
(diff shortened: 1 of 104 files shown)


TESTING.md
```diff
@@ -470,6 +471,40 @@ To run these tests, you go inside `test-artifacts` and execute:
 
 The execution then proceeds as explained in [Binary testing](#Binary-testing).
 
+## Conditional running of tests
+
+The whole test suite takes several hours to run. If some errors happen during the run, we need to clean up and try again
+from the beginning, which is not always convenient.
+There are a few tags that help us gain some control on the flow:
+
+* `-vcd-pre-post-checks`    Perform checks before and after tests (false)
```
lvirbalas, line 480: Please add some words and/or formatting describing what kind of "checks" - to emphasize this flag, so it's clear for a new user that this is a global setting for the whole section (if I got it right?).

lvirbalas (approved):
One ask for docs and LGTM otherwise!

Didainius (approved):

Issue #658, opened 2021-03-22
vcd_edgegateway: Entity not found when more than 25 edge gatways exist


CmdrSharp:
### Terraform Version
Terraform v0.13.5

### Affected Resource(s)
- vcd_edgegateway

Unsure whether this could affect other resources as well.

### Terraform Configuration Files
```json
{
    "resource": {
        "vcd_network_routed": {
            "NetworkNameCensored": {
                "org": "OrgNameCensored",
                "vdc": "VdcNameCensored",
                "name": "NetworkNameCensored",
                "gateway": "IPCensored",
                "netmask": "255.255.255.240",
                "shared": true,
                "dns1": null,
                "dns2": null,
                "dns_suffix": null,
                "edge_gateway": "EdgeNameCensored",
                "static_ip_pool": {
                    "start_address": "IPCensored",
                    "end_address": "IPCensored"
                }
            }
        }
    }
}
```

### Debug Output
Leaving the only relevant output here
```
Error: [routed network read] error retrieving edge gateway PBX-SEFRE140-TESG-Arvika-OAM: [ENF] entity not found
```

### Panic Output
Panic only occurs on creation, not on import. On creating, it would produce the following
```
vcd_network_routed.NameCensored: Creating...
Error: rpc error: code = Unavailable desc = transport is closing

panic: edge gateway 'EdgeCensored' not found: unable to find edge gateway: [ENF] entity not found
```

### Expected Behavior
Network should have been created/imported as the edge gateway does indeed exist.

### Actual Behavior
Error or panic resulting in the resource not being created/imported.

### Steps to Reproduce
1. Attempt to work toward a dependant resource that is outside of the 25 list limit of fetched edge gateways.

### Important Factoids
I believe this is a bug in [go-vcloud-director](https://github.com/vmware/go-vcloud-director/issues/363) and have as such submitted an issue there as well. 
I just figure that this should probably be tracked here as well.

### References
- [go-vcloud-director #363](https://github.com/vmware/go-vcloud-director/issues/363)

Didainius:
@CmdrSharp ,
Thanks for this issue - not a fix, but - if you don't use DHCP I'd suggest to use `vcd_network_routed_v2` which shouldn't have such issue.

CmdrSharp:
@Didainius Thanks for the suggestion! How would other existing resources handle a switch to this? Would they require replacing or are the resources compatible somehow?

dataclouder:
Fixed in PR #660

PR #660, opened 2021-03-23
Fix issue 658 - incomplete retrieval of edge gateways

dataclouder:
When more than 25 edge gateways are present, we get
an "Entity not found" error for most edge gateway operations (Issue #658)


lvirbalas (changes requested):
Looks good, but please name the PR **describing** the issue as we usually do.

Didainius (approved):
Thanks! Even if older function is used somewhere we should be good.

dataclouder:
> Looks good, but please name the PR **describing** the issue as we usually do.

Fixed

lvirbalas (approved):
LGTM!

CmdrSharp:
@dataclouder Thanks for the quick fix! Do you know if there's any chance we might see a new release covering this any time soon? We're affected by the bug at the moment so I'm just curious :)

dataclouder:
@CmdrSharp 
I don't know about the next release timeline. Perhaps @Didainius  or @lvirbalas  can comment.

You could, however, build your own plugin  by running this command in the master branch:

```
make install
```

lvirbalas:
We just had 3.2.0 released, next one shall be around mid-Summer.

Issue #633, opened 2021-01-28
vApp 'description' not working


kokkerhout:
### Terraform Version
Terraform v0.14.5

### Affected Resource(s)
- vcd_vapp

### Terraform Configuration Files
```hcl
terraform {
  required_providers {
    vcd = {
      source  = "vmware/vcd"
      version = ">=3.1.0"
    }
  }
  required_version = ">= 0.14.2"
}

resource "vcd_vapp" "vapps" {
  for_each = var.vapps

  name                 = each.key
  org                  = each.value.org
  vdc                  = each.value.vdc_name
  description          = each.value.description

}

```

### Debug Output
https://gist.github.com/kokkerhout/c1987cf94472b913ab59f22a2991b64b

### Expected Behavior
When I define "description" it should add the description in the vApp

### Actual Behavior
The 'description' option in the module 'vcd_vapp' doesn't do anything. So everytime I run terraform apply it tells me it wants to add a description for the vApp because it fails to do it everytime. 

### Steps to Reproduce
1. `Add value for 'description'`
2. `terraform apply`
3. `Message it wants to add description`
3. `apply completes but no description in the dashboard`
4. `terraform apply`
5. `Message it wants to add description`
6. `apply completes but no description in the dashboard`

Didainius:
Hello,
I have tried two things:
* Add a description during initial "create" phase
* Add a description later on (apply an update).

Here is my example HCL (excerpt, because other things don't really matter here):

```hcl

resource "vcd_vapp_org_network" "tf-net" {
  vapp_name         = vcd_vapp.web.name
  org_network_name  = vcd_network_isolated.tf-net.name
  
  depends_on = [
    vcd_vapp.web,
  ]
}

resource "vcd_vapp_vm" "web" {
  count = 5

  vapp_name     = vcd_vapp.web.name
  name          = "web1-${count.index}"
  description = "VM-${count.index}"

  catalog_name  = "cat-dainius"
  template_name = "photon-hw11"

  memory        = 2048
  cpus          = 1

  network {
    type               = "org"
    name               = vcd_vapp_org_network.tf-net.org_network_name
    ip_allocation_mode = "POOL"
  }

  depends_on = [
    # vcd_catalog_item.myNewCatalogItem,
    vcd_vapp.web,
  ]
}
```

Here is the output when I skipped `description` initially, but then added it later (because it inherited it from VM template initially):
```shell
 vcd_vapp_vm.web[0] will be updated in-place
[...]

PR #664, opened 2021-04-28
Add handling of vApp name and description

dataclouder:
<s>* vApp `name` is now updatable. Changing it doesn't force a resource rebuild.</s> (reverted, because of possible side effects on VMs)
* vApp `description` was ignored up to now. It was not added to new vApps or updated when requested. Now it can be inserted and changed.

Solves issue #633

### DIFF
--- a/vcd/resource_vcd_vapp.go
+++ b/vcd/resource_vcd_vapp.go
@@ -152,7 +152,7 @@ func resourceVcdVAppUpdate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("error finding VApp: %#v", err)
 	}
 
-	if d.HasChange("name") || d.HasChange("description") {
+	if d.HasChange("description") {
 		err = vapp.UpdateNameDescription(d.Get("name").(string), d.Get("description").(string))
 		if err != nil {
 			return fmt.Errorf("error updating VApp: %s", err)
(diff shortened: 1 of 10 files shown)


vcd/resource_vcd_vm_affinity_rule_test.go
```diff
@@ -389,7 +389,7 @@ func testAccCheckVmAffinityRuleDestroy(rule *govcd.VmAffinityRule, orgName, vdcN
 // makeEmptyVapp creates a given vApp without any VM
 func makeEmptyVapp(vdc *govcd.Vdc, name string) (*govcd.VApp, error) {
 
-	err := vdc.ComposeRawVApp(name)
+	err := vdc.ComposeRawVApp(name, name)
```
lvirbalas, line 392: Shouldn't the second parameter be `description`?

dataclouder: changed the function signature to include an explicit description

vcd/resource_vcd_vapp.go
```diff
@@ -151,6 +152,12 @@ func resourceVcdVAppUpdate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("error finding VApp: %#v", err)
 	}
 
+	if d.HasChange("name") || d.HasChange("description") {
```
Didainius, line 155: I guess this can be removed just to avoid confusion:
```suggestion
	if d.HasChange("description") {
```

Didainius (approved):
Thanks!

One non impacting comment which could improve readability. Up to you to choose whether it should be fixed.

vcd/resource_vcd_vapp.go
dataclouder: changed

lvirbalas (approved):

Issue #622, opened 2021-01-13
Create new resource for NSX-T firewall management


Didainius:
NSX-T edge gateway firewall requires new resource and this task is to track this work.

raymonddejong:
If this is a new resource a technical requirement would be this resource supports NSX-T Policy API. In Policy API this is just the creation of a Group in which you specify IP(s), IP Set(s) and such. Documentation for the NSX-T provider with examples here: https://registry.terraform.io/providers/vmware/nsxt/latest/docs/resources/policy_group

Didainius:
> If this is a new resource a technical requirement would be this resource supports NSX-T Policy API. In Policy API this is just the creation of a Group in which you specify IP(s), IP Set(s) and such. Documentation for the NSX-T provider with examples here: https://registry.terraform.io/providers/vmware/nsxt/latest/docs/resources/policy_group

We are consuming VCD API which does NSX-T configuration/API calls by itself. I believe it does use Policy API because VCD does have IP Sets, Security groups, and Application Profiles for firewall configurations.

systems-levigo:
Thanks for the merge request!
We also need the implementation for NSX-T and would therefore be very happy about the merge.

Didainius:
> Thanks for the merge request!
> We also need the implementation for NSX-T and would therefore be very happy about the merge.

Work is in progress about it. We already have code for prerequisites ( Security groups, IP sets and Application Port profiles.) (some are in PRs).

PR #668, opened 2021-05-11
Add vcd_nsxt_ip_set support

Didainius:
This PR continues to add NSX-T functionality and introduces resource and datasource `vcd_nsxt_ip_set` for NSX-T IP Set handling as part of #622

IP sets are groups of objects to which the firewall rules apply. Combining multiple objects into IP sets helps reduce the total number of firewall rules to be created.

Additionally:
* It adds a System user check for test `TestAccVcdNsxtStandaloneVmTemplate` and `TestAccVcdNsxtVAppRawAllNsxtNetworks` because they cannot create imported network.
* FIxes a conditional check to skip tests in `TestAccDataSourceNotFound`(it would skip all tests when Org user is used)


vcd/resource_vcd_nsxt_ip_set_test.go
```diff
@@ -0,0 +1,295 @@
...
+  name = "{{.EdgeGw}}"
+}
+`
+
+const testAccNsxtIpSetEmpty = testAccNsxtIpSetPrereqs + `
+resource "vcd_nsxt_ip_set" "set1" {
+  org = "{{.Org}}"
+  vdc = "{{.NsxtVdc}}"
+
+  edge_gateway_id = data.vcd_nsxt_edgegateway.existing.id
```
lvirbalas, line 233: Total nit, but consider instead `existing` to use `existing_gw` just to depict the type when quickly scanning the code.

vcd/resource_vcd_nsxt_standalone_vm_test.go
```diff
@@ -19,6 +19,15 @@ func init() {
...
+		return
+	}
+
+	vcdClient := createTemporaryVCDConnection()
+	if !vcdClient.Client.IsSysAdmin {
+		t.Skip(t.Name() + " only System Administrator can create Imported networks")
+	}
 
 	if testConfig.Nsxt.Vdc == "" || testConfig.Nsxt.EdgeGateway == "" {
 		t.Skip("Either NSXT VDC or edge gateway not defined")
```
lvirbalas, line 33: I think we should stick to "NSX-T" in messages. Personally, I'd also capitalize the Edge Gateway, just so it's poking in the eyes when you are scanning lines of text.

```suggestion
		t.Skip("Either NSX-T VDC or Edge Gateway not defined")
```

website/docs/d/nsxt_ip_set.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+description: |-
+  Provides a datasource to read NSX-T IP Set. IP sets are groups of objects to which the firewall rules apply. 
+  Combining multiple objects into IP sets helps reduce the total number of firewall rules to be created.
+---
+
+# vcd\_nsxt\_ip\_set
+
+Supported in provider *v3.3+* and VCD 10.1+ with NSX-T backed VDCs.
+
+Provides a datasource to read NSX-T IP Set. IP sets are groups of objects to which the firewall rules apply. Combining
```
lvirbalas, line 14: I believe we have been using "data source" (two words) before?

```suggestion
Provides a data source to read NSX-T IP Set. IP sets are groups of objects to which the firewall rules apply. Combining
```

lvirbalas (commented):
Very nice PR, just a few suggestions.

vcd/resource_vcd_nsxt_ip_set_test.go
Didainius: Changed.

vcd/resource_vcd_nsxt_standalone_vm_test.go
Didainius: Changed.

website/docs/d/nsxt_ip_set.html.markdown
Didainius: True. I slipped here. Now I have changed here and in other places I could grep for docs.

lvirbalas (approved):
LGTM!

dataclouder (approved):

PR #670, opened 2021-05-20
Update resource_vcd_org_vdc.go

fatihbm:
When I use metadata, it sometimes fails to create a vapp. Adding a little lag to the metadata task solves the problem.


Didainius:
Hello @fatihbm ,
We tend to avoid putting sleep unless it is impossible otherwise. Which version are you hitting this on?

Didainius:
@fatihbm ,
I am closing this for now. If you have more details about this issue - feel free to share and reopen. But as I mentioned above - we look very carefully and try to find anything bad before putting in a `sleep`

PR #677, opened 2021-06-15
Roles management

dataclouder:
* **New Resource:** `vcd_role` for provider and tenant role management 
* **New Resource:** `vcd_global_role` for provider role management 
* **New Resource:** `vcd_rights_bundle` for provider role management
* **New Data source:** `vcd_right` for provider and tenant role management 
* **New Data source:** `vcd_role` for provider and tenant role management
* **New Data source:** `vcd_global_role` for provider role management 
* **New Data source:** `vcd_rights_bundle` for provider role management

Please see [Roles management](https://github.com/dataclouder/terraform-provider-vcd/blob/roles-management/website/docs/guides/roles_management.html.markdown) for an overview of how the various entities work together.


vcd/resource_vcd_rights_bundle.go
```diff
@@ -0,0 +1,313 @@
...
+}
+
+func genericRightsBundleRead(ctx context.Context, d *schema.ResourceData, meta interface{}, origin, operation string) diag.Diagnostics {
+	vcdClient := meta.(*VCDClient)
+
+	rightsBundleName := d.Get("name").(string)
+
+	rightsBundle, err := vcdClient.Client.GetRightsBundleByName(rightsBundleName)
+	if err != nil {
+		if origin == "resource" {
```
Didainius, line 120: Similar safeguard as in globalRole
```suggestion
		if govcd.ContainsNotFound(err) && origin == "resource" {
```

vcd/resource_vcd_rights_bundle.go
```diff
@@ -0,0 +1,313 @@
...
+
+	publishAll := false
+	if rightsBundle.RightsBundle.PublishAll != nil {
+		publishAll = *rightsBundle.RightsBundle.PublishAll
+	}
+	_ = d.Set("publish_to_all_tenants", publishAll)
+	for _, tenant := range tenants {
+		registeredTenants = append(registeredTenants, tenant.Name)
+	}
+	if !*rightsBundle.RightsBundle.PublishAll {
```
Didainius, line 160: Should this be such to again avoid nil pointer de-reference risk (just as above)
```suggestion
	if !publishAll {
```

vcd/resource_vcd_rights_bundle.go
```diff
@@ -0,0 +1,313 @@
...
+func resourceRightsBundleRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	return genericRightsBundleRead(ctx, d, meta, "resource", "read")
+}
+
+func genericRightsBundleRead(ctx context.Context, d *schema.ResourceData, meta interface{}, origin, operation string) diag.Diagnostics {
+	vcdClient := meta.(*VCDClient)
+
+	rightsBundleName := d.Get("name").(string)
+
+	rightsBundle, err := vcdClient.Client.GetRightsBundleByName(rightsBundleName)
```
Didainius, line 118: Wouldn't it be safer to "GetById" for resource (I did not comment but sounds like it is the same for other "read" functions). Imagine a scenario where someone changed `name` in UI. ID would still be valid, but name - would not. The result:
* GetByName - we report resource is lost and propose to create new one
* GetById - we report `name` has changed

Didainius (changes requested):
Looks very nice and clean. Very good read in "Roles management".
I've got a few queries/concerns.

Didainius (approved):
Thanks for fixed. LGTM

PR #681, opened 2021-06-21
Vapp creation fix

dataclouder:
Replace `vdc.ComposeRawVApp` with `vdc.CreateRawVApp`


Didainius (approved):
Thanks - I had a few test errors about incorrect storage profiles data before `master` merge. Now these tests pass as it should have included my fix coming together with NAT rules

Issue #584, opened 2020-10-26
Update Org-User docs to include system users


dataclouder:
[The documentation for `vcd_org_user`](https://registry.terraform.io/providers/vmware/vcd/latest/docs/resources/org_user) does not specify that we can just as well create system user. 
We can, indeed, create such users. The only limitation is that we have the right privileges to do so.
The current resource allows us to create system users in the `System` org, using the roles available for such organization.

For example

```hcl
resource "vcd_org_user" "test_user_admin" {
  org            = "System"
  name           = "another_user_admin"
  password       = "change-me-soon"
  #role           = "Multisite System"
  role           = "System Administrator"
  take_ownership = true
```

Related to Issue #564 

CC @lvirbalas

dataclouder:
Added in PR #682

PR #682, opened 2021-06-26
Rebrand docs

dataclouder:
* Change "vCloud Director" or "vCloudDirector" to "VMware Cloud Director" in all docs.
* Add 10.3 to supported VCD versions.
* Address Issue #584 (Update Org-User docs to include system users)


Didainius (approved):

website/docs/r/org_user.html.markdown
```diff
@@ -48,11 +52,52 @@ resource "vcd_org_user" "test_user_vapp_author" {
...
+```
+
+
 ## Argument Reference
 
 The following arguments are supported:
 
-* `org` - (Optional) The name of organization to which the user belongs. Optional if defined at provider level.
+* `org` - (Optional) The name of organization to which the user belongs. Optional if defined at provider level. If we 
+  want to create an user at provider level, use use "System" as org name.
```
lvirbalas, line 100:
```suggestion
  want to create a user at provider level, use "System" as org name.
```

lvirbalas (changes requested):
Very nice, just one fix needed.

lvirbalas (approved):

PR #690, opened 2021-07-02
Add examples folder

dataclouder:
Adds a folder with examples from recent releases


Didainius (approved):

lvirbalas (changes requested):
Let's only use the 2nd folder type (i.e. let's not show versions):

> If a set of features is spread across several release, the directory will be called feature_set_name, and the files inside will be called x.x-feature_name.tf

We don't need to show version numbers here, especially given that we can't be consistent with the folder naming and some of the examples may span multiple versions.

In other words, let's make this folder simpler, without history. it's OK to look the supported versions in the docs. We also want our users to use the latest version.

dataclouder:
> Let's only use the 2nd folder type (i.e. let's not show versions):
> [...]
> In other words, let's make this folder simpler, without history. it's OK to look the supported versions in the docs. We also want our users to use the latest version.

Folder renamed.

lvirbalas (approved):
Nice!

Issue #696, opened 2021-07-19
Catalog deletion failure returns as success


dhruvpatelaj:
Hi there,

The scenario where the bug occurs is when I create a vcd_catalog and upload a local ovf file using vcd_catalog_item. Lets say for whatever reason, kernel dies or you manually (ctrl + c). The file is still uploading. Currently my script calls `terraform destroy` before any actions to remove any existing code. 

So if the file is still uploading, the terraform destroy call will try to delete the catalog. VCD will fail in deleting that catalog as there is a pending template upload, but terraform will return it as deleted. So, now the catalog still exists in VCD, but it no longer exists in terraform .tfstate. Therefore, we would have to manually go and delete the catalog.

The only way to work around this is to let the file upload finish. Should terraform not fail to delete the catalog too?

### Terraform Version
Terraform v1.0.2

### Affected Resource(s)
Please list the resources as a list, for example:
- vcd_catalog
- vcd_catalog_item

### Expected Behavior
Terraform should have returned `failed to delete catalog`

### Actual Behavior
Terraform succeeded in deleting catalog, but in actuality, VCD denied the request as there was a pending ovf file uploading within the catalog

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. wait until OVF file starts uploading. Basically until after the catalog is created.
2. `ctrl + c`
3. `terraform destroy` before OVF file has finished uploading

dataclouder:
Thanks for this clear report.
I will try to reproduce. The described behavior should not happen.

dhruvpatelaj:
Were you able to replicate?

dataclouder:
Reproduced by using this sequence of events:
1. add a catalog
2. add a catalog item
3. run `terraform state rm vcd_catalog_item.$catalog_item_name`
4. run `terraform destroy`

The destruction operation succeeds, but the catalog (and its item) remain in place.

dataclouder:
The behavior is fixed in PR #698.
* Starting the `destroy` while the catalog item is still being imported, the catalog deletion now fails because a catalog object is busy with another operation (just like in the web interface);
* Starting the `destroy` when the catalog item has been fully imported, and both `delete_force` and `delete_recursive` were set, the deletion will succeed

Issue #648, opened 2021-03-01
vcd_org_vdc: adding a storage profile requires the vdc to be replaced


carmine73:
### Terraform Version
```
Terraform v0.14.7
+ provider registry.terraform.io/vmware/vcd v3.1.0
```
### Affected Resource(s)
- vcd_org_vdc 

### Terraform Configuration Files
I have input values like the following:
```
input = {
  [. . .]
  sdc_storage           = [
    {
      type      = "fc"
      limit     = 102400  # MB, 0 means unlimited
      default   = true
    },
  ]
  [. . .]
}
```
The org vdc has a dynamic block for storage profiles:
```
## org-vdc
resource "vcd_org_vdc" "vdc1" {
  name              = local.vcd_org_vdc.name
  description       = local.vcd_org_vdc.descr
  org               = vcd_org.org1.name

  allocation_model  = local.vcd_org_vdc.alloc
  cpu_speed         = local.vcd_org_vdc.cpu_speed
  cpu_guaranteed    = local.vcd_org_vdc.cpu_guaranteed
  memory_guaranteed = local.vcd_org_vdc.mem_guaranteed
  network_pool_name = local.vcd_org_vdc.net_pool
  provider_vdc_name = local.vcd_org_vdc.pvdc

  compute_capacity {
    cpu {
      allocated = local.vcd_org_vdc.cpu_alloc
      limit     = local.vcd_org_vdc.cpu_limit
    }

    memory {
      allocated = local.vcd_org_vdc.mem_alloc
      limit     = local.vcd_org_vdc.mem_limit
    }
  }

  # dynamic storage profile (they could be more than one)
  dynamic "storage_profile" {
    for_each = [ for storage in local.vcd_org_vdc.storage : {
      name    = storage.name
      limit   = storage.limit
      default = storage.default
    } ]

    content {
      name    = storage_profile.value.name
      limit   = storage_profile.value.limit
      default = storage_profile.value.default
    }
  }

  network_quota = 10

  enabled                  = local.vcd_org_vdc.enabled
  enable_thin_provisioning = false
  enable_fast_provisioning = true
  delete_force             = true
  delete_recursive         = true
}
```
### Steps to Reproduce
1. `terraform plan`
2. `terraform apply`
3. Change the value of the input to have:
```
input = {
  [. . .]
  sdc_storage           = [
    {
      type      = "fc"
      limit     = 102400  # MB, 0 means unlimited
      default   = true
    },
    {
[...]

PR #698, opened 2021-07-28
Fix storage profile update

dataclouder:
* Fix Issue #648 
* Fix Issue #696 

In addition to the problem identified in Issue 648, this PR resolves the problem where storage profiles could be chosen from a Provider VDC different from the one that was chosen as parent for the Org VDC.


vcd/resource_vcd_org_vdc.go
```diff
@@ -571,41 +572,162 @@ func resourceVcdVdcUpdate(d *schema.ResourceData, meta interface{}) error {
...
+		}
+	}
+
+	// 7. Delete unwanted storage profiles
+	for _, spCombo := range removeStorageProfiles {
+		err := adminVdc.RemoveStorageProfileWait(spCombo.reference.Name)
+		if err != nil {
+			return fmt.Errorf("error removing storage profile %s: %s", spCombo.reference.Name, err)
+		}
+	}
```
vbauzysvmware, line 728: Have you checked how we handle case when storage profile is already use, for e.g. by VM and you try to update VDC by removing it from the list assigned to VDC?

vbauzysvmware (approved):

.changes/v3.4.0/698-bug-fixes.md
```diff
@@ -0,0 +1,2 @@
+* Fix Issue #696: Catalog deletion failure returns as success [GH-698]
+* Fix Issue #648 `vcd_org_vdc`: adding a storage profile requires the vdc to be replaced [GH-698]
```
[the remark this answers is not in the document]
dataclouder: The number (GH-698) there should be the PR where the issue is solved

vcd/resource_vcd_org_vdc.go
dataclouder: We don't. Here's a quote from my own comment in the issue:

> this could be a problem if the storage profile was used for any object that belongs to the VDC. My plan is - rather than over-engineering and figure out if removal is possible - to attempt removal, and let the API complain if the storage profile cannot be removed.

vcd/resource_vcd_org_vdc.go
```diff
@@ -571,41 +572,162 @@ func resourceVcdVdcUpdate(d *schema.ResourceData, meta interface{}) error {
...
+	return resourceVcdVdcRead(d, meta)
+}
+
+func updateStorageProfileDetails(vcdClient *VCDClient, adminVdc *govcd.AdminVdc, storageProfile *types.Reference, storageConfiguration map[string]interface{}) error {
+	util.Logger.Printf("updating storage profile %#v", storageProfile)
+	uuid, err := govcd.GetUuidFromHref(storageProfile.HREF, true)
+	if err != nil {
+		return fmt.Errorf("error parsing VDC storage profile ID : %s", err)
+	}
+	vdcStorageProfileDetails, err := vcdClient.GetStorageProfileByHref(storageProfile.HREF)
```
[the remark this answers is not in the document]
dataclouder: It is not a new one. It's the same storage profile, but with more information

vcd/resource_vcd_org_vdc.go
vbauzysvmware: That's should be ok, if API complains with reasonable message ;-) that happens not every time.

vcd/resource_vcd_org_vdc.go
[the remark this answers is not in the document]
vbauzysvmware: I got wrong idea, I thought this update changes storageProfile ref to passed one.

.changes/v3.4.0/698-bug-fixes.md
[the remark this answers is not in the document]
vbauzysvmware: got it ;-)

vcd/resource_vcd_org_vdc_nsxt_test.go
```diff
@@ -14,19 +14,20 @@ func TestAccVcdOrgVdcNsxt(t *testing.T) {
...
+		"AllocationModel":            "ReservationPool",
+		"ProviderVdc":                testConfig.VCD.NsxtProviderVdc.Name,
+		"NetworkPool":                testConfig.VCD.NsxtProviderVdc.NetworkPool,
+		"Allocated":                  "1024",
+		"Reserved":                   "1024",
+		"Limit":                      "1024",
+		"LimitIncreased":             "1100",
+		"AllocatedIncreased":         "1100",
+		"ProviderVdcStorageProfile":  testConfig.VCD.NsxtProviderVdc.StorageProfile,
+		"ProviderVdcStorageProfile2": "", // The configuration file is missing the second storage profile for NSXT.
```
lvirbalas, line 28:
```suggestion
		"ProviderVdcStorageProfile2": "", // The configuration file is missing the second storage profile for NSX-T
```

vcd/org_vdc_common_test.go
```diff
@@ -333,3 +387,14 @@ resource "vcd_org_vdc" "{{.VdcName}}" {
   {{.FlexMemoryOverheadKey}} {{.equalsChar}} {{.FlexMemoryOverheadValueUpdate}}
 }
 `
+
+const simpleStorageProfile = `
+  #START_STORAGE_PROFILE
```
lvirbalas, line 392: We haven't used this approach before, right? This needs clarification (comment, better const name) or even better simplification. How did we do updates in other tests? I don't recall such string replacements.

dataclouder: It's a new approach for a new situation.
Using this, I can add a storage profile (but only when we have two of them defined in the configuration), and later on replace the labels with comment markers, to effectively remove the added item.

lvirbalas (approved):
LGTM!

Didainius (approved):

PR #708, opened 2021-09-03
Prevent invalid space in the base64 encoded authentication string in the token scripts specific for Linux

megian:
If the auth string "$user@$org:$password" gets above 76
characters, the GNU coreutils base64 utility on Linux
adds a space into the string, causing invalid
authentication string.

Because different base64 implementation do have different parameters
there is no easy way to dinguish between.

Explain script:

user=long
password=thisisaverylongpasswordandcausesbase64startaddaspaceafteradefaultof76
org=myorg
IP=myip

echo;echo "Causes an empty space in the variable"
echo "$user@$org:$password"
auth=$(echo -n "$user@$org:$password" |base64)
echo $auth

echo;echo "Base64 -w 0 (on Linux) fixes it"
options=""
os=$(uname -s)
is_linux=$(echo "$os" | grep -i linux)
if [ -n "$is_linux" ]
then
  options="-w 0"
fi
echo "$user@$org:$password"
auth=$(echo -n "$user@$org:$password" |base64 $options)
echo $auth

echo;echo "OpenSSL base64 fixes it (on probably MacOS and Linux)"
echo "$user@$org:$password"
auth=$(echo -n "$user@$org:$password" | openssl base64 -A)
echo $auth

Signed-off-by: Gabriel Mainberger <gabriel.mainberger@vshn.net>

### DIFF
--- a/.changes/v3.4.0/708-improvements.md
+++ b/.changes/v3.4.0/708-improvements.md
@@ -1,0 +1,1 @@
+* Prevent invalid space in the base64 encoded authentication string in the token scripts specific for Linux [GH-708]
--- a/scripts/get_token.sh
+++ b/scripts/get_token.sh
@@ -13,7 +13,15 @@ then
     exit 1
 fi
 
-auth=$(echo -n "$user@$org:$password" |base64)
+options=""
+os=$(uname -s)
+is_linux=$(echo "$os" | grep -i linux)
+if [ -n "$is_linux" ]
+then
+  options="-w 0"
+fi
+
+auth=$(echo -n "$user@$org:$password" |base64 $options)
 
 curl -I -k --header "Accept: application/*;version=32.0" \
     --header "Authorization: Basic $auth" \
--- a/scripts/runtest.sh
+++ b/scripts/runtest.sh
@@ -258,7 +258,15 @@ function make_token {
     echo "missing url from configuration file. Can't retrieve token"
     exit 1
   fi
-  auth=$(echo -n "$user@$sysorg:$password" |base64)
+
+  options=""
+  os=$(uname -s)
+  is_linux=$(echo "$os" | grep -i linux)
+  if [ -n "$is_linux" ]
+  then
+    options="-w 0"
+  fi
+  auth=$(echo -n "$user@$sysorg:$password" |base64 $options)
 
   echo "# Connecting to $url ($sysorg)"
   curl --silent --head --insecure \
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -173,7 +173,15 @@ then
     exit 1
 fi
 
-auth=$(echo -n "$user@$org:$password" | base64)
+options=""
+os=$(uname -s)
+is_linux=$(echo "$os" | grep -i linux)
+if [ -n "$is_linux" ]
+then
+  options="-w 0"
+fi
+
+auth=$(echo -n "$user@$org:$password" |base64 $options)
 
 curl -I -k --header "Accept: application/*;version=32.0" \
     --header "Authorization: Basic $auth" \


Didainius:
Hello @megian ,
First question? Are you using Linux?
I have quickly compared `base64` utility manual in MacOS and in Linux and I see the difference:

Linux
```
~$ echo "TEST" | base64 -w0
VEVTVAo=
```

MacOS
```
~$ echo "TEST" | base64 -w0
base64: invalid option -- w
```


We'd need to think of making this script compatible with macOS as many people use it that way.
As per manual it looks that Linux does this, while macOS does not:

Linux excerpt
```
       -w, --wrap=COLS
              wrap encoded lines after COLS character (default 76).  Use 0 to disable line wrapping
```

MacOS excerpt
```
     -b count
     --break=count        Insert line breaks every count characters.  Default is 0, which generates an unbro-
                          ken stream.
```

megian:
@Didainius Yes, I do. Pipelines doing terraform plan/apply are as well Linux based.

Didainius:
> @Didainius Yes, I do. Pipelines doing terraform plan/apply are as well Linux based.

Ok, but we'd need a solution compatible with both OS'es

megian:
Probably using `openssl` instead?

```
user=long
password=thisisaverylongpasswordandcausesbase64startaddaspaceafteradefaultof76
org=myorg
IP=myip

echo;echo "Causes an empty space in the variable"
echo "$user@$org:$password"
auth=$(echo -n "$user@$org:$password" |base64)
echo $auth

echo;echo "Base64 -w 0 fixes it"
echo "$user@$org:$password"
auth=$(echo -n "$user@$org:$password" | base64 -w 0)
echo $auth

echo;echo "OpenSSL enc base64 fixes it"
echo "$user@$org:$password"
auth=$(echo -n "$user@$org:$password" | openssl enc -A -base64)
echo $auth
```

dataclouder:
I think an easier fix is by checking the OS
```
user=long
password=thisisaverylongpasswordandcausesbase64startaddaspaceafteradefaultof76
org=myorg
IP=myip

options=""
os=$(uname -s)
is_linux=$(echo "$os" | grep -i linux)
if [ -n "$is_linux" ]
then
  options="-w 0"
fi
echo "$user@$org:$password"
auth=$(echo -n "$user@$org:$password" |base64 $options)
echo "<$auth>"
```

megian:
Wouldn't say easier, but at least with less dependencies. Fully agree that OpenSSL is not to be expected available on any system or pipeline.

```
user=long
password=thisisaverylongpasswordandcausesbase64startaddaspaceafteradefaultof76
org=myorg
IP=myip

echo;echo "Causes an empty space in the variable"
echo "$user@$org:$password"
auth=$(echo -n "$user@$org:$password" |base64)
echo $auth

echo;echo "Base64 -w 0 (on Linux) fixes it"
options=""
os=$(uname -s)
is_linux=$(echo "$os" | grep -i linux)
if [ -n "$is_linux" ]
then
  options="-w 0"
fi
echo "$user@$org:$password"
auth=$(echo -n "$user@$org:$password" |base64 $options)
echo $auth

echo;echo "OpenSSL base64 fixes it (on probably MacOS and Linux)"
echo "$user@$org:$password"
auth=$(echo -n "$user@$org:$password" | openssl base64 -A)
echo $auth
```

megian:
Updated the PR so it should match your expectation.

megian:
How about checking for the GNU base64 instead of Linux
```
user=long
password=thisisaverylongpasswordandcausesbase64startaddaspaceafteradefaultof76
org=myorg
IP=myip

echo;echo "Base64 -w 0 (for GNU base64) fixes it"
options=""
is_gnu_base64=$(base64 --version | grep -i GNU)
if [ -n "$is_gnu_base64" ]
then
  options="-w 0"
fi
echo "$user@$org:$password"
auth=$(echo -n "$user@$org:$password" |base64 $options)
echo $auth
```

Would make it more robust on systems using the GNU version as well.

dataclouder (approved):

dataclouder:
> How about checking for the GNU base64 instead of Linux

Multi-platform scripts are hard to get right

```
$ uname -s
Darwin
$ base64 --version
base64: unrecognized option `--version'
```

megian:
> Multi-platform scripts are hard to get right

Argh. OK then checking for Linux is probably better.

Didainius:
One more ask - if you can add a changelog (to .changes/v3.4.0/708-improvements.md).

Thanks!

lvirbalas (approved):
Thanks for contributing! Just need the changelog entry as per @Didainius comment.

megian:
Added the CHANGELOG entry in the same commit.

.changes/v3.4.0/708-improvements.md
```diff
@@ -0,0 +1 @@
+* Remove invalid space in the authentication string on Linux [GH-708]
```
Didainius, line 1: Don't want to be picky, but could you just be a bit more specific this note that it only affects token scripts and not the provider codebase itself. Just want to be transparent to whoever reads release notes.

ThanksĄ

megian: Is this more clear now to you?

Didainius: Yep, thanks!

Didainius (approved):

Didainius:
Thanks - that is now merged to master.

PR #711, opened 2021-09-15
Add support for NSX-T segment backed networks in 10.3

Didainius:
VCD 10.3 introduces new type of NSX-T External network - Segment backed external network. This PR adds support for this feature and adds related tests.

Additionally there is some fine tuning done for upgrade tests due to Go 1.17 and last version having non 0  patch version (3.3.1).

### DIFF
--- a/vcd/resource_vcd_external_network_v2.go
+++ b/vcd/resource_vcd_external_network_v2.go
@@ -178,6 +178,11 @@ func resourceVcdExternalNetworkV2Create(d *schema.ResourceData, meta interface{}
 	vcdClient := meta.(*VCDClient)
 	log.Printf("[TRACE] external network V2 creation initiated")
 
+	err := externalNetworkV2ValidateApiVersionFeatures(d, vcdClient)
+	if err != nil {
+		return err
+	}
+
 	netType, err := getExternalNetworkV2Type(vcdClient, d, "")
 	if err != nil {
 		return fmt.Errorf("could not get network data: %s", err)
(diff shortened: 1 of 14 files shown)


vcd/resource_vcd_external_network_v2.go
```diff
@@ -152,12 +162,12 @@ func resourceVcdExternalNetworkV2() *schema.Resource {
...
 			},
 			"nsxt_network": &schema.Schema{
-				Type:         schema.TypeSet,
+				Type:         schema.TypeList,
 				Optional:     true,
 				ExactlyOneOf: []string{"vsphere_network", "nsxt_network"},
 				MaxItems:     1,
 				ForceNew:     true,
-				Description:  "Reference to NSX-T Tier-0 router and manager",
+				Description:  "Reference to NSX-T Tier-0 router or Segment and manager",
```
lvirbalas, line 170: Nit: does "Segment" deserve being capitalized? Alternatively, should "Router" and "Manager" be too (though then it becomes a question for all fields' descriptions :) )?

website/docs/r/external_network_v2.html.markdown
```diff
@@ -20,7 +20,7 @@ requires at least VCD *10.0+*. It supports both NSX-T and NSX-V backed networks
 
 Supported in provider *v3.0+*.
 
-## Example Usage (NSX-T backed external network)
+## Example Usage (NSX-T Tier 0 router backed external network)
```
lvirbalas, line 23: In headings it's easier to read if we capitalize the keywords:
```suggestion
## Example Usage (NSX-T Tier 0 Router backed External Network)
```

website/docs/r/external_network_v2.html.markdown
```diff
@@ -70,6 +70,60 @@ resource "vcd_external_network_v2" "ext-net-nsxt" {
...
+  name = "nsxManager"
+}
+
+resource "vcd_external_network_v2" "ext-net-nsxt-segment" {
+  name        = "nsxt-external-network"
+  description = "First NSX-T segment backed network"
+
+  nsxt_network {
+    nsxt_manager_id   = data.vcd_nsxt_manager.main.id
+    nsxt_segment_name = "existing-nsxt-segment"
```
lvirbalas, line 86: Maybe let's add a comment above this line explaining more about this field. I think it would be easier for the user  to notice the difference from the example above when just scanning the examples, as they are very similar.

vcd/resource_vcd_external_network_v2.go
Didainius: Fixed. Made it lower case here.

website/docs/r/external_network_v2.html.markdown
Didainius: Adjusted, It is upper case now.

website/docs/r/external_network_v2.html.markdown
Didainius: Agreed. I have put a note right in this section. Here is how it looks in docs:

image

lvirbalas (approved):
LGTM!

dataclouder (approved):

PR #714, opened 2021-09-16
Send Media reference alongside CreateVM (fix boot image media connection)

thiagoarrais:
I am hitting a problem with standalone VM creation (via `resource "vcd_vm"`) where the boot image media won't get connected.

This PR fixes it by sending the boot image media reference directly inside the `CreateVmParams` root element. This was tested against VMware Cloud Director version 10.3.0.18295834

### DIFF
--- a/vcd/resource_vcd_vapp_vm.go
+++ b/vcd/resource_vcd_vapp_vm.go
@@ -2509,8 +2509,8 @@ func addEmptyVm(d *schema.ResourceData, vcdClient *VCDClient, org *govcd.Org, vd
 				NetworkConnectionSection:  recomposeVAppParamsForEmptyVm.CreateItem.NetworkConnectionSection,
 				VmSpecSection:             recomposeVAppParamsForEmptyVm.CreateItem.VmSpecSection,
 				GuestCustomizationSection: recomposeVAppParamsForEmptyVm.CreateItem.GuestCustomizationSection,
-				Media:                     mediaReference,
 			},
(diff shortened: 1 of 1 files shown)


Didainius:
@thiagoarrais,
Sorry it stands here forever but we are cautious of not breaking  VM and this gives immediate pushback in our heads. While the PR is a one line one - most often such PRs can impact something and we obviously lack a test here...

To begin with we need to dig deeper into the problem and we will need a test (we'll help with it) to avoid breaking it in future and having a testable solution.

A few questions to begin with:
* You mention resource `vcd_vm`. Does it happen only in `vcd_vm` or in `vcd_vapp_vm` as well (the reason I am asking is because the code is largely shared between these two resources and we must be sure we're not breaking anything)?
* You mentioned VCD version `10.3.0.18295834` - do you know if it worked before/on other versions?
* Are you able to share a code snippet (and give a short description) to replicate the problem?
* Did you have any workaround to make this work before this patch or is it not possible at all?

Thanks!

timstoop:
Just tested this patch and it solves my issue as described in #749.

dataclouder (approved):
Tested with standalone VM and vapp_vm. Both work

lvirbalas (approved):
Thanks for contributing!

Didainius (approved):

vbauzysvmware (approved):

Didainius:
Thanks @thiagoarrais , and sorry for taking it longer

PR #716, opened 2021-09-23
Primary NIC in VM can be removed cold way only

Didainius:
Latest versions of VCD 10.2.2.1 and 10.3 do not allow to hot remove Primary NIC from a powered on VM and it requires a power off. `vcd_vapp_vm` and `vcd_vm` still try to hot remove a NIC and this returns an error.
This PR patches VM update so that Primary NIC removal is done in cold fashion (powers off VM).

This test `TestAccVcdVAppVmMultiNIC` triggers the problem before patch.

### DIFF
--- a/.changes/v3.4.0/716-bug-fixes.md
+++ b/.changes/v3.4.0/716-bug-fixes.md
@@ -1,1 +1,1 @@
-* Primary NIC for `vcd_vapp_vm` and `vcd_vapp` is done in cold fashion [GH-716]
+* Primary NIC removal for `vcd_vapp_vm` and `vcd_vapp` is done in cold fashion [GH-716]
--- a/.changes/v3.4.0/716-bug-fixes.md
+++ b/.changes/v3.4.0/716-bug-fixes.md
@@ -1,0 +1,1 @@
+* Primary NIC for `vcd_vapp_vm` and `vcd_vapp` is done in cold fashion [GH-716]
--- a/vcd/resource_vcd_vapp_vm.go
+++ b/vcd/resource_vcd_vapp_vm.go
@@ -904,7 +904,10 @@ func resourceVmHotUpdate(d *schema.ResourceData, meta interface{}, vmType typeOf
 	}
 
 	// due the bug in VCD 10.1 hot update possible for adding new or update existing network, removing of network has to be done with cold update
-	if d.HasChange("network") && !isNetworkRemovedInVcd101(d, meta) {
+	// There are a few cases when hot NIC updates are not possible:
+	// * VCD 10.1 does not allow to remove NICs
+	// * Primary NIC cannot be removed on a powered on VM
+	if d.HasChange("network") && !isNetworkRemovedInVcd101(d, meta) && !isPrimaryNicRemoved(d) {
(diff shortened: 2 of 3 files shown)


vcd/resource_vcd_vapp_vm.go
```diff
@@ -904,7 +904,10 @@ func resourceVmHotUpdate(d *schema.ResourceData, meta interface{}, vmType typeOf
 	}
 
 	// due the bug in VCD 10.1 hot update possible for adding new or update existing network, removing of network has to be done with cold update
-	if d.HasChange("network") && !isNetworkRemovedInVcd101(d, meta) {
+	// There are a few cases when hot NIC updates are not possible:
+	// * VCD 10.1 does not allow to remove NICs
+	// * Primary NIC cannot be removed on a powered on VM
```
lvirbalas, line 909: Should this state:
```suggestion
	// * Primary NIC cannot be removed on a powered on VM as of VCD 10.2.2.1 and 10.3
```

Just to confirm, is there any supported VCD version where hot removal still works?

lvirbalas (changes requested):
I can't find the changelog :)

vcd/resource_vcd_vapp_vm.go
lvirbalas: Resolving this conversation, as we remove support for hot primary NIC removal for **all** VCD versions.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -904,7 +904,10 @@ func resourceVmHotUpdate(d *schema.ResourceData, meta interface{}, vmType typeOf
...
+	if d.HasChange("network") && !isNetworkRemovedInVcd101(d, meta) && !isPrimaryNicRemoved(d) {
 		networkConnectionSection, err := networksToConfig(d, vdc, vapp, vcdClient)
 		if err != nil {
 			return fmt.Errorf("unable to setup network configuration for update: %s", err)
```
lvirbalas, line 913: Does the user see a descriptive enough error to understand that removal failed, because VM is powered on?

.changes/v3.4.0/716-bug-fixes.md
```diff
@@ -0,0 +1 @@
+* Primary NIC for `vcd_vapp_vm` and `vcd_vapp` is done in cold fashion [GH-716]
```
lvirbalas, line 1: There's a word "removal" missing.

BTW, it sounds as if we are powering off the VM by our selves. Is this the case?

lvirbalas (changes requested):

.changes/v3.4.0/716-bug-fixes.md
Didainius: > There's a word "removal" missing.

Oops. Fixed.



> BTW, it sounds as if we are powering off the VM by our selves. Is this the case?

Yes. That has always been the case for cold operations. And before we introduced support for hot operations (cpu, memory, nic) it was always the case.

lvirbalas (approved):
LTGM!

dataclouder (approved):

PR #719, opened 2021-09-29
Change reference links with final docs format

dataclouder:
With the change to self-service releases, the documentation links previously used to refer from resources to data sources or vice-versa became broken.

We are fixing them by using the full path instead of the old relative path.


lvirbalas (approved):

Didainius (approved):
🤞

PR #726, opened 2021-10-07
Add vcd_nsxt_alb_settings resource and data source

Didainius:
This PR is a start of NSX-T ALB management in Tenant view (**it still requires `System` user to change it - just like Edge Gateway creation requires it**) and introduces `vcd_nsxt_alb_settings` resource and data source. The configuration is quite minimal (at least for now).

image

Note. Tests with tag `alb` have passed on 10.1 (ALB not supported), 10.2 and 10.3 - acceptance, org user acceptance and binary.

### DIFF
--- a/vcd/datasource_vcd_nsxt_alb_settings_test.go
+++ b/vcd/datasource_vcd_nsxt_alb_settings_test.go
@@ -9,8 +9,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 )
 
-// TestAccVcdNsxtAlbGeneralSettingsDS assumes that NSX-T ALB is not configured and General Settings shows "Inactive"
-func TestAccVcdNsxtAlbGeneralSettingsDS(t *testing.T) {
+// TestAccVcdNsxtAlbSettingsDS assumes that NSX-T ALB is not configured and General Settings shows "Inactive"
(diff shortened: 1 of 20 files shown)


vcd/datasource_vcd_nsxt_alb_settings_test.go
```diff
@@ -0,0 +1,71 @@
...
+	// This test requires access to the vCD before filling templates
+	// Thus it won't run in the short test
+	if vcdShortTest {
+		t.Skip(acceptanceTestsSkipped)
+		return
+	}
+
+	vcdClient := createTemporaryVCDConnection()
+	if vcdClient.Client.APIVCDMaxVersionIs("< 35.0") {
+		t.Skip(t.Name() + " requires at least API v35.0 (vCD 10.2+)")
```
lvirbalas, line 25: Officially, capitalization is "VCD" after the name change to VMware Cloud Director. I see you are using "VCD" below in this PR already :)

lvirbalas (approved):
LGTM! Only one small naming nit.

vcd/datasource_vcd_nsxt_alb_settings_test.go
Didainius: Yes. This is actually something that gets carried over many places as we usually copy this boilerplate. I changed all occurrences in my PR, but might need to tackle other leftovers in future.

PR #680, opened 2021-06-18
Add support for bus_type nvme

Gnitset:
We're using nvme for our disks and need support to resize them in our deployment.

Right not that is not supported, but this fix should make it work.
It works for us.

PR #739, opened 2021-11-09
Add support for NVMe bus type

Didainius:
Closes #680 
This PR actually builds on top of great work in #680, but it got stuck with a few undressed comments. 

Note. Tests with tags `vapp` and `vm` have passed on 10.2.2, 10.3.1


lvirbalas (approved):
Great we're finalizing this one, thanks!!

dataclouder (approved):

vbauzysvmware (approved):

PR #742, opened 2021-11-12
Use API tokens for authentication

dataclouder:
API tokens are introduced in VCD 10.3.1.
With this PR, we can use API tokens instead of other credentials to authenticate provider or tenants.

There is a new property `"api_token"` in the provider that we can use to authenticate. The provider will transparently replace that token with a refreshed bearer token, and operations can continue as usual.
There are a few restrictions, though. When using an API token, modification operations with users, roles, global roles, and rights bundles are not permitted.


website/docs/index.html.markdown
```diff
@@ -219,20 +247,27 @@ The following arguments are used to configure the VMware Cloud Director Provider
...
   * `token` allows to specify token in [`token`](#token) field.
+  * `api_token` allows to specify an API token.
   
 * `token` - (Optional; *v2.6+*) This is the token that can be used instead of username
    and password (in combination with field `auth_type=token`). When this is set, username and
    password will be ignored, but should be left in configuration either empty or with any custom
    values. A token can be specified with the `VCD_TOKEN` environment variable.
    Both a (deprecated) authorization token or a bearer token (*v3.1+*) can be used in this field.
 
+* `api_token` - (Optional; *v3.5+*) This is the API token that a System or organization administrator can create and
```
lvirbalas, line 265: I think we need to add additional explanation how `token` differs from `api_token`. We have two tokens now :) Not sure in which field it fits better though.

dataclouder: The text has some explanation and the example provides even more information.
We can't supply all the documentation here, but I added some wording to differentiate the two items.

lvirbalas: Thanks, it's clearer now!

lvirbalas (approved):
LGTM!

vbauzysvmware (approved):

Didainius (approved):
Thanks - it worked in my tested cases.

PR #748, opened 2021-11-20
Add API token test and fix previous token tests

dataclouder:
With the introduction of API tokens, we became more strict on
auth_type: when using tokens, it is now mandatory to state
which authentication type is being used.

Note: this test should have been included in PR #742, but got discarded by mistake

### DIFF
--- a/vcd/auth_test.go
+++ b/vcd/auth_test.go
@@ -4,6 +4,7 @@
 package vcd
 
 import (
+	"os"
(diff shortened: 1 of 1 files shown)


vbauzysvmware (approved):

Didainius (approved):
Test passed

lvirbalas (approved):

PR #760, opened 2021-12-22
Include SDK with weight patch for `ChangeCPUCountWithCore` function

Didainius:


Didainius:
Other PR will include govcd fix.

PR #761, opened 2021-12-22
Fixes for release 3.5.0

dataclouder:
Some resources are not available in 10.1, and we skip the relevant tests
* Skip Datasource test for vcd_vdc_group and vcd_library_certificate
* Skip Resource List test for vcd_vdc_group

Fix wrong references in TestAccVcdVAppVmMultiNIC

### DIFF
--- a/vcd/resource_vcd_vapp_vm_multinetwork_test.go
+++ b/vcd/resource_vcd_vapp_vm_multinetwork_test.go
@@ -263,7 +263,7 @@ resource "vcd_vapp" "{{.VAppName}}" {
   vdc = "{{.Vdc}}"
 
   name       = "{{.VAppName}}"
-  depends_on = ["vcd_network_routed.net", "vcd_network_routed.net2"]
+  depends_on = [vcd_network_routed.net, vcd_network_routed.net2]
 }
 
 resource "vcd_vapp_network" "vappIsolatedNet" {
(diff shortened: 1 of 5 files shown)


vcd/resource_vcd_vapp_vm_multinetwork_test.go
```diff
@@ -547,7 +547,7 @@ resource "vcd_vapp_vm" "{{.VMName}}" {
 
   network {
     type               = "org"
-    name               = vcd_network_routed.net.name
+    name               = vcd_vapp_org_network.vappAttachedNet.org_network_name
```
lvirbalas, line 550: What is the reason behind all these changes from Routed to vApp Org network?

dataclouder: The reason for the failures was that we had the wrong reference for the network name.
Both references in the end result to the same name, but, when we refer to the vapp network, Terraform will wait for the VM to be destroyed before destroying the vapp network. 
When we instead refer to the Org network, Terraform will try destruction in parallel. Sometimes the operation succeeds, and sometimes it fails.

dataclouder: Also explained in the [commit](https://github.com/vmware/terraform-provider-vcd/pull/761/commits/bc3df22c798114603be17589616738684303de0d).

Didainius: To add - this actually dates back to when we introduced `vcd_vapp_org_network` resource instead of doing "attachment magic" inside `vcd_vapp_vm`. Back then we tuned our examples and code in many places, however it looks there are still places left.

lvirbalas: Thank you, this, indeed, reminds me of very similar issues with dependencies I had long time ago.

lvirbalas (approved):

Didainius (approved):

vbauzysvmware (approved):

Issue #372, opened 2019-10-30
Implement meta data support for available resources


vbauzys:
A several resource already support meta data. Most of existing resources have also metadata ability.
Aim is add metadata ability to all possible resource.

Catalog
CatalogItem - implemented
Disk
Media - implemented
Network(Nsxv)
Organization
vApp - implemented
vAppTemplate
VDC - implemented
vdcStorageProfle
edge gtw(NSXV)
providerVDC
pvdcStorageProfle - 
VM - implemented

lvirbalas:
Please provide an exact list of resources where metadata is still missing and should be added.

acederlund:
Hi,

I've found that the vcd_org object is lacking metadata properties, and would be greatful to get support for this since we're using it for integration with billing systems :)

So far, I've found these other resources capable of metadata, but either lacks metadata properties or is lacking resourcetypes:
- Provider VDC (no object here, not sure it's needed either)
- Organizations
- Catalogs (tho there is support for both catalog items and media items)
- Storage Policies

There might be more, but this is what I've found so far.

dataclouder:
Hi @acederlund,
Currently we don't have plans to add provider VDC and storage policy resources or data sources. 

Regarding metadata,  we want to add metadata handling to Orgs, catalogs, networks, but it won't happen in the present release.

Instead, we're working on a system that uses metadata for filtering data sources, even for resources that can't modify metadata. A prototype will soon be available.

lvirbalas:
> Currently we don't have plans to add provider VDC

To be more specific: we have PVDC in the pipeline, but not in the near term plans.

mikeletux:
@vbauzysvmware catalog implemented as per #780

carmine73:
Any step forward for `vcd_org` metadata?

adambarreiro:
@carmine73 Yes, currently we're preparing [a PR](https://github.com/vmware/terraform-provider-vcd/pull/790) to add this feature.

bogi0704:
Are different User Access Levels ("Read/Write","Read only","Hidden") also in scope for the metadata fields. We would like to use hidden Attributes as an msp for our internal automations and processes, but also enable our customers to set read/write attributes for processes that they should be able to influence.

vbauzysvmware:
> @carmine73 Yes, currently we're preparing [a PR](https://github.com/vmware/terraform-provider-vcd/pull/790) to add this feature.

[...]

PR #780, opened 2022-01-31
Add catalog resource and datasource metadata enhancement

mikeletux:
This PR targets #372 regarding metadata on catalog resource and datasource.

With this enhancement, now when creating a Catalog using `vcd_catalog` it is possible to include metadata. An example would be:
```
resource "vcd_catalog" "test-catalog" {
  org = "my_org" 
  
  name               = "my_catalog"
  description        = "This is just a test catalog"

  delete_force      = "true"
  delete_recursive  = "true"

  metadata = {
    catalog  = "catalog Metadata v2"
    catalog2 = "catalog Metadata2 v2"
    catalog3 = "catalog Metadata3"
  }
}
```

Also, when using `vcd_catalog` datasource, also metadata regarding a catalog will be retrieved in the compute field `metadata`.

### DIFF
--- a/vcd/resource_vcd_catalog.go
+++ b/vcd/resource_vcd_catalog.go
@@ -3,6 +3,7 @@ package vcd
 import (
 	"context"
 	"fmt"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
 	"log"
 	"strings"
 
--- a/vcd/resource_vcd_catalog.go
+++ b/vcd/resource_vcd_catalog.go
@@ -388,14 +388,14 @@ func createOrUpdateAdminCatalogMetadata(d *schema.ResourceData, meta interface{}
 			}
 		}
 		for _, k := range toBeRemovedMetadata {
-			err := catalog.DeleteMetadata(k)
+			err := catalog.DeleteMetadataEntry(k)
 			if err != nil {
 				return fmt.Errorf("error deleting metadata: %s", err)
 			}
 		}
 		// Add new metadata
 		for k, v := range newMetadata {
-			_, err = catalog.AddMetadata(types.MetadataStringValue, k, v.(string))
+			err = catalog.AddMetadataEntry(types.MetadataStringValue, k, v.(string))
 			if err != nil {
 				return fmt.Errorf("error adding metadata: %s", err)
 			}
--- a/vcd/resource_vcd_catalog.go
+++ b/vcd/resource_vcd_catalog.go
@@ -3,9 +3,10 @@ package vcd
 import (
 	"context"
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"log"
 	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
 
@@ -86,6 +87,13 @@ func resourceVcdCatalog() *schema.Resource {
 				Sensitive:   true,
 				Description: "An optional password to access the catalog. Only ASCII characters are allowed in a valid password.",
 			},
+			"metadata": {
+				Type:        schema.TypeMap,
+				Optional:    true,
+				Description: "Key and value pairs for catalog metadata",
+				// For now underlying go-vcloud-director repo only supports
+				// a value of type String in this map.
+			},
 		},
 	}
 }
@@ -128,6 +136,11 @@ func resourceVcdCatalogCreate(ctx context.Context, d *schema.ResourceData, meta 
 		if err != nil {
 			return diag.FromErr(err)
 		}
+	}
+
+	err = createOrUpdateAdminCatalogMetadata(d, meta)
+	if err != nil {
+		return diag.Errorf("error adding media item metadata: %s", err)
 	}
 
 	log.Printf("[TRACE] Catalog created: %#v", catalog)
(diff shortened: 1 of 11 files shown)


vcd/resource_vcd_catalog.go
```diff
@@ -130,6 +138,11 @@ func resourceVcdCatalogCreate(ctx context.Context, d *schema.ResourceData, meta
 		}
 	}
 
+	err = createOrUpdateAdminCatalogMetadata(d, meta)
+	if err != nil {
+		return diag.Errorf("error adding media item metadata: %s", err)
```
vbauzysvmware, line 143: error message needs update:) `media item`

vcd/resource_vcd_catalog.go
```diff
@@ -195,6 +208,18 @@ func genericResourceVcdCatalogRead(d *schema.ResourceData, meta interface{}) err
 		dSet(d, "preserve_identity_information", false)
 		dSet(d, "password", "")
 	}
+
+	metadata, err := adminCatalog.GetMetadata()
+	if err != nil {
+		log.Printf("[DEBUG] Unable to find media item metadata: %s", err)
```
vbauzysvmware, line 214: error message needs update:) `media item`

vbauzysvmware (changes requested):
first scan :)

website/docs/r/catalog.html.markdown
```diff
@@ -62,6 +62,7 @@ source [vcd_storage_profile](/providers/vmware/vcd/latest/docs/data-sources/stor
 * `cache_enabled` - (Optional, *v3.6+*) Enable early catalog export to optimize synchronization. Default is `false`.
 * `preserve_identity_information` - (Optional, *v3.6+*) Enable include BIOS UUIDs and MAC addresses in the downloaded OVF package. Preserving the identity information limits the portability of the package and you should use it only when necessary. Default is `false`.
 * `password` - (Optional, *v3.6+*) An optional password to access the catalog. Only ASCII characters are allowed in a valid password.
+* `metadata` - (Optional; *v3.6+*) Key value map of metadata to assign
```
lvirbalas, line 65: Nit, but we want to be consistent:
```suggestion
* `metadata` - (Optional; *v3.6+*) Key value map of metadata to assign.
```

vcd/resource_vcd_catalog.go
```diff
@@ -86,6 +87,13 @@ func resourceVcdCatalog() *schema.Resource {
 				Sensitive:   true,
 				Description: "An optional password to access the catalog. Only ASCII characters are allowed in a valid password.",
 			},
+			"metadata": {
+				Type:        schema.TypeMap,
+				Optional:    true,
+				Description: "Key and value pairs for catalog metadata",
+				// For now underlying go-vcloud-director repo only supports
+				// a value of type String in this map.
```
lvirbalas, line 95: Hm. Do you mean that we can't set a number?

More than here, should we state that this is strings-only in the [website/docs/r/catalog.html.markdown](https://github.com/vmware/terraform-provider-vcd/pull/780/files#diff-bf2145122feda798ffa55bd1e73beda9a95863a60aada75b248fcf9e9cc5b327) ?

CC: @vbauzysvmware

lvirbalas (changes requested):

vcd/resource_vcd_catalog.go
mikeletux: Changed! Typo because copy/pasting ;)

vcd/resource_vcd_catalog.go
mikeletux: Same as above. Changed all occurrences.

website/docs/r/catalog.html.markdown
mikeletux: Dot added!

vcd/resource_vcd_catalog.go
mikeletux: Let me give some light in this thing.

As per the PR that I have opened in go-vcloud SDK vmware/go-vcloud-director#430 , now metadata methods allow passing the metadata typeValue. Before this, it was hardcoded the stringType. Now it can be anything. (from four types, only stringType and numberType have been tested. DateType and booleanType are not. This thing is documented in the code). Please check https://github.com/mikeletux/go-vcloud-director/blob/f134041848100263630e762d599f4a37457b3b14/govcd/metadata.go#L124-L127

The thing is, as per Terraform schema limitations, we don't want to have additional depth in the schema itself. That's why we keep a map and all metadata entries are enforced to Strings, instead of allowing the user choose the type.

vbauzysvmware: Let's from terraform perspective use only String. If there will be requirements for other types when we start brainstorming how to solve this. IMHO

vbauzysvmware (approved):
LGTM

lvirbalas (changes requested):

adambarreiro (approved):

website/docs/d/catalog.html.markdown
```diff
@@ -46,6 +46,7 @@ The following arguments are supported:
 * `publish_enabled` - (*v3.6+*) Enable allows to publish a catalog externally to make its vApp templates and media files available for subscription by organizations outside the Cloud Director installation. Default is `false`.
 * `cache_enabled` - (*v3.6+*) Enable early catalog export to optimize synchronization. Default is `false`.
 * `preserve_identity_information` - (*v3.6+*) Enable include BIOS UUIDs and MAC addresses in the downloaded OVF package. Preserving the identity information limits the portability of the package and you should use it only when necessary. Default is `false`.
+* `metadata` -  (*v3.6+*) Key value map of metadata.
```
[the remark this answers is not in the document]
mikeletux: Fixed!

vcd/resource_vcd_catalog.go
lvirbalas: @mikeletux

mikeletux: I don't think is really needed, since the only option so far from Terraform side of things is strings. I'd leave it like that as @vbauzysvmware says and if in the future there is any need, we can think about it.

lvirbalas (approved):

vcd/resource_vcd_catalog.go
Didainius: This was discussed a lot, but I vote for leaving strings only in Terraform (because we have a `TypeMap`, to accommodate types, we'd have to create blocks of structures accepting type and value for each record). IMO that is less convenient than a map unless it is really needed.

But I suggest to remove the comment above (here and in data source), because govcd does support other types now and the comment itself is misleading

Didainius (approved):

Issue #759, opened 2021-12-22
Can't set `Enable IP Masquerade` for vcd_vapp_nat_rules


tcinbis:
Hey everyone,

The title pretty much describes the issue. When I'm setting `enable_ip_masquerade`  to true it is correctly shown in the terraform plan output, but after I have applied the plan the option is disabled in vCloud when checking it via the UI.
When running `terraform apply` again it also doesn't pickup the change and reports that the infrastructure matches the configuration.

Am I missing something here?

Thanks for your help!

### Terraform Version
```
Terraform v1.1.1
on windows_amd64
+ provider registry.terraform.io/hashicorp/null v3.1.0
+ provider registry.terraform.io/vmware/vcd v3.4.0
```

```
VMware Cloud Director version: 10.2.2.17855680
```

### Affected Resource(s)
- vcd_vapp_nat_rules

### Terraform Configuration Files

```hcl
provider "registry.terraform.io/hashicorp/null" {
  version = "3.1.0"
  hashes = [
    "h1:SFT7X3zY18CLWjoH2GfQyapxsRv6GDKsy9cF1aRwncc=",
  ]
}

provider "registry.terraform.io/vmware/vcd" {
  version     = "3.4.0"
  constraints = "~> 3.4.0"
  hashes = [
    "h1:YAGmqCFJiJJr/shs7P9UKVp3+LsJ8HBr9tYUnP3ZrRA=",
  ]
}
```

### Debug Output
https://gist.github.com/tcinbis/bffddec898be2e31376ec9324de182a9

### Expected Behavior
The `Enable IP Masquerade` option in the NAT setting should be enabled.

### Actual Behavior
`Enable IP Masquerade` is not enabled.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### User Access rights 
- User that Terraform uses for authentication has admin rights.

tcinbis:
So I poked around a bit with PowerCLI and read through the Go code for the NAT configuration. And what I find interesting is at [line 156-157](https://github.com/vmware/terraform-provider-vcd/blob/master/vcd/resource_vcd_vapp_nat_rules.go#L157). 

My interpretation of the code snippet is the following. By default the policy is set to `allowTrafficIn` and is changed to 
 `allowTraffic` if `enable_ip_masquerade` is **false**.
[...]

PR #784, opened 2022-02-10
Fix enable_ip_masquerade handling in vcd_vapp_nat_rules resource

vbauzysvmware:
Fix for: https://github.com/vmware/terraform-provider-vcd/issues/759

The issue was that treating values was vice versa. On reading and for creating/updating.
The test is still correct and didn't need an update. That's the case when read isn't right, then everything looks good :)

Also updated vcd_vapp_nat_rules to not use deprecated schema.Resource fields for create/update/delete/import, now it uses CreateContext, UpdateContext, ReadContext, DeleteContext.

Also functions return the [diag.Diagnostics struct](https://learn.hashicorp.com/tutorials/terraform/provider-debug) instead of plain Go errors, to adjust their signature to the new CreateContext, UpdateContext, ReadContext, DeleteContext fields.

### DIFF
--- a/.changes/v3.6.0/784-bug-fixes.md
+++ b/.changes/v3.6.0/784-bug-fixes.md
@@ -1,0 +1,1 @@
+* Fix Issue #759 where enable_ip_masquerade handling in vcd_vapp_nat_rules resource wasn't correct [GH-784]
--- a/vcd/resource_vcd_vapp_nat_rules.go
+++ b/vcd/resource_vcd_vapp_nat_rules.go
@@ -1,7 +1,9 @@
 package vcd
 
 import (
+	"context"
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"log"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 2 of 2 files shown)


adambarreiro (approved):

lvirbalas (changes requested):
- Please add a sentence to the PR description about the diag changes ("the what and why").
- The fix itself seems trivial, however surprising we had the bug there. Can you add a test that covers this?

vbauzysvmware:
> * Please add a sentence to the PR description about the diag changes ("the what and why").
> 
>     * The fix itself seems trivial, however surprising we had the bug there. Can you add a test that covers this?

Added note about additional change.
As mentioned in the description, the Test covers the issue and nothing additional to add is possible, due that when read was compromised then nothing will help. The compromised read function was confirming incorrect state.

lvirbalas:
> Added note about additional change.
Thanks, but "Also update deprecated method usages" is too abstract. Let's be more precise. Maybe something like here (in the middle of description):
https://github.com/vmware/terraform-provider-vcd/pull/786

vbauzysvmware:
> > Added note about additional change.
> 
> Thanks, but "Also update deprecated method usages" is too abstract. Let's be more precise. Maybe something like here (in the middle of description): #786

added

Didainius (approved):
Tested manually and it worked well now.

lvirbalas (approved):

mikeletux (approved):
Tested manually and it work well!

tcinbis:
Thanks to everyone working on this issue for resolving it so quickly! 🥳

Since we are currently depending on a workaround which I would like to get rid of soon, I am curious, if there is some timeline for when the next release with this fix?

Issue #365, opened 2019-10-28
Add catalog details represeting properties as computed fields


vbauzysvmware:
We can enhance catalog with additional fields which provides more details about catalog.
For e.g.  isShared, isPublished, ownerName, numberOfVAppTemplates,numberOfMedia, etc.

mikeletux:
This feature will be available in v3.6.0.
Feature merge in https://github.com/vmware/terraform-provider-vcd/pull/800

PR #800, opened 2022-03-04
Add catalog details represeting properties as computed fields

mikeletux:
This PR aims to address the enhancements proposed by @vbauzysvmware on github issue #365 . 

## Description
According to the issue above mentioned, 7 new computed fields have been added to `catalog` resource and datasource. 

## Detailed description
The new computed fields for `catalog` resource and datasource are:

* `catalog_version` - Version number from this catalog.
* `owner_name` - Owner of the catalog.
* `number_of_vapp_templates` - Number of vApp templates available in this catalog.
* `number_of_media` - Number of media items available in this catalog.
* `is_shared` - Indicates if the catalog is shared.
* `is_published` - Indicates if this catalog is shared to all organizations.
* `publish_subscription_type` - Shows if the catalog is published, if it is a subscription from another one or none of those.

This PR relays on https://github.com/vmware/go-vcloud-director/pull/450

### DIFF
--- a/vcd/datasource_vcd_catalog.go
+++ b/vcd/datasource_vcd_catalog.go
@@ -85,12 +85,17 @@ func datasourceVcdCatalog() *schema.Resource {
 			"is_shared": {
 				Type:        schema.TypeBool,
 				Computed:    true,
-				Description: "Indicates if the catalog is shared.",
+				Description: "True if this catalog is shared.",
 			},
 			"is_published": {
 				Type:        schema.TypeBool,
 				Computed:    true,
-				Description: "Indicates if the catalog is published.",
+				Description: "True if this catalog is shared to all organizations.",
+			},
+			"publish_subscription_type": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "PUBLISHED if published externally, SUBSCRIBED if subscribed to an external catalog, UNPUBLISHED otherwise.",
 			},
 			"filter": &schema.Schema{
 				Type:        schema.TypeList,
@@ -177,12 +182,13 @@ func datasourceVcdCatalogRead(ctx context.Context, d *schema.ResourceData, meta 
 		return diag.Errorf("There was an issue when retrieving the catalog records - %s", err)
 	}
 
-	dSet(d, "catalog_version", catalog.AdminCatalog.VersionNumber)
+	dSet(d, "catalog_version", catalogRecords[0].Version)
 	dSet(d, "owner_name", catalogRecords[0].OwnerName)
 	dSet(d, "number_of_vapp_templates", catalogRecords[0].NumberOfVAppTemplates)
 	dSet(d, "number_of_media", catalogRecords[0].NumberOfMedia)
 	dSet(d, "is_published", catalogRecords[0].IsPublished)
 	dSet(d, "is_shared", catalogRecords[0].IsShared)
+	dSet(d, "publish_subscription_type", catalogRecords[0].PublishSubscriptionType)
 
 	return nil
 }
--- a/vcd/datasource_vcd_catalog.go
+++ b/vcd/datasource_vcd_catalog.go
@@ -82,11 +82,6 @@ func datasourceVcdCatalog() *schema.Resource {
 				Computed:    true,
 				Description: "Number of Medias this catalog contains.",
 			},
-			"date_created": {
-				Type:        schema.TypeString,
-				Computed:    true,
-				Description: "Date when the catalog was created",
-			},
 			"filter": &schema.Schema{
 				Type:        schema.TypeList,
 				MaxItems:    1,
@@ -172,8 +167,6 @@ func datasourceVcdCatalogRead(ctx context.Context, d *schema.ResourceData, meta 
 	}
 
 	dSet(d, "number_of_vapp_templates", len(numberOfVAppTemplates))
-
-	dSet(d, "date_created", catalog.Catalog.DateCreated)
 
 	numberOfMedia, err := catalog.QueryMediaList()
 	if err != nil {
--- a/vcd/datasource_vcd_catalog.go
+++ b/vcd/datasource_vcd_catalog.go
@@ -134,9 +134,8 @@ func datasourceVcdCatalogRead(ctx context.Context, d *schema.ResourceData, meta 
 	adminOrg, err = vcdClient.VCDClient.GetAdminOrgByName(orgName)
 
 	if err != nil {
-		log.Printf("[DEBUG] Org %s not found. Setting ID to nothing", orgName)
-		d.SetId("")
-		return nil
+		log.Printf("[DEBUG] Org %s not found.", orgName)
+		return diag.Errorf("Org %s not found.", orgName)
 	}
 	log.Printf("[TRACE] Org %s found", orgName)
 
@@ -149,8 +148,14 @@ func datasourceVcdCatalogRead(ctx context.Context, d *schema.ResourceData, meta 
 	}
 	if err != nil {
 		log.Printf("[DEBUG] Catalog %s not found. Setting ID to nothing", identifier)
-		d.SetId("")
 		return diag.Errorf("error retrieving catalog %s: %s", identifier, err)
+	}
+
+	// Catalog record is retrieved to get the owner name, number of vApp templates and medias, and if the catalog is shared and published
+	catalogRecords, err := adminOrg.FindCatalogRecords(catalog.AdminCatalog.Name)
+	if err != nil {
+		log.Printf("[DEBUG] Unable to retrieve catalog record: %s", err)
+		return diag.Errorf("There was an issue when retrieving the catalog records - %s", err)
 	}
 
 	metadata, err := catalog.GetMetadata()
@@ -173,13 +178,6 @@ func datasourceVcdCatalogRead(ctx context.Context, d *schema.ResourceData, meta 
 	err = d.Set("metadata", getMetadataStruct(metadata.MetadataEntry))
 	if err != nil {
 		return diag.Errorf("There was an issue when setting metadata into the schema - %s", err)
-	}
-
-	// Catalog record is retrieved to get the owner name, number of vApp templates and medias, and if the catalog is shared and published
-	catalogRecords, err := adminOrg.FindCatalogRecords(catalog.AdminCatalog.Name)
-	if err != nil {
-		log.Printf("[DEBUG] Unable to find catalog record: %s", err)
-		return diag.Errorf("There was an issue when retrieving the catalog records - %s", err)
 	}
 
 	dSet(d, "catalog_version", catalogRecords[0].Version)
--- a/vcd/datasource_vcd_catalog.go
+++ b/vcd/datasource_vcd_catalog.go
@@ -82,6 +82,11 @@ func datasourceVcdCatalog() *schema.Resource {
 				Computed:    true,
 				Description: "Number of Medias this catalog contains.",
 			},
+			"date_created": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Date when the catalog was created",
+			},
 			"filter": &schema.Schema{
 				Type:        schema.TypeList,
 				MaxItems:    1,
@@ -167,6 +172,8 @@ func datasourceVcdCatalogRead(ctx context.Context, d *schema.ResourceData, meta 
 	}
 
 	dSet(d, "number_of_vapp_templates", len(numberOfVAppTemplates))
+
+	dSet(d, "date_created", catalog.Catalog.DateCreated)
 
 	numberOfMedia, err := catalog.QueryMediaList()
 	if err != nil {
--- a/vcd/datasource_vcd_catalog.go
+++ b/vcd/datasource_vcd_catalog.go
@@ -2,6 +2,7 @@ package vcd
 
 import (
 	"context"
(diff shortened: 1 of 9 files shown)


mikeletux:
Changelog as well as docs still need to be put in place.

vcd/resource_vcd_catalog.go
```diff
@@ -182,6 +212,14 @@ func genericResourceVcdCatalogRead(d *schema.ResourceData, meta interface{}) err
 		return fmt.Errorf("error retrieving catalog %s : %s", d.Id(), err)
 	}
 
+	// catalog user view is retrieved to get the number of vApp templates and medias
```
lvirbalas, line 215: Polishing nit:

```suggestion
	// Catalog user view is retrieved to get the number of vApp templates and medias
```

lvirbalas (approved):
One polishing nit and LGTM!

adambarreiro (approved):

vcd/datasource_vcd_catalog.go
```diff
@@ -62,6 +62,36 @@ func datasourceVcdCatalog() *schema.Resource {
 				Computed:    true,
 				Description: "Key and value pairs for catalog metadata",
 			},
+			"catalog_version": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "Catalog version number.",
+			},
+			/*"owner_name": { // owner_name is not available because catalog user view doesn't have this value
```
Didainius, line 70: Just checking for my understanding - is it just AdminCatalog which has it?

vcd/datasource_vcd_catalog.go
```diff
@@ -139,5 +170,33 @@ func datasourceVcdCatalogRead(ctx context.Context, d *schema.ResourceData, meta
...
+
+	dSet(d, "number_of_media", len(numberOfMedia))
+
+	dSet(d, "is_published", catalog.Catalog.IsPublished)
+
+	controlAccessParams, err := catalog.GetAccessControl(false)
+	if err != nil {
+		return diag.Errorf("error retrieving catalog access control: %s", err)
+	}
+	if controlAccessParams.IsSharedToEveryone || controlAccessParams.AccessSettings != nil {
```
Didainius, line 195: Just double checking. Are we guaranteed the catalog is shared if `controlAccessParams.AccessSettings != nil`?

Didainius, line 195: Just double checking. Are we guaranteed the catalog is shared if `controlAccessParams.AccessSettings != nil`?

vcd/resource_vcd_catalog.go
```diff
@@ -182,6 +212,14 @@ func genericResourceVcdCatalogRead(d *schema.ResourceData, meta interface{}) err
...
+	catalog, err := adminOrg.GetCatalogById(d.Id(), false)
+	if err != nil {
+		log.Printf("[DEBUG] Unable to find catalog. Removing from tfstate")
+		d.SetId("")
+		return fmt.Errorf("error retrieving catalog %s : %s", d.Id(), err)
```
Didainius, line 220: As you are working on this code there is a good opportunity for a fix here and in the  `adminCatalog, err := adminOrg.GetAdminCatalogByNameOrId(d.Id(), false)` error handling a few lines above.

In our provider there are a few historical quirks with how it behaves when a resource or data source is not found (mainly in older resources). The rule of thumb should be:
* **Resource not found** during a read - we must do a `d.SetId("")` and `return nil` (not an error). The reason for that is that if a resource has dissapeared (maybe someone delete it via UI), the consumer will just get an offer to create a new one during `terraform plan`. 
 
Now ideally - we'd like to distinguish between a "resource not found" error and any other error (API timeout, etc). For that reason we have a special [sentinel error](https://earthly.dev/blog/golang-errors/#defining-sentinel-errors) `govcd.ErrorEntityNotFound`  and also two helper functions `govcd.IsNotFound` and `govcd.ContainsNotFound`. Whenever we do SDK code, we aim to provide the `govcd.ErrorEntityNotFound` in the returned error when a resource is not found. You need to _double check_ that `adminOrg.GetAdminCatalogByNameOrId` and `adminOrg.GetCatalogById` both return that `govcd.ErrorEntityNotFound` error when needed.
  
* **Data source not found** during a read - error must be returned to user as that would break downstream operations anyway.

Below is one example how we handle it for best user experience.

Excerpt from resource `read` for NSX-T Edge Gateway. See how it conditionally checks for `govcd.ContainsNotFound(err)` to check if the resource is not found - then it resets the ID and returns nil. But if the error is anything else - it will return it to user as we don't want to remove the resource from statefile if for example a connection timeout occured during terraform run.
```go
	edge, err := org.GetNsxtEdgeGatewayById(d.Id())
	if err != nil {
		if govcd.ContainsNotFound(err) {
			d.SetId("")
			return nil
		}
		return diag.Errorf("could not retrieve NSX-T Edge Gateway: %s", err)
	}
```

Excerpt for data source code (a lot simpler as you can see):
```go
        edge, err := vdc.GetNsxtEdgeGatewayByName(d.Get("name").(string))
	if err != nil {
		return diag.FromErr(fmt.Errorf("could not retrieve NSX-T edge gateway: %s", err))
	}
```

Didainius, line 220: As you are working on this code there is a good opportunity for a fix here and in the  `adminCatalog, err := adminOrg.GetAdminCatalogByNameOrId(d.Id(), false)` error handling a few lines above.

In our provider there are a few historical quirks with how it behaves when a resource or data source is not found (mainly in older resources). The rule of thumb should be:
* **Resource not found** during a read - we must do a `d.SetId("")` and `return nil` (not an error). The reason for that is that if a resource has dissapeared (maybe someone delete it via UI), the consumer will just get an offer to create a new one during `terraform plan`. 
 
Now ideally - we'd like to distinguish between a "resource not found" error and any other error (API timeout, etc). For that reason we have a special [sentinel error](https://earthly.dev/blog/golang-errors/#defining-sentinel-errors) `govcd.ErrorEntityNotFound`  and also two helper functions `govcd.IsNotFound` and `govcd.ContainsNotFound`. Whenever we do SDK code, we aim to provide the `govcd.ErrorEntityNotFound` in the returned error when a resource is not found. You need to _double check_ that `adminOrg.GetAdminCatalogByNameOrId` and `adminOrg.GetCatalogById` both return that `govcd.ErrorEntityNotFound` error when needed.
  
* **Data source not found** during a read - error must be returned to user as that would break downstream operations anyway.

Below is one example how we handle it for best user experience.

Excerpt from resource `read` for NSX-T Edge Gateway. See how it conditionally checks for `govcd.ContainsNotFound(err)` to check if the resource is not found - then it resets the ID and returns nil. But if the error is anything else - it will return it to user as we don't want to remove the resource from statefile if for example a connection timeout occured during terraform run.
```go
	edge, err := org.GetNsxtEdgeGatewayById(d.Id())
	if err != nil {
		if govcd.ContainsNotFound(err) {
			d.SetId("")
			return nil
		}
		return diag.Errorf("could not retrieve NSX-T Edge Gateway: %s", err)
	}
```

Excerpt for data source code (a lot simpler as you can see):
```go
        edge, err := vdc.GetNsxtEdgeGatewayByName(d.Get("name").(string))
	if err != nil {
		return diag.FromErr(fmt.Errorf("could not retrieve NSX-T edge gateway: %s", err))
	}
```

Didainius (changes requested):
Lucky number #800 :)

This looks good. I have a few queries and an ask before I get to manual test

vcd/resource_vcd_catalog.go
```diff
@@ -217,6 +255,33 @@ func genericResourceVcdCatalogRead(d *schema.ResourceData, meta interface{}) err
 		return err
 	}
 
+	dSet(d, "catalog_version", adminCatalog.AdminCatalog.Catalog.VersionNumber)
+	// dSet(d, "owner_name", adminCatalog.AdminCatalog.Owner.User.Name)
+
+	vAppTemplates, err := catalog.QueryVappTemplateList()
```
vbauzysvmware, line 261: I think it's to heavy to call all templates and medias to know the numbers. Catalog record (from query) has these numbers:
```
<CatalogRecord name="cat-vb" description="vb catalog" isPublished="false" isShared="false" creationDate="2022-02-21T13:13:37.525Z" orgName="vb" ownerName="system" numberOfVAppTemplates="2"
```

vbauzysvmware (changes requested):

lvirbalas (approved):

vcd/datasource_vcd_catalog.go
mikeletux: Following @vbauzysvmware recommendation, after retrieving CatalogRecord, I can grab the owner info from there. Now is set, please check :)

vcd/datasource_vcd_catalog.go
mikeletux: I overcomplicated this. Now I'm grabbing this field from Catalog record, much straightforward. Please give it a look!

mikeletux: Repeated comment

vcd/resource_vcd_catalog.go
mikeletux: Repeated comment.

vcd/resource_vcd_catalog.go
mikeletux: @vbauzysvmware please check now. I used your suggestion and now things are simpler and more understandable!

vcd/resource_vcd_catalog.go
```diff
@@ -182,6 +212,13 @@ func genericResourceVcdCatalogRead(d *schema.ResourceData, meta interface{}) err
 		return fmt.Errorf("error retrieving catalog %s : %s", d.Id(), err)
 	}
 
+	// Catalog record is retrieved to get the number of vApp templates, medias and owner name
+	adminCatalogRecord, err := adminOrg.FindAdminCatalogRecords(adminCatalog.AdminCatalog.Name) // Is this returning one even though there are catalogs with similar names?
```
vbauzysvmware, line 216: why in datasource `QueryCatalogList` and here `FindAdminCatalogRecords` is used?

mikeletux: In datasource we do not retrieve the admin view from a catalog. Also there is no function for finding catalog records as it exists for admin catalog.
In the resource, since the only way to create a catalog is the admin view, we can use that function as it is available.

vbauzysvmware: We had investigated and agreed that there is a need a little polish with methods in govcd part.

mikeletux: @vbauzysvmware already open a PR for this in go-vcd. Please check https://github.com/vmware/go-vcloud-director/pull/450
With that, we will be able to get specific catalogRecord by catalog name.

Also, to make the code more consistent, both System Administrator and Org Administrator can query the same way AdminCatalog, so what I've done is to change the `Catalog` struct being used in the datasource to `AdminCatalog` struct. This way code is more consistent.

vbauzysvmware: Thank you

vcd/resource_vcd_catalog.go
mikeletux: Very good insight @Didainius . Let me check the functions in go-vcd and update them if needed!

mikeletux: already implemented @Didainius , please check 😄

mikeletux:
Also upgraded catalog datasource so org attribute is optional, so users can use from now on either the org at provider level or the one at datasource level.

vcd/resource_vcd_catalog.go
Didainius: Looks great! Thank you

vbauzysvmware:
One more thing I would like to have: read part of setting values to the state of data source and source seems are very similar. We could make one function and reuse it. We have quite often that in our provider.

mikeletux:
> One more thing I would like to have: read part of setting values to the state of data source and source seems are very similar. We could make one function and reuse it. We have quite often that in our provider.

@vbauzysvmware Done!
I added a new function called `setCatalogRecordValuesToSchema` that it's being used in both the resource and the datasource!

vcd/datasource_vcd_catalog.go
```diff
@@ -124,20 +153,26 @@ func datasourceVcdCatalogRead(ctx context.Context, d *schema.ResourceData, meta
...
+		dSet(d, "cache_enabled", catalog.AdminCatalog.PublishExternalCatalogParams.IsCachedEnabled)
+		dSet(d, "preserve_identity_information", catalog.AdminCatalog.PublishExternalCatalogParams.PreserveIdentityInfoFlag)
 	}
 
 	err = d.Set("metadata", getMetadataStruct(metadata.MetadataEntry))
 	if err != nil {
 		return diag.Errorf("There was an issue when setting metadata into the schema - %s", err)
 	}
 
+	err = setCatalogRecordValuesToSchema(d, adminOrg, catalog.AdminCatalog.Name)
```
vbauzysvmware, line 172: usually we call this like this: setCatalogData

vbauzysvmware (approved):

vcd/datasource_vcd_catalog.go
mikeletux: Done ✅

Didainius (approved):
Thanks!

Issue #775, opened 2022-01-19
Allow vcd_org to enable/disable publish to external VCD ORG and subscribe


vbauzysvmware:
Add new properties to `vcd_org` which allow toggle `Publish external catalogs` and `Subscribe to external catalogs` features
![image](https://user-images.githubusercontent.com/41415459/150147452-d8ddd6da-2294-4205-91aa-b881f26fb7ff.png)

mikeletux:
Closing this issue as this feature will be included in version 3.6

PR #803, opened 2022-03-08
Add support for "Publish external catalogs" and "Subscribe to external catalogs " in VCD Org 

mikeletux:
This PR is related #775 

# Description
This PR aims to give support to both `datasource_vcd_org`  and `resource_vcd_org` to "Publish external catalogs" and "Subscribe to external catalogs" options.

# Detailed description
The goal of this PR is add support to `resource_vcd_org` to change the boolean values from "Publish external catalogs" and "Subscribe to external catalogs" using `resource_vcd_org` via its new two schema entries `can_publish_external_catalogs`and `can_subscribe_external_catalogs`.  
  
Also, these two new schema entries will be available in `datasource_vcd_org` for reading the current value.  
  
Acceptance test `TestAccVcdOrgFull` has been modified to accommodate these new values. 
This new feature have been tested using acceptance tests and also tested manually.

### DIFF
--- a/vcd/resource_vcd_org.go
+++ b/vcd/resource_vcd_org.go
@@ -282,7 +282,6 @@ func getSettings(d *schema.ResourceData) *types.OrgSettings {
 
 // Deletes org
 func resourceOrgDelete(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
-	var diags diag.Diagnostics
 
 	//DELETING
 	vcdClient := m.(*VCDClient)
@@ -318,13 +317,13 @@ func resourceOrgDelete(ctx context.Context, d *schema.ResourceData, m interface{
 		log.Printf("[DEBUG] Error deleting org %s: %s", orgName, err)
 		return diag.FromErr(err)
 	}
+
 	log.Printf("[TRACE] Org %s deleted", orgName)
-	return diags
+	return nil
 }
 
 // Update the resource
 func resourceOrgUpdate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
-	var diags diag.Diagnostics
 
 	vcdClient := m.(*VCDClient)
 
@@ -371,7 +370,7 @@ func resourceOrgUpdate(ctx context.Context, d *schema.ResourceData, m interface{
 	}
 
 	log.Printf("[TRACE] Org %s updated", orgName)
-	return diags
+	return nil
 }
 
 // setOrgData sets the data into the resource, taking it from the provided adminOrg
@@ -440,7 +439,6 @@ func setOrgData(d *schema.ResourceData, adminOrg *govcd.AdminOrg) error {
 
 // Retrieves an Org resource from vCD
 func resourceOrgRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
-	var diags diag.Diagnostics
 
 	vcdClient := m.(*VCDClient)
 
@@ -470,7 +468,7 @@ func resourceOrgRead(ctx context.Context, d *schema.ResourceData, m interface{})
 	if err != nil {
 		log.Printf("[DEBUG] Org %s not found. Setting ID to nothing", identifier)
 		d.SetId("")
-		return diags
+		return nil
 	}
 	log.Printf("[TRACE] Org with id %s found", identifier)
 	d.SetId(adminOrg.AdminOrg.ID)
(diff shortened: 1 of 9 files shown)


website/docs/r/org.html.markdown
```diff
@@ -60,6 +60,8 @@ The following arguments are supported:
 * `delay_after_power_on_seconds` - (Optional) - Specifies this organization's default for virtual machine boot delay after power on. Default is `0`.
 * `vapp_lease` - (Optional; *v2.7+*) - Defines lease parameters for vApps created in this organization. See [vApp Lease](#vapp-lease) below for details. 
 * `vapp_template_lease` - (Optional; *v2.7+*) - Defines lease parameters for vApp templates created in this organization. See [vApp Template Lease](#vapp-template-lease) below for details.
+* `can_publish_external_catalogs` - (Optional; *v3.6+*) - True if this organization is allowed to publish external catalogs. Default is `false`.
+* `can_subscribe_external_catalogs` - (Optional; *v3.6+*) - True if this organization is allowed to subscribe to external catalogs. Default is `false`.
```
lvirbalas, line 64: What about grouping these two fields with the above `can_publish_catalogs` (i.e. moving these lines up)?

website/docs/d/org.html.markdown
```diff
@@ -60,6 +60,8 @@ The following arguments are supported:
 * `delay_after_power_on_seconds` - Specifies this organization's default for virtual machine boot delay after power on.
 * `vapp_lease` - (*v2.7+*) - Defines lease parameters for vApps created in this organization. See [vApp Lease](#vapp-lease) below for details. 
 * `vapp_template_lease` - (*v2.7+*) - Defines lease parameters for vApp templates created in this organization. See [vApp Template Lease](#vapp-template-lease) below for details.
+* `can_publish_external_catalogs` - (*v3.6+*) - True if this organization is allowed to publish external catalogs.
+* `can_subscribe_external_catalogs` - (*v3.6+*) - True if this organization is allowed to subscribe to external catalogs.
```
lvirbalas, line 64: What about grouping these two fields with the above `can_publish_catalogs` (i.e. moving these lines up)?

lvirbalas (changes requested):

website/docs/r/org.html.markdown
mikeletux: Done!

website/docs/d/org.html.markdown
mikeletux: Done!

vcd/resource_vcd_org.go
```diff
@@ -297,20 +316,21 @@ func resourceOrgDelete(d *schema.ResourceData, m interface{}) error {
 	err = adminOrg.Delete(deleteForce, deleteRecursive)
 	if err != nil {
 		log.Printf("[DEBUG] Error deleting org %s: %s", orgName, err)
-		return err
+		return diag.FromErr(err)
 	}
 	log.Printf("[TRACE] Org %s deleted", orgName)
-	return nil
+	return diags
```
Didainius, line 322: Is there any reason we have defined `var diags diag.Diagnostics` and then use it in return (but nowhere else). Couldn't we simply return `nil`?

vcd/resource_vcd_org.go
```diff
@@ -342,16 +362,16 @@ func resourceOrgUpdate(d *schema.ResourceData, m interface{}) error {
...
 	err = task.WaitTaskCompletion()
 	if err != nil {
 		log.Printf("[DEBUG] Error completing update of Org %s : %s", orgName, err)
-		return fmt.Errorf("error completing update of Org %s", err)
+		return diag.Errorf("error completing update of Org %s", err)
 	}
 
 	log.Printf("[TRACE] Org %s updated", orgName)
-	return nil
+	return diags
```
Didainius, line 374: Is there any reason we have defined `var diags diag.Diagnostics` and then use it in return (but nowhere else). Couldn't we simply return `nil`?

mikeletux: Same as above! 😄

vcd/resource_vcd_org.go
Didainius: I think the reason why they do it is that they utilize this new pattern of `diag` (which can accumulate errors and warnings) instead of `error`, but until we make use of it, the `return diags` (which is always `nil` so far) instead of just `nil` looks to be a bit more confusing (requiring extra scan)

mikeletux: I agree with you @Didainius . Let's keep code consistent as we are not using it. I'll revert this to the previous pattern!

vbauzysvmware (approved):
LGTM

lvirbalas (approved):

Didainius (approved):

adambarreiro (approved):
LGTM!

PR #811, opened 2022-03-30
Add metadata support for storage profile datasource

adambarreiro:
Created by #372

:warn: Depends on https://github.com/vmware/go-vcloud-director/pull/454

This PR adds metadata support (read) for the storage profile datasource. This was tested manually, as there is no `vcd_storage_profile` resource to manipulate metadata. The manual test is as follows:

- Go to the target VDC as System administrator, Provider view.
- Select a target storage profile
- Click on "Metadata" and add some string value.
- Execute a Terraform apply with the datasource. Metadata should have been retrieved.

### DIFF
--- a/.changes/v3.6.0/811-improvements.md
+++ b/.changes/v3.6.0/811-improvements.md
@@ -1,0 +1,1 @@
+* Data source `vcd_storage_profile` supports `metadata` so this field can be populated when reading VDC storage profiles. [GH-811]
--- a/go.mod
+++ b/go.mod
@@ -9,3 +9,5 @@ require (
 	github.com/kr/pretty v0.2.1
 	github.com/vmware/go-vcloud-director/v2 v2.15.0-alpha.9
 )
+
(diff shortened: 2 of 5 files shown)


dataclouder (approved):

mikeletux (approved):

lvirbalas (approved):

vbauzysvmware (approved):

vbauzysvmware:
could you check, maybe in config-gen we could vpshere provider to add metadata in storage profile :)

Didainius (approved):

adambarreiro:
> could you check, maybe in config-gen we could use vpshere provider to add metadata in storage profile :)

There seems to be a resource for storage policies: https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs/resources/vm_storage_policy
But if we managed to create, still we'd need to add metadata through VCD.

vbauzysvmware:
> > could you check, maybe in config-gen we could use vpshere provider to add metadata in storage profile :)
> 
> There seems to be a resource for storage policies: https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs/resources/vm_storage_policy But if we managed to create, still we'd need to add metadata through VCD.

Yes, you are right, even we could import resource we still add metadata on vpshere level, not VCD

Issue #753, opened 2021-11-30
Add dnsServers to vcd_nsxt_network_dhcp


forestEagle:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment
* 

### Description

Hi,
Since the vCloud Director web ui (and API) now supports adding the DNS option to the NSX-T Segment DHCP server, I'd like to see this implemented in the Terraform resource.

Thanks :)

### New or Affected Resource(s)

* vcd_nsxt_network_dhcp

nakasix:
Hello,

Yes this feature would be very useful for me too :)

`vcd_nsxt_network_dhcp => https://{vcd_host}/tenant/{tenant}/domain-networking/org-networks/urn:vcloud:network:{network_id}/ipam/dhcp-nsxt/dhcp-general`

Also, is it possible to provide DNS functionality for the NSX-T gateway itself?

`vcd_nsxt_edgegateway => https://{vcd_host}/provider/cloud/edge-gateways/cloud/urn:vcloud:gateway:{gateway_id}/ip-management/dns`

Regards,

mikeletux:
Hello @forestEagle and @nakasix ,

I've been working on this feature. The PR is already in place. Please check https://github.com/vmware/terraform-provider-vcd/pull/830

This feature will be aimed for next release v3.7.0. Please keep in mind we release once every quarter.
As soon as the PR that implements this feature is merged, I'll proceed to close this issue.

@nakasix Regarding implementing DNS functionality for the NSX-T gateway, please open up another issue as I'd like to keep this one for the original feature.

Thanks!

mikeletux:
Folks, PR merged into main.
This feature will be available in next release v3.7.0. 

I'm closing this issue. 🚀

PR #830, opened 2022-04-21
Add dns_servers option to vcd_nsxt_network_dhcp resource

mikeletux:
This PR relays on https://github.com/vmware/go-vcloud-director/pull/465

# Description
On issue #753 it is suggested to implement DNS servers to `vcd_nsxt_network_dhcp`, as it is available from VCD 10.3.1.

# Detailed description
This PR introduces the possibility of adding `dns_servers` to the `vcd_nsxt_network_dhcp`resource.
Since `dns_servers` is a feature introduced on VCD 10.3.1 (API v36.1), if these are set and the VCD version the user is working on is below that, it will error. 
 
Successfully tested on VCD 10.3.2 with system administrator as well as org administrator

### DIFF
--- a/vcd/resource_vcd_nsxt_network_dhcp_test.go
+++ b/vcd/resource_vcd_nsxt_network_dhcp_test.go
@@ -35,6 +35,22 @@ func TestAccVcdOpenApiDhcpNsxtRouted(t *testing.T) {
 		return
 	}
 
+	vcdClient := createTemporaryVCDConnection(true)
+	vcdVersionIsLowerThan1031 := func() (bool, error) {
+		if vcdClient != nil && vcdClient.Client.APIVCDMaxVersionIs(">= 36.1") {
+			return false, nil
+		}
+		return true, nil
+	}
+
+	// This case is specific for VCD 10.3.1 onwards since dns servers are not present in previous versions
+	var configText2 string
+	if vcdClient != nil && vcdClient.Client.APIVCDMaxVersionIs(">= 36.1") {
+		params["FuncName"] = t.Name() + "-step2"
+		configText2 = templateFill(testAccRoutedNetDhcpStep3, params)
+		debugPrintf("#[DEBUG] CONFIGURATION for step 2: %s", configText2)
+	}
+
 	resource.Test(t, resource.TestCase{
 		ProviderFactories: testAccProviders,
 		PreCheck:          func() { testAccPreCheck(t) },
@@ -48,9 +64,6 @@ func TestAccVcdOpenApiDhcpNsxtRouted(t *testing.T) {
 						"start_address": "7.1.1.100",
 						"end_address":   "7.1.1.110",
 					}),
-					resource.TestCheckResourceAttr("vcd_nsxt_network_dhcp.pools", "dns_servers.#", "2"),
-					resource.TestCheckResourceAttr("vcd_nsxt_network_dhcp.pools", "dns_servers.0", "8.8.8.8"),
-					resource.TestCheckResourceAttr("vcd_nsxt_network_dhcp.pools", "dns_servers.1", "8.8.4.4"),
 				),
 			},
 			{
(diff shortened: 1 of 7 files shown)


lvirbalas (approved):

vcd/resource_vcd_nsxt_network_dhcp_test.go
```diff
@@ -130,5 +138,7 @@ resource "vcd_nsxt_network_dhcp" "pools" {
     start_address = "7.1.1.130"
     end_address   = "7.1.1.140"
   }
+
+  dns_servers = ["8.8.4.4", "8.8.8.8"]
```
adambarreiro, line 142: So if I understood correctly, this means that order matters for this attribute, right?

mikeletux: That is right!
Those are the DNS servers that are being server through DHCP. Normally you will use the first one and have the second one as backup.

Also in the VCD UI you can see the order if you change it.

adambarreiro: Great, thanks!

lvirbalas, line 142: Actually, we could use these in the examples (more private DNS):

```suggestion
  dns_servers = ["1.1.1.1", "1.0.0.1"]
```

mikeletux: Ok, I'll change that!

vcd/resource_vcd_nsxt_network_dhcp_test.go
```diff
@@ -111,6 +117,8 @@ resource "vcd_nsxt_network_dhcp" "pools" {
     start_address = "7.1.1.100"
     end_address   = "7.1.1.110"
   }
+
+  dns_servers = ["8.8.8.8", "8.8.4.4"]
```
Didainius, line 121: Will the test work if I run it on version lower than 10.3.1?

mikeletux: You are right. The test it's going to fail for versions lower than 10.3.1. Should I create the `dns_servers` var conditionally depending version?

Didainius: Two options:
* Either another test with 10.3.1 specific features
* Either you can add a conditional "TestStep" (using `TestStep.SkipFunc`). Just have in mind that calling `templateFill` in the setup phase also creates a binary test definition so it also needs to be skipped for older versions

vcd/resource_vcd_nsxt_network_dhcp_test.go
mikeletux: @lvirbalas already changed!

vcd/resource_vcd_nsxt_network_dhcp_test.go
mikeletux: @Didainius I've created another test step that has set `TestStep.SkipFunc`, so if version is lower than 10.3.1 that testStep is not executed.
Also I conditionally create the config text using templateFill if vcd version is equal or greater than 10.3.1

vcd/resource_vcd_nsxt_network_dhcp_test.go
```diff
@@ -35,6 +35,22 @@ func TestAccVcdOpenApiDhcpNsxtRouted(t *testing.T) {
...
+	vcdVersionIsLowerThan1031 := func() (bool, error) {
+		if vcdClient != nil && vcdClient.Client.APIVCDMaxVersionIs(">= 36.1") {
+			return false, nil
+		}
+		return true, nil
+	}
+
+	// This case is specific for VCD 10.3.1 onwards since dns servers are not present in previous versions
+	var configText2 string
+	if vcdClient != nil && vcdClient.Client.APIVCDMaxVersionIs(">= 36.1") {
```
vbauzysvmware, line 48: this check isn't needed I think, as vcdVersionIsLowerThan1031 will manage the skip part

vbauzysvmware (approved):

vcd/resource_vcd_nsxt_network_dhcp_test.go
mikeletux: I did that check because @Didainius suggested to avoid calling `templateFill` function as per his comment https://github.com/vmware/terraform-provider-vcd/pull/830#discussion_r855130417

Didainius (approved):

PR #834, opened 2022-04-25
Fix description of rights in role and rights bundle resource and data source

adambarreiro:
I was playing around with **role** & **rights_bundle** data sources and resources and found a small but probably misleading "typo" in the docs when I tried to use `concat` in two right lists, which is not possible as they're sets (one needs to use `setunion` instead). Not a big deal but could be annoying for someone.

This simple PR tunes it for a more accurate description.

Let me know if a changelog entry is needed.


dataclouder (approved):

Didainius (approved):

Didainius:
Changelog note :)

lvirbalas (approved):

mikeletux (approved):

vbauzysvmware (approved):
:)

PR #837, opened 2022-04-27
TestAccVcdCatalogPublishedToExternalOrg workaround to avoid having this test faling

mikeletux:
This PR is not associated with any issue

# Description
This PR instroduces a code snippet that avoids the test TestAccVcdCatalogPublishedToExternalOrg from failing due to catalog publishing settings not set appropriately. 

# Detailed description
Since there is a bug in VCD that randomly changes catalog publishing settings at org level, we introduce this workaround that does the following:
- If executed as a sys admin, it sets the right parameters to org level to avoid the test from failing due to that.
- If executed as org admin, since it cannot change those parameters, it checks if all parameters are enabled. If so, it execute the test. If there are not set appropriately, the test is skipped.

Signed-off-by: Miguel Sama <msama@vmware.com>

### DIFF
--- a/vcd/resource_vcd_catalog_test.go
+++ b/vcd/resource_vcd_catalog_test.go
@@ -256,6 +256,38 @@ func TestAccVcdCatalogPublishedToExternalOrg(t *testing.T) {
 		"PreserveIdentityInformationUpdate2": false,
 	}
 
+	// This code snippet is to avoid having the org catalog publishing settings set to disable.
+	// There are some bugs in VCD that disable those options. This code snippet will be removed
+	// as soon as those bugs are solved.
+	vcdClient := createTemporaryVCDConnection(false)
+	adminOrg, err := vcdClient.GetAdminOrg(testConfig.VCD.Org)
+	if err != nil {
+		t.Errorf("couldn't retrieve the adminOrg for setting workaround for VCD bug - %s", err)
+	}
+
+	if vcdClient.Client.IsSysAdmin {
+		adminOrg.AdminOrg.OrgSettings.OrgGeneralSettings.CanPublishCatalogs = true
+		adminOrg.AdminOrg.OrgSettings.OrgGeneralSettings.CanPublishExternally = true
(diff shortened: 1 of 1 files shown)


vcd/resource_vcd_catalog_test.go
```diff
@@ -256,6 +256,38 @@ func TestAccVcdCatalogPublishedToExternalOrg(t *testing.T) {
 		"PreserveIdentityInformationUpdate2": false,
 	}
 
+	// This code snippet is to avoid having the org catalog publishing settings set to disable.
+	// There are some bugs in VCD that disable those options. This code snippet will be removed
+	// as soon as those bugs are solved.
```
vbauzysvmware, line 261: add TODO tag here :)

vbauzysvmware (approved):

vcd/resource_vcd_catalog_test.go
mikeletux: Done! Thanks for the suggestion!

lvirbalas (approved):

vcd/resource_vcd_catalog_test.go
```diff
@@ -256,6 +257,43 @@ func TestAccVcdCatalogPublishedToExternalOrg(t *testing.T) {
...
 
+	if vcdShortTest {
+		t.Skip(acceptanceTestsSkipped)
+		return
+	}
+
+	// TODO - This code snippet is to avoid having the org catalog publishing settings set to disable.
+	// There are some bugs in VCD that disable those options. This code snippet will be removed
+	// as soon as those bugs are solved.
+	vcdClient := createTemporaryVCDConnection(false)
```
Didainius, line 268: You can make this work even when run in Org user mode by using `createSystemTemporaryVCDConnection` instead of `createTemporaryVCDConnection`.

mikeletux: @Didainius changed and tested. It works with both sys admin and org admin

adambarreiro (approved):
Great 🚀

Didainius (approved):

PR #844, opened 2022-05-03
Upgrade Terraform plugin SDK

vbauzysvmware:
* library update
* fixing issues like this (`* Check 10/23 error: vcd_nsxv_firewall_rule.vms: list or set attribute 'source.0.org_networks' must be checked by element count key (source.0.org_networks.#) or element value keys (e.g. source.0.org_networks.0). Set element value checks should use TestCheckTypeSet functions instead.`) in tests.

Tested with VCD 10.3.2


Didainius (approved):
Thanks - I am good with PR as long as tests passed here.

lvirbalas (approved):
Thanks!

mikeletux (approved):
LGTM!

adambarreiro (approved):
LGTM!

PR #845, opened 2022-05-04
Add Security Tags resource

mikeletux:
There is not issue associated with this PR.

# Description
This PR brings to the Terraform provider a new resource to create security tags on VCD.

# Detailed description
With this PR, the Terraform provider for VCD introduces the possibility of creating security tags via new `vcd_security_tag` resource.
With this resource, a new security tag can be created and can be associated to several VMs. An example would be:
```
resource "vcd_security_tag" "my_tag" {
  name = "test-tag"
  vm_ids = [vcd_vm.my-vm-one.id, vcd_vm.my-vm-two.id]
}
```
Where `name` is the security tag name and `vm_ids` a list of VM IDs that this security tag is going to be applied to.

Alongside this resource it comes acceptance tests as well as Terraform documentation on how to use this new resource.


website/docs/r/security_tag.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+
+* `name` - (Required) The name of the security tag.
+* `vm_ids` - (Required) List of VM IDs that the security tag is going to be applied to.
+
+# Importing
+
+~> **Note:** The current implementation of Terraform import can only import resources into the state.
+It does not generate configuration. [More information.](https://www.terraform.io/docs/import/)
+
+An existing security tag can be [imported][docs-import] into this resource
```
vbauzysvmware, line 39: don't need comment about dot

website/docs/r/security_tag.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+via supplying the full dot separated path to the security tag. An example is
+below:
+
+```
+terraform import vcd_security_tag.my-tag my-security-tag-name
+```
+
+NOTE: the default separator (.) can be changed using Provider.import_separator or variable VCD_IMPORT_SEPARATOR
```
vbauzysvmware, line 47: don't need to comment about the dot

vbauzysvmware (commented):
I will play manually to check some things.

website/docs/r/security_tag.html.markdown
mikeletux: Removed!

website/docs/r/security_tag.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+  vm_ids = [vcd_vm.my-vm-one.id, vcd_vm.my-vm-two.id]
+}
+```
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the security tag.
+* `vm_ids` - (Required) List of VM IDs that the security tag is going to be applied to.
+
```
Didainius, line 33: We don't usually put `## Attribute ` section, but could you just mention here that the `id` of `vcd_security_tag` is set to `name` because no real ID for security tag exists.

website/docs/r/security_tag.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+# Importing
+
+~> **Note:** The current implementation of Terraform import can only import resources into the state.
+It does not generate configuration. [More information.](https://www.terraform.io/docs/import/)
+
+An existing security tag can be [imported][docs-import] into this resource. An example is below:
+
+```
+terraform import vcd_security_tag.my-tag my-security-tag-name
```
Didainius, line 42: The question here remains - how do I specify `org` as I can see in UI that creating a `tag` in one organization does not show it in other one.

vcd/resource_vcd_security_tag.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"context"
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+)
+
+func resourceVcdOpenApiSecurityTag() *schema.Resource {
```
Didainius, line 11: IMO you can drop the `OpenApi` substring for function names as it doesn't give much value in this context and make it more readable.

```suggestion
func resourceVcdSecurityTag() *schema.Resource {
```

vcd/resource_vcd_security_tag.go
```diff
@@ -0,0 +1,113 @@
...
+	securityTag := &types.SecurityTag{
+		Tag:      securityTagName,
+		Entities: convertSchemaSetToSliceOfStrings(d.Get("vm_ids").(*schema.Set)),
+	}
+	_, err := vcdClient.UpdateSecurityTag(securityTag)
+	if err != nil {
+		return diag.Errorf("error when setting up security tags - %s", err)
+	}
+
+	d.SetId(securityTagName)
```
Didainius, line 54: For background - we had (and still have in some places) quite a long history of settings `id` fields with names that bit us. Could you just put a comment that Security Tags don't have real `ID` and this is the reason we use names. Our future selves might thank for it.

vcd/resource_vcd_security_tag.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return resourceVcdOpenApiSecurityTagRead(ctx, d, meta)
+}
+
+func resourceVcdOpenApiSecurityTagRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	vcdClient := meta.(*VCDClient)
+
+	securityTagName := d.Id()
+	taggedEntities, err := vcdClient.GetSecurityTaggedEntities(fmt.Sprintf("tag==%s", securityTagName))
+	if err != nil {
```
Didainius, line 64: I must remind the case of handling the error if the object is deleted outside of Terraform 

https://github.com/vmware/terraform-provider-vcd/pull/800#discussion_r825335060

Probably the most fluent way to achieve this would be to have a dedicated function `vcdClient.GetSecurityTaggedEntityByName` which returns `govcd.ErrorEntityNotFound` if that exact item is not found.

vcd/resource_vcd_security_tag.go
```diff
@@ -0,0 +1,113 @@
...
+	return &schema.Resource{
+		CreateContext: resourceVcdOpenApiSecurityTagCreate,
+		ReadContext:   resourceVcdOpenApiSecurityTagRead,
+		UpdateContext: resourceVcdOpenApiSecurityTagCreate,
+		DeleteContext: resourceVcdOpenApiSecurityTagDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceVcdOpenApiSecurityTagImport,
+		},
+
+		Schema: map[string]*schema.Schema{
```
Didainius, line 21: I think we're missing `org` field as I can create a security tag in one org, but don't see it in other - that means that it is somehow bound to a particular Org I created it in.

website/docs/r/security_tag.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+Provides a VMware Cloud Director Security Tag resource. This can be
+used to assign security tag to VMs.
+---
+
+# vcd\_security\_tag
+
+Provides a VMware Cloud Director Security Tag resource. This can be
+used to assign security tag to VMs.
+
+Supported in provider *v3.7+* and requires VCD 10.3.1+
```
Didainius, line 15: ?

```suggestion
Supported in provider *v3.7+* and requires VCD 10.3.0+
```

vcd/resource_vcd_security_tag.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return diag.Errorf("error retrieving tagged entities - %s", err)
+	}
+
+	readEntities := make([]string, len(taggedEntities))
+	for i, entity := range taggedEntities {
+		readEntities[i] = entity.ID
+	}
+
+	d.Set("vm_ids", convertStringsToTypeSet(readEntities))
```
Didainius, line 73: For a complicated structures (not single fields) we usually handle the error - in fact it has uncovered some bugs in the past. Also `coverity` security scanner would complain about unhandled error unless we use `dSet`

website/docs/r/security_tag.html.markdown
mikeletux: Add a comment about it

website/docs/r/security_tag.html.markdown
mikeletux: Changed so now an org need to be specified

vcd/resource_vcd_security_tag.go
mikeletux: Function name changed

vcd/resource_vcd_security_tag.go
mikeletux: Add a comment on code that highlight that

vcd/resource_vcd_security_tag.go
mikeletux: My fault. Added a new function on govcd called `Org.GetAllSecurityTaggedEntityByName` that wraps `Org.GetAllSecurityTaggedEntities` and returns `ErrorEntityNotFound` if nothing was found.
Also sets the Id to empty if that error is received.

website/docs/r/security_tag.html.markdown
mikeletux: Changed. I mixed things up with another feature that was only available for 10.3.1+

vcd/resource_vcd_security_tag.go
mikeletux: Now those errors are handled!

website/docs/r/security_tag.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+```
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `name` - (Required) The name of the security tag.
+* `vm_ids` - (Required) List of VM IDs that the security tag is going to be applied to.
+
+**Important**:
```
Didainius, line 35: I think you can use the `Note` syntax instead of `**Important**` which is quite common in terraform docs

```-> The ID of `vcd_security_tag` is set to its name since VCD behind the scenes doesn't create an ID.```

Would render to (you can check it in https://registry.terraform.io/tools/doc-preview):
image

File: vcd/resource_vcd_security_tag.go [hunk 12]
```diff
@@ -0,0 +1,167 @@
...
+	org, err := vcdClient.GetOrgByName(orgName)
+	if err != nil {
+		return nil, fmt.Errorf(errorRetrievingOrg, orgName)
+	}
+
+	filter := url.Values{
+		"filter": []string{"tag==" + securityTag},
+	}
+
+	taggedEntities, err := org.GetAllSecurityTaggedEntities(filter)
```
Didainius, line 148: Probably would be more convenient to use the new function `GetAllSecurityTaggedEntitiesByName`

vcd/resource_vcd_security_tag_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"strings"
+	"testing"
+)
+
+func TestAccVcdSecurityTag(t *testing.T) {
```
Didainius, line 11: This test crashes on versions that don't support Security tags (<10.3). (needs a conditional skip)
```
=== RUN   TestAccVcdSecurityTag
    resource_vcd_security_tag_test.go:38: Step 1/2 error: Error running apply: exit status 1

        Error: error when setting up security tags - error getting minimum required API version: endpoint '1.0.0/securityTags' requires API version to support at least '36.0'. Maximum supported version in this instance: '35.0'

          with vcd_security_tag.testaccvcdsecuritytag-tag1,
          on terraform_plugin_test.tf line 21, in resource "vcd_security_tag" "testaccvcdsecuritytag-tag1":
          21: resource "vcd_security_tag" "testaccvcdsecuritytag-tag1" {

    testing_new.go:70: Error running post-test destroy, there may be dangling resources: error retrieving Org datacloud security tags
--- FAIL: TestAccVcdSecurityTag (187.40s)
FAIL
```

vcd/resource_vcd_security_tag_test.go
```diff
@@ -0,0 +1,170 @@
...
+	tag1 := strings.ToLower(t.Name() + "-tag1") // security tags are always lowercase in serverside
+	tag2 := strings.ToLower(t.Name() + "-tag2")
+
+	var params = StringMap{
+		"Org":          testConfig.VCD.Org,
+		"Vdc":          testConfig.VCD.Vdc,
```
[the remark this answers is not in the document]
mikeletux: You were right, changed!

website/docs/r/security_tag.html.markdown
```diff
@@ -0,0 +1,51 @@
...
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `name` - (Required) The name of the security tag.
+* `vm_ids` - (Required) List of VM IDs that the security tag is going to be applied to.
+
+-> The ID of `vcd_security_tag` is set to its name since VCD behind the scenes doesn't create an ID.
+
+# Importing
+
+~> **Note:** The current implementation of Terraform import can only import resources into the state.
+It does not generate configuration. [More information.](https://www.terraform.io/docs/import/)
```
vbauzysvmware, line 40: Please return comments about `.` usage as now org.securitytag is used :)

vbauzysvmware (approved):

vcd/resource_vcd_security_tag_test.go
```diff
@@ -0,0 +1,191 @@
...
+  os_type          = "sles10_64Guest"
+  hardware_version = "vmx-14"
+  catalog_name     = "{{.Catalog}}"
+  boot_image       = "{{.Media}}"
+}
+
+resource "vcd_security_tag" "{{.SecurityTag1}}" {
+  org    = "{{.Org}}"
+  name   = "{{.SecurityTag1}}"
+  vm_ids = [vcd_vapp_vm.{{.VmName}}.id]
```
lvirbalas, line 102: In the test below, I like how you test **two** security tags for the same VM, esp. as issues usually show up with >1 entities.

Similarly here, could we test this single SecurityTag1, but with **two** VMs in vm_ids?

website/docs/r/security_tag.html.markdown
mikeletux: Done!

vcd/resource_vcd_security_tag_test.go
mikeletux: @lvirbalas I've updated the test to check that security tags can be applied to a several VMs at the same time.
Also I've written the additional function `testAccCheckSecurityTagOnVMCreated` which checks that a tag has been applied successfully to a VM.

lvirbalas: Perfect, thanks!

lvirbalas (approved):

adambarreiro:
Just a nit-like suggestion, to rename the PR to something more specific like _"Add Security Tags resource"_

adambarreiro (approved):
Great work 💪

PR #849, opened 2022-05-11
Add independent disk sharing_type missing value

dataclouder:
The independent disk `sharing_type` property can have two active modes, which are already provided, but there is also an inactive mode (`None`) which was not recognized by the validation function.


Didainius (approved):

vbauzysvmware (approved):
LGTM. I tried to create disk with "None" and it worked. Made things simpler :)

lvirbalas (approved):

adambarreiro (approved):
LGTM! 👍🏻

mikeletux (approved):

PR #852, opened 2022-05-13
Add CloudInit Guest Customization example guide

Didainius:
This PR adds an example to Guest Customization docs in our Guides section. It demonstrates verified execution of a script using CloudInit mechanism in Ubuntu 22.04.

### DIFF
--- a/website/docs/guides/vm_guest_customization.html.markdown
+++ b/website/docs/guides/vm_guest_customization.html.markdown
@@ -138,7 +138,7 @@ sudo echo "Command executed as 'root' user" >> /tmp/setup.log
 ```
 
 VM definition might be different, but key here is `guest_properties`. It will supply needed data for
-CloudInit (including the shell script in file `script.sh`)
+CloudInit (including the base64 encoded shell script in file `script.sh` as required by CloudInit)
 
 ```hcl
 resource "vcd_vapp_vm" "guest-vm" {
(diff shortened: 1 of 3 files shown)


dataclouder (approved):

website/docs/guides/vm_guest_customization.html.markdown
```diff
@@ -105,3 +105,63 @@ After applying the configuration one should be able to login using `core` user a
...
+supplying CloudInit user data.
+
+```sh
+#!/bin/sh
+echo "Command executed as 'ubuntu' user" >> /tmp/setup.log
+sudo echo "Command executed as 'root' user" >> /tmp/setup.log
+```
+
+VM definition might be different, but key here is `guest_properties`. It will supply needed data for
+CloudInit (including the shell script in file `script.sh`)
```
lvirbalas, line 141: Additionally, maybe add a sentence about using base64encode?

lvirbalas (approved):
One suggestion and LGTM! (Read in doc preview tool).

adambarreiro (approved):
Nice!

vbauzysvmware (approved):

website/docs/guides/vm_guest_customization.html.markdown
Didainius: Additionally mentioned base64 encoded file is requirement by CloudInit

mikeletux (approved):

PR #841, opened 2022-05-02
NSX-T Edge Gateway Child resource support for VDC Groups

Didainius:
WIP

PR #854, opened 2022-05-18
vcd_nsxt_alb_edgegateway_service_engine_group fix when no `vdc` is specified in `provider` level 

Didainius:
After PR #841 adding support for a bunch of resources a bug still persisted in `vcd_nsxt_alb_edgegateway_service_engine_group` dependency on VDC field. This PR adds a test for replication of it (commit 5ae834e) it and introduces a fix (commit 67b4157). 

Special thanks goes to @adambarreiro  for finding it.

Acceptance tests pass on 10.2 and 10.3 using tags `network nsxt vdcGroup alb`
Binary tests passed on 10.2 using tag `alb`


vbauzysvmware (approved):

adambarreiro (approved):
Thanks for the fix!

mikeletux (approved):

PR #858, opened 2022-05-27
Add Resource NSX-T Route Advertisement

mikeletux:
This PR relies on govcd PR https://github.com/vmware/go-vcloud-director/pull/478

## Description
This PR brings the possibility of adding route advertisement to an NSX-T Edge Gateway.

## Detailed description 
This new resource implements a new resource for NSX-T Edge Gateway, that allows to publish networks to the NSX-T Tier-0 Gateway. Please bear in mind that for this resource to work, the NSX-T Edge Gateway option "Use dedicate Tier-0 Gateway" must be active, otherwise this resource will fail on its creation.

An example HCL file where this resource is used could be:
```
data "vcd_org_vdc" "my_vdc" {
  org = "my-org"
  name = "my-vdc"
}

data "vcd_nsxt_edgegateway" "my_edge_gateway" {
  owner_id = data.vcd_org_vdc.my_vdc.id
  name     = "my-nsxt-edge-gateway"
}

resource "vcd_nsxt_route_advertisement" "my_route_advertisement" {
  edge_gateway_id = data.vcd_nsxt_edgegateway.my_edge_gateway.id
  enabled = true
  subnets = ["192.168.1.0/24", "192.168.2.0/24"]
}
```

As seen in the HCL code snippet, route advertisement will be enabled on NSX-T Edge Gateway and a couple of subnets will be advertised: 192.168.1.0/24 and 192.168.2.0/24.


website/docs/r/nsxt_route_advertisement.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+  enabled         = true
+  subnets         = ["192.168.1.0/24"]
+}
+```
+
+## Example Usage 2 (Enable route advertisement and publish 192.168.1.0/24 and 192.168.2.0/24)
+
+```hcl
+data "vcd_org_vdc" "my_vdc" {
+  org  = "my-org"
```
vbauzysvmware, line 39: same here

website/docs/r/nsxt_route_advertisement.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful
+  when connected as sysadmin working across different organisations.
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level.
```
vbauzysvmware, line 61: depends on other part answers, but seems isn't needed

vcd/resource_vcd_nsxt_route_advertisement.go
```diff
@@ -0,0 +1,168 @@
...
+				ForceNew: true,
+				Description: "The name of organization to use, optional if defined at provider " +
+					"level. Useful when connected as sysadmin working across different organizations",
+			},
+			"vdc": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "The name of VDC to use, optional if defined at provider level",
+				Deprecated:  "Edge Gateway will be looked up based on 'edge_gateway_id' field",
```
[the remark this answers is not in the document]
lvirbalas: Good catch, @vbauzysvmware 

@mikeletux Do I understand right that there is no need for `vdc` field for this resource, because it is traversed from the required EG?

If yes, what is the reason to add this field?

CC: @Didainius

website/docs/r/nsxt_route_advertisement.html.markdown
```diff
@@ -0,0 +1,80 @@
+---
+layout: "vcd"
+page_title: "VMware Cloud Director: vcd_nsxt_route_advertisement"
+sidebar_current: "docs-vcd-resource-nsxt_route_advertisement"
+description: |-
+Provides a VMware Cloud Director resource for setting route advertisement in a NSX-T Edge Gateway.
```
lvirbalas, line 6: Nit:

```suggestion
Provides a VMware Cloud Director resource for setting route advertisement in an NSX-T Edge Gateway.
```

Reference: https://www.grammar.com/a-vs-an-when-to-use/

website/docs/r/nsxt_route_advertisement.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+---
+
+# vcd\_nsxt\_route\_advertisement
+
+Provides a VMware Cloud Director resource for setting route advertisement in a NSX-T Edge Gateway.
```
lvirbalas, line 11: Nit:

```suggestion
Provides a VMware Cloud Director resource for setting route advertisement in an NSX-T Edge Gateway.
```

website/docs/r/nsxt_route_advertisement.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+
+~> **Note:** This resource requires a NSX-T Edge Gateway. Also, for this resource to work appropriately, the option "Dedicate Tier-0 Gateway" must be enabled. Otherwise, it won't work.
```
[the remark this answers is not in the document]
lvirbalas: @mikeletux Please be more specific in place of "it won't work".

lvirbalas (changes requested):

vcd/resource_vcd_nsxt_route_advertisement.go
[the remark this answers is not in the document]
mikeletux: Removed VDC as you were right, it's not needed for this resource. 

Take into account that for import is still there, as the URL is org.vdc-or-vdc-group.gateway-name as we don't use IDs for edge GW

vcd/resource_vcd_nsxt_route_advertisement.go
```diff
@@ -0,0 +1,168 @@
...
+func resourceVcdNsxtRouteAdvertisementCreateUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	vcdClient := meta.(*VCDClient)
+	vcdClient.lockParentEdgeGtw(d)
+	defer vcdClient.unLockParentEdgeGtw(d)
+
+	var subnets []string
+	enableRouteAdvertisement := d.Get("enabled").(bool)
+	subnetsFromSchema, subnetsArgumentSet := d.GetOk("subnets")
+
+	if subnetsArgumentSet {
```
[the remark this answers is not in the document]
mikeletux: changed it for `ok`

vcd/resource_vcd_nsxt_route_advertisement.go
```diff
@@ -0,0 +1,168 @@
...
+
+	d.SetId(edgeGateway.EdgeGateway.ID)
+
+	return resourceVcdNsxtRouteAdvertisementRead(ctx, d, meta)
+}
+
+func resourceVcdNsxtRouteAdvertisementRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	vcdClient := meta.(*VCDClient)
+
+	orgName := d.Get("org").(string)
```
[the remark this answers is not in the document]
mikeletux: I changed for the helper function you said.
The thing is, doing it that way is a little inefficient since we are retrieving the whole Org just to get its name, but I agree its much safer

vcd/resource_vcd_nsxt_route_advertisement.go
```diff
@@ -0,0 +1,168 @@
...
+
+	return nil
+}
+
+func resourceVcdNsxtRouteAdvertisementDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	vcdClient := meta.(*VCDClient)
+	vcdClient.lockParentEdgeGtw(d)
+	defer vcdClient.unLockParentEdgeGtw(d)
+
+	orgName := d.Get("org").(string)
```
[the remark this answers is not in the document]
mikeletux: changed as per the previous comment

vcd/resource_vcd_nsxt_route_advertisement.go
```diff
@@ -0,0 +1,168 @@
...
+	nsxtEdge, err := vcdClient.GetNsxtEdgeGatewayById(orgName, d.Id())
+	if err != nil {
+		return diag.Errorf("error retrieving NSX-T Edge Gateway: %s", err)
+	}
+
+	err = nsxtEdge.DeleteNsxtRouteAdvertisement(true)
```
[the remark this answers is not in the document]
mikeletux: I can change govcd to always sent the context. VCD UI always sends it, but there are a lot of functions/methods in govcd that allows devs to choose if user ctx is sent or not

vcd/resource_vcd_nsxt_route_advertisement_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+data "vcd_nsxt_edgegateway" "{{.EdgeGw}}" {
+  owner_id = data.vcd_org_vdc.{{.NsxtVdc}}.id
+  name     = "{{.EdgeGw}}"
+}
+
+resource "vcd_nsxt_route_advertisement" "testing" {
+  org = "{{.Org}}"
+  edge_gateway_id = data.vcd_nsxt_edgegateway.{{.EdgeGw}}.id
+  enabled = {{.Enabled}}
```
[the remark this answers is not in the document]
mikeletux: A new test case to check false has been added

website/docs/r/nsxt_route_advertisement.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+
+## Example Usage 1 (Enable route advertisement and publish 192.168.1.0/24)
+
+```hcl
+data "vcd_org_vdc" "my_vdc" {
+  org  = "my-org"
```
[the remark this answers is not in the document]
mikeletux: Added

website/docs/r/nsxt_route_advertisement.html.markdown
mikeletux: added as well

website/docs/r/nsxt_route_advertisement.html.markdown
mikeletux: fixed

vcd/resource_vcd_nsxt_route_advertisement.go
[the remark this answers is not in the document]
vbauzysvmware: You can add better or additional helper function, or deal with provider lvl ORG also.

vcd/resource_vcd_nsxt_route_advertisement.go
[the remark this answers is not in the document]
vbauzysvmware: @Didainius what do you think?

website/docs/r/nsxt_route_advertisement.html.markdown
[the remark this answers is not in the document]
mikeletux: No, you're right. The error that the resource gives is not enough to understand what is happening. I'll handle this situation. Will comment once it's in place

vcd/resource_vcd_nsxt_route_advertisement.go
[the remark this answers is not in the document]
Didainius: Let's just think this way - is there any use case to execute this function without sending tenant context?

vcd/resource_vcd_nsxt_route_advertisement.go
```diff
@@ -0,0 +1,184 @@
...
+					Type: schema.TypeString,
+				},
+			},
+		},
+	}
+}
+
+func resourceVcdNsxtRouteAdvertisementCreateUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	vcdClient := meta.(*VCDClient)
+	vcdClient.lockParentEdgeGtw(d)
```
Didainius, line 56: It is hard to get clarifications, but some resources require a lock on VDC Group when the Edge Gateway is a member of VDC Group. (https://github.com/vmware/terraform-provider-vcd/blob/main/vcd/resource_vcd_nsxt_ip_set.go#L80-L89)

Could you possibly tune-up test `TestAccVcdNsxVdcGroupCompleteMigration` to include route advertisement configuration as well, when the Edge Gateway is in VDC Group. Other option is to create some separate test, but we need to be sure that this interoperates well when the Edge Gateway is in VDC Group - I've been hit by that as well.

vcd/resource_vcd_nsxt_route_advertisement_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"regexp"
+	"strconv"
+	"testing"
+)
+
+func TestAccVcdNsxtRouteAdvertisement(t *testing.T) {
+	preTestChecks(t)
+	skipNoNsxtConfiguration(t)
+
+	isRouteAdvertisementEnable := true
```
Didainius, line 19: It would be too luxurious to skip this test always as our environments do not have a dedicated Edge Gateway yet. The best here would be to invoke `govcd` and temporarily change Edge Gateway to Dedicated with a `defer` function to set it back so that other tests don't brake.

Something similar to https://github.com/vmware/terraform-provider-vcd/blob/main/vcd/resource_vcd_nsxt_alb_pool_test.go#L780-L791

vcd/resource_vcd_nsxt_route_advertisement.go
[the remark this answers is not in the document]
mikeletux: It works as well... but since UI sends it, I'd rather give the option to send it or not.

vcd/resource_vcd_nsxt_route_advertisement.go
```diff
@@ -0,0 +1,168 @@
...
+				Description: "The name of organization to use, optional if defined at provider " +
+					"level. Useful when connected as sysadmin working across different organizations",
+			},
+			"vdc": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "The name of VDC to use, optional if defined at provider level",
+				Deprecated:  "Edge Gateway will be looked up based on 'edge_gateway_id' field",
+			},
```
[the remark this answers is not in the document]
mikeletux: I've removed it already 😄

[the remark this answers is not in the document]
Didainius: Yeah, I started reviewing before someone else saw it :)

lvirbalas (approved):

vcd/resource_vcd_nsxt_route_advertisement.go
mikeletux: @Didainius so I've added a conditional lock like other resources have so in case the NSX-T edge gateway is attached to a vdc group it will lock the group, and if that's not the case it'll lock the edge gateway otherwise.

Also please check that I've added a new test called `TestAccVcdNsxtRouteAdvertisementVdcGroup` that checks route advertisement fits good with vdc groups. I have tested just one case because it's exactly the same as testing it in a normal edge gateway, so we save some testing time.

Also this PR introduces new two fields `testConfig.Nsxt.VdcGroup` and `testConfig.Nsxt.VdcGroupEdgeGateway` to testConfig to use 😄  I'll update the PR description to reflect that.

Didainius (commented):
This looks good - thanks for a adding the VDC Group test. 
A few tiny bits and a data source code part should be all thats needed.

website/docs/r/nsxt_route_advertisement.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+}
+
+resource "vcd_nsxt_route_advertisement" "my_route_advertisement" {
+  edge_gateway_id = data.vcd_nsxt_edgegateway.my_edge_gateway.id
+  enabled         = true
+  subnets         = ["192.168.1.0/24"]
+}
+```
+
+## Example Usage 2 (Enable route advertisement and publish 192.168.1.0/24 and 192.168.2.0/24)
```
vbauzysvmware, line 35: I think second example is enough:) as their are almost identical

vbauzysvmware (commented):
LGTM and I think the data source is missing?

vcd/resource_vcd_nsxt_route_advertisement.go
```diff
@@ -0,0 +1,216 @@
...
+	vcdClient := meta.(*VCDClient)
+
+	// Handling locks on a route advertisement is conditional. There are two scenarios:
+	// * When the parent Edge Gateway is in a VDC - a lock on parent Edge Gateway must be acquired
+	// * When the parent Edge Gateway is in a VDC Group - a lock on parent VDC Group must be acquired
+	// To find out parent lock object, Edge Gateway must be looked up and its OwnerRef must be checked
+	// Note. It is not safe to do multiple locks in the same resource as it can result in a deadlock
+	parentEdgeGatewayOwnerId, _, err := getParentEdgeGatewayOwnerId(vcdClient, d)
+	if err != nil {
+		return diag.Errorf("[routed network create v2] error finding parent Edge Gateway: %s", err)
```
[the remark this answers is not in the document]
mikeletux: Addressed!

vcd/resource_vcd_nsxt_route_advertisement_test.go
```diff
@@ -0,0 +1,368 @@
...
+		"NsxtVdcGroup":      testConfig.Nsxt.VdcGroup,
+		"EdgeGwVdcGroup":    testConfig.Nsxt.VdcGroupEdgeGateway,
+		"Enabled":           strconv.FormatBool(isRouteAdvertisementEnable),
+		"Subnet1Cidr":       subnet1,
+		"NsxtImportSegment": testConfig.Nsxt.NsxtImportSegment,
+	}
+
+	params["FuncName"] = t.Name() + "-step4"
+	configText4 := templateFill(testAccNsxtRouteAdvertisementCreationVDCGroup, params)
+	debugPrintf("#[DEBUG] CONFIGURATION for step 4: %s", configText4)
```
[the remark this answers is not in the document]
mikeletux: Changed context to its tests 😄

website/docs/r/nsxt_route_advertisement.html.markdown
mikeletux: Got rid of example1 !

website/docs/r/nsxt_route_advertisement.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+~> **Note:** The current implementation of Terraform import can only import resources into the state.
+It does not generate configuration. [More information.](https://www.terraform.io/docs/import/)
+
+An existing NSX-T route advertisement can be [imported][docs-import] into this resource
+via supplying the full dot separated path for NSX-T Edge Gateway. An example is below:
+
+[docs-import]: https://www.terraform.io/docs/import/
+
+```
+terraform import vcd_nsxt_route_advertisement.my-route-advertisement my-org.my-org-vdc-org-vdc-group-name.my-edge-gw
```
dataclouder, line 56:
```suggestion
terraform import vcd_nsxt_route_advertisement.my-route-advertisement my-org.my-org-vdc-or-vdc-group-name.my-edge-gw
```

vcd/datasource_vcd_nsxt_route_advertisement.go
```diff
@@ -0,0 +1,73 @@
...
+			},
+			"enabled": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Define if route advertisement is active",
+			},
+			"subnets": {
+				Type:        schema.TypeSet,
+				Computed:    true,
+				Description: "Set of subnets that will be advertised to Tier-0 gateway. Leaving it empty means none",
```
Didainius, line 33:
```suggestion
				Description: "Set of subnets that will be advertised to Tier-0 gateway. Empty means none",
```

vcd/datasource_vcd_nsxt_route_advertisement.go
```diff
@@ -0,0 +1,73 @@
...
+			},
+			"edge_gateway_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "NSX-T Edge Gateway ID in which route advertisement is located",
+			},
+			"enabled": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Define if route advertisement is active",
```
Didainius, line 28:
```suggestion
				Description: "Defines if route advertisement is active",
```

vcd/datasource_vcd_nsxt_route_advertisement.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return diag.Errorf("error while retrieving route advertisement - %s", err)
+	}
+
+	dSet(d, "enabled", routeAdvertisement.Enable)
+
+	subnetSet := convertStringsToTypeSet(routeAdvertisement.Subnets)
+	err = d.Set("subnets", subnetSet)
+	if err != nil {
+		return diag.Errorf("error while setting subnets argument")
```
Didainius, line 67:
```suggestion
		return diag.Errorf("error while setting subnets argument: %s", err)
```

vcd/resource_vcd_nsxt_route_advertisement_test.go
```diff
@@ -0,0 +1,368 @@
...
+			{
+				Config: configText1,
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestMatchResourceAttr("vcd_nsxt_route_advertisement.testing", "id", regexp.MustCompile(`^urn:vcloud:gateway:.*$`)),
+					resource.TestCheckResourceAttr("vcd_nsxt_route_advertisement.testing", "enabled", strconv.FormatBool(isRouteAdvertisementEnable)),
+					resource.TestCheckResourceAttr("vcd_nsxt_route_advertisement.testing", "subnets.#", "1"),
+					resource.TestCheckResourceAttr("vcd_nsxt_route_advertisement.testing", "subnets.0", subnet1),
+				),
+			},
+			{
```
Didainius, line 73: Can you pull in additional test step which check that data source functions and returns the same fields which were set. You can look for inspiration how to do this in many places, but here is one I remember doing so:
https://github.com/vmware/terraform-provider-vcd/blob/main/vcd/resource_vcd_nsxt_alb_edgegateway_service_engine_group_test.go#L177

adambarreiro (approved):
Nice work, thanks!

vcd/datasource_vcd_nsxt_route_advertisement.go
mikeletux: Changed, thanks!

vcd/datasource_vcd_nsxt_route_advertisement.go
mikeletux: Changed in resource and datasource, thanks!

vcd/datasource_vcd_nsxt_route_advertisement.go
mikeletux: Changed in resource and datasource, thanks!

vcd/resource_vcd_nsxt_route_advertisement_test.go
mikeletux: All right, I've added a new test step for testing datasource. This new test step is the number 3. The former number 3 has been renamed to 4. 

The test step can be found here: https://github.com/mikeletux/terraform-provider-vcd/blob/c1690c2df11da7fa331913ce10e64e456c7a81e6/vcd/resource_vcd_nsxt_route_advertisement_test.go#L87-L97

The template being used comes with the `# skip-binary-test` directive to avoid including it in the binary test as suggested: https://github.com/mikeletux/terraform-provider-vcd/blob/c1690c2df11da7fa331913ce10e64e456c7a81e6/vcd/resource_vcd_nsxt_route_advertisement_test.go#L155-L160

website/docs/r/nsxt_route_advertisement.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+sidebar_current: "docs-vcd-resource-nsxt_route_advertisement"
+description: |-
+Provides a VMware Cloud Director resource for setting route advertisement in an NSX-T Edge Gateway.
+---
+
+# vcd\_nsxt\_route\_advertisement
+
+Provides a VMware Cloud Director resource for setting route advertisement in an NSX-T Edge Gateway.
+
+~> **Note:** This resource requires an NSX-T Edge Gateway. Also, for this resource to work appropriately, the option "Dedicate Tier-0 Gateway" must be enabled. Otherwise, route advertisement creation will fail.
```
vbauzysvmware, line 13: maybe worth to make reference to `vcd_nsxt_edgegateway.dedicate_external_network` for option "Dedicate Tier-0 Gateway ?

Didainius (approved):
Thanks for all the fixes

PR #856, opened 2022-05-18
Add CSE guide to install it with the VCD Provider

adambarreiro:
Signed-off-by: abarreiro <abarreiro@vmware.com>

PR #876, opened 2022-06-29
Specify that only TKGm clusters are supported in CSE guide

adambarreiro:
During #856 there was a lack of precision about the type of clusters supported by the CSE guide and examples created. This tiny PR just adds/modifies pieces of text to specify that only TKGm clusters are supported.


lvirbalas (approved):
Perfect.

Didainius (approved):

PR #882, opened 2022-07-18
Update CSE guide with some hints and improvements in the process

adambarreiro:
## Description

This PR complements #856 and #876 with extra hints and steps that were not taken into account in the mentioned changes.

- Mentions TKGm version used in downloaded OVAs in the example, to let customers to have more context where the OVA comes from.
- Onboards the service account role mentioned in [CSE documentation](https://vmware.github.io/container-service-extension/cse3_0/CSE_INSTALL_PREREQUISITES.html) to be used instead of system administrator for CSE installation.

PS: The list of rights for the role is taken from [here](https://github.com/vmware/container-service-extension/blob/master/container_service_extension/common/constants/server_constants.py#L596).

## Tests

- [x] Example HCL finishes correctly in VCD 10.3.2
- [x] Cluster successfully created
- [x] Tested with CSE 3.1.4


examples/container-service-extension-3.1.x/3.7-cse-3.1.x-install.tf
```diff
@@ -230,6 +230,110 @@ resource "vcd_catalog" "cat-cse" {
   delete_recursive = "true"
 }
 
+# Create a role at provider level to be used to manage CSE (recommended).
+# You can create a `vcd_org_user` that uses this role and configure the Terraform provider to use it instead of System Administrator (recommended).
+# Here we continue to use System Administrator for simplicity and to configure CSE in one run.
```
lvirbalas, line 235: Hm. The last sentence is a bit confusing, i.e.: why we recommend to use Service Role, but then do not use it ourselves?

examples/container-service-extension-3.1.x/3.7-cse-3.1.x-install.tf
```diff
@@ -230,6 +230,110 @@ resource "vcd_catalog" "cat-cse" {
   delete_recursive = "true"
 }
 
+# Create a role at provider level to be used to manage CSE (recommended).
```
lvirbalas, line 233: As this is a new item, I suggest we name it in the first sentence:

```suggestion
# Create a service role at provider level to be used to manage CSE (recommended).
```

examples/container-service-extension-3.1.x/3.7-cse-3.1.x-install.tf
adambarreiro: Actually your comment gave me an idea, I'll see if it works and it could leave the HCL in a more understandable state. (I'll try to use `alias`).
https://www.terraform.io/language/providers/configuration

adambarreiro: Done, I refactored the HCL to use Terraform alias feature. I think it's nicer this way:
- Avoids the confusing comment and taking for granted that all customers will understand what to do.
- Preserves the "single-run" approach, just one `terraform apply` and that's it.

The idea here is that we have two provider configs, one aliased as `administrator` with Sysadmin powers, and one without alias that uses the CSE Service account. The `administrator` one creates this Service Account and some other required items (Orgs, VDCs, ALBs, etc) while the CSE Service account one creates the remaining elements.

I tested it and it worked, but I'll re-test again with a fresh VCD as I did some last-minute changes and I want to be 100% sure it works.

examples/container-service-extension-3.1.x/3.7-cse-3.1.x-install.tf
```diff
@@ -25,12 +26,142 @@ terraform {
   }
 }
 
+# We'll use the System Administrator only to create the service role, the service user to manage CSE and the minimum
```
lvirbalas, line 29: I think we should explain that this is provider alias (some users may miss it). In other words, provide some context on the approach.

examples/container-service-extension-3.1.x/3.7-cse-3.1.x-install.tf
```diff
@@ -25,12 +26,142 @@ terraform {
...
+# objects required (Organization, VDC, ALB, etc). The remaining items will be created by this service user.
+
 provider "vcd" {
```
lvirbalas, line 32: It may be only me, but I'd like to see both providers next to each other, as otherwise it's easy to miss that there are two. Maybe we should move this vcd.administrator provider right below the default one?

examples/container-service-extension-3.1.x/3.7-cse-3.1.x-install.tf
lvirbalas: I like the approach - have made some suggestions in-line. Main train of thought was: many users may not be aware of `alias` and we should try to make it really clear (in the structure of HCL and comments) that there are two providers.

> The only "disadvantage" is that the provider = vcd.administrator expression is quite repetitive, but it's a minor thing imho. We could do it the other way around and having a provider = vcd.cse-service-account, but I preferred to make it clear where admin grants were used.

I was having the same thoughts when reviewing. The question is which is "more default" in this use case? Does the configuration work without administrator rights at all? If not, it feels like administrator should be the default, while the service account - an option, esp. given that we create the service account. But these are just thoughts for discussion. 

One way to weigh the decision is to answer the question: do you have more resources requiring vcd.administrator provider or the service account provider? I.e. which would lead to less references.

adambarreiro: Switched to having the System Admin as the `default` alias, and the CSE Service Account as the `cse-service-account` alias. Feels less cluttered now.

examples/container-service-extension-3.1.x/3.7-cse-3.1.x-install.tf
```diff
@@ -25,12 +26,142 @@ terraform {
...
+
+resource "vcd_org_user" "cse-service-account" {
+  provider = vcd.administrator
+
+  name     = var.service-account-user
+  password = var.service-account-password
+  role     = vcd_role.cse-service-role.name
+}
+
+# Use the created CSE service account to perform all CSE related operations.
```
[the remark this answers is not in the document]
adambarreiro: Changed!

examples/container-service-extension-3.1.x/3.7-cse-3.1.x-install.tf
adambarreiro: Done, let me know your thoughts.

examples/container-service-extension-3.1.x/3.7-cse-3.1.x-install.tf
adambarreiro: Done, I've re-ordered it a bit + some changed explanations.

adambarreiro:
I've updated also the Guide with this alias approach, to have more consistency between the example HCL code and the Guide.

examples/container-service-extension-3.1.x/3.7-cse-3.1.x-install.tf
```diff
@@ -25,13 +26,143 @@ terraform {
   }
 }
 
+# Create a service role at provider level with the minimum rights required to install and manage CSE items.
+# This role will be used by the CSE Service Account defined below.
+
+resource "vcd_role" "cse-service-role" {
```
adambarreiro, line 32: Taken from https://github.com/vmware/container-service-extension/blob/master/container_service_extension/common/constants/server_constants.py#L596

examples/container-service-extension-3.1.x/3.7-cse-3.1.x-install.tf
lvirbalas: It's perfect now :) Thanks for this improvement!

lvirbalas: I'm resolving this conversation, but tagging @Didainius , so he's aware of this change.

lvirbalas (approved):

adambarreiro:
Improved `null_resource.cse-install-script` as doing a `terraform plan` without any change in infrastructure was showing changes, as the resource was forced to always re-create. Now it's only re-creating when a required field is changed.

Didainius (approved):
Thanks!

PR #870, opened 2022-06-17
Update insert media locking to be concise with internal disk locking

vbauzys:
Ref: https://github.com/vmware/terraform-provider-vcd/issues/868

* update deprecated methods
* update insert media locking mechanism to match internal disk locking to avoid race conditions

Signed-off-by: Vaidotas Bauzys <vbauzys@vmware.com>

PR #884, opened 2022-07-26
Update insert media locking to be concise with internal disk locking

Didainius:
This PR builds on top of #870.
The only difference it has is it uses different configuration values to use NSX-T VDC backed storage media for test.

The test passes now when having a proper config:
```
=== RUN   TestAccVcdMediaLockTest
--- PASS: TestAccVcdMediaLockTest (1140.56s)
PASS
```


lvirbalas (approved):
Woohoo! Thank you!

adambarreiro (approved):

PR #880, opened 2022-07-07
Add 'vcd_nsxt_edgegateway_bgp_neighbor' for managing BGP Neighbors

Didainius:
This PR continues work of #861 and #879 and completes full coverage of NSX-T Edge Gateway BGP configuration (3x resources and 3x data sources in total for complete BGP configuration management +1 more resource for route advertisement which is already in `main` #858)

This last PR adds BGP neighbor configuration resource and data source `vcd_nsxt_edgegateway_bgp_neighbor`.

Tested on 10.2.0, 10.2.2, 10.3.3.

**Note.** Once all PRs are merged, additional tests will be done to integrate all components together. This was not done now to avoid making huge PRs containing all code.

PR #879, opened 2022-07-04
Add resource and data source to manage BGP IP Prefix Lists

Didainius:
This PR continues to build on #861 to cover complete BGP configuration (as requested in #755) on NSX-T Edge Gateways.
This PR adds BGP IP Prefix List managemet resource and data source `vcd_nsxt_edgegateway_bgp_ip_prefix_list`

**Note**. For non network savy guys the `greater_than_or_equal_to` and `less_than_or_equal_to` fields in `ip_prefix` block look confusing. VCD itself docs do have a note about it but could do a bit better. I was looking for better explanations and found this one to be explaining it good -> https://community.cisco.com/t5/routing/bgp-prefix-list/m-p/986282/highlight/true#M81916 . 

VCD Docs ->  https://docs.vmware.com/en/VMware-Cloud-Director/10.3/VMware-Cloud-Director-Service-Provider-Admin-Portal-Guide/GUID-F2A3BC91-036A-4E29-A1C9-6EAB8602562E.html

PR #861, opened 2022-06-03
Add NSX-T Edge Gateway BGP Configuration resource and data source

Didainius:
This PR introduces BGP Configuration resource and data source for NSX-T Edge Gateways (as requested in #755).
Resource and data source in this PR will reflect BGP "Configuration" tab in UI.

image

Additional fixes:
* `TestAccDataSourceNotFound/vcd_nsxt_route_advertisement`
```
    datasource_not_found_test.go:93: Step 1/1, expected an error with pattern, no match on: Error running pre-apply refresh: exit stat
us 1

        Error: error while retrieving route advertisement - error in HTTP GET request: BAD_REQUEST - [ 1674343e-bfca-4cde-b467-606d3a6
e79d8 ] Route advertisement is not supported for edge nsxt-gw-datacloud because its not dedicated to an external network.

          with data.vcd_nsxt_route_advertisement.not-existing,
          on terraform_plugin_test.tf line 3, in data "vcd_nsxt_route_advertisement" "not-existing":
           3: data "vcd_nsxt_route_advertisement" "not-existing" {
```
* `TestAccVcdOrgVdcAccessControl` (2 fixes)
```
=== RUN   TestAccVcdOrgVdcAccessControl
    resource_vcd_org_vdc_access_control_test.go:51: Step 2/3 error: Error running apply: exit status 1
        
        Error: open org_user_pwd.txt: no such file or directory
        
          with vcd_org_user.testaccvcdorgvdcaccesscontrol,
          on terraform_plugin_test.tf line 2, in resource "vcd_org_user" "testaccvcdorgvdcaccesscontrol":
           2: resource "vcd_org_user" "testaccvcdorgvdcaccesscontrol" {
        
        
        Error: open org_user_pwd.txt: no such file or directory
        
          with vcd_org_user.testaccvcdorgvdcaccesscontrol2,
          on terraform_plugin_test.tf line 9, in resource "vcd_org_user" "testaccvcdorgvdcaccesscontrol2":
           9: resource "vcd_org_user" "testaccvcdorgvdcaccesscontrol2" {
        
    testing_new.go:84: Error running post-test destroy, there may be dangling resources: expected to have VDC sharing settings set to none and got something else
--- FAIL: TestAccVcdOrgVdcAccessControl (10.40s)
```

PR #887, opened 2022-07-28
Add BGP component integration tests

Didainius:
After merging #861, #879, #880 as separate code chunks one piece was still missing - it is testing that all these components work together. This PR brings **only new tests** which integrate all 3 components into one piece.


lvirbalas (approved):
Great to see these new tests! BTW, how much time all the BGP tests for this feature added to the suite (approx.)?

Didainius:
> Great to see these new tests! BTW, how much time all the BGP tests for this feature added to the suite (approx.)?

Roughly 25-44 minutes depending on env speed.

adambarreiro (approved):
Nice, thanks!

PR #890, opened 2022-08-01
Add CDS to supported versions list

adambarreiro:
This tiny PR just adds CDS to the supported versions list of the documentation.

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -26,6 +26,8 @@ The following Cloud Director versions are supported by this provider:
 * 10.3
 * 10.4
 
+Also Cloud Director Service (CDS) is supported.
+
 ## Example Usage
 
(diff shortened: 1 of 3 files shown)


website/docs/index.html.markdown
```diff
@@ -26,6 +26,8 @@ The following Cloud Director versions are supported by this provider:
 * 10.3
 * 10.4
 
+Also Cloud Director Service (CDS) is supported.
```
lvirbalas, line 29: Do we need to specify backing VCD version here or not really useful?

Didainius (approved):

website/docs/index.html.markdown
lvirbalas: After more discussion, this statement seems great and simple for users to understand (i.e. that we support same CDS' VCD versions as VCD on-prem).

PR #911, opened 2022-09-26
Fix bug that causes Compute Policies to fail on VCD 10.3.0 and 10.3.1

adambarreiro:
## Description

This PR amends the code added in PR #904, where the VM Placement Policies were onboarded to the Provider.

The problem that this PR fixes is that in VCD 10.3.0 and 10.3.1, vGPU Policies do not exist, hence the filter `isVgpuPolicy` causes an error when fetching VM Placement Policies.

To fix this situation, I added two auxiliary functions, `getVgpuFilter`, that returns a `isVgpuPolicy` filter option only if API version is greater or equal than 36.2 (VCD 10.3.2). Second function is `getVgpuFilterToPrepend` which is similar to the first one but it adds a `;` to the right (The `;` acts as a logical AND). This second one is intended to be used to prepend this vGPU filtering to an existing filter string.

## RCA

The issue was identified when executing the acceptance tests in VCD 10.3.0.
PR #904 was tested with VCD 10.3.3 and 10.4.0, which already had vGPU Policy support (this was introduced in VCD 10.3.2), so this issue got undetected.

### DIFF
--- a/vcd/resource_vcd_vm_placement_policy.go
+++ b/vcd/resource_vcd_vm_placement_policy.go
@@ -453,8 +453,8 @@ func getVgpuFilterToPrepend(vcdClient *VCDClient, isVgpu bool) string {
 }
 
 // getFriendlyErrorIfVmPlacementPolicyAlreadyExists is intended to be used when a VM Placement Policy already exists and the provider
-// tries to create another one with the same name. When this happens, VCD discloses a lot of unnecessary information to the user,
-// so this function simplifies the message.
+// tries to create another one with the same name. When this happens, VCD discloses a lot of unnecessary information to the user that is
+// hard to read and understand, so this function simplifies the message.
 // Note: This function should not be needed anymore once VCD 10.4.0 is discontinued.
 func getFriendlyErrorIfVmPlacementPolicyAlreadyExists(vmPlacementPolicyName string, err error) error {
 	if err != nil && strings.Contains(err.Error(), "already exists") {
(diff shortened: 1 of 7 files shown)


vcd/resource_vcd_vm_placement_policy.go
```diff
@@ -365,7 +366,7 @@ func getVmPlacementPolicy(d *schema.ResourceData, meta interface{}, policyId str
 	computePolicy, err = vcdClient.GetVdcComputePolicyV2ById(policyId)
 	if err != nil {
 		queryParams := url.Values{}
-		queryParams.Add("filter", fmt.Sprintf("name==%s;isSizingOnly==false;isVgpuPolicy==false", policyId))
+		queryParams.Add("filter", fmt.Sprintf("%sname==%s;isSizingOnly==false", getVgpuFilterToPrepend(vcdClient, false), policyId))
```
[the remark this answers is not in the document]
adambarreiro: Actually we have the tests to detect this issue. The problem was **the environment** in which I ran them in #904, I only used 10.3.3 and 10.4.0 where this issue could be reproduced in 10.3.0 and 10.3.1 with the current tests.

So I'd say it's more related to running the tests in a more extensive set of environments.

lvirbalas (approved):

vcd/resource_vcd_vm_placement_policy.go
```diff
@@ -439,3 +440,37 @@ func setVmPlacementPolicy(_ context.Context, d *schema.ResourceData, vcdClient *
 	log.Printf("[TRACE] VM Placement Policy read completed: %s", policy.Name)
 	return nil
 }
+
+// getVgpuFilterToPrepend gets a vGPU Policy filter set to `isVgpu` if API version of target VCD is greater than 36.2 (VCD 10.3.2).
```
dataclouder, line 444: Shouldn't all these functions be in go-vcloud-director?
(getVgpuFilterToPrepend, getVgpuFilter, getFriendlyErrorIfVmPlacementPolicyAlreadyExists)

dataclouder (commented):
The tests work well, but there is something to discuss

vcd/resource_vcd_vm_placement_policy.go
adambarreiro: I've moved `getFriendlyErrorIfVmPlacementPolicyAlreadyExists` to go-vcloud-director.

Didainius (approved):

dataclouder (approved):

PR #918, opened 2022-10-12
Fix flaky test TestAccVcdNsxVdcGroupResources, trim certificate newlines

Didainius:
* `TestAccVcdNsxVdcGroupResources` test import sometimes fails due to `vcd_nsxt_ipsec_vpn_tunnel` changing states as the test progresses. The solution is to ignore validation of these fields on import test.
```
=== RUN   TestAccVcdNsxVdcGroupResources
    resource_vcd_vdc_group_common_test.go:667: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

        (map[string]string) (len=2) {
         (string) (len=18) "ike_service_status": (string) (len=11) "NEGOTIATING",
         (string) (len=6) "status": (string) (len=4) "DOWN"
        }


        (map[string]string) (len=2) {
         (string) (len=18) "ike_service_status": (string) "",
         (string) (len=6) "status": (string) ""
        }
--- FAIL: TestAccVcdNsxVdcGroupResources (277.36s)
```
* <s>`TestAccVcdOrgGroup` was failing because of inaccessible LDAP container. It was because test `TestAccVcdNsxtDynamicSecurityGroupIntegration` removed "Allow all traffic" firewall rule. It is normal, because `vcd_nsxt_firewall` destroy operation removed preprovisioned firewall rule. The soluation was to skip this test for binary tests and restore firewall rule after test run using `resetFirewall` function. </s>
* Certificates in `test-resources` directory have their trailining newlines removed as some VCD versions trim them causing inconsistent plan.


lvirbalas (approved):

adambarreiro (approved):
LGTM

Issue #898, opened 2022-08-09
VDC VM placement policies


cbotha:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected Resource(s)

* vcd_org_vdc

Desperately need the capability to add pre-created VM placement policies to VDCs.
UI flow = Resources > Organisation VDCs > Select VDC Name > Policies > VM Placement > ADD policy 

```

### References

<!---
Information about referencing Github Issues: https://help.github.com/articles/basic-writing-and-formatting-syntax/#referencing-issues-and-pull-requests

Are there any other GitHub issues (open or closed) or pull requests that should be linked here? Vendor blog posts or documentation? For example:

--->

* Issue #0000

adambarreiro:
Hi @cbotha, thanks for the suggestion. We will be implementing this new feature in PR #904.

Best regards.

Piedro1:
Hi, glad this feature is also made available. Will it be possible to list and choose a placement policy as a tenant administrator? I would like to test this feature if possible.

adambarreiro:
Hi @cbotha @Piedro1, the request suggested here has been merged to the main branch. It is still not officially released, but you can build the Provider from the source code and try it out. We can help with that if required.

Please note that the complete journey for VM Placement Policies is not 100% complete though, we still need to onboard them to the VM resource/data source to be able to use them. What we merged so far enables users to assign VM Placement Policies to VDCs, creating VM Placement Policies and fetching them.

wizardmatas:
Hi, @adambarreiro , thanks for release.

Could you please share, if new release can assign placement policies to vms and vapps, or current release is only capable of creating policies?

If current release is only for policy management, do you plan, to add policy management to vapp and vms ?

Thanks!

cbotha:
Hi

Thanks for the update.

I tried the new provider this morning, but I can’t seem to get it working.
In my case I would like to add an existing placement policy to a new VDC upon creation.

I added the following to my plan:

data "vcd_provider_vdc" "pvdc" {
  name = local.tenant.vdc_pvdc_name
}

[...]

PR #922, opened 2022-10-25
Add placement policy to vcd_vapp_vm and vcd_vm resources and data sources

adambarreiro:
### Description

This PR:
- Requires https://github.com/vmware/go-vcloud-director/pull/513
- Complements #904, #911
- Completes #898 so it could be closed once merged.

The proposed changes onboard the **new** attribute **placement_policy_id** to `vcd_vapp_vm` and `vcd_vm` resource and data source. This attribute complements **sizing_policy_id**: It assigns the referenced VM Placement Policy to the specified VM.

This VM Placement Policy can be assigned, updated or removed, although it can be removed only if the VM has already a VM Sizing Policy, as the VM **must have** either a VM Sizing Policy or a VM Placement Policy (if the VDC has any assigned, of course). If one tries to remove both, operation should fail.

Both VM Sizing Policy and VM Placement Policy need to be updated in the same API call, as both are "VDC Compute Policies". Otherwise, one or another would get unassigned from the VM.

### Use cases

- A VM picks the Compute Policy automatically from the VDC default on creation:
```
resource "vcd_vm" "foo" {
  ...
}
```
The state should populate either `sizing_policy_id` or `placement_policy_id` with the Compute Policy ID.

- A VM may have **only** a VM Sizing Policy. In this case, it can't be unassigned (in UI you usually can choose between either System Default or another one)

```
resource "vcd_vm" "foo" {
  ...
  sizing_policy_id = vcd_vm_sizing_policy.sizing.id
}
```

- A VM may have **only** a VM Placement Policy. In this case, if we had already created the VM with a Sizing Policy, we need to explicitly set an empty value in the Sizing Policy to unassign it, as it's an Optional+Computed value which would get ignored otherwise.

```
resource "vcd_vm" "foo" {
  ...
  sizing_policy_id = "" # It is computed so the attribute needs this explicit empty value
  placement_policy_id = vcd_vm_placement_policy.placement.id
}
```

- A VM may have both VM Sizing Policy and VM Placement Policy.

```
resource "vcd_vm" "foo" {
  ...
  sizing_policy_id = vcd_vm_sizing_policy.sizing.id
  placement_policy_id = vcd_vm_placement_policy.placement.id
}
```

- A VM must have either a VM Sizing Policy or a VM Placement Policy. Having none should fail:

```
resource "vcd_vm" "foo" {
  ...
  sizing_policy_id = ""
  # placement_policy_id = vcd_vm_placement_policy.placement.id
}
```
should give this error:
```
│ Error: error updating compute policy: either sizing policy ID or placement policy ID is needed
```

### Issues

In the original code, `sizing_policy_id` was Computed and Optional. To preserve compatibility, this didn't change, but adds some complexity/quirks to the way that the combination of sizing_policy_id + placement_policy_id work. The main problem is that **sizing_policy_id** does not trigger an Update if one puts the following snippets **and nothing else changed in config**:

```
  sizing_policy_id = null
  sizing_policy_id = ""
  # sizing_policy_id = "urn:vcloud:..."
```

In the Refresh stage, Terraform will say that the state matches the configuration, unless we put:

```
    sizing_policy_id = "  "  # some spaces there
```

In the code we explicitly trim the spaces so this should be good. I didn't find any way to force an Update when one tries to explicitly set an empty value to this attribute. Terraform simply ignores it during Refresh, as if you didn't specify anything at all.

Also this is not really a big deal as this situation (trying to remove the sizing policy when there is not a placement policy and nothing else changed in config) should always trigger an error (at least one policy is required, if the VDC has assigned policies). So unless someone wants to make the Apply fail, she should never face this scenario.

In versions <= v3.7.0 this was not an issue because we did not support VM Placement Policies, so one could not (and should not) remove the VM Sizing Policy once set. Now, with VM Placement Policies, one can remove the Sizing Policy.

### Tests

- [x] Acceptance tests pass
- [ ] Binary tests pass
- [ ] Upgrade tests pass
- Complementary manual tests:
   - [x] Create a VM in a VDC with several policies and a default sizing policy. It should be created with just the default sizing policy in state. Placement policy is nil.
   - [x] Create a VM in a VDC with several policies and a default placement policy. It should be created with just the default placement policy in state. Sizing policy is empty and Placement policy is not.
   - [x] Create a VM with a defined VM Placement Policy and VM Sizing Policy. State should have both.
   - [x] Update previous VM, removing the VM Placement Policy entirely. Placement Policy should be removed and not appear in state.
   - [x] Update previous VM, putting `sizing_policy_id = " "`. It should error as it needs at least one policy.
   - [x] Change to different VM Placement Policies / VM Sizing Policies. State should show the changes.


vcd/resource_vcd_vapp_vm_tools.go
```diff
@@ -92,34 +92,30 @@ func lookupStorageProfile(d *schema.ResourceData, vdc *govcd.Vdc) (*types.Refere
...
-// does not hold technical details, only references)
-func lookupComputePolicy(d *schema.ResourceData, vcdClient *VCDClient) (*types.VdcComputePolicy, *types.ComputePolicy, error) {
-	var sizingPolicy *types.VdcComputePolicy
-	var vmComputePolicy *types.ComputePolicy
-	if value, ok := d.GetOk("sizing_policy_id"); ok {
-		vdcComputePolicy, err := vcdClient.Client.GetVdcComputePolicyById(value.(string))
+// lookupComputePolicy returns two types of compute policy structures associated to the value of the given Compute Policy attribute:
+// * types.VdcComputePolicyV2 contains ID and all technical configuration which that policy provides
+// * types.Reference structure is used in VM creation to assign a Compute Policy (and does not hold technical details, it's only a reference)
+func lookupComputePolicy(d *schema.ResourceData, vcdClient *VCDClient, computePolicyAttribute string) (*types.VdcComputePolicyV2, *types.Reference, error) {
```
Didainius, line 98: I like the return of `*types.Reference`!

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1390,17 +1417,35 @@ func resourceVmHotUpdate(d *schema.ResourceData, meta interface{}, vmType typeOf
...
+	sizingId, newSizingId := d.GetChange("sizing_policy_id")
+	placementId, newPlacementId := d.GetChange("placement_policy_id")
+
+	sizingPolicyChanged := d.HasChange("sizing_policy_id")
+	placementPolicyChanged := d.HasChange("placement_policy_id")
+	if !sizingPolicyChanged {
+		// As sizing_policy_id is Computed+Optional, the only way to unset it should be to write `sizing_policy_id = ""`
+		// in the HCL. However, when this is done, Terraform SDK doesn't detect this change: d.HasChange() returns false,
+		// d.GetChange returns both old values and d.Get returns old value, hence `sizingPolicyChanged` will be always false.
+		// We need to inspect the raw HCL to get the correct value.
```
Didainius, line 1429: Can you possibly now that this is the case with `terraform-plugin-sdk v2.24.0`. This can change in future and should make it easier to look for new/improved solutions

Didainius (commented):
Pass 1

vcd/datasource_vcd_vapp_vm.go
```diff
@@ -375,6 +375,12 @@ func vcdVmDS(vmType typeOfVm) map[string]*schema.Schema {
 			Computed:    true,
 			Description: "VM sizing policy ID.",
 		},
+		"placement_policy_id": {
+			Type:        schema.TypeString,
+			Optional:    true,
```
lvirbalas, line 380: I'm reading the PR description and from "Use cases" section it seems as if you can't create a VM without neither the sizing_policy_id nor placement_policy_id. Maybe worth stating that option in PR description?

website/docs/r/vapp_vm.html.markdown
```diff
@@ -329,7 +357,8 @@ example for usage details.
 * `cpu_hot_add_enabled` - (Optional; *v3.0+*) True if the virtual machine supports addition of virtual CPUs while powered on. Default is `false`.
 * `memory_hot_add_enabled` - (Optional; *v3.0+*) True if the virtual machine supports addition of memory while powered on. Default is `false`.
 * `prevent_update_power_off` - (Optional; *v3.0+*) True if the update of resource should fail when virtual machine power off needed. Default is `false`.
-* `sizing_policy_id` (Optional; *v3.0+*, *vCD 10.0+*) VM sizing policy ID. Has to be assigned to Org VDC using `vcd_org_vdc.vm_sizing_policy_ids` and `vcd_org_vdc.default_vm_sizing_policy_id`.
+* `sizing_policy_id` (Optional; *v3.0+*, *vCD 10.0+*) VM sizing policy ID. Has to be assigned to [Org VDC](/providers/vmware/vcd/latest/docs/resources/org_vdc) using `vcd_org_vdc.vm_sizing_policy_ids` and optionally `vcd_org_vdc.default_compute_policy_id`.
+* `placement_policy_id` (Optional; *v3.8+*) VM placement policy ID. Has to be assigned to [Org VDC](/providers/vmware/vcd/latest/docs/resources/org_vdc) using `vcd_org_vdc.vm_placement_policy_ids` and optionally `vcd_org_vdc.default_compute_policy_id`.
```
lvirbalas, line 361: I'm comparing these docs to the elaborate PR description. Thinking that we may want to expand here for the users, e.g.: possibly worth adding sentences about how these fields relate to each other? Now it's not clear that there's an implicit relationship between them in terms of being defined and removed (here we state that both are just optional).

lvirbalas (approved):
Looks great! Just would be good to expand the docs on these two fields.

vcd/resource_vcd_vapp_vm.go
adambarreiro: Done!

vcd/datasource_vcd_vapp_vm.go
adambarreiro: Done!

website/docs/r/vapp_vm.html.markdown
adambarreiro: Done, I've added this:
```
```

Didainius (approved):
Should be good as long as the tests pass.

dataclouder (approved):

Issue #611, opened 2021-01-04
vcd provider problem with resources deleted outside of terraform


pasikarkkainen:
Hello,

Current version of Terraform vcd provider (v3.1.0) does not seem to work in the scenario if some of the vCD resources (vApps) are deleted from vCD manually / outside of terraform. vcd provider just gives a generic error and does not work. One needs to figure out the missing vCD resource somehow, and then manually delete it from the terraform state to get the vcd provider working again.

Earlier version of vcd provider (v2.3.0) did work OK in this scenario, ie. it noticed the resource (vApp/VM) is missing from vCD, and re-provisioned it to fix the situation automatically.

### Terraform Version
terraform v0.12.29
terraform vcd provider v3.1.0

### Affected Resource(s)
- vcd_vapp
- vcd_vapp_vm

### Expected Behavior
- Terraform vcd provider should show an error with the actual missing resource name/ID mentioned.
- Terraform vcd provider should re-provision the missing resources.

### Actual Behavior
- Terraform vcd provider gives a generic error message:
- ```Error: error finding Vapp: [ENF] entity not found``` 
- .. and refuses to do anything. terraform "plan", "apply" or "destroy" actions don't work in this scenario.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply` and get some vapps/VMs provisioned to vCD.
2. Delete one of the vApps manually from vCD UI, or by using for example vcd-cli.
3. Run `terraform plan`, `terraform apply` or `terraform destroy` again and it'll fail with a generic error which doesn't mention which resource is not found / missing.

### Important Factoids
Currently running vCD 10.1.2

murakest:
Hello,
when is it expected this issue to be fixed?

dataclouder:
We have no timeline for this issue yet. We are aware of it and we'll look into it when we clear the current backlog

durehed:
@dataclouder when, approximately, will the current backlog be cleared?

dataclouder:
Sorry, can't provide any detail yet. We are currently focused on different priorities

mightydok:
+1

de1ayer:
+1

claudineimatos:
Hi, I really consider this problem a major. How may it be prioritized or at least remediated?
+1

mij0r:
+1

danielfeiler:
Hi,
I have the same problem here with the following Versions:
Terraform v1.0.7
[...]

PR #925, opened 2022-10-27
Fix read operations for deleted items

### LINKED ISSUE #855: Terrafom operation fail if VM /  vApp with VMs containing vcd_internal_disk is deleted from outside TF
Hello,

Current version of Terraform vcd provider (v3.6.0) does not seem to work in the scenario if some of the vCD resources (vApps or VMs with configured internal Disks) are deleted from vCD manually / outside of terraform. vcd provider just gives a generic error and does not work. To work around the issue in the case of vApp deletion it is necessary to delete vcd_vapp_network  and vcd_vm_internal_disk ressources from the terraform state file.
In the case of VM or internal disk deletion  it is necessary to delete the vcd_vm_internal_disk ressources from the terraform state file.

### Terraform Version
terraform v1.19
terraform vcd provider v3.6.0

### Affected Resource(s)
vcd_vapp_network.vapp-net
vcd_vm_internal_disk

### Expected Behavior
Terraform vcd provider should show an error with the actual missing resource name/ID mentioned.
Terraform vcd provider should re-provision the missing resources.

### Actual Behavior
If the whole vApp is deleted from outside terraform the terraform vcd provider gives following  error message:
```
│ Error: error finding Vapp: [ENF] entity not found
│
│   with vcd_vapp_network.vapp-net,
│   on main.tf line 53, in resource "vcd_vapp_network" "vapp-net":
│   53: resource "vcd_vapp_network" "vapp-net" {
```
After removing the network from the from the terraform state file the following error message apears 
```
Error: [getVm] failed to get vApp: [ENF] entity not found
│
│   with vcd_vm_internal_disk.REPROVM01-Disk01,
│   on main.tf line 89, in resource "vcd_vm_internal_disk" "REPROVM01-Disk01":
│   89: resource "vcd_vm_internal_disk" "REPROVM01-Disk01" {
```
After deleting the internal disk from terraform state, the plan and apply operation succeded.

If only the internal disk is deleted from the VM the following error apears during a terraform plan

```
Error: cannot get internal disk - VM doesn't have internal disks
│
│   with vcd_vm_internal_disk.REPROVM01-Disk01,
│   on main.tf line 89, in resource "vcd_vm_internal_disk" "REPROVM01-Disk01":
│   89: resource "vcd_vm_internal_disk" "REPROVM01-Disk01" {
│
```
after removing the internal disk from terraform state, terraform will re-apply the internal disk correctly.

If the whole VM is deleted the following error is displayed:

```
 Error: [getVm] failed to get VM: [ENF] entity not found
│
│   with vcd_vm_internal_disk.REPROVM01-Disk01,
│   on main.tf line 89, in resource "vcd_vm_internal_disk" "REPROVM01-Disk01":
│   89: resource "vcd_vm_internal_disk" "REPROVM01-Disk01" {
│
```
after removing the internal disk from terraform state, terraform will re-apply the VM and the internal disk correctly.

 terraform "plan", "apply" or "destroy" actions don't work in this scenario.

###Steps to Reproduce

terraform apply and get some vapps/VMs with internal disks provisioned to vCD.
Delete one of the vApps manually from vCD UI, or by using for example vcd-cli.
Run terraform plan, terraform apply or terraform destroy again and it'll fail with the described errors

1. 
- terraform apply and get some vapps/VMs with internal disks provisioned to vCD.
- Delete one of the vApps manually from vCD UI, or by using for example vcd-cli.
- Run terraform plan, terraform apply or terraform destroy again and it'll fail with the described errors
2.
- terraform apply and get some vapps/VMs with internal disks provisioned to vCD.
- Delete one of the internal disks from a VM manually from vCD UI, or by using for example vcd-cli.
- Run terraform plan, terraform apply or terraform destroy again and it'll fail with the described errors
3.
- terraform apply and get some vapps/VMs with internal disks provisioned to vCD.
- Delete one of the in VMs manually from vCD UI, or by using for example vcd-cli.
- Run terraform plan, terraform apply or terraform destroy again and it'll fail with the described errors

Important Factoids
Currently running vCD 10.3.0
A terraform config file is attached to this post. I have used this config to reproduce the described behavior

[main.tf.txt](https://github.com/vmware/terraform-provider-vcd/files/8715512/main.tf.txt)

Didainius:
Closes #855 

References: 
https://github.com/vmware/terraform-provider-vcd/issues/611
https://github.com/vmware/terraform-provider-vcd/pull/783
https://github.com/vmware/terraform-provider-vcd/pull/451

This is a continuation of #451 from the previous time.

tl;dr - this PR fixes `refresh` errors when a resource (or in some cases its parent) is no longer found. Instead of error the resource is removed from state which does not brick user workflow. The PR itself is mainly adding tests with just tiny changes. 95% of code are tests.

### Background 

As we have a somewhat automatic `ErrorEntityNotFound` test for Data Sources, I was looking for a smart solution for resources (just like I tried in #451 some time ago). However, due to the nature of the problem, there is no trivial solution. It is easy to trigger a "data source not found" error by just pointing it at a non-existing resource, but the resource requires to have the item created and then make it disappear before triggering another `plan/apply`.

Ideally, Terraform acceptance test framework could accept fake state config injection (`d *schema.ResourceData`), but it does not. (https://github.com/hashicorp/terraform-plugin-sdk/issues/309)

### Solution 

As there is no clean way (and not requiring big engineering efforts with constant maintenance cost) to make a universal solution I have tackled this problem manually - scanned all our existing resource code and looked where `read` has missing `d.SetId("")` when lookup error contains sentinel error `ErrorEntityNotFound`. Also, I found some remaining community issues reporting specific errors. This is the list of items that this PR fixes read for:
- [x] vcd_edgegateway_settings
- [x] vcd_vapp_network
- [x] vcd_vm_internal_disk
- [x] vcd_nsxv_dhcp_relay
- [x] vcd_vapp_static_routing
- [x] vcd_vapp_nat_rules
- [x] vcd_vapp_firewall_rules
- [x] vcd_vapp_access_control
- [x] vcd_nsxt_alb_edgegateway_service_engine_group
- [x] vcd_org_vdc
- [x] vcd_org_user
- [x] vcd_external_network

The way I approached it is that the removal from state `d.SetId("")` would happen only if the resource (or where it is not applicable, its immediate parent disappears). Altering behavior while traversing hierarchy for entity (what happens if a parent Org for VM is no longer available) is a deeper question that I don't think should be addressed now as it looks more like a nonexistent/rare error which could then explicitly be handle by using `terraform state rm` command.

For all the newly fixed items I have added relevant tests where they employ `PreConfig` attribute in `resource.TestStep` to directly use go-vcloud-director SDK and remove the item without Terraform knowledge. The TestStep will also use attributes `PlanOnly: true` and `ExpectNonEmptyPlan: true`.They will only run in acceptance testing mode (not binary or upgrade, because we cannot trigger SDK code in the middle of binary test. And there is no need). 

For code fixes it was mainly adding such snippet to previously existing error handling code:
```go
if err != nil { # previously existing line
  if govcd.ContainsNotFound(err) {
    d.SetId("")
    return nil
  }
  return diag.FromErr(err)    # previously existing line
} # previously existing line
```

Additionally in some cases I had to split `read` for resource and datasource and adding the `origin` field.

Specifically for #855, I have adjusted the code to avoid failing when the VM owning `vcd_internal_disk` is no longer available. It makes sense as `vcd_internal_disk` is actually part of VM itself.

### Testing
For the resources which I fixed in this iteration, I have added tests to check this particular scenario. They all have suffix `*ResourceNotFound`. They only run in acceptance test mode (not binary) as they need to run SDK code in the `PreConfig` step. I have also run a complete acceptance test suite on 10.4.0

### Docs
Additionally, I have added a section about handling Read in our [Coding Guidelines.](https://github.com/vmware/terraform-provider-vcd/pull/925/files#diff-fc045f5811272d70206705e0edc6ae36ddc392f1b798b945118db1a649cf6ea0)


vcd/resource_vcd_external_network_test.go
```diff
@@ -169,3 +169,85 @@ resource "vcd_external_network" "{{.ExternalNetworkName}}" {
...
+	preTestChecks(t)
+
+	// This test invokes go-vcloud-director SDK directly
+	if vcdShortTest {
+		t.Skip(acceptanceTestsSkipped)
+		return
+	}
+
+	if !usingSysAdmin() {
+		t.Skip("TestAccVcdExternalNetworkBasic requires system admin privileges")
```
adambarreiro, line 183: It's referencing another test, so we could either put `TestAccVcdExternalNetworkResourceNotFound` or maybe better:
```suggestion
		t.Skip(t.Name() + " requires system admin privileges")
```

adambarreiro (commented):
First read, just a little thing:

CODING_GUIDELINES.md
```diff
@@ -73,3 +71,69 @@ the resource in the documentation (`website/docs/d/resource_list.html.markdown`)
...
+  if err != nil {
+  	return nil, diag.Errorf("entity %s not found: %s", identifier, err)
+  }
+...
+}
+```
+
+### Handling Terraform Resource read
+
+Resource is an entity that is created and managed by Terraform.It can happen that a resource created
```
lvirbalas, line 101:
```suggestion
Resource is an entity that is created and managed by Terraform. It can happen that a resource created
```

lvirbalas (approved):
This is a great workflow cleanup. Thanks!

CODING_GUIDELINES.md
Didainius: Fixed.

CODING_GUIDELINES.md
```diff
@@ -73,3 +71,69 @@ the resource in the documentation (`website/docs/d/resource_list.html.markdown`)
 Every feature in the provider must include testing. See
 [TESTING.md](https://github.com/vmware/terraform-provider-vcd/blob/main/TESTING.md)
 for more info.
+
+## Handling Terraform Read of dissapeared (removed by other means than Terraform) entities
```
adambarreiro, line 75:
```suggestion
## Handling Terraform Read of disappeared (removed by other means than Terraform) entities
```

CODING_GUIDELINES.md
```diff
@@ -73,3 +71,69 @@ the resource in the documentation (`website/docs/d/resource_list.html.markdown`)
...
+
+There are some specific about handling entities in read code (handled in `schema.Resource.Read`)
```
adambarreiro, line 77:
```suggestion
There are some specifics about handling entities in read code (handled in `schema.Resource.Read`)
```

adambarreiro (approved):
Looks great to me! Just a few little typos

dataclouder (approved):

CODING_GUIDELINES.md
Didainius: Fixed.

CODING_GUIDELINES.md
Didainius: Fixed.

PR #931, opened 2022-11-15
Add vApp Template support for vcd_vapp_vm and vcd_vm resource and data sources

adambarreiro:
### Description

This PR:

- Builds on top of #922
- Requires https://github.com/vmware/go-vcloud-director/pull/520

The proposed changes add the following attributes to vcd_vapp_vm and vcd_vm resource and data source:

- `vapp_template_id`: The URN of the vApp Template to use. Deprecates `template_name`. The idea is to use the newest `vcd_catalog_vapp_template` data source to fetch this unique URN.
- `boot_image_id`: The URN of the media to use as boot image. Deprecates `boot_image`. The idea is to use `vcd_catalog_media` datasource to fetch this unique URN.

As the URNs are unique, the need of `catalog_name` is no more, so it has been deprecated as well.

This PR deprecates **vcd_catalog_item** as well.

### Not included

This PR doesn't change resource `vcd_inserted_media`, which will remain in backlog as it requires to use IDs instead of names.


vcd/resource_vcd_vapp_vm_test.go
```diff
@@ -317,6 +317,147 @@ resource "vcd_vapp_vm" "{{.VmName2}}" {
 }
 `
 
+// TestAccVcdVmAndVAppVmWithIds tests that vApp VMs and standalone VMs can be created using the
+// `template_id` and `boot_image_id` attributes.
+// TODO: Ideally, we should refactor all tests to use `template_id` and `boot_image_id` and create a test
+// for the deprecated fields to avoid regressions.
+func TestAccVcdVmAndVAppVmWithIds(t *testing.T) {
```
adambarreiro, line 324: Let me know wether I should remove this test, as we have `TestAccVcdSubscribedCatalog`

dataclouder: I think it's good to have

vcd/resource_vcd_catalog_vapp_template.go
```diff
@@ -377,6 +377,11 @@ func findVAppTemplate(d *schema.ResourceData, vcdClient *VCDClient, origin strin
 			if err != nil {
 				return nil, err
 			}
+			// This checks that the vApp Template is synchronized in the catalog
+			_, err = vcdClient.QueryVAppTemplateById(vAppTemplate.VAppTemplate.ID)
+			if err != nil {
+				return nil, fmt.Errorf("the found vApp Template %s is not synchronized in the catalog", vAppTemplate.VAppTemplate.ID)
```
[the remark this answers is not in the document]
adambarreiro: I'll revisit this part, thanks for your inputs!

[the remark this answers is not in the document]
adambarreiro: About

> One way to detect this case and do the right thing is by fetching the catalog ID (which does not change) and check whether the vApp template ID has changed. Not sure we want to make it so complex here, but at least we should know that it can happen.

The idea is to get the vApp Template by Catalog ID + Name and compare Old ID == New ID?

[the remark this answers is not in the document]
dataclouder: Yes. As I said, it is complicated. Not sure whether we want to introduce this check here.
Testing this behavior is particularly difficult to reproduce in Terraform infrastructure. You have this issue only when `vcd_subscribed_catalog.make_local_copy` is false and the catalog has synchronized the list of items, but not their contents.

CC: @Didainius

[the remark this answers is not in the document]
dataclouder: Yes. The only thing we need to be careful about is avoiding a temporary vApp template ID to get into Terraform state, as it might be difficult to get a clean state after that.

dataclouder (approved):

Didainius (approved):

website/docs/r/vapp_vm.html.markdown
```diff
@@ -343,9 +449,8 @@ The following arguments are supported:
 * `vdc` - (Optional; *v2.0+*) The name of VDC to use, optional if defined at provider level
 * `vapp_name` - (Required) The vApp this VM belongs to.
 * `name` - (Required) A name for the VM, unique within the vApp 
-* `computer_name` - (Optional; *v2.5+*) Computer name to assign to this virtual machine. 
-* `catalog_name` - (Optional; *v2.9+*) The catalog name in which to find the given vApp Template or media for `boot_image`.
-* `template_name` - (Optional; *v2.9+*) The name of the vApp Template to use
+* `computer_name` - (Optional; *v2.5+*) Computer name to assign to this virtual machine.
+* `template_id` - (Optional; *v3.8+*) The URN of the vApp Template to use. You can fetch it using a [`vcd_catalog_vapp_template`](/providers/vmware/vcd/latest/docs/data-sources/catalog_vapp_template) data source.
```
lvirbalas, line 453: CC: @dataclouder @Didainius

One reservation about naming this field: `template_id` or `vapp_template_id`?

Thoughts:
- `template_id` is more consistent with the previous `template_name` and is shorter to write
- `vapp_template_id` is more consistent with `vapp_name` and `catalog_vapp_template` data source, which makes it more intuitive without reading the docs

E.g. how it would look with `vapp_template_id`:

```hcl
data "vcd_catalog" "cat-where-is-template" {
  org  = "test"
  name = "cat-where-is-template"
}

data "vcd_catalog_vapp_template" "vappWithMultiVm" {
  org        = "test"
  catalog_id = data.vcd_catalog.cat-where-is-template.id
  name       = "vappWithMultiVm"
}

resource "vcd_vapp_vm" "secondVM" {
  vapp_name           = vcd_vapp.web.name
  vapp_template_id    = data.vcd_catalog_vapp_template.vappWithMultiVm.id
  vm_name_in_template = "secondVM" # Specifies which VM to use from the template
  name                = "secondVM"
  computer_name       = "db-vm"
  memory              = 512
  cpus                = 2
  cpu_cores           = 1
}
```

What do you think?

lvirbalas: We discussed that at some point also `vapp_name` will be changed to `vapp_id`, then `vapp_template_id` would be even more consistent:

```hcl
resource "vcd_vapp_vm" "secondVM" {
  vapp_id             = ...
  vapp_template_id    = data.vcd_catalog_vapp_template.vappWithMultiVm.id
  vm_name_in_template = "secondVM" # Specifies which VM to use from the template
  name                = "secondVM"
  computer_name       = "db-vm"
  memory              = 512
  cpus                = 2
  cpu_cores           = 1
}
```

adambarreiro: Changed to `vapp_template_id`!

lvirbalas: @adambarreiro Saw your refactoring commit, thanks!
Maybe change the four occurrences of `templateId` to `vappTemplateId` in code too?

lvirbalas (approved):
This is important change. Let's write it up in a blog post too, @adambarreiro

janbaer:
How can I migrate my TF state to the new version without destroying and recreating all of my VMs? This breaking change is really dangerous.

adambarreiro:
> How can I migrate my TF state to the new version without destroying and recreating all of my VMs? This breaking change is really dangerous.

Thanks for your feedback. I'd say that in this case (existing VMs) it's better to stay with current `template_name` attribute, while using `vapp_template_id` for the new ones. The `template_name` didn't change its behavior and still works, the `Deprecated` warning is more of a heads up to switch to `vapp_template_id` whenever possible.
We will try to find a proper solution for this scenario that guarantees that VMs don't get recreated.

PR #932, opened 2022-11-16
Prerelease test fixes

Didainius:
These are mainly test fixes, but there was one bug which is required to fix the tests:
`vcd_vapp` power handling had error in a scenario when one creates a powered on vApp and wants to power it down later on - this case was simply not covered.

Test fixed below:

* `TestAccVcdNsxtDynamicSecurityGroupIntegration` fails on VCD <10.3.2 because it uses earlier unsupported `comment` field. It has no purpose in this particular test so it is removed in file `resource_vcd_nsxt_dynamic_security_group_test.go`
* VCD 10.3.0 `TestAccVcdVAppVm_4types_PowerState` reports `POWERED_OFF` state for vApp instead of `MIXED` when a child VM is in vApp is powered off. Fix is to check for both cases.
* `TestAccVcdVdcGroupResourceAsOrgUser` failed on newly introduced `edge_cluster_id` field read. Normally, a `vcd_org_vdc` resource is created by sysadmin, but in this case the test jugles and switches to Org user mode after the resources are created, therefore it failed. 

The following tests are amended to power off vApp before removing vApp networks as it is not allowed. This has been done in a few ways - somewhere it is keep the vApp powered off (after bug fixed), in some cases it is powering of the VM, because a VM cannot be powered on in a powered off vApp.
* `TestAccVcdVAppEmptyVm` ensure that vApp is powered off before destroy phase to avoid triggering an error when removing vApp network from a powered on vApp.
* `TestAccVcdVAppMultiVmInTemplate`
* `TestAccVcdNsxtVAppRawAllNsxtNetworks`
* `TestAccVcdVAppVmCreateCustomizationFalse`
* `TestAccVcdVAppVmCreateCustomization`
* `TestAccVcdVAppVmUpdateCustomization`
* `TestAccVcdNsxvEdgeFirewallRuleVms`
* `TestAccVcdVAppVmDhcpWait`
* `TestAccVcdVAppRaw_Basic`
* `TestAccVcdVAppHotUpdateVm`
* `TestAccVcdVAppVmMultiNIC`
* `TestAccVcdVAppVm_Basic`
* `TestAccVcdVAppVm_Clone`

These tests are patched to be skipped when running in Org user mode:
* `TestAccVcdStandaloneVmWithVmSizing`
* `TestAccVcdSubscribedCatalog`
* `TestAccVcdVAppVm_4types_sizing_min`
* `TestAccVcdVAppVm_4types_sizing_max`
* `TestAccVcdVAppVm_4types_sizing_cpu_only`
* `TestAccVcdVAppAndVmWithComputePolicies`

### DIFF
--- a/.changes/v3.8.0/932-bug-fixes.md
+++ b/.changes/v3.8.0/932-bug-fixes.md
@@ -1,0 +1,2 @@
+* Fixed `resource/vcd_vapp` that would prevent to Power off vApp when previous state was
+  `power_on=true` [GH-932]
--- a/.changes/v3.8.0/932-bug-fixes.md
+++ b/.changes/v3.8.0/932-bug-fixes.md
@@ -1,2 +1,2 @@
-* Fixed `resource/vcd_vapp` that would prevent to Power off vApp when previous state was
+* Fixed a bug in `resource/vcd_vapp` that would prevent to Power off vApp when previous state was
   `power_on=true` [GH-932]
--- a/vcd/resource_vcd_nsxt_vapp_raw_test.go
+++ b/vcd/resource_vcd_nsxt_vapp_raw_test.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"regexp"
 	"testing"
-	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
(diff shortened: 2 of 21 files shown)


.changes/v3.8.0/932-bug-fixes.md
```diff
@@ -0,0 +1,2 @@
+* Fixed `resource/vcd_vapp` that would prevent to Power off vApp when previous state was
```
adambarreiro, line 1:
```suggestion
* Fixed a bug in `resource/vcd_vapp` that would prevent to Power off vApp when previous state was
```

adambarreiro (approved):
Thanks for fixing all of those tests!

lvirbalas (approved):
Just in case (about the newer stuff): no more tests to change with the new field's rename from 'template_id' to 'vapp_template_id'?

vcd/resource_vcd_nsxt_vapp_raw_test.go
```diff
@@ -41,14 +40,22 @@ func TestAccVcdNsxtVAppRawAllNsxtNetworks(t *testing.T) {
...
 	params["FuncName"] = t.Name() + "-step2"
-	configText2 := templateFill(testAccCheckVcdNsxtVAppRaw_basicCleanup, params)
+	params["VappPowerOn"] = "false"
+	configText2 := templateFill(testAccCheckVcdNsxtVAppRaw_basic, params)
+
+	// params["FuncName"] = t.Name() + "-step3"
+	// configText3 := templateFill(testAccCheckVcdNsxtVAppRaw_basicCleanup, params)
 	debugPrintf("#[DEBUG] CONFIGURATION: %s\n", configText)
+	debugPrintf("#[DEBUG] CONFIGURATION: %s\n", configText2)
+	// debugPrintf("#[DEBUG] CONFIGURATION: %s\n", configText3)
```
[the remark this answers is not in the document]
Didainius: Cleaned up.

.changes/v3.8.0/932-bug-fixes.md
Didainius: Done.

Didainius:
> Just in case (about the newer stuff): no more tests to change with the new field's rename from 'template_id' to 'vapp_template_id'?

I do not expect such. I have rescanned code for `template_id` entries and couldn't find any.

PR #941, opened 2022-11-24
Update CSE guide

adambarreiro:
This PR adapts the CSE guide and the runnable example for v3.8.0 of the VCD Provider.

Tested with VCD 10.3.3 and it worked.


Didainius (approved):

dataclouder (approved):

lvirbalas:
Just for clarity: why this PR removed a single `catalog_item_metadata` entry, but added _six_ `metadata_entry` in place of it?

adambarreiro:
> Just for clarity: why this PR removed a single `catalog_item_metadata` entry, but added _six_ `metadata_entry` in place of it?

Old metadata was a TypeMap of string->string. So you could only have the following:
```
metadataKey1 = "metadataValue1"
metadataKey2 = "metadataValue2"
...
```
Behind the scenes, the assigned type was `MetadataStringValue` with user access `READWRITE` and `GENERAL` domain.

With the new metadata, which is a TypeSet, you can set everything (key, value, type, user access and domain), that's why instead of one map instance we have a **set** of six entries.

Issue #944, opened 2022-11-29
Datasource vcd_catalog not working for shared catalog with tenant users


NilsBusche:
### Terraform Version
```
Terraform v1.3.5
on linux_amd64
+ provider registry.terraform.io/vmware/vcd v3.8.0
```

### Affected Resource(s)
Please list the resources as a list, for example:
- Datasource vcd_catalog

### Terraform Configuration Files

```hcl
data "vcd_catalog" "catalog" {
  org = "org-1234"
  name = "catalog-abcd"
}

output "catalog_id" {
  value = data.vcd_catalog.catalog.id
}
```

### Expected Behavior
Catalog ID should be returned as output, even if a tenant specific user is used.

### Actual Behavior
Catalog cannot be found:
```
data.vcd_catalog.catalog: Reading...
╷
│ Error: unable to retrieve catalog record - [ENF] entity not found
│
│   with data.vcd_catalog.catalog,
│   on main.tf line 11, in data "vcd_catalog" "catalog":
│   11: data "vcd_catalog" "catalog" {
│
╵
```

### Steps to Reproduce
1. `terraform apply`

### User Access rights 
Used user is a user which has only access to one specific tenant, not a system/provider user

### Important Factoids
- The used catalog is a catalog shared by one organization to others organizations
- The given `org` value is the name of the org where the original catalog is located.
- Using a user with system/provider access rights the code above is working as expected
- Changing the `org` value to an org the catalog is shared to does not work, even with system/provider access, but this might be the expected behavior
- The datasource is very important for us to use the new attribute `vapp_template_id` in the `vapp_vm`

dataclouder:
The reason for the failure is this:
* when we search for the catalog, I try to retrieve the org. As a tenant, I don't have access to the sharing org, and the search fails (org not found)
* If I put the current org, the search again fails because the catalog parent is not the current org. (catalog not found)

A possible solution, which, BTW, would also solve the case where we have the catalog shared with the same name from multiple organizations, is to add a field `sharing_org` to use instead of `org` in the data source.

```hcl
data "vcd_catalog" "catalog" {
  sharing_org = "org-1234"
  name           = "catalog-abcd"
}
```
 
Would this solution be acceptable?

dataclouder:
The way I see it working is this:
When you make a data source with `org` it will fail like it does now, but the error message will contain information about the catalog belonging to another org, and that you should use `sharing_org` instead.
[...]

PR #949, opened 2022-12-01
Fix issue 944 - allow org users to read shared catalogs

dataclouder:
Address issue #944 

*  Fix title of catalog operations guide
*  Add functions to find a catalog without its Org

    These changes fix the problem of a catalog that was shared
    to the current org, but a tenant could not retrieve it
    for lack of privileges.
    It also fixes the case when multiple catalogs with the same
    name are available in the organization.
    The new functions search for catalogs without need of accessing
    the sharing organization.
    Also added a new  'is_local' property to all catalog resources and data sources.
    This property tells whether the catalog was created in the current org
    or not.

When there are multiple catalogs with the same name in one Org, and users attempt retrieving a data source, there are two possibilities:

1. They know the name of the sharing organization, from where the catalog comes. So they can use that name in the `org` field. The data source gets retrieved as if it were local.
2. They don't know the original organization name. So they attempt to retrieve the data source, and get an error message that will help retrieving the right one:

```
│ Error: [catalog read DS] error retrieving catalog cat-datacloud: no records found for catalog 'cat-datacloud' from org 'datacloud-1'
│ There are catalogs with the same name from other orgs: [datacloud datacloud-2]
│
│   with data.vcd_catalog.catalog-1,
│   on config.tf line 32, in data "vcd_catalog" "catalog-1":
│   32: data "vcd_catalog" "catalog-1" {
│
```

**Side effect**: There is still a problem, which existed before, but has been made official by an appropriate error message in this PR: when searching a data source by filter (not by name), the filter only works if the Org containing the catalog is reachable by the user. Thus, it won't work for a shared catalog.


vcd/datasource_vcd_catalog.go
```diff
@@ -141,34 +148,96 @@ func datasourceVcdCatalog() *schema.Resource {
...
+	if orgName == "" {
+		orgName = vcdClient.Org
+	}
+	return orgName
+}
+
+func getCatalogFromResource(catalogName string, d *schema.ResourceData, meta interface{}) (*govcd.AdminCatalog, error) {
+	vcdClient := meta.(*VCDClient)
+
+	orgName := getOrgName(vcdClient, d)
```
adambarreiro, line 162: Not sure if I'm missing something, but I think this could be simplified to
```suggestion
	orgName, err := vcdClient.GetOrgNameFromResource(d)
```
then `getOrgName` is no longer needed.

vcd/datasource_vcd_catalog.go
```diff
@@ -141,34 +148,96 @@ func datasourceVcdCatalog() *schema.Resource {
...
 	)
 
 	if !nameOrFilterIsSet(d) {
 		return diag.Errorf(noNameOrFilterError, "vcd_catalog")
 	}
 
-	adminOrg, err := vcdClient.GetAdminOrgFromResource(d)
-	if err != nil {
-		return diag.Errorf(errorRetrievingOrg, err)
+	orgName := getOrgName(vcdClient, d)
```
adambarreiro, line 211: Same as previous comment

adambarreiro (approved):

Didainius (approved):

vcd/datasource_vcd_catalog.go
```diff
@@ -141,34 +148,87 @@ func datasourceVcdCatalog() *schema.Resource {
...
+		orgNames = append(orgNames, cr.OrgName)
+		if cr.OrgName == orgName {
+			catalogRecord = cr
+			break
+		}
+	}
+	if catalogRecord == nil {
+		message := fmt.Sprintf("no records found for catalog '%s' from org '%s'", catalogName, orgName)
+		if len(orgNames) > 0 {
+			message = fmt.Sprintf("%s\nThere are catalogs with the same name from other orgs: %v", message, orgNames)
```
lvirbalas, line 184: This is a tiny suggestion to change "from org" into "in org", purely for clarity:

```suggestion
		message := fmt.Sprintf("no records found for catalog '%s' in org '%s'", catalogName, orgName)
		if len(orgNames) > 0 {
			message = fmt.Sprintf("%s\nThere are catalogs with the same name in other orgs: %v", message, orgNames)
```

Or maybe you had some specific architecture-driven reason for using the "from" which I'm missing?

dataclouder: The first `from` should be changed to `in` as it conveys the meaning of belonging to the current org.
The second one should stay, as it is referring to catalog "shared from" other organizations. It may be clearer if I add the word "shared" before "from".

dataclouder: On second thought, better use `in` everywhere, as we are not sure which Org was picked by the user

.changes/v3.8.1/949-bug-fixes.md
```diff
@@ -0,0 +1 @@
+* Fix issue #944 shared catalog datasource not working [GH-949]
```
lvirbalas, line 1: This sounds really harsh :) (as if it was not working at all). Would you add more context like "when" it was not working?

lvirbalas (approved):

PR #965, opened 2022-12-20
RDE (Runtime Defined Entities) framework support with Runtime Defined Entity Interfaces resource and data source (1/3)

adambarreiro:
## Overview

This PR depends on https://github.com/vmware/go-vcloud-director/pull/527

This PR adds a new resource and data source **vcd_rde_interface** to be able to _create_, _read_, _update_ and _delete_ Defined Interfaces, and also to _import_ existing ones into Terraform state.

## Design details

To define a Defined Interface unequivocally, one needs `vendor`, `nss` and `version`:
- **Same** 3-tuple of (`vendor`, `nss`,`version`) can't have a **different** name. Otherwise it fails with `RDE_INTERFACE_ALREADY_EXISTS`.
- **Different** 3-tuples of (`vendor`, `nss`,`version`)  can have the **same** name.

An example of a valid scenario:

```
{
  "name": "sameName",
  "nss": "nss1",
  "version": "1.0.0",
  "vendor": "vendor1"
}
{
  "name": "sameName",
  "nss": "nss2",
  "version": "1.0.0",
  "vendor": "vendor2"
}
```
Because the 3-tuple (`vendor`, `nss`,`version`) is different.

An example of a wrong scenario:

```
{
  "name": "someName",
  "nss": "nss1",
  "version": "1.0.0",
  "vendor": "vendor1"
}
{
  "name": "differentName",
  "nss": "ns11", // Wrong: same nss
  "version": "1.0.0", // Wrong: same version
  "vendor": "vendor1" // Wrong: same vendor
}
```

Creation of `differentName` will fail as the 3-tuple is being reused.

- Because of the constraint above, the data source requires only the `vendor`, `nss` and `version`. In other words, we can't use `name` to fetch Defined Interfaces for the data source because it can be non-unique.

- In the resource, all attributes except `name` have `ForceNew: true`, this is because the PUT endpoint in VCD API ignores all changes in fields except for `name`. Also, if `name` is not sent during the PUT, API fails with a very ugly error.

- To import an existing Defined Interface from VCD into Terraform state, one needs to build an URN like `vendor.nss.version`. As version follows SemVer, the identifier results in something like `vmware.kubernetes.1.0.0` which can be visually odd. The usage of `.` as separator can be changed with `VCD_IMPORT_SEPARATOR`, but as with the data source, one requires the version to unequivocally identify the Defined Interface.

## Disclaimer

The new resource **vcd_rde_interface** does **NOT** add behaviors support in this PR. A "behavior" is an extension of the capabilities of the Defined Interfaces that would require an extra resource (something like **vcd_rde_interface_behavior**). These behaviors can be mapped in the Runtime Defined Entity types `hooks map[string]string` property, where this map supports PRE_CREATED, POST_CREATION, etc as keys and behavior IDs coming from Defined Interfaces as values.

## RDE saga:

- RDE types PR: #973
- RDEs PR: #977


.changes/v3.9.0/965-features.md
```diff
@@ -0,0 +1,2 @@
+* **New Resource:** `vcd_rde_interface` to manage the Defined Interfaces required to use RDE types [GH-965]
```
adambarreiro, line 1: I had two options here:

- Name it with a more "canonical" `vcd_defined_interface`
- Name it with a grouping `rde` prefix: `vcd_rde_interface`
- A mix of both: `vcd_rde_defined_interface`

I chose the second option but I have my doubts as it could be confusing.

vcd/datasource_vcd_rde_interface_test.go
```diff
@@ -0,0 +1,58 @@
...
+		ProviderFactories: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: configText,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(interfaceName, "namespace", params["Namespace"].(string)),
+					resource.TestCheckResourceAttr(interfaceName, "version", params["Version"].(string)),
+					resource.TestCheckResourceAttr(interfaceName, "vendor", params["Vendor"].(string)),
+					resource.TestCheckResourceAttr(interfaceName, "name", "Kubernetes"), // Name is always the same
+					resource.TestCheckResourceAttr(interfaceName, "id", fmt.Sprintf("urn:vcloud:interface:%s:%s:%s", params["Vendor"].(string), params["Namespace"].(string), params["Version"].(string))),
```
adambarreiro, line 43: Not sure if I should use a `regexp.MustCompile()` with `resource.TestMatchResourceAttr()` as `fmt.Sprintf("urn:vcloud:interface:%s:%s:%s", params["Vendor"].(string), params["Namespace"].(string), params["Version"].(string)))` is assuming too much knowledge on how Defined Interfaces identifiers work internally. But in the other hand, this check is much more strict.

.changes/v3.9.0/965-features.md
lvirbalas, line 1: Trying to suggest something self-explanatory for first-time users. Maybe like this?

```suggestion
* **New Resource:** `vcd_rde_interface` to manage Runtime Defined Entity Interfaces which are required for using Runtime Defined Entity (RDE) types [GH-965]
```

.changes/v3.9.0/965-features.md
```diff
@@ -0,0 +1,2 @@
...
+* **New Data Source:** `vcd_rde_interface` to fetch existing Defined Interfaces [GH-965]
```
lvirbalas, line 2: (This comment is not about change log, but an ask for the PR description.)

I suggest moving the "An example of a valid scenario" above the invalid example. Also, putting an inline comment into the invalid example that "it is wrong". This is to avoid accidental copy/pastes of the false example.

website/docs/r/rde_interface.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `vendor` - (Required) The vendor of the Defined Interface.
+* `namespace` - (Required) A unique namespace associated with the Defined Interface.
+* `version` - (Required) The version of the Defined Interface. Must follow [semantic versioning](https://semver.org/) syntax.
```
lvirbalas, line 34: We need the "3-tuple key" explanation in the main section of this doc page. Something similar to what you have in the PR description.

lvirbalas (changes requested):

website/docs/d/rde_interface.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+layout: "vcd"
+page_title: "VMware Cloud Director: vcd_rde_interface"
+sidebar_current: "docs-vcd-data-source-rde-interface"
+description: |-
+   Provides the capability of fetching an existing Defined Interfaces from VMware Cloud Director.
+---
+
+# vcd\_rde\_interface
+
+Provides the capability of fetching an existing Defined Interface from VMware Cloud Director.
```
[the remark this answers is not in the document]
adambarreiro: Yeah, makes totally sense. I was a bit biased as in API docs they're simply called Defined Interfaces and the name got stuck in my head.
I've renamed "Defined Interface" to "Runtime Defined Entity Interface"/ "RDE Interface".

website/docs/r/rde_interface.html.markdown
adambarreiro: Done!

vcd/resource_vcd_rde_interface_test.go
```diff
@@ -0,0 +1,135 @@
+//go:build rde || ALL || functional
```
Didainius, line 1: I think you're missing `rde` tag  in:
* config_test.go
* provider_test.go


```
go test -tags rde -v -timeout 200m
# github.com/vmware/terraform-provider-vcd/v3/vcd [github.com/vmware/terraform-provider-vcd/v3/vcd.test]
./config_test.go:840:3: undefined: testAccProvider
./config_test.go:841:3: undefined: testAccProviders
./config_test.go:843:12: undefined: testAccProvider
./config_test.go:1277:11: undefined: testAccProvider
./config_test.go:1300:11: undefined: testAccProvider
./config_test.go:1518:15: undefined: createTemporaryVCDConnection
./datasource_vcd_rde_interface_test.go:26:2: undefined: testParamsNotEmpty
./datasource_vcd_rde_interface_test.go:39:22: undefined: buildMultipleProviders
./resource_vcd_rde_interface_test.go:23:2: undefined: testParamsNotEmpty
./resource_vcd_rde_interface_test.go:46:22: undefined: testAccProviders
./resource_vcd_rde_interface_test.go:46:22: too many errors
FAIL    github.com/vmware/terraform-provider-vcd/v3/vcd [build failed]
```

vcd/datasource_vcd_rde_interface_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"testing"
+)
+
+func TestAccVcdRdeDefinedInterfaceDS(t *testing.T) {
+	preTestChecks(t)
+	skipIfNotSysAdmin(t)
+
+	// This is a Defined Interface that comes with VCD out of the box
```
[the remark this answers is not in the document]
adambarreiro: Good point, I'll check.

vcd/resource_vcd_rde_interface_test.go
adambarreiro: Thanks, I forgot to check in this repo, should be fixed now.

vcd/resource_vcd_rde_interface.go
```diff
@@ -0,0 +1,181 @@
...
+
+	return vcdClient.VCDClient.GetDefinedInterface(vendor, nss, version)
+}
+
+func resourceVcdRdeInterfaceUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	di, err := getDefinedInterface(d, meta)
+	if govcd.ContainsNotFound(err) {
+		log.Printf("[DEBUG] Defined Interface no longer exists. Removing from tfstate")
+		return nil
+	}
```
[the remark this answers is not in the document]
adambarreiro: True! Removed.

Didainius (approved):

.changes/v3.9.0/965-features.md
adambarreiro: Changed.

lvirbalas (approved):
My only reservation is whether `nss` (field name coming from the API) is not too cryptic for the new users.

vcd/datasource_vcd_rde_interface_test.go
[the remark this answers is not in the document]
adambarreiro: May I suggest to postpone the validation to #973 or #977? I'm having difficulties in testing this in CDS.

website/docs/r/rde_interface.html.markdown
```diff
@@ -0,0 +1,84 @@
...
+## Attribute Reference
+
+* `readonly` - Specifies if the RDE Interface can be only read.
+
+## Importing
+
+~> **Note:** The current implementation of Terraform import can only import resources into the state. It does not generate
+configuration. [More information.][docs-import]
+
+An existing RDE Interface can be [imported][docs-import] into this resource via supplying its vendor, nss and version, which
```
adezxc, line 61: Nit, but could we say namespace in this context? Or maybe put the variables in `backticks`?

adezxc (approved):

website/docs/r/rde_interface.html.markdown
adambarreiro: Makes sense, as I used backticks just a few paragraphs above.
I'd change it to namespace, but as the parameter is `nss` I'll just added the backticks.

vcd/datasource_vcd_rde_interface_test.go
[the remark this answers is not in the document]
lvirbalas: At least to me, sounds good.

dataclouder (approved):

PR #973, opened 2023-01-02
RDE (Runtime Defined Entities) framework support with RDE Type resource and data source (2/3)

adambarreiro:
## Overview

- This PR builds on top of #965
- This PR requires https://github.com/vmware/go-vcloud-director/pull/545

This PR adds a new resource and data source **vcd_rde_type** to be able to create, read, update and delete Runtime Defined Entity types, and also to import existing ones into Terraform state.

A RDE type specifies what a RDE looks like. Once created, you can create as many instances of the RDE type as you want. The RDE types may need of the usage of **vcd_rde_interface** references (#965)

## Design details

- Same as RDE Interfaces implemented in #965, in the data source, to fetch a RDE Type unequivocally one needs the `vendor`, `nss` and `version`. This is because this 3-tuple assembles the RDE Type URN/ID. The `name` here is not as important as in other data sources we have, it acts more like a "friendly" way to refer to the RDE Type, but one needs to be careful as **different** 3-tuples (vendor, namespace, version) **can have the same name**.
  In other words, we can't use `name` to fetch RDE type for the data source because it can be non-unique.

- Due to a current buggy behaviour I experienced using RDE Type API that causes a race condition, every RDE Type operation that writes something into VCD DB must be executed in mutual exclusion. The mutex key is still to be decided, right now is just a generic key that forces all writes to be sequential. If the key is improved, performance could be improved as well.

- Some attributes are immutable, meaning that doing a PUT to the URN with changed values doesn't have any effect, despite being answered with a 204. In consequence, these immutable attributes have `ForceNew: true` to force re-creation.

- To import an existing RDE Type from VCD into Terraform state, one needs to build an URN like `vendor.namespace.version`. As version follows SemVer, the identifier results in something like `vmware.kubernetes.1.0.0` which can be visually odd. The usage of `.` as separator can be changed with `VCD_IMPORT_SEPARATOR`, but as with the data source, one requires the version to unequivocally identify the RDE type.

## Tests run

- [x] go test -tags rde -v -run "TestAccVcdRdeTypeDS" on VCD 10.4.1
- [x] go test -tags rde -v -run "TestAccVcdRdeType" on VCD 10.4.1
- [x] go test -tags rde -v -run "TestAccVcdRdeTypeDS" on VCD 10.3.0
- [x] go test -tags rde -v -run "TestAccVcdRdeType" on VCD 10.3.0
- [x] go test -tags functional -v -run "TestAccDataSourceNotFound" on VCD 10.4.1
- [x] go test -tags functional -v -run "TestAccDataSourceNotFound" on VCD 10.3.0
- [x] ./test-binary.sh names 'vcd.TestAccVcdRde*.tf' on VCD 10.4.1
- [x] ./test-binary.sh names 'vcd.TestAccVcdRde*.tf' on VCD 10.3.0
- [x] Run all acceptance tests on VCD 10.3.0
- [x] Run all acceptance tests on VCD 10.4.1

## RDE saga:

- Defined Interfaces PR: #965
- RDEs PR: #977


vcd/resource_vcd_rde_type_test.go
```diff
@@ -0,0 +1,241 @@
...
+		"ProviderVcdSystem":   providerVcdSystem,
+		"ProviderVcdOrg1":     providerVcdOrg1,
+		"Nss":                 "nss",
+		"Version":             "1.0.0",
+		"Vendor":              "vendor",
+		"Name":                t.Name(),
+		"Description":         "Created by " + t.Name(),
+		"InterfaceReferences": "vcd_rde_interface.rde_interface1.id",
+		"SchemaPath":          getCurrentDir() + "/../test-resources/rde_type.json",
+		"SchemaUrl":           "https://raw.githubusercontent.com/adambarreiro/terraform-provider-vcd/add-rde-support-2/test-resources/rde_type.json", // FIXME
```
adambarreiro, line 30: Reminder to set final URL here. Or if you have any other suggestion, let me know

go.mod
```diff
@@ -59,3 +59,5 @@ require (
 	google.golang.org/grpc v1.50.1 // indirect
 	google.golang.org/protobuf v1.28.1 // indirect
 )
+
```
lvirbalas, line 62: Changelog entry missing?

vcd/datasource_vcd_rde_type.go
```diff
@@ -0,0 +1,74 @@
...
+			},
+			"nss": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "A unique namespace associated with the Runtime Defined Entity type",
+			},
+			"version": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The version of the Runtime Defined Entity type. The version string follows semantic versioning rules",
```
lvirbalas, line 26: Just a thought: maybe we should add in the three Descriptions above that these three properties compose a single unique key?

vcd/datasource_vcd_rde_type.go
```diff
@@ -0,0 +1,74 @@
...
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The description of the Runtime Defined Entity type",
+			},
+			"external_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "An external entity's id that this definition may apply to",
```
lvirbalas, line 54:
```suggestion
				Description: "An external entity's ID that this definition may apply to",
```

vcd/resource_vcd_rde_type.go
```diff
@@ -0,0 +1,337 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The vendor name for the Runtime Defined Entity Type",
+			},
+			"nss": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "A unique namespace associated with the Runtime Defined Entity Type",
```
lvirbalas, line 43: Same thought as above: maybe we should add in the three Descriptions above that these three properties compose a single unique key?

vcd/resource_vcd_rde_type.go
```diff
@@ -0,0 +1,337 @@
...
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The description of the Runtime Defined Entity Type",
+			},
+			"external_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "An external entity's id that this definition may apply to",
```
lvirbalas, line 82:
```suggestion
				Description: "An external entity's ID that this definition may apply to",
```

vcd/resource_vcd_rde_type_test.go
lvirbalas: We need your bot to capture "https://raw.githubusercontent.com" links :)

website/docs/r/rde_type.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+---
+
+# vcd\_rde\_type
+
+Provides the capability of creating, updating, and deleting Runtime Defined Entity types in VMware Cloud Director.
+Requires system administrator privileges.
+
+Supported in provider *v3.9+*
+
+## Example Usage with a schema file
```
lvirbalas, line 16:
```suggestion
## Example Usage with a local schema file
```

website/docs/r/rde_type.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+  vendor        = "vmware"
+  nss           = "vcd"
+  version       = "4.5.6"
+  name          = "My VMware RDE Type"
+  interface_ids = [data.vcd_rde_interface.my_interface.id]
+  schema        = file("${path.module}/schemas/my-type-schema.json")
+}
+```
+
+## Example Usage with a URL that contains a schema file
```
lvirbalas, line 35:
```suggestion
## Example Usage with a URL hosting the schema file
```

website/docs/r/rde_type.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+  name          = "My VMware RDE Type"
+  interface_ids = [data.vcd_rde_interface.my_interface.id]
+  schema_url    = "https://just.an-example.com/schemas/my-type-schema.json"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
```
lvirbalas, line 57: Looks like we need a similar message here about the 3-tuple as you had here:
https://github.com/vmware/terraform-provider-vcd/blob/main/website/docs/r/rde_interface.html.markdown#argument-reference

website/docs/r/rde_type.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+* `vendor` - (Required) The vendor of the Runtime Defined Entity Type.
+* `nss` - (Required) A unique namespace associated with the Runtime Defined Entity Type.
+* `version` - (Required) The version of the Runtime Defined Entity Type. Must follow [semantic versioning](https://semver.org/) syntax.
+* `name` - (Required) The name of the Runtime Defined Entity Type.
+* `description` - (Optional) The description of the Runtime Defined Entity Type.
+* `interface_ids` - (Optional) The set of [Defined Interfaces](/providers/vmware/vcd/latest/docs/resources/rde_interface) that this Runtime Defined Entity Type will use.
+* `schema` - (Optional) A string that specifies a valid JSON schema. It can be retrieved with functions such as `file`, `templatefile`... Either `schema` or `schema_url` is required.
```
lvirbalas, line 64: What is the thought behind having the three dots here: "`templatefile`..."?

website/docs/r/rde_type.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+* `schema_url` - (Optional) The URL that points to a valid JSON schema. Either `schema` or `schema_url` is required.
+  If `schema_url` is used, the downloaded schema will be computed in the `schema` attribute.
+* `external_id` - (Optional) An external entity's id that this Runtime Defined Entity Type may apply to.
```
lvirbalas, line 67:
```suggestion
* `external_id` - (Optional) An external entity's ID that this Runtime Defined Entity Type may apply to.
```

website/docs/r/rde_type.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+   Provides the capability of creating, updating, and deleting Runtime Defined Entity types in VMware Cloud Director.
+---
+
+# vcd\_rde\_type
+
+Provides the capability of creating, updating, and deleting Runtime Defined Entity types in VMware Cloud Director.
+Requires system administrator privileges.
+
+Supported in provider *v3.9+*
+
```
lvirbalas, line 15: Would we want to add a note about operations on RDE type being run in serial due to API? Or is that too much for the user for this use case? How long does it take to create these types in practice?

CC: @dataclouder @Didainius @adezxc

lvirbalas (changes requested):
Just some suggestions, looks great overall!

go.mod
adambarreiro: Yeah, indeed. Added!
Thanks!

website/docs/r/rde_type.html.markdown
adambarreiro: I don't think it's needed, even they run sequentially they're created almost immediately. It is not like VMs, for example, which require some time to be created.

go.mod
adambarreiro: Yeah, indeed. Added!
Thanks!

website/docs/r/rde_type.html.markdown
adambarreiro: I don't think it's needed, even they run sequentially they're created almost immediately. It is not like VMs, for example, which require some time to be created.

website/docs/r/rde_type.html.markdown
adambarreiro: Maybe I should put `etc` instead.

vcd/datasource_vcd_rde_type.go
adambarreiro: Done, also in `rde_interface`.

vcd/datasource_vcd_rde_type.go
adambarreiro: Changed, thanks!

vcd/resource_vcd_rde_type.go
adambarreiro: Done, same in `rde_interface`.

vcd/resource_vcd_rde_type.go
adambarreiro: Changed, thanks!

website/docs/r/rde_type.html.markdown
adambarreiro: Changed, thanks!

website/docs/r/rde_type.html.markdown
adambarreiro: Changed, thanks!

website/docs/r/rde_type.html.markdown
adambarreiro: Added.

website/docs/r/rde_type.html.markdown
adambarreiro: Changed, thanks!

website/docs/r/rde_type.html.markdown
Didainius: I was going to say we don't need that, but just thinking if there can ever be such situation that RDE types are created from different Terraform instances (then locks will not work).

Not sure if we're too cautious here.

vcd/resource_vcd_rde_type.go
```diff
@@ -0,0 +1,337 @@
...
+			return nil, fmt.Errorf("could not download JSON schema from url %s: %s", url, err)
+		}
+	} else {
+		jsonSchema = d.Get("schema").(string)
+	}
+
+	var unmarshalledJson map[string]interface{}
+	err = json.Unmarshal([]byte(jsonSchema), &unmarshalledJson)
+	if err != nil {
+		return nil, err
```
Didainius, line 157: the error message could give some context ("failed unmarshalling schema: %s" or something similar) here as it will be returned bare without any context if it fails here

vcd/resource_vcd_rde_type.go
```diff
@@ -0,0 +1,337 @@
...
+// fileFromUrlToString checks that the given url is correct and points to a given file type,
+// if so it downloads its contents and returns it as string.
+func fileFromUrlToString(url, fileType string) (string, error) {
+	if !strings.HasSuffix(url, fileType) {
+		return "", fmt.Errorf("it was expecting the URL to point to a %s file but it was %s", fileType, url)
+	}
+	buf := new(strings.Builder)
+
+	// #nosec G107 -- The URL needs to come from a variable for this purpose
+	resp, err := http.Get(url)
```
Didainius, line 183: It is sort of stupid here, but you need to at least initialize client with timeout as it might hang forever https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779

Or maybe access the same initialized HTTP client from  go-vcloud-director `vcdClient.Client.Http.Get`

vcd/resource_vcd_rde_type.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			util.Logger.Printf("[ERROR] fileFromUrlToString: Could not close HTTP response body: %s", err)
+		}
+	}()
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("could not get file from URL %s, got status %s", url, resp.Status)
+	}
+	_, err = io.Copy(buf, resp.Body)
```
Didainius, line 195: Just an option - you could use `io.ReadAll(resp.Body)` for simplicity

vcd/structure.go
```diff
@@ -200,6 +203,42 @@ func contains(sliceToSearch []string, searched string) bool {
...
+
+// areMarshaledJsonEqual compares that two marshaled JSON strings are equal or not. Returns an error if something
+// wrong happens when compacting both for comparison.
+func areMarshaledJsonEqual(json1, json2 []byte) (bool, error) {
+	if !json.Valid(json1) {
+		return false, fmt.Errorf("not a valid JSON: '%s'", json1)
+	}
+	if !json.Valid(json2) {
+		return false, fmt.Errorf("not a valid JSON: '%s'", json2)
+	}
```
Didainius, line 228: Would possibly be good to differentiate these errors somehow (even "not a valid JSON 1", "not a valid JSON 2") could work as it would help troubleshooting should such an error be ever hit

vcd/structure.go
```diff
@@ -200,6 +203,42 @@ func contains(sliceToSearch []string, searched string) bool {
...
+
+	var unmarshaledJson1, unmarshaledJson2 interface{}
+	err := json.Unmarshal(json1, &unmarshaledJson1)
+	if err != nil {
+		return false, fmt.Errorf("could not unmarshal JSON '%s': %s", json1, err)
```
Didainius, line 233: Similar request as above to differentiate error messages for easier troubleshooting :)

Didainius (changes requested):
Looks good - a few inline comments

website/docs/r/rde_type.html.markdown
lvirbalas: Too small probability for now, IMHO.

website/docs/r/rde_type.html.markdown
lvirbalas: Thanks. Maybe also "with functions" -> "with Terraform functions"?

vcd/resource_vcd_rde_type.go
lvirbalas: FYI: @adezxc (just for reference)

lvirbalas (approved):

vcd/resource_vcd_rde_type.go
adambarreiro: Fixed, thanks!

vcd/resource_vcd_rde_type.go
adambarreiro: Changed!

vcd/structure.go
adambarreiro: True, changed.

vcd/structure.go
adambarreiro: True, changed.

Didainius (approved):
Thanks.
I have tried to create 200 of `vcd_rde_type` - it all worked. While it is sequential - it is quite fast.

Also imitated an error for schema_url which worked good:
```
vcd_rde_type.my_rde_type[66]: Creation complete after 15s [id=urn:vcloud:type:vmware:vcd:4.66.6]
╷
│ Error: could not download JSON schema from url http://localhost:9000/rde_type.json: Get "http://localhost:9000/rde_type.json": read tcp 127.0.0.1:64098->127.0.0.1:9000: read: connection reset by peer
│ 
│   with vcd_rde_type.my_rde_type[6],
│   on vcd.TestAccVcdRdeDefinedInterface.tf line 38, in resource "vcd_rde_type" "my_rde_type":
│   38: resource "vcd_rde_type" "my_rde_type" {
│ 
╵
```

vcd/resource_vcd_rde_type.go
```diff
@@ -0,0 +1,344 @@
...
+				Computed:    true,
+				Description: "True if the Runtime Defined Entity Type cannot be modified",
+			},
+		},
+	}
+}
+
+// hasJsonValueChanged tells Terraform whether the JSON schema set in HCL configuration (which can have whatever identation and other quirks)
+// matches the obtained JSON from VCD.
+func hasJsonValueChanged(key, oldValue, newValue string, _ *schema.ResourceData) bool {
```
adezxc, line 107: I'm not sure how it's done in other files, but would it make sense to have it in a more general file like `structure.go`?

adezxc (approved):

vcd/resource_vcd_rde_type.go
```diff
@@ -0,0 +1,344 @@
...
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The name of the Runtime Defined Entity Type",
+			},
+			"vendor": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringMatch(regexp.MustCompile(`(?i)^[a-z0-9_-]+$`), "only alphanumeric characters, underscores and hyphens allowed"),
```
dataclouder, line 39: Since this validation is used several times, we should make a permanent one in `validate_func.go`

vcd/resource_vcd_rde_type.go
```diff
@@ -0,0 +1,344 @@
...
+			"readonly": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "True if the Runtime Defined Entity Type cannot be modified",
+			},
+		},
+	}
+}
+
+// hasJsonValueChanged tells Terraform whether the JSON schema set in HCL configuration (which can have whatever identation and other quirks)
```
dataclouder, line 105:
```suggestion
// hasJsonValueChanged tells Terraform whether the JSON schema set in HCL configuration (which can have whatever indentation and other quirks)
```

vcd/resource_vcd_rde_type.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+	return nil
+}
+
+// resourceVcdRdeTypeImport is responsible for importing the resource.
+// The following steps happen as part of import
+// 1. The user supplies `terraform import _resource_name_ _the_id_string_` command
+// 2. `_the_id_string_` contains a dot formatted path to resource as in the example below
+// 3. The functions splits the dot-formatted path and tries to lookup the object
+// 4. If the lookup succeeds it set's the ID field for `_resource_name_` resource in state file
```
dataclouder, line 321:
```suggestion
// 4. If the lookup succeeds it sets the ID field for `_resource_name_` resource in state file
```

vcd/structure.go
```diff
@@ -200,6 +203,42 @@ func contains(sliceToSearch []string, searched string) bool {
...
+	compactedJson := new(bytes.Buffer)
+	err = json.Compact(compactedJson, rawJson)
+	if err != nil {
+		return "", err
+	}
+	return compactedJson.String(), nil
+}
+
+// areMarshaledJsonEqual compares that two marshaled JSON strings are equal or not. Returns an error if something
+// wrong happens when compacting both for comparison.
```
dataclouder, line 221: We are actually not compacting the two items, but encoding them

vcd/structure.go
```diff
@@ -200,6 +203,42 @@ func contains(sliceToSearch []string, searched string) bool {
...
+func areMarshaledJsonEqual(json1, json2 []byte) (bool, error) {
```
dataclouder, line 222: It would be worth adding a unit test for both `areMarshaledJsonEqual` and `jsonToCompact`

vcd/resource_vcd_rde_type.go
adambarreiro: True, I refactored it into the `validateAlphanumericWithUnderscoresAndHyphens` function.

vcd/resource_vcd_rde_type.go
adambarreiro: Fixed, thanks for spotting that!

vcd/resource_vcd_rde_type.go
adambarreiro: Fixed, thanks again for spotting that!

vcd/structure.go
adambarreiro: True, that was some sort of leftover that I forgot to change. Thanks for spotting it!

vcd/structure.go
adambarreiro: Added a new file `structure_unit_test.go` with a couple of unit tests.

website/docs/r/rde_type.html.markdown
adambarreiro: Done!

vcd/resource_vcd_rde_type_test.go
adambarreiro: Changed, time to merge.

PR #977, opened 2023-01-04
RDE (Runtime Defined Entities) framework support with RDE resource and data source (3/3)

adambarreiro:
### `computed_entity` vs `input_entity`/`input_entity_url`

There is a common use case for RDEs where they are used by 3rd party components that perform continuous updates on them, which are expected and desired. This conflicts with Terraform way of working, as doing a `terraform plan` would reveal the required actions to remove every single change done by those 3rd party tools, which we don't want in this case.

To add compatibility with this scenario, there are two important arguments, `input_entity` and `input_entity_url`,
and one important computed attribute, `computed_entity`.

If the RDE is intended to be managed **only and exclusively** by Terraform, the contents of the input JSON should always match with those retrieved into `computed_entity`. Otherwise, only `computed_entity` will reflect the current state of the RDE in VCD, whereas `input_entity` and `input_entity_url` will only specify the RDE contents that are needed either on creation or in a deliberate update that will cause the RDE contents to be **completely overridden**.

As per this last point, one needs to be careful when updating `input_entity` or `input_entity_url`, as Terraform will apply
whatever changes were done, ignoring the real state from `computed_entity`. To perform a real update, one
needs to check the contents of the `computed_entity` and do some diff with the original input.

### `resolve` and `resolve_on_destroy`

RDEs must be resolved to be used or deleted, and this operation can be done either by Terraform with `resolve=true`, or by a 3rd party actor that will do it behind the scenes (`resolve=false`). To cover this last scenario, the RDE resource also has an argument `resolve_on_destroy` so Terraform can destroy the RDE if it was not resolved by anyone.

### Non-unique imports

RDEs uniqueness is only granted by their ID. This means that the combination of `vendor`, `nss`, `version` and `name` does not guarantee its uniqueness. One can have multiple RDEs with the same values of `vendor`, `nss`, `version` and `name`.

In order to have a decent implementation of the `terraform import` operation, this PR implements three ways of importing:
- `list@vendor.nss.version.name`: This is the way to list all available RDE identifiers of the given type and name. Import will exit without performing any other operation. This way the users can somehow decide which RDE they want to import (if the ID gives some information to them, that is)
- `rde-id`: Import by RDE ID, which is unique.
- `vendor.nss.version.name.position` import by type, name and **position**. Position is a natural number that starts from 1.

## RDE saga:

- Defined Interfaces PR: #965
- RDE types PR: #973


vcd/resource_vcd_rde_test.go
```diff
@@ -0,0 +1,566 @@
...
+		"ProviderSystem": providerVcdSystem,
+		"ProviderOrg1":   providerVcdOrg1,
+		"Nss":            "nss",
+		"Version":        "1.0.0",
+		"Vendor":         "vendor",
+		"Name":           t.Name(),
+		"Resolve":        false,
+		"SchemaPath":     getCurrentDir() + "/../test-resources/rde_type.json",
+		"EntityPath":     getCurrentDir() + "/../test-resources/rde_instance.json",
+		"EntityUrl":      "https://raw.githubusercontent.com/adambarreiro/terraform-provider-vcd/add-rde-support-3/test-resources/rde_instance.json", // FIXME
```
adambarreiro, line 32: Reminder to change this URL to its final destination

website/docs/d/rde.html.markdown
```diff
@@ -0,0 +1,54 @@
+---
+layout: "vcd"
+page_title: "VMware Cloud Director: vcd_rde"
+sidebar_current: "docs-vcd-data-source-rde"
+description: |-
+   Provides the capability of reading an existing Runtime Defined Entity in VMware Cloud Director.
```
lvirbalas, line 6: PR description still says "Data source not implemented" :)

website/docs/r/rde.html.markdown
```diff
@@ -0,0 +1,198 @@
...
+* `input_entity` - (Optional) A string that specifies a valid JSON for the RDE. It can be retrieved with functions such as `file`, `templatefile`... Either `input_entity` or `input_entity_url` is required.
+* `input_entity_url` - (Optional) The URL that points to a valid JSON for the RDE. Either `input_entity` or `input_entity_url` is required.
+* `external_id` - (Optional) An external input_entity's ID that this Runtime Defined Entity may have a relation to.
+
+## Attribute Reference
+
+The following attributes are supported:
+
+* `computed_entity` - The real state of this RDE in VCD.
+* `entity_in_sync` - It's `true` when `computed_entity` is equal to either `input_entity` or the contents of `input_entity_url`,
```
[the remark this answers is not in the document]
adambarreiro: That's a good point that actually I forgot to also add this to the tests.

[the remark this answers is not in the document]
adambarreiro: Answering to your original question, if the JSON referenced by `input_entity_url` changes, `entity_in_sync` will be false and `computed_entity` will reflect those changes.

I improved the following:
- Added the checks for `entity_in_sync` which I forgot.
- Checking that a change in the VCD contents should update `entity_in_sync` to false also for an RDE created with `input_entity_url`.

website/docs/r/rde.html.markdown
```diff
@@ -0,0 +1,198 @@
...
+In other words:
+
+~> When you want to update an RDE and `entity_in_sync` is `false`, you should always merge the contents
+of `computed_entity` and `input_entity` to avoid overriding the whole entity by mistake with an old value. 
+
+## RDE resolution
+
+When a RDE is created, its `state` will be `PRE_CREATED`, which means that the entity JSON was not validated against the
+[RDE Type](/providers/vmware/vcd/latest/docs/resources/rde_type) schema. After resolution, `state` should be either `RESOLVED`
+or `RESOLUTION_ERROR` if the input JSON doesn't match the schema.
```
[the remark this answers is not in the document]
lvirbalas: I see. Reading the paragraph below one question comes up:

> The RDE must be resolved at some point to be used or deleted, and this operation can be done either by Terraform with `resolve=true`, or by a 3rd party actor that will do it behind the scenes (`resolve=false`).

1. If there's a schema mismatch, why can we mark it resolved? Doesn't it counter the purpose of the field?
2. If the schema was matched but later, does resolve become `true` automatically? Does Terraform pick that up?

website/docs/r/rde.html.markdown
```diff
@@ -0,0 +1,198 @@
...
+
+The RDE must be resolved at some point to be used or deleted, and this operation can be done either by Terraform with
+`resolve=true`, or by a 3rd party actor that will do it behind the scenes (`resolve=false`).
```
lvirbalas, line 123: Why would a 3rd party actor would set resolve to `false`?

adambarreiro: I think the text is misleading, I didn't mean that, I have rephrased it.
What I meant is that one needs to leave `resolve=false` if the responsible of resolving it is an external actor outside Terraform (so Terraform leaves the RDE in `PRE_CREATED` state and it's up to this actor to resolve).
Not sure whether it does capture this idea now with the most recent change, let me know.

website/docs/r/rde.html.markdown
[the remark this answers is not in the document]
adambarreiro: 1) You can mark it resolved as you need to validate it against the schema in every case. So maybe there's a schema mismatch but you are not aware of it (it's not like you do resolve it deliberately knowing it has errors, it's more that a mistake/typo can happen and resolution would fail). In other words, you don't know there's a schema mismatch until you resolve it and get the feedback from VCD (we can think that "resolve"="validate" and "resolved"="validated", but I guess it's was not named like that for a good reason).

2) That's a good point. Right now, it doesn't, as it is acting more like an "imperative" argument rather than a "desired state" argument. The "state" in this case would be the purely Computed attribute `state`, which will be `RESOLVED` if it's resolved, even if this sort of imperative argument `resolve` was `false` (this case corresponds to an external actor who resolved it outside Terraform).

Not sure if I explained myself correctly, let me know if I can clarify if not.

website/docs/r/rde.html.markdown
```diff
@@ -0,0 +1,199 @@
...
+* `rde_type_id` - (Required) The ID of the [RDE Type](/providers/vmware/vcd/latest/docs/data-sources/rde_type) to instantiate.
+* `name` - (Required) The name of the Runtime Defined Entity.
+* `resolve` - (Required) If `true`, the Runtime Defined Entity will be resolved by this provider. If `false`, it won't be
+  resolved and must be either done by an external component or with an update. The Runtime Defined Entity can't be
+  deleted until the input_entity is resolved by either party, unless `resolve_on_removal=true`.
+* `resolve_on_removal` - (Optional) If `true`, the Runtime Defined Entity will be resolved before it gets deleted, to forcefully delete it. Otherwise, destroy will fail if it is not resolved. It is `false` by default.
+* `input_entity` - (Optional) A string that specifies a valid JSON for the RDE. It can be retrieved with functions such as `file`, `templatefile`... Either `input_entity` or `input_entity_url` is required.
+* `input_entity_url` - (Optional) The URL that points to a valid JSON for the RDE. Either `input_entity` or `input_entity_url` is required.
```
Didainius, line 140: I think an important thing to mention here is that this will be downloaded on **every read** operation and it will break terraform operations if these contents are no longer present on the remote site. Maybe even hinting that if a user cannot guarantee `input_entity_url` it is safer to use `input_entity`

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,437 @@
...
+	vcdClient := meta.(*VCDClient)
+	rde, err := getRde(d, vcdClient, "resource")
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	if d.Get("resolve_on_removal").(bool) {
+		err = rde.Resolve()
+		if err != nil {
+			return diag.FromErr(err)
```
Didainius, line 338: Detailed error should aid user IMO

```suggestion
			return diag.Errorf("could not resolve the Runtime Defined Entity before removal '%s': %s", rde.DefinedEntity.Name, err)
```

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,437 @@
...
+			"rde_type_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The Runtime Defined Entity Type ID",
+			},
+			"external_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
```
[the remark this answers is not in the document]
adambarreiro: Yes, exactly.

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,437 @@
...
+	}
+
+	// We save the ID immediately as the Resolve operation can fail, but the RDE is already created. If this happens,
+	// it should go to the Update operation instead.
+	d.SetId(rde.DefinedEntity.ID)
+
+	if d.Get("resolve").(bool) {
+		err = rde.Resolve()
+		if err != nil {
+			return diag.Errorf("could not resolve the Runtime Defined Entity '%s' of type '%s': %s", name, rdeTypeId, err)
```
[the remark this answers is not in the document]
adambarreiro: Then lines 144 to 146 are false, right? My assumption is wrong.

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,437 @@
...
+				ExactlyOneOf:          []string{"input_entity_url", "input_entity"},
+				DiffSuppressFunc:      hasJsonValueChanged,
+				DiffSuppressOnRefresh: true,
+			},
+			"computed_entity": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "A computed representation of the actual Runtime Defined Entity JSON retrieved from VCD. Useful to see the actual entity contents if it is being changed by a third party in VCD",
+			},
+			"owner_id": {
```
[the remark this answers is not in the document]
dataclouder: `owner_user_id` or even the simpler `user_id` look good to me

vcd/resource_vcd_rde.go
adambarreiro: Nice, thanks!

website/docs/r/rde.html.markdown
adambarreiro: Good point. I added the suggested note and also in the rde_type documentation (which has also this same combination with a different name)

vcd/resource_vcd_rde.go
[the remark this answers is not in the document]
adambarreiro: Changed to `owner_user_id` (It needs to highlight that is the **owner**, so I discarded the shorter `user_id`)

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,437 @@
...
+	jsonSchema, err := getRdeJson(vcdClient, d)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	rde, err := rdeType.CreateRde(types.DefinedEntity{
+		Name:       name,
+		ExternalId: d.Get("external_id").(string),
+		Entity:     jsonSchema,
+	}, &tenantContext)
```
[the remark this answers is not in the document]
adambarreiro: Yeah, that's right. I simplified it

vcd/resource_vcd_rde.go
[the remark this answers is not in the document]
Didainius: Thanks!

vcd/resource_vcd_rde.go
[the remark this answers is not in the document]
Didainius: It is good to store `ID` (line 146) as otherwise we would have an orphan resource (nothing would be stored in state, while the entity would be created in VCD), but when we have `d.SetId` and then return non `nil` error - Terraform will automatically `taint` that resource so that it is recreated on next step

website/docs/r/rde.html.markdown
```diff
@@ -0,0 +1,201 @@
...
+## Attribute Reference
+
+The following attributes are supported:
+
+* `computed_entity` - The real state of this RDE in VCD.
+* `entity_in_sync` - It's `true` when `computed_entity` is equal to either `input_entity` or the contents of `input_entity_url`,
+  meaning that the computed RDE retrieved from VCD is synchronized with the input RDE.
+* `owner_user_id` - The ID of the [Organization user](/providers/vmware/vcd/latest/docs/resources/org_user) that owns this Runtime Defined Entity.
+* `org_id` - The ID of the [Organization](/providers/vmware/vcd/latest/docs/resources/org) to which the Runtime Defined Entity belongs.
+* `state` - If the specified JSON in either `input_entity` or `entity_url` is correct, the state will be `RESOLVED`, otherwise it will be `RESOLUTION_ERROR`. If an input_entity in an `RESOLUTION_ERROR` state, it will require to be updated to a correct JSON to be usable.
```
adezxc, line 154: I might be wrong, but should it be more like this?
```suggestion
* `state` - If the specified JSON in either `input_entity` or `entity_url` is correct, the state will be `RESOLVED`, otherwise it will be `RESOLUTION_ERROR`. If an input_entity results in a `RESOLUTION_ERROR` state, it will require to be updated to a correct JSON to be usable.
```

vcd/resource_vcd_rde_test.go
```diff
@@ -0,0 +1,553 @@
...
+`
+
+const testAccVcdRdeDS = testAccVcdRde3 + `
+# skip-binary-test - Contains data source referencing a resource
+data "vcd_rde" "existing_rde1" {
+  provider = {{.ProviderSystem}}
+
+  rde_type_id        = vcd_rde_type.rde_type.id
+  name               = "{{.Name}}file-updated"
+}
```
adezxc, line 438: nit
```suggestion
  rde_type_id = vcd_rde_type.rde_type.id
  name        = "{{.Name}}file-updated"
```

vcd/resource_vcd_rde_test.go
```diff
@@ -0,0 +1,553 @@
...
+
+data "vcd_rde" "existing_rde2" {
+  provider = {{.ProviderOrg1}}
+
+  rde_type_id        = vcd_rde_type.rde_type.id
+  name               = "{{.Name}}file-updated"
```
adezxc, line 444: nit
```suggestion
  rde_type_id = vcd_rde_type.rde_type.id
  name        = "{{.Name}}file-updated"
```

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,435 @@
+package vcd
+
+import (
```
adezxc, line 3: Could you run `goimports -w` on this and other newly created files?

adezxc (changes requested):

website/docs/r/rde.html.markdown
adambarreiro: Indeed, thanks!

vcd/resource_vcd_rde_test.go
adambarreiro: Thanks for spotting! Fixed.

vcd/resource_vcd_rde_test.go
adambarreiro: Thanks for spotting! Fixed.

vcd/resource_vcd_rde.go
adambarreiro: Done, changed all RDE fileset.

vcd/resource_vcd_rde.go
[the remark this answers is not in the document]
adambarreiro: Got it, thanks! I was thinking of improving the error message, but I guess that it's clear enough (if the docs were read)

website/docs/r/rde.html.markdown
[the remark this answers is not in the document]
lvirbalas: Thank you for an in depth explanation, I think it's good. @dataclouder @Didainius @adezxc is the "resolve" keyword and the whole resolution workflow clear enough from your perspective?

lvirbalas (approved):

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,435 @@
...
+				Type: schema.TypeBool,
+				Description: "If `true`, the Runtime Defined Entity will be resolved by this provider. If `false`, it won't be" +
+					"resolved and must be either done by an external component or with an update. The Runtime Defined Entity can't be" +
+					"deleted until the entity is resolved.",
+				Required: true,
+			},
+			"resolve_on_removal": {
+				Type: schema.TypeBool,
+				Description: "If `true`, the Runtime Defined Entity will be resolved before it gets deleted, to forcefully delete it." +
+					"Otherwise, destroy will fail if it is not resolved.",
```
dataclouder, line 91:
```suggestion
				Description: "If `true`, the Runtime Defined Entity will be resolved before it gets deleted, to ensure forced deletion." +
					"Destroy will fail if it is not resolved.",
```

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,435 @@
...
+				Default:  false,
+				Optional: true,
+			},
+			"state": {
+				Type: schema.TypeString,
+				Description: "When created it will be in PRE_CREATED state. If the specified JSON in either `input_entity` or `input_entity_url` is correct, the state will be RESOLVED," +
+					"otherwise it will be RESOLUTION_ERROR. If an entity in an RESOLUTION_ERROR state, it will require to be updated to a correct JSON to be usable",
```
dataclouder, line 98:
```suggestion
					"otherwise it will be RESOLUTION_ERROR. If an entity is in a RESOLUTION_ERROR state, it will require to be updated to a correct JSON to be usable",
```

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,435 @@
...
+	}
+	return nil
+}
+
+// resourceVcdRdeImport is responsible for importing the resource.
+// The following steps happen as part of import
+// 1. The user supplies `terraform import _resource_name_ _the_id_string_` command
+// 2. `_the_id_string_` contains a dot formatted path to resource as in the example below
+// 3. The functions splits the dot-formatted path and tries to lookup the object
+// 4. If the lookup succeeds it set's the ID field for `_resource_name_` resource in state file
```
dataclouder, line 352:
```suggestion
// 4. If the lookup succeeds it sets the ID field for `_resource_name_` resource in state file
```

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,435 @@
...
+		if len(listAndVendorSplit) != 2 {
+			return nil, helpError
+		}
+		return nil, printList(listAndVendorSplit[1], resourceURI[1], resourceURI[2], resourceURI[3])
+	case 5: // ie: VCD_IMPORT_SEPARATOR="_" vendor_nss_1.2.3_name_1
+		rde, err = getRdeInPosition(resourceURI[0], resourceURI[1], resourceURI[2], resourceURI[3], resourceURI[4])
+		if err != nil {
+			return nil, err
+		}
+	case 6: // ie: VCD_IMPORT_SEPARATOR="_" list@vendor_nss_1.2.3_name
```
dataclouder, line 416: It says "case 6", but the example has only 4 items.
It would be 6 items with the default import separator

vcd/resource_vcd_rde_duplicate_test.go
```diff
@@ -0,0 +1,167 @@
...
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resourceFieldsEqual(rde2Tenant, rde3Tenant, []string{"id"}),
+					resourceFieldsEqual(rde1System, rde3Tenant, []string{"id", "org", "org_id"}),
+					// We cache some IDs to use it on later steps
+					cachedIds[0].cacheTestResourceFieldValue(rde2Tenant, "id"),
+					cachedIds[1].cacheTestResourceFieldValue(rde3Tenant, "id"),
+				),
+			},
+			// We use a data source to fetch the RDE that belongs to System.
+			// Despite it is duplicated (it has same name, type, etc), it should work as we can unequivocally fetch it
```
dataclouder, line 65:
```suggestion
			// Despite being duplicated (it has same name, type, etc), it should work as we can unequivocally fetch it
```

vcd/resource_vcd_rde_duplicate_test.go
```diff
@@ -0,0 +1,167 @@
...
+			// thanks that is the only one present in the System organization.
```
dataclouder, line 66:
```suggestion
			// since it is the only one present in the System organization.
```

vcd/resource_vcd_rde_test.go
```diff
@@ -0,0 +1,536 @@
...
+	// addRightsToTenantUser needs to happen before the client has been created, otherwise retrieving
+	// RDEs will fail. Thus, we need to invalidate existing client cache and start a new one
+	cachedVCDClients.reset()
+
+	resource.Test(t, resource.TestCase{
+		ProviderFactories: buildMultipleProviders(),
+		CheckDestroy:      testAccCheckRdeDestroy(rdeType, rdeFromFile, rdeFromUrl),
+		Steps: []resource.TestStep{
+			// Preconfigure the test: We fetch an existing interface and create an RDE Type.
+			// Creating an RDE Type creates an associated Rights Bundle that is NOT published to all tenants by default.
```
dataclouder, line 87:
```suggestion
			// Creating an RDE Type results in the creation of an associated Rights Bundle that is NOT published to all tenants by default.
```

website/docs/d/rde.html.markdown
```diff
@@ -0,0 +1,54 @@
...
+
+Provides the capability of reading an existing Runtime Defined Entity in VMware Cloud Director.
+
+-> VCD allows to have multiple RDEs of the same [RDE Type](/providers/vmware/vcd/latest/docs/resources/rde_type) with
+the same name, meaning that the data source will not be able to fetch a RDE in this situation, as this data source
+can only retrieve **unique RDEs**.
+
+Supported in provider *v3.9+*
+
+## Example Usage with a JSON file
```
dataclouder, line 19: I don't see the JSON file in the example

dataclouder (commented):
First pass

website/docs/r/rde.html.markdown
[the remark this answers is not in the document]
Didainius: good for me as well

Didainius (approved):
I have tried it and approving, but have a look at binary tests as they need some `skip` directives - for example `vcd.TestAccVcdRde-Resolve.tf` binary test will not be skipped, but it will fail

vcd/config.go
```diff
@@ -82,6 +82,11 @@ const (
 // When the Client() function is called with the same parameters, it will return
 // a cached value instead of connecting again.
 // This makes the Client() function both deterministic and fast.
+//
+// WARNING: Cached clients need to be evicted by calling cacheStorage.reset() after the rights of the associated
```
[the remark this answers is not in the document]
adambarreiro: Mmm, I think the test may be confusing as we invalidate the VCD client cache at the beginning*, but it should be done as the last step of `addRightsToTenantUser`, so I have moved it there. Now it shows in a clearer way that clients must be evicted from cache after the rights are updated.

This makes sense as you need to force a re-login after rights are modified, to refresh them in the session.

* and it worked like this as the tenant provider is not created until `addRightsToTenantUser` is completed, so there's no tenant client until that step.

website/docs/r/rde.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+  namespace = "tech1"
+  version   = "1.2.3"
+}
+
+resource "vcd_rde" "my_rde" {
+  org         = "my-org"
+  rde_type_id = data.vcd_rde_type.my-type.id
+  name        = "My custom RDE"
+  resolve     = true
+  # Functions are evaluated eagerly, so the file must exist and not be a reference to a created Terraform resource
```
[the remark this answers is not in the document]
adambarreiro: I wanted to state that functions are evaluated before the dependency tree is built, hence the file cannot be created "on-the-fly" by a resource. Maybe I should simply rephrase the whole sentence into this statement, as "eager/lazy" probably are not the best words to describe it.

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,435 @@
...
+			},
+			"org_id": {
+				Type:        schema.TypeString,
+				Description: "The organization of the Runtime Defined Entity",
+				Computed:    true,
+			},
+			"resolve": {
+				Type: schema.TypeBool,
+				Description: "If `true`, the Runtime Defined Entity will be resolved by this provider. If `false`, it won't be" +
+					"resolved and must be either done by an external component or with an update. The Runtime Defined Entity can't be" +
```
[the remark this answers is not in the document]
adambarreiro: Fixed, thanks!

vcd/resource_vcd_rde.go
adambarreiro: Fixed (also in docs), thanks!

vcd/resource_vcd_rde.go
adambarreiro: Fixed (also in docs), thanks!

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,435 @@
...
+		return diag.Errorf("could not delete the Runtime Defined Entity: %s", err)
+	}
+	return nil
+}
+
+// resourceVcdRdeImport is responsible for importing the resource.
+// The following steps happen as part of import
+// 1. The user supplies `terraform import _resource_name_ _the_id_string_` command
+// 2. `_the_id_string_` contains a dot formatted path to resource as in the example below
+// 3. The functions splits the dot-formatted path and tries to lookup the object
```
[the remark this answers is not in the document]
adambarreiro: Fixed, thanks!

vcd/resource_vcd_rde.go
adambarreiro: Fixed, thanks!

vcd/resource_vcd_rde_duplicate_test.go
adambarreiro: Fixed, thanks!

website/docs/r/rde.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+As per this last point, one needs to be careful when updating `input_entity` or `input_entity_url`, as Terraform will apply
+whatever changes were done, ignoring the real state from `computed_entity`. To perform a real update, one
+needs to check the contents of the `computed_entity` and do some diff with the original input.
+
+In other words:
+
+~> When you want to update an RDE and `entity_in_sync` is `false`, you should always merge the contents
+of `computed_entity` and `input_entity` to avoid overriding the whole entity by mistake with an old value. 
+
+## RDE resolution
```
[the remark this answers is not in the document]
adambarreiro: Also fixed

vcd/resource_vcd_rde_test.go
adambarreiro: > Just thinking: is it possible to put it in a GIST?

Yes, but somehow I feel that a Gist would be detached from this repository, in the sense that it would be owned by a user and not by this repository, which is making use of it.

> Maybe extend the bot to capture "githubusercontent.com" and auto-generate these?

Yes, could be useful.

vcd/resource_vcd_rde.go
```diff
@@ -0,0 +1,435 @@
...
+			"resolve_on_removal": {
+				Type: schema.TypeBool,
+				Description: "If `true`, the Runtime Defined Entity will be resolved before it gets deleted, to forcefully delete it." +
+					"Otherwise, destroy will fail if it is not resolved.",
+				Default:  false,
+				Optional: true,
+			},
+			"state": {
+				Type: schema.TypeString,
+				Description: "When created it will be in PRE_CREATED state. If the specified JSON in either `input_entity` or `input_entity_url` is correct, the state will be RESOLVED," +
```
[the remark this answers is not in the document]
adambarreiro: I've rephrased it to include a bit an explanation, both here and in data source (idem for docs)

vcd/resource_vcd_rde.go
adambarreiro: Good catch, it is indeed the case 4. I've fixed it.

website/docs/d/rde.html.markdown
dataclouder: just use the description that fits the example

vcd/resource_vcd_rde_test.go
lvirbalas: > Yes, but somehow I feel that a Gist would be detached from this repository

+1 for this. Let's have it in the repo.

vcd/config.go
```diff
@@ -82,6 +82,11 @@ const (
...
+// logged user change. Otherwise, retrieving or manipulating objects that require the new rights could return 403
+// forbidden errors. For example, adding read rights of a RDE Type to a specific user requires a cacheStorage.reset()
+// afterwards, to force a re-authentication. If this is not done, the cached client won't be able to read this RDE Type.
```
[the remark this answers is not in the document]
adambarreiro: Same as above, changed the location of the reset call in the test, to make it more precise.

dataclouder (approved):

adezxc (approved):

Issue #966, opened 2022-12-20
Crossplane provider


eremid:
Hello,

We are working on a Crossplane provider built with Upjet framework. I’m just start testing and I’ve some issues to share with you like :
* string parameters in documentation but bool in provider code.
* remove unused parameters like show_upload_progress in documentation (or adding it in provider code ?).
Examples :
* https://github.com/FrangipaneTeam/terraform-provider-vcd/commit/6ba2e07836b515bd954252034631a63584b44e42
* https://github.com/FrangipaneTeam/terraform-provider-vcd/commit/dd2b8041113f319cb989d17676ea0f1693449dfb

Reading your Code Contribution Flow, this issue is open to ask if it is possible to contribute towards the goal of having a working crossplane provider and how we can contribute ?

Thanks

dataclouder:
Hi
Thanks for your interest in our project.
Contributions are welcome. However, when it comes to documentation, we would like to keep the contributions to the matter of fixing errors.
Thus, removing an unneeded parameter in an example is a very useful fix, but changing every `*` to `-` or `_` is not, as it doesn't solve any problems, and it would make the rest of our documentation inconsistent.

Similarly, changing `"true"` to `true` is not necessary. Although stylistically it's better to use the unquoted `true` or `false`, Terraform doesn't complain and will [automatically convert from Boolean to string when needed](https://developer.hashicorp.com/terraform/language/expressions/types#type-conversion).
Should such a change become necessary in the future, we would prefer changing all of them at once using our own coding tools.

Summing up: if there are changes that affect the usefulness of the documentation, by all means do it, but if you feel suggesting formatting enhancements, be aware that we may not be able to dedicate review time to large stylistic changes with hundreds of lines involved.

WRT 
> it is possible to contribute towards the goal of having a working crossplane provider 

please explain what you have in mind, and how the current `terraform-provider-vcd` should be involved.
In other words, if the current code could be enhanced to be used in a crossplane provider, please outline what such enhancements should be.

eremid:
> Contributions are welcome. However, when it comes to documentation, we would like to keep the contributions to the matter of fixing errors.

OK.

> Similarly, changing "true" to true is not necessary.

It’s better for the crossplane provider, I detail this point below.

I’m using [upjet](https://github.com/upbound/upjet) code generator to build the crossplane provider. The crossplane APIs and kubernetes CRD generated from the terraform provider code missmatch with the example generated (based on the terraform provider documentation). 
For example :
[...]

PR #972, opened 2023-01-02
Fix catalog access

dataclouder:
This PR does the following:
* Allows accessing catalogs and related entities without having a full Org object to retrieve them, thus fixing access to shared catalogs (Issue #960)
* Add `catalog_id` to `vcd_catalog_media` resource and data source to allow its handling similar to `vcd_catalog_vapp_template`
* Add test framework for testing with multiple orgs (See `TESTING.md` for details)
* Use the multiple orgs framework to test shared and subscribed catalogs access
* Add test for subscribing a catalog from vSphere content library (Issue #967)

PR #978, opened 2023-01-05
doc: improve documentation

eremid:
string parameter to bool parameter if type is schema.TypeBool in code

FIX Issue #966


dataclouder:
Please note that these same changes were already applied in PR #972

eremid:
Ok, thanks

PR #981, opened 2023-01-10
Patch `TestAccVcdNsxVdcGroupCompleteMigration` import steps to ignore status fields

Didainius:
`TestAccVcdNsxVdcGroupCompleteMigration` _import_ checks sometimes fail due to VPN status changes as test steps progress. To patch this import steps will ignore these fields as it is not possible to predict them.

```
=== RUN   TestAccVcdNsxVdcGroupCompleteMigration
    resource_vcd_vdc_group_common_test.go:105: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

        (map[string]string) (len=2) {
         (string) (len=18) "ike_service_status": (string) (len=11) "NEGOTIATING",
         (string) (len=6) "status": (string) (len=4) "DOWN"
        }


        (map[string]string) (len=2) {
         (string) (len=18) "ike_service_status": (string) "",
         (string) (len=6) "status": (string) ""
        }
--- FAIL: TestAccVcdNsxVdcGroupCompleteMigration (289.85s)
```


adambarreiro (approved):
Thanks!

dataclouder (approved):

PR #983, opened 2023-01-12
Add leftovers removal and clean up tests

dataclouder:
* Add leftovers removal to test suite (it will run automatically)
    * `vcd/config_test.go`
    * `vcd/remove_leftovers_test.go` (new file)
    * `scripts/get_test_resource_names.go` (new file)
    * `scripts/runtest.sh`
    * `TESTING.md`
* Remove check for existing catalog before and after suite (it was added before we had a way of populating VCDs with automated operations)
    * `vcd/config_test.go` 
* Remove redundant `// +build` tags, which are not needed since we use Go 1.19+ by default
    * All files `vcd/*_.test.go` 

New testing options:

```
-vcd-silent-leftovers-removal              Omit details during removal of leftovers (false)
-vcd-skip-leftovers-removal                Do not attempt removal of leftovers at the end of the test suite (false)
```

**NOTES:**

1. The feature removal handles the following resources, which are the most likely ones to be left behind:
    * `vcd_org`
    * `vcd_org_vdc`
    * `vcd_catalog`
    * `vcd_catalog_vapp_template`
    * `vcd_catalog_media`
    * `vcd_network_*`
    * `vcd_vapp`
    * `vcd_vm`

2. The file `scripts/get_resource_names.go` gets copied to `./vcd/test-artifacts` during `make test-binary`, and can be used to generate a list of the resources created during tests. There shouldn't be a need to run it again, except when we notice that some test resources are left behind after a failed test.

3. to run the removal without another test run (for example, when the failing test was running in another machine) execute the following:

```
go test -tags functional -v -run NoSuchTest # or any other non-existing test name
```


vcd/resource_vcd_vapp_vm_multi_test.go
```diff
@@ -1,5 +1,4 @@
-//go:build multivm || ALL || functional
-// +build multivm ALL functional
+//go:build multivm || functional
```
adambarreiro, line 1: So, isn't `ALL` required in this case?

dataclouder: Not required, as the main test will work all the same. We had a recent case of ALL vs functional and we determined that ALL doesn't always do what it means :)
But ALL will be needed anyway. Adding it.

scripts/get_resource_names.go
```diff
@@ -0,0 +1,229 @@
+// get_resource_names collects the resource names from the test files in test-artifacts
```
lvirbalas, line 1: Can this script be named `get_test_resource_names.go`, so it's instantly clear it's about testing? Actual, as we have non testing items in this folder as well.

scripts/get_resource_names.go
```diff
@@ -0,0 +1,229 @@
...
+// By default, it skips the names starting with 'Test' or 'test', as they are otherwise
+// recognised and handled during the leftover removal operation.
+// The resulting list has to be added to `remove_leftover_test.go`
```
lvirbalas, line 4: The file is `...leftovers...go` (plural).

scripts/get_resource_names.go
```diff
@@ -0,0 +1,229 @@
...
+
+// Run the code in ./vcd/test-artifacts, possibly after running `make test-binary-prepare` in the top directory
```
lvirbalas, line 6: I can't find `./vcd/test-artifacts`.

scripts/get_resource_names.go
```diff
@@ -0,0 +1,229 @@
...
+// go run get_resource_names.go | vim -
```
lvirbalas, line 7: This is the first Go file in a "scripts" folder, though it is not a script. Feels a bit misplaced. This is probably less because its a to-be-compiled Go application, but more because it's purpose is hard to grasp in isolation. Maybe add a line on in what format it returns the result?

scripts/get_resource_names.go
dataclouder: Renamed

scripts/get_resource_names.go
dataclouder: fixed

scripts/get_resource_names.go
dataclouder: `./vcd/test-artifacts` gets created during `make test-binary`

scripts/get_resource_names.go
dataclouder: The result is described already in the main function, but I added it at the top of the file as well.
As for the misplaced file, it is a standalone program, and can't stay with the rest of the vcd files.
Since it needs to be used in `./vcd/test-artifacts`, it is hosted in `./scripts`, from where it gets copied to the appropriated place during `make test-binary`, same as the other files we use for that purpose.

Didainius (approved):
Thanks. I briefly played with it - looks to be working. I am fine and we can improve when there is a need.

adambarreiro (approved):

lvirbalas (approved):

PR #984, opened 2023-01-13
Add support for DNS settings in NSX-T backed external networks

### LINKED ISSUE #964: Add DNS field for external networks backed by NSX-t Segment
n external networks backed by nsx-t vlan segment , we don't have an option to add DNS servers unlike vlan backed by vds one.
as a result guest customization set no DNS servers in the OS level

We have confirm it's a UI miss, as it is available in the openAPI  https:///cloudapi/1.0.0/externalnetworks/<urn: id>
however applying the terraform script again remove this setting back to null.

so would u please update vcd_external_network_v2 to accept DNS setting for NSX-t as it's access for NSX-v version

image

Didainius:
Closes #964 

External network V2 resource can handle NSX-V and NSX-T based external networks (using the same API), however, DNS fields (dns1, dns2, dns_suffix) were only available for NSX-V based networks in UI. We also had a Terraform schema validation to prevent setting these fields for NSX-T networks.

Apparently, there are some use cases for having these fields set for NSX-T networks as well (DNS settings can be inherited for direct networks).

This PR - improves tests to test these fields and also removes schema validation. It turns out that the `dns_suffix` can be set, but is never returned by API which is a VCD bug. It should be fixed in future versions, but that test part remains commented for now as it doesn't work up to VCD 10.4.1.


website/docs/r/external_network_v2.html.markdown
```diff
@@ -179,9 +179,9 @@ The following arguments are supported:
 * `gateway` - (Required) Gateway of the network.
 * `prefix_length` - (Required) Network prefix.
 * `static_ip_pool` - (Required) IP ranges used for static pool allocation in the network.  See [IP Pool](#ip-pool) below for details.
-* `dns1` - (Optional) Primary DNS server. **Only valid** for NSX-V networks.
-* `dns2` - (Optional) Secondary DNS server. **Only valid** for NSX-V networks.
-* `dns_suffix` - (Optional) A FQDN for the virtual machines on this network. **Only valid** for NSX-V networks.
+* `dns1` - (Optional) Primary DNS server.
```
dataclouder, line 182: Do we need to indicate that NSX-T networks support starts with 3.9.0?

Didainius: > Do we need to indicate that NSX-T networks support starts with 3.9.0?

Yes. I added a note about it.

dataclouder (approved):

.changes/v3.9.0/948-improvements.md
```diff
@@ -0,0 +1,2 @@
+* `vcd_external_network_v2` allows setting DNS fields `dns1`, `dns2` and `dns_suffix` for NSX-T
+  backed entities [GH-948]
```
lvirbalas, line 2: Would it be possible to list the use case in here as well? Or are there more than one?

vcd/resource_vcd_external_network_v2.go
```diff
@@ -47,23 +47,17 @@ var networkV2IpScope = &schema.Resource{
...
 			ValidateFunc: validation.IsIPAddress,
-			// Only NSX-V allows configuring DNS
-			ConflictsWith: []string{"nsxt_network"},
 		},
 		"dns_suffix": {
 			Type:        schema.TypeString,
 			Optional:    true,
 			Description: "DNS suffix",
-			// Only NSX-V allows configuring DNS
-			ConflictsWith: []string{"nsxt_network"},
```
lvirbalas, line 66: As we are now lifting the validation you had previously for a reason, do we open up a possibility for misuse? If yes, does the API throw an error in that case and is the error descriptive enough?

Didainius: Yes. It was stated, that this was supposed to be only for NSX-V networks, however, it is not the case. Note, that as I stated in PR description - `dns_suffix` field still has a bug in VCD and it is not being set, but this is to be fixed in future VCD versions.

.changes/v3.9.0/948-improvements.md
Didainius: So far there is one known situation - I have added it now.

lvirbalas (approved):

adambarreiro (approved):
LGTM!

Issue #459, opened 2020-02-24
Add resources to create NSX-T security tags and manage assignment


jpbuecken:
Hi,

could you please add support to create security tags and manage it members:

This is the corresponding vCD documentation:

https://docs.vmware.com/en/vCloud-Director/9.5/com.vmware.vcloud.tenantportal.doc/GUID-681D5E9C-8DD1-485C-A007-DFD2C1AABB70.html

I'm not sure about the best way to implement management of membership. AWS, OCI and Openstack implement this as part of the port/network interface or of the VM/instance resource.
But vCD follow another approach in its GUI. Straight forward is to configure membership in the resource which create it.
Or you implement an attachment resource which attach a security tag to a member (VM, Network, ....)
All three ways will have its advantage and disadvantage.

### Terraform Version
ALL

### Affected Resource(s)
vcd_security_tag (suggestion)
vcd_security_tag_assignment (suggestion)

### Terraform Configuration Files
```hcl
resource "vcd_vapp_vm" "web1" {
[...]
}
resource "vcd_vapp_vm" "web2" {
[...]
}
resource "vcd_security_tag" "webrules" {
   name = "applicationset1"
   description = "Rules for application 1"
}

resource "vcd_security_tag_assignment" "webassign" {
   tagid = vcd_security_tag.webrules.id
   # array of instance ids
   members = [ vcd_vapp_vm.web1.id, vcd_vapp_vm.web2.id ]
}
```

Didainius:
Hi,
What is your end goal? Is it dynamic security groups for fire-walling?

jpbuecken:
Our product catalog allows you to order a server with an "intended use".
e.g. if you choose SAP or Webserver we can extend the firewall ruleset for further monitoring as a day one operation during installation.

Content and audit conform documentation of rules in the security group is done by team A. 
OS management is team B. 
The security tag feature allows team A to provide pre-defined rulesets which team B can use.

storbel:
@Didainius 
IS there a way to create dynamic groups with VCD? 
sound that "vcd_security_tag" is not yet implemented with vcd 3.51, have any clue when this will be implemented ?

Didainius:
@storbel ,
We're heads down for VDC group and general NSX-T support so these are definitely priorities however I can't yet promise if security tags will appear in the next release.

Didainius:
An update on this.

@jpbuecken , @storbel  - you had ideas so I'd like to share where we're going and possibly get your feedback:

[...]

PR #1006, opened 2023-02-23
Vm security tags

### LINKED ISSUE #894: vcd_vapp_vm: Add feature to add security tag during create
### Community Note

### Description

With provider 3.7.0 you added the possibility to manage vcd_nsxt_dynamic_security_group and vcd_security_tag, thanks for that.

But I need help for the following use case.
I think the following is a typical usage of the script feature of the guestcustomization

During the create phase of a VM, you may run an update of your operation system or establish a connection to your configuration managment (CM) and the CM will deploy further OS settings and applications.
The problem: To run updates or deploy a configuration, you already need firewall rules in place. Assume those rules are provided via security tags.

So you need the following process:
Create the VM but do not run guest customization / keep it power off
Add VM id to security tag
PowerOn VM and run guest customization

* vcd_vapp_vm

### Terraform Configuration (if it applies)

```hcl
resource "vcd_vapp_vm" "vm" {
[...]
security_tags = ["tag1","tag2",...]
power_on                 = true
[...]
customization {
[...]
   initscript                 = <<EOF
#!/bin/bash
if [ "$1" == "precustomization" ]; then
  echo "precustomization tasks"
  yum -y update
elif [ "$1" == "postcustomization" ]; then
  # callout configuration managment
EOF
}
[...]
```

Additional information:
- So the api call for security_tags = ["tag1","tag2",...] must run before api call power_on  = true is executed. 
- Our NSX team told me the terraform NSX-T provider does it this way (Manage Tag at VM)
- how about /1.0.0/securityTags/vm/{id} ?

### References

* Issue #459

bogi0704:
Fixes #894
I have added the capability to create `security_tags` using the `vcd_vapp_vm` ressource.
Honestly the hard lifting was already done, I have just added some logic to use the existing functions to handle entity specific security tags.
I have not added any automatic tests, since I am not really sure if every argumentsshould be tested. The current tests do not cover all possible arguments. My manual tests all worked though.


lvirbalas:
Maybe I'm missing something, but have you seen this:
https://registry.terraform.io/providers/vmware/vcd/latest/docs/resources/security_tag
https://github.com/vmware/terraform-provider-vcd/pull/845

bogi0704:
I am trying to sign my old commits, but it does not work...

@lvirbalas Yes, I know that that ressource exists. But we have a setup where we manage all the network, gateway, firewall etc. settings in one TF stack, while we are doing VMs in another. With that use-case it is impossible to use the `vcd_security_tag` ressource if your VM is doing some bootstrapping right after booting. That is why this is our preferred way of setting the security tags.

Didainius:
@bogi0704 ,
> I am trying to sign my old commits, but it does not work...

It is a bit ugly to get back. What I used to do in such cases is squash the commits into one (I usually use `git rebase -i _AFTER_COMMIT_HASH_) and sign it, then force push it back to the branch

vcd/resource_vcd_vapp_vm.go
```diff
@@ -617,6 +617,13 @@ func vmSchemaFunc(vmType typeOfVm) map[string]*schema.Schema {
 			Computed:    true, // As it can get populated automatically by VDC default policy
 			Description: "VM placement policy ID. Has to be assigned to Org VDC.",
 		},
+		"security_tags": {
```
Didainius, line 620: `read`  function of VM is shared with a data source `vcd_vapp_vm` so if you could also add the same field to `datasource_vcd_vapp_vm.go`. Terraform data source will crash otherwise by trying to set non-existing field.

In data source the field should be `computed` only:
```go
	"security_tags": {
			Type:        schema.TypeSet,
			Computed:    true,
			Description: "Security tags assigned to this VM",
			Elem:        &schema.Schema{Type: schema.TypeString},
		},
```

Didainius (commented):
Thanks for the PR @bogi0704,

I have left a comment inline, but I'd like to clarify on your use case so that it is useful for you, but also doesn't break workflows for other people.

What would be your end to end use case. Would you use this completely without `vcd_security_tag`?

How would you "bootstrap" your firewall? Would it be dynamic security group with some pattern matching? (because a security tag cannot exist without any VMs due to VCD behavior)

I remember there was this discussion when we implemented `vcd_security_tag` -> https://github.com/vmware/terraform-provider-vcd/issues/459#issuecomment-1122035797

With current code as it stands for someone to handle env using `vcd_security_tag`, he would also need to match relevant tags in the `vcd_vapp_vm`/`vcd_vm` as otherwise these resources would contradict each other (`vcd_security_tag` would add value, `vcd_vapp_vm` will attempt to remove it because it is not set). We could overcome this by setting `Optional: true` + `Computed: true` for the `security_tags` field in VM. When it is `Computed`, not having the field set will just populate it as attributes, but if it is set - it will generate diff. 

The same could happen for `vcd_security_tag` if new `vcd_vapp_vm` resources use a tag, but when creating this resource - at least one `vm_ids`  value is required (that is why we require at least 1 value), otherwise it would not create it. 

We should be able to make it suitable while navigating Terraform limitations, but just want to understand your workflow

bogi0704:
> What would be your end to end use case. Would you use this completely without `vcd_security_tag`?

Pretty much, yes.

> How would you "bootstrap" your firewall? Would it be dynamic security group with some pattern matching? (because a security tag cannot exist without any VMs due to VCD behavior)

Exactly. We are creating dynamic security groups where we are matching tags to define group memberships. We are using those groups in our firewall settings. This way we can use the same subnet for machines that require specific network connectivity while booting and for machines that don't.

> With current code as it stands for someone to handle env using `vcd_security_tag`, he would also need to match relevant tags in the `vcd_vapp_vm`/`vcd_vm` as otherwise these resources would contradict each other (`vcd_security_tag` would add value, `vcd_vapp_vm` will attempt to remove it because it is not set). We could overcome this by setting `Optional: true` + `Computed: true` for the `security_tags` field in VM. When it is `Computed`, not having the field set will just populate it as attributes, but if it is set - it will generate diff.

I was actually trying to address that issue by setting `Optional: true` + `Computed: true` as you suggested. I had to set `Computed: false` because when the attribute `security_tags` is not set at all in HCL, the provider does not care to remove them. So you might have been able to create the tags, but unsetting the attribute `security_tags` leads to no action being performed by terraform. That way you would be unable to unassign the security_tags from the VM and your firewalls would still let traffic flow, because the tags are still set.
After thinking about it some more I actually liked the idea of setting `Computed: false` more, because it is more consistent. Because of this you pretty much have to think carefully which tags you want to create by setting `security_tags` in the ressource `vcd_vapp_vm` and which by using the `vcd_security_tag` ressource. 
Also: If the platform supports it, why not also support it via the terraform provider. Let the user decide which way he should take. IMHO: It is better to set the tags via `security_tags` in the VM ressource.

Didainius:
> After thinking about it some more I actually liked the idea of setting Computed: false more, because it is more consistent. Because of this you pretty much have to think carefully which tags you want to create by setting security_tags in the ressource vcd_vapp_vm and which by using the vcd_security_tag ressource.
> Also: If the platform supports it, why not also support it via the terraform provider. Let the user decide which way he should take. IMHO: It is better to set the tags via security_tags in the VM ressource.

I agree with you but there are a few problems with that:
* it will break user workflows when they use `vcd_security_tag` (all VMs will prompt to remove tags after provider update)
* we enforce a new field for users that might have already altered by tags other means (`vcd_security_tag` or manually)

Ideally, we want to add support both - for setting and removing all tags in `vcd_vapp_vm`, but in a non-intrusive way - that is if someone doesn't want to use `security_tags` field (and have security tags set in some other way). That is the ugliest part of doing Terraform development - finding a way to overcome Terraform schema limitations :) 

I am going to do a few experiments to see if we can somehow accommodate it all

bogi0704:
> I agree with you but there are a few problems with that:
> 
>     * it will break user workflows when they use `vcd_security_tag` (all VMs will prompt to remove tags after provider update)
> 
>     * we enforce a new field for users that might have already altered by tags other means (`vcd_security_tag` or manually)
> 

You are right, I didn't think about that...

> I am going to do a few experiments to see if we can somehow accommodate it all

Thank you very much. I will also try to do some tests and see if I can find a way around that.

Didainius:
I was experimenting with `Optional: true` + `Computed: true` which is pretty much the only schema way of 
introducing a new field gracefully, without enforcing it.

I was hoping to have the difference between setting empty `security_tags` (`security_tags = []`) and
not having them set at all, but Terraform does not count this as a real difference (plan changes are
not reported at all, therefore we can't rely it on trigerring any operation)

As a result I think the best we can do is to ask user for an explicit value of an empty string
`security_tags = [""]`. When setting this, Terraform does detect drift and we can explicitly check
that if a single empty string is supplied as value - all existing security tags can be removed.

This is a bit of a hack around schema , but should accommodate all cases. Let me know what are your
thoughts on such approach?

bogi0704:
> As a result I think the best we can do is to ask user for an explicit value of an empty string `security_tags = [""]`. When setting this, Terraform does detect drift and we can explicitly check that if a single empty string is supplied as value - all existing security tags can be removed.

So if `security_tags = ["testtag"]` have been set once and the user wants to remove all `security_tags`:
Rather than removing/uncommenting the tags like this `#security_tags = ["testtag"]` the user must set the tags to an emtpy string `security_tags = [""]` for the tags to actually be removed, right?

What if we find a way to tell the `vcd_vapp_vm` ressource how to work with existing `vcd_security_tag`. I was thinking of maybe introducing a boolean flag, like `override_existing_security_tags`. I don't know if this is possible, but setting `override_existing_security_tags = true` would lead to `Computed: false` in the schema. If that is not possible we could maybe use that flag to set the default value of `security_tags` to `security_tags = [""]` rather than `security_tags = []`. We would have a solution where the user would be able to decide which way is the preferred way of setting `tags`. 

That workaround could maybe be a v2 of this feature since it would mean additional work. Having a way to set the tags at all is still a huge win :)

Didainius:
> So if security_tags = ["testtag"] have been set once and the user wants to remove all security_tags:
Rather than removing/uncommenting the tags like this #security_tags = ["testtag"] the user must set the tags to an emtpy string security_tags = [""] for the tags to actually be removed, right?

I would expect so - we still might need to get to full implementation to see if there are any ugly sides of that (its a hack after all).

> What if we find a way to tell the vcd_vapp_vm ressource how to work with existing vcd_security_tag. I was thinking of maybe introducing a boolean flag, like override_existing_security_tags. I don't know if this is possible, but setting override_existing_security_tags = true would lead to Computed: false in the schema.

Hm. You could try, but I doubt this is possible because resource schema should get registered to Terraform before actually evaluating its field values.

> If that is not possible we could maybe use that flag to set the default value of security_tags to security_tags = [""] rather than security_tags = []. We would have a solution where the user would be able to decide which way is the preferred way of setting tags.

Would you find this valuable, given that we still have `Optional: true` + `Computed: true`?

bogi0704:
> Would you find this valuable, given that we still have `Optional: true` + `Computed: true`?

At least I think that it would be valuable, because it would behave like a user is intuitively thinking about it.

Didainius:
I drifted from this feature, but we need to take the decision. So what are your thoughts about it?
Which version do we aim to test (as there can still be unexpected outcomes with Terraform limitations)?

We need to use the `Optional+Computed` so that we introduce the feature gradually instead of enforcing it.

The options we have now are:
1. `Default` Terraform behavior `security_tags = ["tag1", "tag2"]`. When this field is removed - tags are not removed - they only stop being managed. A possible workaround for users could be to set 1 invalid tag which doesn't make sense at all if they want to isolate VMs network
2. Try to use `security_tags = [""] ` explicitly for deletion (this needs experimentation but might cause trouble during `read` phase)
3. Try to use default behavior of `security_tags` (as in option 1), but additionally have a field `override_existing_security_tags = true`. However, I doubt this can change schema behavior (schema is initialized before this code). It could maybe remove all tags when explicitly set from `false` to `true` because otherwise Terraform simply will not trigger an `update` and no actions will happen



As for my outlook now the safest is `1` (because that is how Terraform schema naturally works), however, it has this limitation for tag removal.

Didainius:
> sorry for my late response. I agree to option `1` since you think that it would be the safest. We should just clarify the limitations in the documentation.
> Using Terraform modules I would get around the limitations when removing the tags.

Thanks. So I guess the only thing you need to do in code is to make the `Optional+Computed=true` for `security_tags` field in the resource.

Regarding the documentation, we need a sentence that removing the `security_tags` field from HCL configuration will not remove tags, but stops managing them

lvirbalas:
My main comment is that we need to have explanations and cross-links in docs on how this new field "works" with the existing `security_tag` resource. We need to describe that in both the new doc page and the existing one of `security_tag`.

bogi0704:
Yes, I agree. I hope I will be able to do it this week and make a proposal for the doc texts.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1064,6 +1079,12 @@ func createVmFromTemplate(d *schema.ResourceData, meta interface{}, vmType typeO
 		return nil, fmt.Errorf("error refreshing VM %s : %s", vmName, err)
 	}
 
+	// Handle VM Security Tags settings
+	err = createOrUpdateVmSecurityTags(d, vm)
+	if err != nil {
+		return nil, fmt.Errorf("[VM create] error creating security tags for VM %s : %s", vm.VM.Name, err)
+	}
+
```
Didainius, line 1087: In fact, this one and https://github.com/vmware/terraform-provider-vcd/pull/1006/files#diff-d888384050e0ea29e45bdfbcd476b8c09c90ec98ab5c9bfbcfe5491760497847R1425-R1430 are not needed. 

The way it works is `genericResourceVmCreate` conditionally calls `createVmFromTemplate` or `createVmEmpty` based on the type of VM and then finalizes by performing "common" actions for all VMs. Security tags are common and it is sufficient to have it here https://github.com/vmware/terraform-provider-vcd/pull/1006/files#diff-d888384050e0ea29e45bdfbcd476b8c09c90ec98ab5c9bfbcfe5491760497847R801-R808

vcd/resource_vcd_vapp_vm_tools.go
```diff
@@ -1216,3 +1216,25 @@ func updateHardwareVersionAndOsType(d *schema.ResourceData, vm *govcd.VM) error
...
+		if err != nil {
+			return fmt.Errorf("error setting VM Security Tags: %s", err)
+		}
+	} else {
+		_, err = vm.UpdateVMSecurityTags(entitySecurityTags)
+		log.Printf("[DEBUG] security_tags are NOT set %s", entitySecurityTags)
+		if err != nil {
+			return fmt.Errorf("error setting VM Security Tags: %s", err)
+		}
+	}
```
Didainius, line 1238: We don't need this branch as it will always wipe security tags even if `security_tags` are not set

Didainius (commented):
Thanks for changes. I had a refresher look before running tests. There are a few optimization things and this `else` branch needs to be removed as it will always remove tags 
https://github.com/vmware/terraform-provider-vcd/pull/1006/files#diff-005d997aed7b0dfa6e5b99f3ef615083be5aa6cf8e034e09f8ab1d0189d13012R1232-R1238

Didainius:
One more thing - we need a changelog line. This should go into a `.changes/v3.9.0/1006-improvements.md`.
You could see other files for inspiration but this should be something similar to (no need to put exact words, just putting out sample structure :)):
```
* `vcd_vapp_vm` and `vcd_vm` resources support security tag management via new field `security_tags` [GH-1006]
```

lvirbalas:
@bogi0704 NOTE: looking at the dates, to be able to pull this PR into 3.9.0, it has to be finalized by the end of this week. In otger words, there is a sense of urgency now :)

vcd/datasource_vcd_vapp_vm.go
```diff
@@ -383,6 +383,13 @@ func vcdVmDS(vmType typeOfVm) map[string]*schema.Schema {
 			Computed:    true,
 			Description: "VM placement policy ID.",
 		},
+		"security_tags": {
+			Type:        schema.TypeSet,
+			Optional:    true,
```
adambarreiro, line 388: If we only plan to _read_ security tags **in the data source**, I think `Optional` is probably not needed.

website/docs/d/vapp_vm.html.markdown
```diff
@@ -118,6 +118,7 @@ The following arguments are supported:
 * `placement_policy_id` - (*v3.8+*) VM placement policy ID.
 * `status` - (*v3.8+*) The vApp status as a numeric code.
 * `status_text` - (*v3.8+*) The vApp status as text.
+* `security_tags` - (*v3.9+*) List of security tags assigned to this VM.
```
adambarreiro, line 121: nit, just to be more precise (as there's also this `TypeList`):
```suggestion
* `security_tags` - (*v3.9+*) Set of security tags assigned to this VM.
```

website/docs/r/vapp_vm.html.markdown
```diff
@@ -503,6 +503,7 @@ example for usage details.
   using `vcd_org_vdc.vm_placement_policy_ids` (and optionally `vcd_org_vdc.default_compute_policy_id` to make it default).
   In this case, if the placement policy is not set, it will pick the VDC default on creation. It must be set explicitly
   if one wants to update it to another policy (the VM requires at least one Compute Policy), and needs to be set to `""` to be removed.
+* `security_tags` - (Optional; *v3.9+*) List of security tags to be managed by the `vcd_vapp_vm` ressource. Removing `security_tags` will not remove the security tags themselves. They remain unchanged and just stop being managed by `security_tags` in the `vcd_vapp_vm` ressource. This is to keep backwards compatibility with existing security tags that were created by the `vcd_security_tags` ressource.
```
adambarreiro, line 506: nit, just to be more precise (as there's also this `TypeList`):
```suggestion
* `security_tags` - (Optional; *v3.9+*) Set of security tags to be managed by the `vcd_vapp_vm` resource. Removing `security_tags` will not remove the security tags themselves. They remain unchanged and just stop being managed by `security_tags` in the `vcd_vapp_vm` resource. This is to keep backwards compatibility with existing security tags that were created by the `vcd_security_tags` resource.
```

> They remain unchanged and just stop being managed by `security_tags` in the `vcd_vapp_vm` resource

Should we mention what to do to actually delete them? If I understood the conversation with @Didainius, it was `security_tags = [""]`?

vcd/resource_vcd_vapp_vm.go
bogi0704: I had a hunch that it should be sufficient to put it into `genericResourceVmCreate `, but I was not sure. Thanks for clarifying this :)

vcd/resource_vcd_vapp_vm_tools.go
bogi0704: I removed the if-branch and reworked the error checking and logging a little bit.

bogi0704:
I think that I have adressed all your comments. Thanks for helping me out :) 
I tested the new version and it worked for me. I hope your tests will run smoothly as well.

lvirbalas (approved):
Thank you for your contribution!

adambarreiro (approved):

Didainius (approved):
Thank you. Tests on `vm vapp` tags passed with a known unrelated anomaly of:
```
=== CONT  TestAccVcdVmInternalDisk
    testing_new_import_state.go:125: unexpected output type: <nil>
```

(we will look into it separately)

Didainius:
Thank you very much @bogi0704 ,
This is now in main. We will make another PR on top to add specific tests and unless there is something really odd (which I doubt) this will be in `v3.9.0`

PR #1016, opened 2023-03-09
Testing PR: include govcd branch with govcd PR 528

Didainius:
This is not to be merged, but just testing to run Terraform with SDK pull request https://github.com/vmware/go-vcloud-director/pull/528


Didainius:
Was never intended to be merged

PR #1018, opened 2023-03-14
Add RDE metadata

adambarreiro:
## Description
This PR adds metadata support to the `vcd_rde` resource and data source by adding the `metadata_entry` attribute, which is a set.

The peculiar characteristic about this one is that this is the first OpenAPI metadata feature, so it differs a bit from the existing implementation (XML based). Main differences:

- There can be multiple metadata entries with **same key**, but **different namespace**.
- Types are reduced in comparison to XML (there's no date, for instance).
- Each entry has an ID that has to be respected by the Provider during Terraform operations (this is guaranteed by `getOpenApiMetadataOperations`).

## Nuances

As `metadata_entry` is a set, doing any modification to an entry shows that all entries will be deleted and re-created during `terraform plan`, which is not true. The code is "smart" enough (`getOpenApiMetadataOperations `) to perform the minimal changes, that is, respects the metadata that is not changed and doesn't touch them during `terraform apply`.

This nuance has been mentioned as a **note** in the docs. More info:

- https://discuss.hashicorp.com/t/typeset-resource-always-want-to-destroy-and-recreate-everything-in-set-when-one-item-has-a-diff/54895
- https://discuss.hashicorp.com/t/typeset-is-not-picking-up-correct-changes/23735/2

## Sample snippet
```hcl
provider "vcd" {
# ...  
}


data "vcd_rde_type" "rde_type" {
  vendor  = "vmware"
  nss     = "tkgcluster"
  version = "1.0.0"
}
resource "vcd_rde" "test-rde" {
  org          = "System"
  rde_type_id  = data.vcd_rde_type.rde_type.id
  name         = "TestAccVcdRdeMetadata"
  input_entity = "{\"foo\":\"bar\"}" # We are just testing metadata so we don't care about entity state
  resolve      = true

  metadata_entry {
    key    = "stringKey1"
    value  = "stringValue1"
    type   = "StringEntry"
    domain = "TENANT"

    readonly   = false
    persistent = false
  }
  metadata_entry {
    key    = "numberKey1"
    value  = "1"
    type   = "NumberEntry"
    domain = "TENANT"

    readonly   = false
    persistent = false
  }
  metadata_entry {
    key    = "boolKey1"
    value  = "false"
    type   = "BoolEntry"
    domain = "TENANT"

    readonly   = false
    persistent = false
  }
  metadata_entry {
    key    = "readOnly1"
    value  = "readOnly1"
    type   = "StringEntry"
    domain = "TENANT"

    readonly   = true
    persistent = false
  }
  metadata_entry {
    key        = "namespace"
    value      = "namespace1"
    type       = "StringEntry"
    domain     = "TENANT"
    namespace  = "namespace1"
    readonly   = false
    persistent = false
  }
  metadata_entry {
    key        = "namespace"
    value      = "namespace2"
    type       = "StringEntry"
    domain     = "TENANT"
    namespace  = "namespace2"
    readonly   = false
    persistent = false
  }
  metadata_entry {
    key    = "provider1"
    value  = "provider1"
    type   = "StringEntry"
    domain = "PROVIDER"

    readonly   = false
    persistent = false
  }
  metadata_entry {
    key    = "persistent1"
    value  = "persistent1"
    type   = "StringEntry"
    domain = "TENANT"

    readonly   = false
    persistent = true
  }
}

```

Remember that RDEs don't have any UI, so these are the endpoints to perform checks in the backend:
```shell
# Get RDEs
https://{{vcd_host}}/cloudapi/1.0.0/entities/types/vmware/tkgcluster/1.0.0/

# Get specific RDE with ID metadata (change {{rde_id}} with the RDE urn from previous call, or get it from Terraform state/output)
https://{{vcd_host}}/cloudapi/1.0.0/entities/{{rde_id}}/metadata
```

To check namespace+key uniqueness, you can add blocks like
```hcl
metadata_entry {
    key        = "namespace" # Repeat the same key
    value      = "namespace3" # Change the value (otherwise TypeSet considers a duplicate entry and does nothing)
    type       = "StringEntry"
    domain     = "TENANT"
    namespace  = "namespace2" # Repeat the same namespace
    readonly   = false
    persistent = false
}
```


vcd/metadata_openapi.go
```diff
@@ -0,0 +1,329 @@
...
+		return nil
+	}
+
+	oldRaw, newRaw := d.GetChange("metadata_entry")
+	metadataToAdd, metadataToUpdate, metadataToDelete, err := getMetadataOperations(oldRaw.(*schema.Set).List(), newRaw.(*schema.Set).List())
+	if err != nil {
+		return fmt.Errorf("could not calculate the needed metadata operations: %s", err)
+	}
+
+	// getMetadataOperations retrieves keys and namespaces merged with a separator, this functions
```
Didainius:
```suggestion
	// getMetadataOperations retrieves keys and namespaces merged with a separator, this function
```

vcd/resource_vcd_rde.go
```diff
@@ -320,6 +339,16 @@ func resourceVcdRdeUpdate(ctx context.Context, d *schema.ResourceData, meta inte
...
+	// Metadata is only supported since v37.0
+	if vcdClient.Client.APIVCDMaxVersionIs(">= 37.0") {
+		err = createOrUpdateOpenApiMetadataEntryInVcd(d, rde)
+		if err != nil {
+			return diag.Errorf("could not create metadata for the Runtime Defined Entity: %s", err)
+		}
+	} else if _, ok := d.GetOk("metadata_entry"); ok {
+		return diag.Errorf("metadata_entry is only supported since VCD 10.4.0")
+	}
+
```
Didainius, line 351: I am thinking now - we don't even support versions < 10.4.0 anymore with next version - maybe we can simplify it.
Comment works for other places and tests as well

Didainius (commented):
I see it is in progress yet, but one thing that caught my eye - I left one comment, but this is throughout this PR - we shouldn't test if VCD < 10.4.0 as 3.11 will not support it all anyway. The error user would get error straight from SDK as we require API to be >= 37.0

vcd/metadata_openapi.go
adambarreiro: Fixed!

vcd/resource_vcd_rde.go
adambarreiro: Done, thanks for the heads up.

vcd/metadata_openapi.go
```diff
@@ -0,0 +1,329 @@
...
+			namespace = metadataEntry["namespace"].(string)
+		}
+
+		value, err := convertOpenApiMetadataValue(metadataEntry["type"].(string), metadataEntry["value"].(string))
+		if err != nil {
+			return nil, fmt.Errorf("error parsing the 'value' attribute '%s' from state: %s", metadataEntry["value"].(string), err)
+		}
+
+		// In OpenAPI, metadata is namespaced, hence it is possible to have same keys but in different namespaces
+		namespacedKey := fmt.Sprintf("%s%%%%%%%s", namespace, metadataEntry["key"].(string))
```
Didainius, line 248: IMO would be good to put a comment line about the separator string (even the detail that each `%` needs to be escaped with `%`) as I was double-checking in [playground](https://go.dev/play/p/G5D3Y0pwBcB) if it really matches https://github.com/vmware/terraform-provider-vcd/pull/1018/files#diff-a2bd1e75e42abfead79ed7048b71ae16920bc8df863d39a4e15ce5f1842e456cR151

vcd/metadata_openapi.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	// getOpenApiMetadataOperations retrieves keys and namespaces merged with a separator, this function
+	// splits the values in two: namespace and key, separately.
+	getKeyAndNamespace := func(namespacedKey string) (string, string, error) {
+		r := strings.Split(namespacedKey, "%%%") // Separator used by getOpenApiMetadataEntryMap
+		if len(r) == 2 {
+			return r[0], r[1], nil
+		}
+		return "", "", fmt.Errorf("bad formatting of metadata map key %s, this is a provider error", namespacedKey)
```
Didainius, line 155: Would be good to clear out this part `this is a provider error` - it is ambiguous. Does it want to say that Terraform provider VCD error?

website/docs/r/rde.html.markdown
```diff
@@ -185,6 +186,50 @@ should have `resolve=false` to avoid being resolved).
...
+    value    = "bar"
+    domain   = "TENANT"
+    readonly = true
+  }
+  metadata_entry {
+    key       = "bar"
+    type      = "NumberEntry"
+    value     = "42"
+    domain    = "TENANT"
+    permanent = true
```
Didainius, line 229:
```suggestion
    persistent = true
```

Didainius (commented):
Some more comments from me. I am testing it.

vcd/metadata_openapi.go
adambarreiro: Yes. Ideally it should never happen, but if it does, I wanted to make it clear that it's an issue in our side, not VCD.

vcd/metadata_openapi.go
adambarreiro: Done, let me know whether it is better now or is still confusing.

vcd/metadata_openapi.go
adambarreiro: Changed to a more concise error.

vcd/metadata_openapi.go
adambarreiro: Actually this suggestion made me think whether this was needed at all or I was overcomplicating things (from some leftovers of previous ideas/issues). Turns out that yes, this was unnecessary at all. So:
- This key generation is simplified, no `%%%` anymore
- The function to retrieve keys is removed.

vcd/metadata_openapi.go
adambarreiro: As per the comment above, I've removed this function, not needed at all as this was overcomplicated.

website/docs/r/rde.html.markdown
```diff
@@ -185,6 +186,55 @@ should have `resolve=false` to avoid being resolved).
 In this last scenario, it is advisable to mark `resolve_on_removal=true` so Terraform can delete the RDE even if it was not
 resolved by anyone.
 
+<a id="metadata"></a>
+## Metadata
+
+-> Due to Terraform limitations, `terraform plan` shows that all metadata entries will be deleted and re-created when performing
+a change in of one or more of them, but applying it will respect every non-changed metadata entry by not performing any modification on
+these.
```
lvirbalas, line 194: I'm trying to simplify and clarify the statement:

```suggestion
-> Due to Terraform limitations, when performing a change in any of the metadata entries, a `terraform plan` shows that all of them will be deleted and re-created. However, the succeeding `terraform apply` will respect the non-changed metadata entries by not modifying them.
```

adambarreiro: Yeah, sounds way better. Changed, thanks!

lvirbalas (approved):

website/docs/r/rde.html.markdown
```diff
@@ -185,6 +186,55 @@ should have `resolve=false` to avoid being resolved).
...
+The `metadata_entry` is a set of metadata entries that have the following structure:
+
+* `key` - (Required) Key of this metadata entry.
+* `value` - (Required) Value of this metadata entry.
+* `type` - (Optional) Type of this metadata entry. One of: `StringEntry`, `NumberEntry`, `BoolEntry`. Defaults to `StringEntry`.
+* `namespace` - (Optional) Namespace of the metadata entry. Allows having multiple entries with same key in different namespaces.
+* `domain` - (Optional) Only meaningful for providers. Allows them to share entries with their tenants. Currently, accepted values are: `TENANT`, `PROVIDER`. Defaults to `TENANT`.
+* `readonly` - (Optional) `true` if the metadata entry is read only. Defaults to `false`.
+* `persistent` - (Optional) `true` if the metadata is persistent. Persistent entries can be copied over on some entity operation
+  (e.g. Creating a copy of a VDC, capturing a vApp to a template, instantiating a catalog item as a VM...). Defaults to `false`.
```
dataclouder, line 207: This example, given that we are in the RDE documentation, is potentially confusing. Is there an example related to RDE that we can give?

website/docs/r/rde.html.markdown
```diff
@@ -185,6 +186,55 @@ should have `resolve=false` to avoid being resolved).
...
+    type     = "StringEntry"
+    value    = "bar"
+    domain   = "TENANT"
+    readonly = true
+  }
+  metadata_entry {
+    key        = "bar"
+    type       = "NumberEntry"
+    value      = "42"
+    domain     = "TENANT"
```
dataclouder, line 233:
```suggestion
    domain     = "TENANT"  # will be also visible to an organization user
```

website/docs/r/rde.html.markdown
```diff
@@ -185,6 +186,55 @@ should have `resolve=false` to avoid being resolved).
...
+  org         = "my-org"
+  rde_type_id = data.vcd_rde_type.my-type.id
+  name        = "My custom RDE"
+  resolve     = true
+  entity_url  = "https://just.an-example.com/entities/custom-rde.json"
+  metadata_entry {
+    key      = "foo"
+    type     = "StringEntry"
+    value    = "bar"
+    domain   = "TENANT"
```
dataclouder, line 226:
```suggestion
    domain   = "PROVIDER"  # will be only visible to the provider 
```

vcd/metadata_openapi_common_test.go
```diff
@@ -0,0 +1,282 @@
...
+	preTestChecks(t)
+	var params = StringMap{
+		"Org":  testConfig.VCD.Org,
+		"Vdc":  testConfig.Nsxt.Vdc,
+		"Name": t.Name(),
+	}
+
+	// This output allows to perform some assertions on the ID inside a TypeSet,
+	// which is impossible to obtain otherwise.
+	outputHcl := `
```
dataclouder, line 30: Any chance we can improve the formatting of this generated code?
Here's an example:
```
resource "vcd_rde" "test_rde" {
  org          = "System"
  rde_type_id  = data.vcd_rde_type.rde_type.id
  name         = "TestAccVcdRdeMetadata"
  input_entity = "{\"foo\":\"bar\"}" # We are just testing metadata so we don't care about entity state
  resolve      = true

		  metadata_entry {
			key         = "stringKey1"
			value       = "stringValue1"
			type        = "StringEntry"
			domain      = "TENANT"

			readonly    = false
            persistent  = false
		  }
		  metadata_entry {
			key         = "numberKey1"
			value       = "1"
			type        = "NumberEntry"
			domain      = "TENANT"

			readonly    = false
            persistent  = false
		  }
		  metadata_entry {
			key         = "boolKey1"
			value       = "false"
			type        = "BoolEntry"
			domain      = "TENANT"

			readonly    = false
            persistent  = false
		  }
		  metadata_entry {
			key         = "readOnly1"
			value       = "readOnly1"
			type        = "StringEntry"
			domain      = "TENANT"

			readonly    = true
            persistent  = false
		  }
		  metadata_entry {
			key         = "namespace"
			value       = "namespace1"
			type        = "StringEntry"
			domain      = "TENANT"
            namespace   = "namespace1"
			readonly    = false
            persistent  = false
		  }
		  metadata_entry {
			key         = "namespace"
			value       = "namespace2"
			type        = "StringEntry"
			domain      = "TENANT"
            namespace   = "namespace2"
			readonly    = false
            persistent  = false
		  }
		  metadata_entry {
			key         = "provider1"
			value       = "provider1"
			type        = "StringEntry"
			domain      = "PROVIDER"

			readonly    = false
            persistent  = false
		  }
		  metadata_entry {
			key         = "persistent1"
			value       = "persistent1"
			type        = "StringEntry"
			domain      = "TENANT"

			readonly    = false
            persistent  = true
		  }
}

		output "metadata_id" {
           value = tolist(vcd_rde.test_rde.metadata_entry)[0].id
        }
```
It doesn't have to be perfect, but a bit more readable would be good

vcd/metadata_openapi.go
```diff
@@ -0,0 +1,325 @@
...
+			}
+			if oldEntry.IsReadOnly != newEntry.IsReadOnly || oldEntry.IsPersistent != newEntry.IsPersistent ||
+				oldEntry.KeyValue.Namespace != newEntry.KeyValue.Namespace || oldEntry.KeyValue.Domain != newEntry.KeyValue.Domain ||
+				oldEntry.KeyValue.Value.Type != newEntry.KeyValue.Value.Type {
+				util.Logger.Printf("[DEBUG] entry with namespace '%s' and key '%s' is being deleted and re-created", oldEntry.KeyValue.Namespace, oldEntry.KeyValue.Key)
+				metadataToRemove = append(metadataToRemove, oldMetadataEntries[newNamespacedKey])
+				metadataToCreate = append(metadataToCreate, newMetadataEntries[newNamespacedKey])
+			} else {
+				metadataToUpdateMap[newNamespacedKey] = newEntry
+			}
```
Didainius, line 205: So the last hook here for me to clarify (and possibly worth a comment in this block) - update can happen only if value `KeyValue.Value.Value` change? Otherwise we remove/create metadata, right? I think you put it in website docs, but sucha comment explaining this code block could help is in future.

vcd/metadata_openapi_common_test.go
```diff
@@ -0,0 +1,282 @@
...
+	withDefaults := templateFill(resourceTemplate, params)
+	debugPrintf("#[DEBUG] CONFIGURATION WithDefaults: %s", withDefaults)
+
+	if vcdShortTest {
+		t.Skip(acceptanceTestsSkipped)
+		return
+	}
+
+	// This is used to validate that metadata IDs don't change despite of an update/delete.
+	cachedId := testCachedFieldValue{}
```
Didainius, line 87: I guess this reflects it better as it is checking for `Value`

```suggestion
	cachedValue := testCachedFieldValue{}
```

Didainius (approved):

vcd/metadata_openapi_common_test.go
adambarreiro: Actually I'm saving the ID of one metadata entry here (so I can test that IDs don't change with subsequent updates/deletes)

vcd/metadata_openapi_common_test.go
adambarreiro: Should be fixed now. I also amended identation in the XML metadata tests

website/docs/r/rde.html.markdown
adambarreiro: Good idea, changed

website/docs/r/rde.html.markdown
adambarreiro: Good idea, changed

website/docs/r/rde.html.markdown
adambarreiro: I need to check, as far as I know RDEs can't be re-used/inherited elsewhere.

vcd/metadata_openapi.go
adambarreiro: Good point, added a couple of comments.

website/docs/r/rde.html.markdown
dataclouder: I see two possibilities:

1. We move this section to a location (such as a file in the `guides`) where the definition can apply to several resources. (BRW, if the API mention VMs, it is possible that we could start dealing with VM metadata using OpenAPI, but it is not the moment to try it).
2. We put a generic description, saying "not applicable to RDE as of v3.11".

dataclouder (approved):

adezxc (approved):

Issue #987, opened 2023-01-25
Add argument for "Share the read-only access to all other organizations" under vcd_catalog_access_control


PabloRapidScale:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Under resource **vcd_catalog_access_control**, you have **shared_with_everyone** as an option that matches the GUI "Share Catalog" option for the Users and Groups tab. But I do not see an equivalent option for "Share the read-only access to all other organizations" under the organization tab. See screenshot below.

![image](https://user-images.githubusercontent.com/117782445/214679565-d99f66c3-1996-4e57-bc3f-662377fd2fd9.png)

It would be great to expose that as a false or true optional argument.

### New or Affected Resource(s)

* vcd_catalog_access_control

### Terraform Configuration (if it applies)

```hcl
resource "vcd_catalog_access_control" "catalog_acl" {
  org                   = var.org_name
  catalog_id         = vcd_catalog.catalog.id
  shared_with_everyone  = true
  everyone_access_level = "ReadOnly"
}
```
### References

* Issue #915

dataclouder:
Implemented in PR #1020

PR #1020, opened 2023-03-17
Add 'read_only_shared_with_other_orgs' sharing property

dataclouder:
Update `vcd_catalog_access_control` with property `read_only_shared_with_other_orgs`, to address Issue #987 

Testing

There is a curious failure that happens in some resources. It has been addressed with a not entirely satisfying retry mechanism.

To see the failure, run
```
VCD_SKIP_RETRY=1 go test  -tags catalog -v -timeout 0 -run 'TestAccVcdCatalogAccessControl2'
```

To see where the retry acts, and the failure would have occurred without it, run
```
VCD_RETRY_VERBOSE=1 go test  -tags catalog -v -timeout 0 -run 'TestAccVcdCatalogAccessControl2'
```
Each occurrence of `attempts=1` (or =`2`) shows an averted failure.


vcd/resource_vcd_catalog_access_control.go
```diff
@@ -237,3 +319,59 @@ func resourceVcdCatalogAccessControlImport(_ context.Context, d *schema.Resource
...
+
+// runWithRetry allows to run a given operation and retry it in case of error until the timeout is reached
+// * operationDescription is a human understandable description of the operation
+// * errorMessage is what to say in case of final error
+// * timeout is for how long we retry in case of failure
+// * preRun is an (optional) operation to run before attempting the operation
+// * operation is the main operation we are running
+func runWithRetry(operationDescription, errorMessage string, timeout time.Duration, preRun func() error, operation func() (any, error)) (any, error) {
+	// ---------------------------------------------------------
+	// TODO: remove this check after PR review
```
dataclouder, line 331: self review: reminder to change this after review

vcd/resource_vcd_catalog_access_control.go
```diff
@@ -237,3 +319,59 @@ func resourceVcdCatalogAccessControlImport(_ context.Context, d *schema.Resource
...
+		if preRun != nil {
+			err = preRun()
+			if err != nil {
+				return nil, err
+			}
+		}
+		result, err = operation()
+		if err == nil {
+			// ---------------------------------------------------------
+			// TODO: convert printing on screen to logging after PR review
```
dataclouder, line 363: self review: reminder to change this after review

website/docs/r/catalog_access_control.html.markdown
```diff
@@ -90,6 +102,7 @@ The following arguments are supported:
   this property must be set to `false`.
 * `everyone_access_level` - (Optional) Access level when the Catalog is shared with everyone (it can only be set to
   `ReadOnly`). Required if `shared_with_everyone` is set.
+* `read_only_shared_with_all_orgs` - (Optional; *v3.11+*) If true, the catalog is shared as read-only with all organizations
```
lvirbalas, line 105: Consistency nit to other bullet items:

```suggestion
* `read_only_shared_with_all_orgs` - (Optional; *v3.11+*) If true, the catalog is shared as read-only with all organizations.
```

vcd/resource_vcd_catalog_access_control.go
```diff
@@ -237,3 +319,59 @@ func resourceVcdCatalogAccessControlImport(_ context.Context, d *schema.Resource
...
 	return []*schema.ResourceData{d}, nil
 }
+
+// runWithRetry allows to run a given operation and retry it in case of error until the timeout is reached
+// * operationDescription is a human understandable description of the operation
+// * errorMessage is what to say in case of final error
+// * timeout is for how long we retry in case of failure
+// * preRun is an (optional) operation to run before attempting the operation
+// * operation is the main operation we are running
+func runWithRetry(operationDescription, errorMessage string, timeout time.Duration, preRun func() error, operation func() (any, error)) (any, error) {
```
lvirbalas, line 329: This could be a public function in some utility file, or close to the query operations?

vcd/resource_vcd_org_user.go
```diff
@@ -233,10 +234,24 @@ func resourceToUserData(d *schema.ResourceData, meta interface{}) (*govcd.OrgUse
 func resourceToOrgUser(d *schema.ResourceData, meta interface{}) (*govcd.OrgUser, *govcd.AdminOrg, error) {
 
 	vcdClient := meta.(*VCDClient)
-	adminOrg, err := vcdClient.GetAdminOrgFromResource(d)
+	sessionInfo, err := vcdClient.Client.GetSessionInfo()
 	if err != nil {
-		return nil, nil, fmt.Errorf("[resourceToOrgUser] error retrieving org: %s", err)
+		return nil, nil, fmt.Errorf("error retrieving session info: %s", err)
 	}
+	result, err := runWithRetry("[resourceToOrgUser]",
```
lvirbalas, line 241: This PR contains almost ten new runWithRetry calls. All of them would benefit from a comment why are we doing this and what failure exactly we are targeting.

Speaking of which, what would happen if there's an error, but the error is not what we expect and it is an error which won't go away after retrying? Won't it cause the plan apply to stuck for the duration of the timeout? Right now it looks like we're "casting a wide net" here, which may catch true errors too. What about narrowing down the `runWithRetry` function to retry only on a specific condition/failure/error message? It doesn't have to be too specific, but could be a substring-match, which narrows down the scope.

lvirbalas (changes requested):

vcd/resource_vcd_catalog_access_control.go
dataclouder: Since it is only used within this repo, I think it should stay private so far.
If we need to use it in other contexts, we can move it to govcd and make it public.

vcd/resource_vcd_org_user.go
dataclouder: I tried to find a root cause, without avail.
If I has found such pattern, probably I would have used a different approach. The usage of `runWithRetry` happens because these failing calls only fail when we deal with shared catalog, and they manifest themselves as temporary access denied incidents, involving catalogs, users, and orgs. In each case, the error wording is different.
As it is now, the scope is the narrowest we could use.

lvirbalas: We had discussed this more, and there's just no clear way to narrow down the scope of these retries. Resolving this thread.

lvirbalas (approved):

website/docs/r/catalog_access_control.html.markdown
dataclouder: fixed

Didainius (approved):
I have tested it manually - setting and resetting it worked.

.changes/v3.9.0/1020-improvements.md
```diff
@@ -0,0 +1 @@
+* Add property `read_only_shared_with_all_orgs` to vcd_catalog_access_control to address Issue #987 [GH-1020]
```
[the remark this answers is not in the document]
dataclouder: Yes. Been brewing this one quite a lot.
Thanks. Removing the old one

vcd/resource_vcd_catalog_access_control.go
```diff
@@ -135,9 +149,21 @@ func resourceVcdCatalogAccessControlCreateUpdate(ctx context.Context, d *schema.
 		EveryoneAccessLevel: everyoneAccessLevel,
 		AccessSettings:      accessSettingsList,
 	}
-	err = catalog.SetAccessControl(&accessControlParams, true)
+	_, err = runWithRetry(sessionText,
+		"error when setting Catalog control access parameters",
+		time.Second*30,
+		nil, //func() error { return catalog.Refresh() },
```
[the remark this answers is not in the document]
dataclouder: fixed

vcd/resource_vcd_catalog_access_control2_test.go
```diff
@@ -0,0 +1,473 @@
...
+	debugPrintf("#[DEBUG] UPDATE CONFIGURATION: %s", updateText)
+	debugPrintf("#[DEBUG] CHECK CONFIGURATION: %s", checkText)
+
+	resourceAC0 := "vcd_catalog_access_control." + acCatalog0
+	resourceAC1 := "vcd_catalog_access_control." + acCatalog1
+	resourceAC2 := "vcd_catalog_access_control." + acCatalog2
+	resourceAC3 := "vcd_catalog_access_control." + acCatalog3
+	resourceAC4 := "vcd_catalog_access_control." + acCatalog4
+	//resourceAC5 := "vcd_catalog_access_control." + acCatalog5
+	//resourceAC6 := "vcd_catalog_access_control." + acCatalog6
```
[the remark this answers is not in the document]
dataclouder: removed

vcd/resource_vcd_org.go
```diff
@@ -89,6 +89,28 @@ func resourceOrg() *schema.Resource {
...
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "Number of VDCs, owned or shared, available to this organization",
+			},
+			"list_of_vdcs": {
+				Type:        schema.TypeSet,
+				Computed:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Description: "List of VDCs, owned or shared, available to this organization",
+			},
```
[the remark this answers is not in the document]
dataclouder: Thanks for the suggestion. 
Added a check on `number_of_catalogs` and `list_of_catalogs` in `resource_vcd_catalog_access_control2_test.go` and a check on `number_of_vdcs` in `org_vdc_common_test.go`

adambarreiro (approved):
LGTM!
I tried with and without retry and got the expected results.

adezxc (approved):
Tested with and without env variables and got the expected errors

vcd/resource_vcd_catalog_access_control.go
```diff
@@ -237,3 +319,59 @@ func resourceVcdCatalogAccessControlImport(_ context.Context, d *schema.Resource
...
+	// ---------------------------------------------------------
+	if os.Getenv("VCD_SKIP_RETRY") != "" {
+		if preRun != nil {
+			err := preRun()
+			if err != nil {
+				return nil, err
+			}
+		}
+		return operation()
+	}
```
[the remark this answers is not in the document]
dataclouder: The environment variable is used when it contains anything except the empty string. In this context, "0" and "1" and "Tooth Fairy" will have the same effect.

PR #996, opened 2023-02-14
Adjust vcd_nsxt_alb_settings to support IPv6 and Transparent mode

Didainius:
This PRs builds on top of https://github.com/vmware/go-vcloud-director/pull/549 and adds two new fields to `vcd_nsxt_alb_settings` resource and data source:
* `ipv6_service_network_specification` (requires VCD 10.4.0+). 
* `is_transparent_mode_enabled` (requires VCD 10.4.1+). This field allows using a "Preserve Client IP" on a Virtual Service. Note. The Virtual Service resource (`vcd_nsxt_alb_virtual_service`) will be updated in separate PRs during this release cycle.

PR #1024, opened 2023-03-21
NSX-T ALB Virtual Service Transparent mode and Pool Firewall Group membership (VCD 10.4.1+)

Didainius:
This PR continues on ALB feature catchup (started in  #996), where `vcd_nsxt_alb_settings` added support for Transparent mode.

This PR adds two new things (both require VCD 10.4.1+):
* `vcd_nsxt_alb_virtual_service` support for Transparent mode (via field `is_transparent_mode_enabled`)
* `vcd_nsxt_alb_pool` Group membership via `member_group_id` field (members of ALB Pool can come from IP Set instead of being manually enabled) 

**Note** There is quite a list of prerequisites to enable Transparent mode for ALB Virtual Services now. They are well outlined in [this blog post](https://fojta.wordpress.com/2022/12/16/new-networking-features-in-vmware-cloud-director-10-4-1/) (section _Transparent Load Balancing_) however, I don't think it is worth putting it all in our docs as it may soon get outdated due to this being fresh technology. I also observed that errors are meaningful when creation fails due to missed requirements.

#### UI views

`vcd_nsxt_alb_virtual_service`

image

`vcd_nsxt_alb_pool`
image

#### Testing 
* Test leftover cleanup support for all ALB resources added. They must _always_ be removed in reverse order as each of them depend on the parent. The cleanup should look similar to this (I am open to discussion if we need to make the indentation somehow different although it does reflect dependencies correctly):
```
....
Start leftovers removal
[vcd_nsxt_alb_controller] TestAccVcdNsxtAlbVirtualService (DELETE)
        [vcd_nsxt_alb_cloud] TestAccVcdNsxtAlbVirtualService-alb-cloud (DELETE)
                [vcd_nsxt_alb_service_engine_group] TestAccVcdNsxtAlbVirtualService-se-group (DELETE)
                        [vcd_nsxt_alb_settings] Testnsxt-gw-datacloud (DELETE)
                                [vcd_nsxt_alb_edgegateway_service_engine_group] TestAccVcdNsxtAlbVirtualService-se-group (DELETE)
                                        [vcd_nsxt_alb_pool] TestAccVcdNsxtAlbVirtualService-pool (DELETE)
                                                [vcd_nsxt_alb_virtual_service] TestAccVcdNsxtAlbVirtualService (DELETE)
[vcd_org] System (keep)
[vcd_org] datacloud (keep)
....
```

Tested on 10.3.0, 10.3.3, 10.4.0, 10.4.1

### DIFF
--- a/scripts/skip-upgrade-tests.txt
+++ b/scripts/skip-upgrade-tests.txt
@@ -263,3 +263,5 @@ vcd.TestAccVcdVAppVmMulti.tf v3.7.0 "vApp network removal from powered on vApp i
 vcd.ResourceSchema-vcd_nsxt_alb_settings.tf v3.8.2 "New fields to support latest ALB features"
 vcd.ResourceSchema-vcd_vapp_org_network.tf v3.8.2 "New field 'reboot_vapp_on_removal'"
 vcd.ResourceSchema-vcd_vapp_network.tf v3.8.2 "New field 'reboot_vapp_on_removal'"
+vcd.ResourceSchema-vcd_nsxt_alb_virtual_service.tf v3.8.2 "New field 'is_transparent_mode_enabled'"
+vcd.ResourceSchema-vcd_nsxt_alb_pool.tf v3.8.2 "New field 'member_group_id'"
--- a/vcd/datasource_vcd_nsxt_alb_pool.go
+++ b/vcd/datasource_vcd_nsxt_alb_pool.go
@@ -113,7 +113,7 @@ func datasourceVcdAlbPool() *schema.Resource {
 			"member_group_id": {
 				Type:        schema.TypeString,
 				Computed:    true,
-				Description: "ID of Firewall Group to use for pool Membership (VCD 10.4.1+)",
+				Description: "ID of Firewall Group to use for Pool Membership (VCD 10.4.1+)",
 			},
 			"health_monitor": {
 				Type:     schema.TypeSet,
--- a/vcd/datasource_vcd_nsxt_alb_pool.go
+++ b/vcd/datasource_vcd_nsxt_alb_pool.go
@@ -110,6 +110,11 @@ func datasourceVcdAlbPool() *schema.Resource {
 					},
 				},
 			},
+			"member_group_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "ID of Firewall Group to use for pool Membership (VCD 10.4.1+)",
+			},
 			"health_monitor": {
 				Type:     schema.TypeSet,
 				Computed: true,
--- a/vcd/datasource_vcd_nsxt_alb_virtual_service.go
+++ b/vcd/datasource_vcd_nsxt_alb_virtual_service.go
@@ -100,6 +100,11 @@ func datasourceVcdAlbVirtualService() *schema.Resource {
 					},
 				},
 			},
+			"is_transparent_mode_enabled": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Preserves Client IP on a Virtual Service when enabled",
+			},
 		},
 	}
 }
--- a/vcd/datasource_vcd_nsxt_alb_virtual_service.go
+++ b/vcd/datasource_vcd_nsxt_alb_virtual_service.go
@@ -103,7 +103,7 @@ func datasourceVcdAlbVirtualService() *schema.Resource {
 			"is_transparent_mode_enabled": {
 				Type:        schema.TypeBool,
 				Computed:    true,
-				Description: "Preserves Client IP on a Virtual Service when enabled",
+				Description: "Preserves Client IP on a Virtual Service when enabled (VCD 10.4.1+)",
 			},
 		},
 	}
--- a/vcd/remove_leftovers_test.go
+++ b/vcd/remove_leftovers_test.go
@@ -265,7 +265,6 @@ func removeLeftovers(govcdClient *govcd.VCDClient, verbose bool) error {
 				}
 			}
 		}
-
 	}
 	// --------------------------------------------------------------
 	// RDE Types
@@ -486,7 +485,6 @@ func removeLeftoversNsxtAlbTenant(govcdClient *govcd.VCDClient, verbose bool) er
 				}
 			}
 		}
-
 	}
 
 	return nil
--- a/vcd/remove_leftovers_test.go
+++ b/vcd/remove_leftovers_test.go
@@ -425,7 +425,7 @@ func removeLeftoversNsxtAlbTenant(govcdClient *govcd.VCDClient, verbose bool) er
 			}
 
 			for _, albEdgeGatewayServiceEngineGroup := range albEdgeGatewayServiceEngineGroups {
-				// Edge Gateay Service Engine Group Assignment does not have names, therefore we use
+				// Edge Gateway Service Engine Group Assignment does not have names, therefore we use
 				// Service Engine Group name as a name for the resource
 				edgeServiceEngineGroupAssignmentName := albEdgeGatewayServiceEngineGroup.NsxtAlbServiceEngineGroupAssignment.ServiceEngineGroupRef.Name
 				toBeDeletedAlbEdgeGatewayServiceEngineGroup := shouldDeleteEntity(alsoDelete, doNotDelete, edgeServiceEngineGroupAssignmentName, "vcd_nsxt_alb_edgegateway_service_engine_group", 4, verbose)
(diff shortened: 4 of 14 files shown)


vcd/remove_leftovers_test.go
```diff
@@ -291,6 +301,195 @@ func removeLeftovers(govcdClient *govcd.VCDClient, verbose bool) error {
...
+		// --------------------------------------------------------------
+		albSetting, err := edgeGateway.GetAlbSettings()
+		if err != nil {
+			return fmt.Errorf("error retrieving NSX-T ALB Edge Gateway settings: %s", err)
+		}
+
+		if albSetting.Enabled {
+			// Real name of Edge Gateway would not be matched, but ALB needs to be deleted therefore adding a prefix Test
+			edgeNameWithPrefix := "Test" + edgeGateway.EdgeGateway.Name
+			toBeDisabledAlbSettings := shouldDeleteEntity(alsoDelete, doNotDelete, edgeNameWithPrefix, "vcd_nsxt_alb_settings", 3, verbose)
```
adambarreiro, line 415: As it matches the `isTest` regexp (we force to have the "Test" prefix), and the resources created by tests are not in the `doNotDelete` slice, wouldn't this always be `true`?

adambarreiro (commented):
Thanks for the improvements, just a few doubts

adezxc (approved):

scripts/skip-upgrade-tests.txt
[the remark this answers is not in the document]
Didainius: This is a file that marks specific tests to skip in `upgrade` test mode `make test-upgrade`.
Whenever we run `upgrade` tests - there are test artifacts being created in `vcd/test-artifacts`. These are snippets to run (you can use `make test-upgrade-prepare` to only generate, but not run them all).

Among many snippets files, it generates `vcd.ResourceSchema-*` files for each resource we have. This is based  on a specific data source we have -> https://registry.terraform.io/providers/vmware/vcd/latest/docs/data-sources/resource_schema

Whenever we have a schema change - this particular test (for that resource) will fail because it will detect that schema of a particular resource was changed when migrating from `PREVIOUS_VERSION` to `VERSION` (we have such files in our root directory of this repo)

You are right about the `prefix_length` field (which would have caused an issue) but because I already added that resource to the list (https://github.com/vmware/terraform-provider-vcd/pull/1024/files#diff-fd4815413ffffe38827635f089c2a2867e3673ed52f0a3435e2247f0aad87976R265) when adding `reboot_vapp_on_removal` field - you didn't need to add it anymore.

I saw that when reviewing your PR, but didn't dig into the details - hopefully this puts things into perspective.

vcd/remove_leftovers_test.go
Didainius: Yes - it will.
Due to the nature of ALB we have all of the resources always match now (we can't keep the controller we need to test we can create it).

The only reason why I tried to follow the pattern for all ALB components is if for whatever reason we decide to keep it (maybe during the development phase).

The ALB Settings itself is even uglier because it is not a separate entity, but only a setting of Edge Gateway

vcd/resource_vcd_nsxt_alb_virtual_service.go
```diff
@@ -256,6 +262,15 @@ func getNsxtAlbVirtualServiceType(d *schema.ResourceData) (*types.NsxtAlbVirtual
 		albVirtualServiceConfig.CertificateRef = &types.OpenApiReference{ID: d.Get("ca_certificate_id").(string)}
 	}
 
+	// Setting transparent mode is only possible in VCD 10.4.1+ (37.1+), throw error otherwise
+	if !d.GetRawConfig().GetAttr("is_transparent_mode_enabled").IsNull() {
```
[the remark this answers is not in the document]
Didainius: Yes. This looks to be working that way. Although we're replacing a `deprecated` function with `experimental` - however it looks to be working :)

vcd/datasource_vcd_nsxt_alb_virtual_service.go
```diff
@@ -100,6 +100,11 @@ func datasourceVcdAlbVirtualService() *schema.Resource {
 					},
 				},
 			},
+			"is_transparent_mode_enabled": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Preserves Client IP on a Virtual Service when enabled",
```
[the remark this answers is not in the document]
Didainius: Yes. This won't harm.

vcd/remove_leftovers_test.go
```diff
@@ -291,6 +301,195 @@ func removeLeftovers(govcdClient *govcd.VCDClient, verbose bool) error {
...
+			// --------------------------------------------------------------
+			filterOnEdgeGateway := url.Values{}
+			filterOnEdgeGateway.Add("filter", fmt.Sprintf("(gatewayRef.id==%s)", edgeGateway.EdgeGateway.ID))
+			albEdgeGatewayServiceEngineGroups, err := govcdClient.GetAllAlbServiceEngineGroupAssignments(filterOnEdgeGateway)
+			if err != nil {
+				return fmt.Errorf("error retrieving NSX-T ALB Edge Gateway Service Engine Group assignments list: %s", err)
+			}
+
+			for _, albEdgeGatewayServiceEngineGroup := range albEdgeGatewayServiceEngineGroups {
+				// Edge Gateay Service Engine Group Assignment does not have names, therefore we use
```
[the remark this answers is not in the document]
Didainius: Fixed.

vcd/remove_leftovers_test.go
adambarreiro: Got it, thanks!

vcd/resource_vcd_nsxt_alb_virtual_service.go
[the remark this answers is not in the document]
adambarreiro: Got it, thanks!

dataclouder (approved):

vcd/datasource_vcd_nsxt_alb_pool.go
```diff
@@ -110,6 +110,11 @@ func datasourceVcdAlbPool() *schema.Resource {
 					},
 				},
 			},
+			"member_group_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "ID of Firewall Group to use for pool Membership (VCD 10.4.1+)",
```
lvirbalas, line 116: Polishing:

```suggestion
				Description: "ID of Firewall Group to use for Pool Membership (VCD 10.4.1+)",
```

vcd/resource_vcd_nsxt_alb_pool.go
```diff
@@ -90,10 +90,17 @@ func resourceVcdAlbPool() *schema.Resource {
...
+				Type:          schema.TypeSet,
+				Optional:      true,
+				Elem:          nsxtAlbPoolMember,
+				Description:   "ALB Pool Members",
+				ConflictsWith: []string{"member_group_id"},
+			},
+			"member_group_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "ID of Firewall Group to use for pool Membership (VCD 10.4.1+)",
```
lvirbalas, line 102: Polishing:

```suggestion
				Description:   "ID of Firewall Group to use for Pool Membership (VCD 10.4.1+)",
```

website/docs/r/nsxt_alb_virtual_service.html.markdown
```diff
@@ -129,6 +129,41 @@ resource "vcd_nsxt_alb_virtual_service" "test" {
...
+  member_group_id = data.vcd_nsxt_ip_set.frontend.id
+}
+
+resource "vcd_nsxt_alb_virtual_service" "test" {
+  org = "my-org"
+
+  name            = "new-virtual-service"
+  edge_gateway_id = data.vcd_nsxt_edgegateway.existing.id
+
+  # Preserve Client IP - can only be enabled in VCD 10.4.1+ (must also be enabled in `vcd_nsxt_alb_settings`)
```
lvirbalas, line 153: BTW, what happens if user enables this setting on lower VCD version?

Didainius: We have validation for it https://github.com/vmware/terraform-provider-vcd/pull/1024/files#diff-818eadb1083a77a7d7e5f532244049aa39ef464d5fd645518386c9d48e85392eR267-R269

Without it, one would get an API error of an unrecognized field on older versions.

PR #1031, opened 2023-03-27
add "reboot_vapp_on_removal" field to test's template

adezxc:
Simple PR that adds `reboot_vapp_on_removal: True` field to the `testAccCheckVappNetwork_basic_ipv6` template as it caused the binary test to fail.

### DIFF
--- a/vcd/resource_vcd_vapp_network_test.go
+++ b/vcd/resource_vcd_vapp_network_test.go
@@ -115,6 +115,7 @@ func TestAccVcdVappNetwork_Isolated_ipv6(t *testing.T) {
 		"orgNetworkForUpdate":         " ",
 		"retainIpMacEnabled":          "false",
 		"retainIpMacEnabledForUpdate": "false",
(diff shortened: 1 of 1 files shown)


Didainius (approved):
Thanks - tests passed. 
You could add this PR to existing issue 1007

You would do it this way:

```
* Add `prefix_length` field to `vcd_vapp_network` as creating IPv6 vApp networks was not supported due to the lack of a suitable subnet representation (Issue #999) [GH-1007, GH-1031]
```

lvirbalas (commented):
Did the test start failing _after_ we introduced `reboot_vapp_on_removal`? If yes, I wonder why, as the parameter is optional?
https://registry.terraform.io/providers/vmware/vcd/latest/docs/resources/vapp_network#reboot_vapp_on_removal

p.s. Also, a nit: please capitalize the first word "Add ..." in the PR title :)

lvirbalas (approved):

PR #1034, opened 2023-03-27
Go 1.20 support for testing scripts

Didainius:
Go 1.20 removed `-i` flag (was deprecated in 1.19) and `make test-binary-prepare` fails on Go 1.20 now. This PR fixes it.

It also uses Go 1.20 in GitHub actions now.

Additionally, it bumps `staticcheck` to latest version `2023.1.3`


lvirbalas (approved):
Thanks! This PR also reminds me that we saw a parameter deprecation in `goreleaser` (was it `--rm-dist`, which will be removed around June?):
https://goreleaser.com/deprecations/

PR #1040, opened 2023-04-03
Enable service account authentication

adezxc:


vcd/provider.go
```diff
@@ -246,6 +246,13 @@ func Provider() *schema.Provider {
 				Description: "The API token used instead of username/password for VCD API operations. (Requires VCD 10.3.1+)",
 			},
 
+			"service_account": {
```
adambarreiro, line 249: Maybe it's too long, but should it be `service_account_token_file`? Or something shorter, but that highlights that it is a file and contains the token.

website/docs/index.html.markdown
```diff
@@ -192,6 +192,30 @@ resource "vcd_network_routed" "net1" {
 
 Note that when connecting with API tokens you can't create or modify users, roles, global roles, or rights bundles.
 
+### Connecting with a Service Account API token
+
+With VCD 10.4.0+, you can connect using a service account API token, as defined in the [documentation](https://blogs.vmware.com/cloudprovider/2022/07/cloud-director-service-accounts.html). Because a new API token is provided on every authentication request, the user is required to provide a file in `json` format with the current API key. e.g:
```
adambarreiro, line 197: Should we state that the file is read+write?

vcd/provider.go
adezxc: Changed. I don't like that it's very long, but it probably provides much more context to the user.

website/docs/index.html.markdown
adezxc: Changed.

adezxc (commented):
Reminder to not forget about changelog entry, either in v3.9.0 or v3.10.0

vcd/provider.go
adambarreiro: Yeah, let's see what @Didainius and @lvirbalas think about it also.

lvirbalas: I vote for `service_account_token_file` in this case too. Mainly because `provider` section does not take a lot of space in overall configuration and it's better to be extra clear here.

website/docs/index.html.markdown
```diff
@@ -192,6 +192,30 @@ resource "vcd_network_routed" "net1" {
...
+```
+
+~> **NOTE:** The service account needs to be in `Active Stage` and it's up to the user to provide the initial API token.
+
+```hcl
+provider "vcd" {
+  user                 = "none"
+  password             = "none"
+  auth_type            = "service_account"
+  api_token            = "token.json"
```
lvirbalas, line 209: Probably it's me and Friday, but shouldn't the example include `service_account_token_file` parameter when `auth_type "service_account"`?

Reference:
https://github.com/vmware/terraform-provider-vcd/pull/1040/files#diff-e96a5cf811e4de4a40ed325399541cc28aef6972ab94e260331cb58898cd6a8bR366-R369

website/docs/index.html.markdown
```diff
@@ -192,6 +192,30 @@ resource "vcd_network_routed" "net1" {
...
+  sysorg               = "System"
+  org                  = var.vcd_org # Default for resources
+  vdc                  = var.vcd_vdc # Default for resources
+  url                  = var.vcd_url
+  max_retry_timeout    = var.vcd_max_retry_timeout
+  allow_unverified_ssl = var.vcd_allow_unverified_ssl
+}
+```
+
```
lvirbalas, line 218: It seems like PR is missing docs of the new fields. I mean, we have example above, but no additions below this place:
https://registry.terraform.io/providers/vmware/vcd/latest/docs#auth_type

vcd/provider.go
```diff
@@ -246,6 +246,13 @@ func Provider() *schema.Provider {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("VCD_SA_TOKEN_FILE", nil),
+				Description: "The Service Account API token file instead of username/password for VCD API operations. (Requires VCD 10.4.0+)",
```
lvirbalas, line 253: With `token`, `api_token`, `saml_adfs_rpt_id` and `service_account_token_file` (what did I miss?) parameters available now, it is becoming increasingly easier to mis-configure.

It seems, we do have runtime validation (is it enough?):
https://github.com/vmware/terraform-provider-vcd/pull/1040/files#diff-e96a5cf811e4de4a40ed325399541cc28aef6972ab94e260331cb58898cd6a8bR366-R369

But it looks like the `auth_type` doc section needs to be improved:
https://registry.terraform.io/providers/vmware/vcd/latest/docs#auth_type

Specifically:
- `saml_adfs` should state which additional parameter to use with it: https://registry.terraform.io/providers/vmware/vcd/latest/docs#saml_adfs
- This section of the docs should describe the new service account option too

CC: @Didainius @adambarreiro @dataclouder

lvirbalas (changes requested):
Unless, I missed it, we need more docs :)

vcd/config_test.go
```diff
@@ -743,7 +744,7 @@ func getVcdVersion(config TestConfig) (string, error) {
 	if vcdClient == nil || err != nil {
 		return "", err
 	}
-	err = ProviderAuthenticate(vcdClient, config.Provider.User, config.Provider.Password, config.Provider.Token, config.Provider.SysOrg, config.Provider.ApiToken)
+	err = ProviderAuthenticate(vcdClient, config.Provider.User, config.Provider.Password, config.Provider.Token, config.Provider.SysOrg, config.Provider.ApiToken, config.Provider.ServiceAccountToken)
```
adambarreiro, line 747: During a second review, I saw this and I thought, why don't we pass `config` to `ProviderAuthenticate` so it has less arguments? As it seems that it can be prone to add more and more (like now) and it's definitely a code smell.

But then I saw that `ProviderAuthenticate` is public, is there any reason for that? As it is in `vcd` package and our provider should not export anything IMHO. It looks like a convenience private function to me.

If we would like to refactor it, maybe we could:
- Instead of passing members of `config.Provider` to `ProviderAuthenticate(vcdClient, c.Provider...)`, make `(c *Config) providerAuthenticate(vcdClient)`, or in other words, make `providerAuthenticate` a method of `Config`.
- As the function is used for testing, maybe we could do a wrapper:
```
func (testConfig *TestConfig) providerAuthenticateForTests(client *govcd.VCDClient) error {
	c := &Config{
		User:            testConfig.Provider.User,
		// ...
	}
	return c.providerAuthenticate(client, c)
}
```

As this is a huge refactoring that affects almost everything, I'm just adding this as a note here, maybe we could just leave a TODO or FIXME or something and tackle this more calmly in the next release.

Thoughts? @lvirbalas @Didainius

vcd/provider.go
Didainius: +1 for `service_account_token_file`

vcd/config_test.go
Didainius: Yeah, agree. It has outgrown original signature. It is just a question of when we do it :)

What I don't remember is why is it public @dataclouder might remember if we need it together with a few other functions `DataSources` and `Resources`.

In general - this is not a library and is not meant to be used as one so we don't care that much about changing signature.

Didainius:
It might be convenient to provide a similar script (https://registry.terraform.io/providers/vmware/vcd/latest/docs#shell-script-to-obtain-a-bearer-token) to get the Service Account active.

I also have a consideration that maybe it would be less friction to accept initial token as an attribute in `provider` section and do token rotation in file further.

adezxc:
> It might be convenient to provide a similar script (https://registry.terraform.io/providers/vmware/vcd/latest/docs#shell-script-to-obtain-a-bearer-token) to get the Service Account active.

I created the script `create_service_account.sh` which goes from creation to active stage, but it's a 100 lines, so it's probably sub optimal to put it in the documentation and a link to the file could be given instead?

> I also have a consideration that maybe it would be less friction to accept initial token as an attribute in provider section and do token rotation in file further.
I agree that it's less friction for the user, but is there a nice way of making it work? How would the provider know about the file the next time it is run? Also I feel like having files as the default just provides more flexibility to the user as to how they would want to make use multiple service account tokens

Didainius:
> I created the script `create_service_account.sh` which goes from creation to active stage, but it's a 100 lines, so it's probably sub optimal to put it in the documentation and a link to the file could be given instead?

Yes. It could probably live in `examples` subdirectory. We may think of a `guide` page in future how to use service accounts, but this is good for now.

The script itself is great - thanks. IMO this makes it so much easier to start with service accounts for those who don't have any infrastructure around them. My only wish (up to discussion) is probably to add `-s` to `curl` commands so that it doesn't spit out these:
```
nter VCD Role in URN format (e.g. urn:vcloud:role:System%20Administrator):
urn:vcloud:role:System%20Administrator
Enter software version (optional):
1
Enter client URI (optional):

Creating service account...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   483    0   313  100   170    345    187 --:--:-- --:--:-- --:--:--   536
```




> I agree that it's less friction for the user, but is there a nice way of making it work? How would the provider know about the file the next time it is run? Also I feel like having files as the default just provides more flexibility to the user as to how they would want to make use multiple service account tokens

Agreed. Can live like that and then see if there is any demand to shape it in future

vcd/config_test.go
adezxc: I added a TODO above the function as a reminder. I'll look into it but thought we should tackle it in a different PR to keep this one small and easier to review

vcd/provider.go
adezxc: I updated the docs, mostly, to stay consistent but this part:

> saml_adfs should state which additional parameter to use with it: https://registry.terraform.io/providers/vmware/vcd/latest/docs#saml_adfs

Do you mean we should mention `saml_adfs_rpt_id` should be used with it?

website/docs/index.html.markdown
adezxc: Fixed.

vcd/provider.go
adezxc: Changed.

website/docs/index.html.markdown
adezxc: Updated.

website/docs/index.html.markdown
```diff
@@ -293,6 +321,13 @@ The following arguments are used to configure the VMware Cloud Director Provider
    environment variable. This token requires at least VCD 10.3.1. There are restrictions to its use, as defined in
    [the documentation](https://docs.vmware.com/en/VMware-Cloud-Director/10.3/VMware-Cloud-Director-Service-Provider-Admin-Portal-Guide/GUID-A1B3B2FA-7B2C-4EE1-9D1B-188BE703EEDE.html)
 
+* `service_account_token_file` - (Optional; *v3.9+*) This is the file that contains a Service Account API token. The
```
adezxc, line 324: Question, do we still try to pull this into v3.9.0 or just move it to v3.10.0? Will keep this unresolved until we reach a decision when it is good to get merged.

website/docs/index.html.markdown
```diff
@@ -293,6 +321,13 @@ The following arguments are used to configure the VMware Cloud Director Provider
...
+   path to the file could be provided as absolute or relative to the working directory. It is used instead of username
+   and password (in combination with `auth_type=service_account_token_file`. The file can also be specified with the 
+   `VCD_SA_TOKEN_FILE` environment variable. This option requires at least VCD 10.4.0.
```
lvirbalas, line 327: Let's move required VCD version to the parenthesis, like here:
https://registry.terraform.io/providers/vmware/vcd/latest/docs/resources/vapp_vm#sizing_policy_id

vcd/provider.go
adezxc: Updated, let me know if the current way is okay.

website/docs/index.html.markdown
adezxc: Got it, introduced a changelog entry.

vcd/provider.go
lvirbalas: @adambarreiro @Didainius would you review this change too:
https://github.com/vmware/terraform-provider-vcd/pull/1040/files#diff-4839085b71de83d7294472f5f8fa6652faf4b0b24df211c8eaf945ac96933aa5R305-R311

lvirbalas (approved):

vcd/provider.go
adambarreiro: LGTM!

Didainius: LGTM

website/docs/index.html.markdown
```diff
@@ -192,6 +192,33 @@ resource "vcd_network_routed" "net1" {
 
 Note that when connecting with API tokens you can't create or modify users, roles, global roles, or rights bundles.
 
+### Connecting with a Service Account API token
+
+With VCD 10.4.0+, you can connect using a service account API token, as defined in the [documentation](https://blogs.vmware.com/cloudprovider/2022/07/cloud-director-service-accounts.html). Because a new API token is provided on every authentication request, the user is required to provide a readable+writable file in `json` format with the current API key. e.g:
+```json
+{"refresh_token":"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"}
+```
+
```
Didainius, line 201: One more thing. Sounds obvious for us, but  should we have a note, highlighting that the file must be treated as secret because it contains auth details.

website/docs/index.html.markdown
```diff
@@ -192,6 +192,33 @@ resource "vcd_network_routed" "net1" {
...
+~> **NOTE:** The service account needs to be in `Active Stage` and 
+it's up to the user to provide the initial API token. A sample shell 
+script for creating, authorizing and activating a VCD Service Account
+can be found in the [repository](https://github.com/vmware/terraform-provider-vcd/blob/service_accounts/scripts/create_service_account.sh)
+
+```hcl
+provider "vcd" {
+  user                       = "none"
+  password                   = "none"
```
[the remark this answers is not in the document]
adezxc: Updated.

vcd/provider.go
```diff
@@ -246,6 +246,13 @@ func Provider() *schema.Provider {
 				Description: "The API token used instead of username/password for VCD API operations. (Requires VCD 10.3.1+)",
 			},
 
+			"service_account_token_file": {
```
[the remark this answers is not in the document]
adezxc: I think we can leave it in case there are users that use multiple fields and only change the `auth_type` as ideally we should have that for all fields but that could possibly be breaking.

scripts/create_service_account.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
+# This script will connect to the vCD using username and password,
+# and show the headers that contain a bearer or authorization token.
```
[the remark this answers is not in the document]
adezxc: Fixed, forgot to change it from the `get_token.sh` script

.changes/v3.9.0/1040-improvements.md
```diff
@@ -0,0 +1 @@
+* Support provider authentication using Service Accounts [GH-1040]
```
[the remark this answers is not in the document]
adezxc: Agree, fixed.

website/docs/index.html.markdown
adezxc: Added, didn't want to add the `~> **NOTE**` so just left a normal line.

vcd/provider.go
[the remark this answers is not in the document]
Didainius: ok, let it be

vcd/auth_test.go
```diff
@@ -304,8 +308,28 @@ func TestAccAuth(t *testing.T) {
...
 		})
-
 	}
+
+	// Testing sending an invalid API token
+	createTestTokenFile(t)
+	testCases = append(testCases, authTestCase{
+		name:        "ServiceAccountTokenFile,AuthType=service_account_token_file",
+		skip:        testConfig.Provider.UseSamlAdfs,
+		skipReason:  "testConfig.Provider.UseSamlAdfs must be false",
```
Didainius, line 318: This needs addressing as it is not Saml

vcd/auth_test.go
```diff
@@ -304,8 +308,28 @@ func TestAccAuth(t *testing.T) {
 			}
 		  `,
 		})
-
 	}
+
+	// Testing sending an invalid API token
```
Didainius, line 313:
```suggestion
	// Testing sending an invalid Service Account token
```

scripts/create_service_account.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+# This script will connect to the vCD using username and password,
```
adambarreiro, line 2: nit
```suggestion
# This script will connect to VCD using username and password,
```

scripts/create_service_account.sh
```diff
@@ -0,0 +1,105 @@
...
+# and go through the steps to create/authorize/activate a service account.
+
```
adambarreiro, line 4: Maybe we could add an early check for `jq`?
Something like
```
! which jq > /dev/null && echo 'Error: No "jq" command found in PATH, please install it' && exit 1
```

adezxc: Nice suggestion, fixed

scripts/create_service_account.sh
adezxc: Fixed.

vcd/auth_test.go
adezxc: Fixed.

adambarreiro (approved):
Nice work!

Didainius (approved):
Token rotation and script to get service account worked in my testing.
Good to approve without an actual test because we do not have Go code to handle service account creation at the moment

vcd/config_test.go
```diff
@@ -61,10 +61,11 @@ func init() {
...
 	Provider struct {
-		User     string `json:"user"`
-		Password string `json:"password"`
-		Token    string `json:"token,omitempty"`
-		ApiToken string `json:"api_token,omitempty"`
+		User                string `json:"user"`
+		Password            string `json:"password"`
+		Token               string `json:"token,omitempty"`
+		ApiToken            string `json:"api_token,omitempty"`
+		ServiceAccountToken string `json:"service_account,omitempty"`
```
dataclouder, line 68:
```suggestion
		ServiceAccountTokenFile string `json:"service_account_token_file,omitempty"`
```
If we implement full support for service accounts, this name will have a different meaning

website/docs/index.html.markdown
```diff
@@ -293,6 +327,16 @@ The following arguments are used to configure the VMware Cloud Director Provider
...
    [the documentation](https://docs.vmware.com/en/VMware-Cloud-Director/10.3/VMware-Cloud-Director-Service-Provider-Admin-Portal-Guide/GUID-A1B3B2FA-7B2C-4EE1-9D1B-188BE703EEDE.html)
 
+* `service_account_token_file` - (Optional; *v3.9+, VCD 10.4+*) This is the file that contains a Service Account API token. The
+   path to the file could be provided as absolute or relative to the working directory. It is used instead of username
+   and password (in combination with `auth_type=service_account_token_file`. The file can also be specified with the 
+   `VCD_SA_TOKEN_FILE` environment variable. There are restrictions to its use, as defined in 
+   [the documentation](https://docs.vmware.com/en/VMware-Cloud-Director/10.4/VMware-Cloud-Director-Service-Provider-Admin-Portal-Guide/GUID-8CD3C8BE-3187-4769-B960-3E3315492C16.html)
+
+* `allow_service_account_token_file` - (Optional; *v3.9+, VCD 10.4+*) When using `auth_type=service_account_token_file`,
+  if set to `false`, will print a warning to the user about the service account token file containing *sensitive information*.
```
dataclouder, line 337:
```suggestion
  if set to `true`, will suppress a warning to the user about the service account token file containing *sensitive information*.
```

scripts/create_service_account.sh
```diff
@@ -0,0 +1,124 @@
...
+
+echo "Please accept the service account access request at $verification_uri"
+echo "The user code is: $user_code"
+
+# Ask the user to write yes to continue
+# If user enters anything other than yes, ask again
+while [ "$answer" != "yes" ]
+do
+    echo "Please enter 'yes' when the service account access was granted: "
```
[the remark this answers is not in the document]
adezxc: I removed this, as I managed to automate the whole process.

scripts/create_service_account.sh
```diff
@@ -0,0 +1,124 @@
...
+# Check if any of the three values is equal to null and exit if so
+if [ "$user_code" = "null" ] || [ "$device_code" = "null" ] || [ "$verification_uri" = "null" ]
+then
+    echo "Error: Could not authorize service account. Check if you have supplied the correct arguments."
+    exit 1
+fi
+
+
+
+echo "Please accept the service account access request at $verification_uri"
```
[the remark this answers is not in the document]
adezxc: Yes, thank you for the suggestion, as I checked this possibility recently in postman, updated.

vcd/auth_test.go
```diff
@@ -304,8 +308,50 @@ func TestAccAuth(t *testing.T) {
...
+		      allow_unverified_ssl       = true
+			}
+		  `,
+		})
 	}
+
+	// Testing sending an invalid Service Account token
+	createTestTokenFile(t)
+	testCases = append(testCases, authTestCase{
+		name:        "ServiceAccountTokenFile,AuthType=service_account_token_file",
```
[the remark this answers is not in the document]
adezxc: updated

vcd/config_test.go
adezxc: updated

website/docs/index.html.markdown
adezxc: updated

dataclouder (approved):

PR #1046, opened 2023-04-15
Create vApp VM Security Tag Tests

adezxc:
WIP

### DIFF
--- a/vcd/resource_vcd_security_tag_test.go
+++ b/vcd/resource_vcd_security_tag_test.go
@@ -173,6 +173,169 @@ resource "vcd_security_tag" "{{.SecurityTag2}}" {
 }
 `
 
+// TestAccVcdVappVmWithSecurityTags tests security_tags field of a vcd_vapp_vm
+// There is still a vcd_security_tag resource which is responsible for managing
+// the tag in VCD as security_tags field only creates and doesn't remove the
+// tag itself if it is removed from a VM to not break other VMs.
+func TestAccVcdVappVmWithSecurityTags(t *testing.T) {
+	tag1 := strings.ToLower(t.Name() + "-tag1") // security tags are always lowercase in serverside
+	tag2 := strings.ToLower(t.Name() + "-tag2")
+	vAppName := t.Name() + "-vapp"
+	vmName := t.Name() + "-vm"
+
+	var params = StringMap{
+		"Org":          testConfig.VCD.Org,
+		"Vdc":          testConfig.Nsxt.Vdc,
+		"vappName":     vAppName,
+		"vmName":       vmName,
+		"computerName": t.Name() + "-vm",
+		"securityTag1": tag1,
+		"securityTag2": tag2,
+		"FuncName":     t.Name(),
+	}
+	testParamsNotEmpty(t, params)
+
+	configText := templateFill(testAccVappVmWithSecurityTags, params)
+	params["FuncName"] = t.Name() + "-mixsecuritytags"
+	configText1 := templateFill(testAccVappVmWithMixedSecurityTags, params)
+	params["FuncName"] = t.Name() + "-onesecuritytag"
+	configText2 := templateFill(testAccVappVmWithOneSecurityTag, params)
+
+	debugPrintf("#[DEBUG] CONFIGURATION: %s\n", configText)
+	if vcdShortTest {
+		t.Skip(acceptanceTestsSkipped)
+		return
+	}
+
+	resourceName := "vcd_vapp_vm." + vmName
+	resource.Test(t, resource.TestCase{
+		ProviderFactories: testAccProviders,
+		CheckDestroy: resource.ComposeTestCheckFunc(
+			testAccCheckSecurityTagDestroy(tag1),
+			testAccCheckSecurityTagDestroy(tag2),
+			testAccCheckVcdVAppVmDestroy(vAppName),
+		),
+		Steps: []resource.TestStep{
+			{
+				Config: configText,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckSecurityTagCreated(tag1),
+					testAccCheckSecurityTagOnVMCreated(tag1, vAppName, vmName),
+					resource.TestCheckTypeSetElemAttr(resourceName, "security_tags.*", tag1),
+					testAccCheckSecurityTagCreated(tag2),
+					testAccCheckSecurityTagOnVMCreated(tag2, vAppName, vmName),
+					resource.TestCheckTypeSetElemAttr(resourceName, "security_tags.*", tag2),
+				),
+			},
+			{
+				Config: configText1,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckSecurityTagCreated(tag1),
+					testAccCheckSecurityTagOnVMCreated(tag1, vAppName, vmName),
+					resource.TestCheckTypeSetElemAttr(resourceName, "security_tags.*", tag1),
+					testAccCheckSecurityTagDestroy(tag2),
+				),
+			},
+			{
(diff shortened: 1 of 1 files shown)


vcd/resource_vcd_security_tag_test.go
```diff
@@ -173,6 +173,170 @@ resource "vcd_security_tag" "{{.SecurityTag2}}" {
...
+}
+
+resource "vcd_security_tag" "{{.securityTag1}}" {
+	name   = "{{.securityTag1}}"
+	vm_ids = [vcd_vapp_vm.{{.vmName}}.id]
+}
+
+resource "vcd_security_tag" "{{.securityTag2}}" {
+	name   = "{{.securityTag2}}"
+	vm_ids = [vcd_vapp_vm.{{.vmName}}.id]
```
Didainius, line 312: indentation looks to be off

vcd/resource_vcd_security_tag_test.go
```diff
@@ -173,6 +173,170 @@ resource "vcd_security_tag" "{{.SecurityTag2}}" {
...
+}
+`
+
+const testAccVappVmWithOneSecurityTag = `
```
Didainius, line 316: I don't see a security tag at all in the definition

vcd/resource_vcd_security_tag_test.go
```diff
@@ -173,6 +173,170 @@ resource "vcd_security_tag" "{{.SecurityTag2}}" {
...
+  hardware_version = "vmx-14"
+  depends_on       = [vcd_vapp.{{.vappName}}]
+
+  security_tags = ["{{.securityTag1}}"]
+}
+
+resource "vcd_security_tag" "{{.securityTag1}}" {
+	name   = "{{.securityTag1}}"
+	vm_ids = [vcd_vapp_vm.{{.vmName}}.id]
+}
```
Didainius, line 277: indentation is off

Didainius (commented):
A few initial comments. 

* The `vcd_security_tag` is not necessary to create tag directly on `vcd_vapp_vm` or `vcd_vm`. We need to test at least such cases:
  * Create a security tag only by `security_tags` field on `vcd_vapp_vm` and `vcd_vm` (to check if it worked - we could either plug in a govcd function, or add data sources of these resources and check their values as we need to validate data sources anyway)
  * Improve `TestAccVcdSecurityTag` by checking if `vcd_security_tag` field gets populated on a `vcd_vapp_vm` resource that is being used in the test
  * Check a scenario when you create a security tag on in `vcd_vapp_vm` and what happens when you delete that VM (is the tag removed completely or left as orphan without any VMs)
  * Check that once you create a tag using `security_tags` on a VM and then remove that field from vm definition - there is still a tag on that VM (this will require using govcd function, probably `testAccCheckSecurityTagOnVMCreated` can be reused)

* We also need to add `vcd.ResourceSchema-vcd_vapp_vm.tf` and `vcd.ResourceSchema-vcd_vm.tf` to `skip-upgrade-tests.txt` because their schema has changed (`security_tags` field added)

Didainius (approved):
LGTM, thanks

adambarreiro (approved):

website/docs/r/vapp_vm.html.markdown
```diff
@@ -503,7 +503,13 @@ example for usage details.
   using `vcd_org_vdc.vm_placement_policy_ids` (and optionally `vcd_org_vdc.default_compute_policy_id` to make it default).
   In this case, if the placement policy is not set, it will pick the VDC default on creation. It must be set explicitly
   if one wants to update it to another policy (the VM requires at least one Compute Policy), and needs to be set to `""` to be removed.
-* `security_tags` - (Optional; *v3.9+*) Set of security tags to be managed by the `vcd_vapp_vm` resource. Removing `security_tags` will not remove the security tags themselves. They remain unchanged and just stop being managed by `security_tags` in the `vcd_vapp_vm` resource. This is to keep backwards compatibility with existing security tags that were created by the `vcd_security_tags` resource. Setting `security_tags` to `[]` will remove the security tags.
+* `security_tags` - (Optional; *v3.9+*) Set of security tags to be managed by the `vcd_vapp_vm` resource.
+  To remove `security_tags` you must set `security_tags = []` and do not remove the attribute. Removing the attribute will cause that the tags remain unchanged and just stop being managed by this resource.
```
dataclouder, line 507:
```suggestion
  To remove `security_tags` you must set `security_tags = []` and do not remove the attribute. Removing the attribute will cause the tags to remain unchanged and just stop being managed by this resource.
```

website/docs/r/vapp_vm.html.markdown
```diff
@@ -503,7 +503,13 @@ example for usage details.
...
+  This is to keep backwards compatibility with existing security tags that were created by the `vcd_security_tags` resource.
```
dataclouder, line 508:
```suggestion
  This is to keep compatibility with existing security tags that were created by the `vcd_security_tags` resource.
```
It is not **backwards** compatibility. It's a separate functionality.
Alternatively:
```suggestion
  This is to be consistent with existing security tags that were created by the `vcd_security_tags` resource.
```

website/docs/r/vapp_vm.html.markdown
```diff
@@ -503,7 +503,13 @@ example for usage details.
...
+
+~> **Note:** Either `security_tags` attribute or [`vcd_security_tag`](/providers/vmware/vcd/latest/docs/resources/security_tag) resource
```
dataclouder, line 510:
```suggestion
~> **Note:** Only one of `security_tags` attribute or [`vcd_security_tag`](/providers/vmware/vcd/latest/docs/resources/security_tag) resource
```

website/docs/r/vapp_vm.html.markdown
```diff
@@ -503,7 +503,13 @@ example for usage details.
...
+  should be used as they would cause a behavioral conflict with each other.
```
dataclouder, line 511:
```suggestion
  should be used. Using both would cause a behavioral conflict.
```

website/docs/r/security_tag.html.markdown
```diff
@@ -14,6 +14,9 @@ used to assign security tag to VMs.
 
 Supported in provider *v3.7+* and requires VCD 10.3.0+
 
+~> **Note:** Either `vcd_security_tag` resource or [`security_tags` attribute from `vcd_vapp_vm`](/providers/vmware/vcd/latest/docs/resources/vapp_vm)
+should be used as they would cause a behavioral conflict with each other.
```
dataclouder, line 18:
```suggestion
~> **Note:** Only one of `vcd_security_tag` resource or [`security_tags` attribute from `vcd_vapp_vm`](/providers/vmware/vcd/latest/docs/resources/vapp_vm)
should be used. Using both would cause a behavioral conflict.
```

website/docs/r/vapp_vm.html.markdown
adezxc: Updated to the alternative

website/docs/r/vapp_vm.html.markdown
adezxc: Updated

dataclouder (approved):

website/docs/r/vapp_vm.html.markdown
adezxc: Updated

website/docs/r/security_tag.html.markdown
adezxc: Updated

website/docs/r/vapp_vm.html.markdown
adezxc: Updated

PR #1048, opened 2023-04-17
Flaky test cleanup

Didainius:
* `TestAccVcdNetworkImportedV2InheritedVdc` - fixed by avoiding comparison of total field count
* `TestAccDataSourceNotFound/vcd_nsxt_edgegateway_qos_profile` - Fixed by skiping on VCD < 10.3.2
* `TestAccVcdOrgFull` should not create disabled Org in some 10.4 VCD versions due to they have a bug. Additionally, create method of `vcd_org` resource has a patch to improve the error message of the bug :
```
│ Error: [org creation] error creating Org org1:
│ this version of VCD (10.4.2.21595985) has a bug that prevents creating a disabled Org
│ If you need to disable it, please create an enabled Org and then disable it using update
│
│   with vcd_org.org1,
│   on config.tf line 27, in resource "vcd_org" "org1":
│   27: resource "vcd_org" "org1" {
```

The original error is as below, but it doesn't give a lot of sense to the user therefore it is not shown (will still appear in go-vcloud-director logs if someone is troubleshooting):
```
resource_vcd_org_test.go:230: Step 1/3 error: Error running apply: exit status 1

        Error: [org creation] error creating Org org2: error instantiating a new Org: API Error: 400: [ a33015b2-92e3-47f0-9500-a95280dfa493 ] Unexpected error.
        unexpected end of subtree [DELETE FROM com.vmware.vcloud.common.model.oauth.oidc.OidcAuthorizationModel o WHERE (o.orgMemberId in ())]
         - unexpected end of subtree [DELETE FROM com.vmware.vcloud.common.model.oauth.oidc.OidcAuthorizationModel o WHERE (o.orgMemberId in ())]

          with vcd_org.org2,
          on terraform_plugin_test.tf line 2, in resource "vcd_org" "org2":
           2: resource "vcd_org" "org2" {
```
* `TestAccVcdVmInternalDisk`  - `import` step returned `testing_new_import_state.go:122: unexpected output type: <nil>`
After a debugging session, I've identified that the import mechanism fails on one of the `output` fields. These `output` fields don't provide any value as their values can be checked the regular `resource.TestCheckResourceAttr` functions.

image

### DIFF
--- a/vcd/resource_vcd_org.go
+++ b/vcd/resource_vcd_org.go
@@ -197,8 +197,12 @@ func resourceOrgCreate(ctx context.Context, d *schema.ResourceData, m interface{
 		log.Printf("[DEBUG] Error creating Org: %s", err)
 		// Some 10.4 VCD versions have a bug that fail when creating a disabled Org
 		if !isEnabled && strings.Contains(err.Error(), "com.vmware.vcloud.common.model.oauth.oidc.OidcAuthorizationModel") {
-			err = fmt.Errorf("%s\n\nThis version of VCD has a bug which prevents creating a disabled Org\n"+
-				"If you need to disable, please create an enabled Org and then disable it using update", err)
+			// getting VCD version for error message but not explicitly failing if it cannot be
+			// retrieved to avoid unnecessary failure path
+			// The error itself is SQL message therefore we override it
+			vcdVersion, _, _ := vcdClient.Client.GetVcdVersion()
+			err = fmt.Errorf("\nthis version of VCD (%s) has a bug that prevents creating a disabled Org\n"+
+				"If you need to disable, please create an enabled Org and then disable it using update", vcdVersion)
 		}
 
 		return diag.Errorf("[org creation] error creating Org %s: %s", orgName, err)
--- a/vcd/resource_vcd_org.go
+++ b/vcd/resource_vcd_org.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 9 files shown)


vcd/resource_vcd_org.go
```diff
@@ -194,6 +195,12 @@ func resourceOrgCreate(ctx context.Context, d *schema.ResourceData, m interface{
 
 	if err != nil {
 		log.Printf("[DEBUG] Error creating Org: %s", err)
+		// Some 10.4 VCD versions have a bug that fail when creating a disabled Org
+		if !isEnabled && strings.Contains(err.Error(), "com.vmware.vcloud.common.model.oauth.oidc.OidcAuthorizationModel") {
+			err = fmt.Errorf("%s\n\nThis version of VCD has a bug which prevents creating a disabled Org\n"+
```
dataclouder, line 200: Should we display the detected VCD version here?

dataclouder (approved):

vcd/resource_vcd_org.go
adambarreiro, line 200: I wonder if the error should come after the clarification? As it is a bit unintelligible Java stacktrace.
cc: @dataclouder @adezxc

vcd/resource_vcd_org_test.go
```diff
@@ -64,6 +64,15 @@ func TestAccVcdOrgFull(t *testing.T) {
 	preTestChecks(t)
 
 	skipIfNotSysAdmin(t)
+
+	createEnabledOrg := false
+	vcdClient := createTemporaryVCDConnection(false)
+	if vcdClient.Client.APIVCDMaxVersionIs("> 37.1") {
+		// TODO revisit once bug is fixed in VCD
```
adambarreiro, line 71: Good reminder!

adambarreiro (approved):
LGTM

adezxc (approved):

vcd/resource_vcd_org.go
Didainius: Final decision is to mask the error (which is too verbose for a regular user) and show our own:
```
│ Error: [org creation] error creating Org org1:
│ this version of VCD (10.4.2.21595985) has a bug that prevents creating a disabled Org
│ If you need to disable it, please create an enabled Org and then disable it using update
│
│   with vcd_org.org1,
│   on config.tf line 27, in resource "vcd_org" "org1":
│   27: resource "vcd_org" "org1" {
```

Didainius: Implemented

PR #901, opened 2022-08-16
Refactor VM "Create" code

Didainius:
The goal is to improve VM Create code and:
* Remove unnecessary power on/power off cycles during VM spinup which should improve guest customization experience
* Simplify legacy code which is hard to maintain (it should additionally improve performance of operations due to reduced API call number)

Architecture fixes:
* **Create** used to call **Update** for a VM creation task. Create is now decoupled from Update.
* Multiple power on and power off operations were executed during create operation. **Power on** is now executed as the **last step** of creation.
* All VM creation structures (`types.XXXX`) are now populated inside `Create` code as historically some of them used a govcd function, while others populated it directly. The problem with such an approach is that one needs to create new SDK functions with every new feature in VM. This is the list of `types` that `Create` code now uses.
   * `types.InstantiateVmTemplateParams` (Standalone VM from template)
   * `types.ReComposeVAppParams` (vApp VM from template)
   * `types.RecomposeVAppParamsForEmptyVm` (Empty vApp VM)
   * `types.CreateVmParams` (Empty Standalone VM)

* `Update` code is not affected by this PR.

What wasn't done:
* The functions, responsible for particular structure, like guest customization, networks, etc - were not changed - some of them were just moved to a separate function instead of being inline.

The new structure of Create code allows users to inject code where it is needed:

Flow diagram for `create` functions of `vcd_vapp_vm` and `vcd_vm`.
* Purple shows the main flow of code for `Create`
* Green blocks signify `govcd.types` definition for each of VM types that are being created.
* Yellow blocks signify where additional code can be hooked up for particular scope

```mermaid
flowchart TD
    S[<b>VM Create</b>]
    S ==> AAAA
    AAAA{Standanlone or vApp VM?}
    AAAA --> |resource <b>vcd_vm</b>| AA{vApp VM or Standalone}
    AAAA --> |resource <b>vcd_vapp_vm</b>| A{vApp VM or Standalone}

    AA[func <b>resourceVcdStandaloneVmCreate</b>] --> B[func <b>genericResourceVmCreate</b>]
    A[func <b>resourceVcdVAppVmCreate</b>] -->|lock parent vApp\n defer unlock| B[func <b>genericResourceVmCreate</b>]

    BB{From <b>template</b> or <b>empty</b> VM}

    B --> BB

    BB -->|isVmFromTemplate| D[func <b>createVmFromTemplate</b>]
    BB -->|isEmptyVm| E[func <b>createVmEmpty</b>]

    DDD([vApp VM from template])
[...]

PR #1057, opened 2023-05-17
Add new argument to Provider configuration to ignore metadata entries

### LINKED ISSUE #959: ignore specific metadata_entries
### Community Note

### Description

I am migrating our use of the now deprecated `metadata` to the new `metadata_entry` block. I have the need to ignore specific `metadata_entries`, because they are managed by other systems rather than terraform. There is for example the `veeam-plugin` for VCD, which itself creates a `metadata_entry` of `veeam-endpoint` -> `https://<URL>` in the `vcd_org` object. The same is the case for `object-storage-extensions`. We also have our own processes which write their metadata to the `vcd_org` or `vcd_org_vdc` objects.

I was able to work around this until now, because of the way the old `metadata`-entries were working. The following workaround was working:

```terraform
lifecycle {
    ignore_changes = [
      metadata["ose-storage-id"], metadata["veeam-endpoint"]
    ]
  }
```

Using the new `metadata_entry`, there is no such way of ignoring specific entries. All the ways I have tried have failed. This is because I am now required to evaluate the ignore_changes values in a more dynamic way, which currently isn't possible. 

All resources supporting `metadata_entry`. 

### Example behaviour

An example behaviour after migrating to `metadata_entry` and performing a plan:

```hcl
# module.*.vcd_org.org will be updated in-place
  ~ resource "vcd_org" "org" {
        id                              = "urn:vcloud:org:65ce7ee1-a2e0-49d0-b21f-2d86121d6f9a"
        - metadata_entry {
          - is_system   = true -> null
          - key         = "veeam-endpoint" -> null
          - type        = "MetadataStringValue" -> null
          - user_access = "READONLY" -> null
          - value       = "\"https://backup.<URL>:9443\"" -> null
        }
```

### References

There is a rather old request to improve this on the terraform core, but it looks like this will not be happening: https://github.com/hashicorp/terraform/issues/3116 

The AWS provider has implemented a workaround for this, see https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/resource-tagging#ignoring-changes-in-all-resources . This would also allow to ignore specific prefixes, which would be even more flexible than the old way I was doing it.

I will not be able to migrate to `metadata_entry`, until this is resolved. I would brick our tenant's ability to use the VCD extensions for veeam, object storage and so on.

adambarreiro:
Closes #959
Requires https://github.com/vmware/go-vcloud-director/pull/581/files

This PR adds a new argument to the Terraform VCD Provider configuration: `ignore_metadata_changes`, which accepts the following sub-arguments:

* `resource_type`: Optional, specifies the resource type which metadata needs to be ignored. If set, the resource type must be one of:
  *"vcd_catalog"*, *"vcd_catalog_item"*, *"vcd_catalog_media"*, *"vcd_catalog_vapp_template"*, *"vcd_independent_disk"*, *"vcd_network_direct"*,
  *"vcd_network_isolated"*, *"vcd_network_isolated_v2"*, *"vcd_network_routed"*, *"vcd_network_routed_v2"*, *"vcd_org"*, *"vcd_org_vdc"*, *"vcd_provider_vdc"*,
  *"vcd_storage_profile"*, *"vcd_vapp"*, *"vcd_vapp_vm"* or *"vcd_vm"*, which are the resources compatible with `metadata_entry`.
  Any other resource type will not have effect.
* `resource_name`: Optional, specifies the name of the object which metadata needs to be ignored. All object types are supported, except for `vdcStorageProfile` which **cannot be filtered by name**.
* `key_regex`: Optional, it is a regular expression that can filter out metadata keys that match. 
* `value_regex`: Optional, it is a regular expression that can filter out metadata keys that match.
* `conflict_action`: Optional, defines what to do if a conflict exists between a `metadata_entry` that is managed
  by Terraform, and it matches the criteria defined in the `ignore_metadata_changes` block, as the metadata will be stored in state but nothing will be done in VCD. If the value is `error`, when this happens, the Plan will fail. When the value is `warn`, it will just give a warning but the Plan will continue, and with the `none` value nothing will be shown. Defaults to `error`.

Either `key_regex` or `value_regex` are required on every `ignore_metadata` block.

## Example

```hcl
provider "vcd" {
  password             = "SuperSecure123"
  url                  = "https://my-awesome-vcd.com"
  org                  = "myOrg"
  allow_unverified_ssl = true
  logging              = true

  # Filters all metadata with key "Environment" or "environment" in all VCD objects with any name.
  ignore_metadata_changes {
    key_regex   = "^[Ee]nvironment$"
  }

  # Filters all metadata with key "NiceMetadataKey" in all VCD objects named "SpecificName".
  ignore_metadata_changes {
    resource_name = "SpecificName"
    key_regex   = "^NiceMetadataKey$"
  }

  # Filters all metadata with values "Yes" in the Organization named "Tatooine".
  ignore_metadata_changes {
    resource_type = "org"
    resource_name = "Tatooine"
    value_regex = "^Yes$"
  }
}
```

Acceptance tests passed on VCD versions: TBD


vcd/metadata.go
```diff
@@ -4,9 +4,85 @@ import (
...
+				"object_name": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: "Ignores metadata from the specific object in VCD named like this argument",
+				},
+				"key_regex": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: "Regular expression of the metadata entry keys to ignore. Either `key_regex` or `value_regex` is required",
+					// Note: AtLeastOneOf can't be used here
```
Didainius, line 35: Could  `ExactlyOneOf` work here if this is not Terraform limitation that it doesn't work? (I see this is being done https://github.com/vmware/terraform-provider-vcd/pull/1057/files#diff-194d724e0d1ac9bf551517de0ed3d86746f2a11776b9e404b13054e31c156fdaR74)

Didainius (commented):
I have done a manual  try for the code. It looks to be working (attempted to manipulate metadata with and then add multiple `ignore_metadata` blocks and seeing how the resource and data source behave.

I have left some thoughts and clarifications inline

vcd/metadata_common_test.go
```diff
@@ -259,6 +263,219 @@ func testMetadataEntryCRUD(t *testing.T, resourceTemplate, resourceAddress, data
...
+		return
+	}
+
+	// We will cache the ID of the created resource after Step 1, so it can be used afterward.
+	cachedId := testCachedFieldValue{}
+
+	testFunc := func(ignoredMetadata []map[string]string, expectedMetadataInVcd int) {
+		var object metadataCompatible
+		resource.Test(t, resource.TestCase{
+			ProviderFactories: map[string]func() (*schema.Provider, error){
```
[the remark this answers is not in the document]
dataclouder: perhaps we can isolate this set of tests behind a tag that is never used unless specifically invoked. Need to experiment

website/docs/index.html.markdown
```diff
@@ -373,6 +373,76 @@ The following arguments are used to configure the VMware Cloud Director Provider
...
+-> To avoid side effects, you should only use this feature to deal with metadata not managed by this provider.
+It is encouraged to avoid ignoring metadata that was created by a Terraform resource, as you could get unwanted updates or inconsistencies
+in the state.
+
+The available sub-attributes for `ignore_metadata` are:
+
+* `object_type` - (Optional) Specifies the type of the object which metadata needs to be ignored. The object types must be declared
+  as defined in the metadata endpoints of the [VCD API documentation](https://developer.vmware.com/apis/1601/vmware-cloud-director). For example,
+  `catalog`, `media`, `disk`, `org`, `catalogItem`, `vAppTemplate`, `vdc`, `network`, `providervdc`, `vApp` (note this one is used
+  for both vApps and VMs), etc.
```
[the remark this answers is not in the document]
lvirbalas: Thoughts:
1. The resource list needs to be exact and specific. It shouldn't be "for example". The reason is that we need to tell users which resources this is actual for.
2. I agree that it would be easier if these types mapped to Terraform resources. It would also be more managable and readable in HCL.

website/docs/index.html.markdown
```diff
@@ -373,6 +373,76 @@ The following arguments are used to configure the VMware Cloud Director Provider
 * `import_separator` - (Optional; *v2.5+*) The string to be used as separator with `terraform import`. By default
   it is a dot (`.`).
 
+* `ignore_metadata` - (Optional; *v3.10+*) Use one or more of these blocks to ignore specific metadata entries.
```
lvirbalas, line 376: I feel we're missing something like "... from being managed by this Terraform provider".

When you first see this feature without context the above is not instantly clear.

Same ask for the variable's description in the schema.

lvirbalas (changes requested):
I'd like to ask to extend docs a bit more to make it instantly clear for the unaware user that it's ignored "by Terraform provider management". I feel fresh user may miss the point at the first glance.

website/docs/index.html.markdown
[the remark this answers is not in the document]
adezxc: I agree with both suggestions, if possible we should try and use resource names instead of API objects, also specificity is very important, or every resource should have that information clearly stated in the documentation.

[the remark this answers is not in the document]
adambarreiro: I had the same thought, as it simplifies the user experience, but maintaining this map ("resource name"->"metadata endpoint") is what pulled me back.
I'll do the changes and we can check.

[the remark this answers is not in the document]
lvirbalas: Maybe we can utilize (or extend) resource list that @dataclouder had implemented earlier for holding this mapping? If the resource list already has some maintainable list/map, it's better to have everything in one place.

vcd/metadata.go
adambarreiro: Could be, but in this case it would cause that you either populate `key_regex` or `value_regex`, if you populate both it would fail, right?
What I wanted is that one should at least use one of them.

website/docs/index.html.markdown
```diff
@@ -373,6 +373,76 @@ The following arguments are used to configure the VMware Cloud Director Provider
...
+* `object_name`- (Optional) Specifies the name of the object which metadata needs to be ignored. All object types are supported, except for
+  `vdcStorageProfile` which **cannot be filtered by name**.
+* `key_regex`- (Optional) A regular expression that can filter out metadata keys that match. Either `key_regex` or `value_regex` are required on each block. 
+* `value_regex`- (Optional) A regular expression that can filter out metadata values that match. Either `key_regex` or `value_regex` are required on each block.
+
+Note that these attributes **are evaluated as a logical `and`**. This means that the snippet below would ignore all metadata entries
+that belong to the specific Organization named "client1" **and** which keys match the regular expression `[Ee]nvironment`:
+
+```hcl
+provider "vcd" {
```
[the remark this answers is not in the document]
adambarreiro: Good point, as it was done precisely for that :)
Added a dummy `vcd_org` with a metadata_entry block to illustrate that.

vcd/metadata_common_test.go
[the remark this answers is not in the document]
adambarreiro: I have added the `skip-binary-test` to all the generated HCL snippets, so it should be good for now.

.changes/v3.10.0/1057-improvements.md
```diff
@@ -0,0 +1,2 @@
+* Add `ignore_metadata` argument to the Provider configuration to be able to specify metadata entries that should **not**
+  be managed by Terraform [GH-1057]
```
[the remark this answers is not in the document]
adambarreiro: Changed.

website/docs/index.html.markdown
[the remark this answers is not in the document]
adambarreiro: Done, I created a small function that handles the mapping.

website/docs/index.html.markdown
adambarreiro: Done

website/docs/index.html.markdown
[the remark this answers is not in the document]
lvirbalas: I like the way it became easy to use with the `resource_type` field and all the familiar resource names!

@dataclouder are you OK with this being a separate metadata-centric function or should this be integrated into the `vcd_resource_list`/`vcd_resource_schema` framework?

vcd/metadata.go
```diff
@@ -92,18 +214,9 @@ func metadataEntryResourceSchema(objectNameInDescription string) *schema.Schema
 	}
 }
 
-// getMetadataEntrySchema returns a schema for the "metadata_entry" attribute, that can be used to
-// build data sources (isDatasource=true) or resources (isDatasource=false). The description of the
-// attribute will refer to the input resource name.
-func getMetadataEntrySchema(resourceNameInDescription string, isDatasource bool) *schema.Schema {
```
lvirbalas, line 98: Just curious, what prompted you to remove this function and use the two specific ones below across the board? :)

adambarreiro: When I implemented that I thought it was a good idea, but after revisiting this function to implement this feature I noticed:
- The bool parameter is hard to understand unless you read and understand the function docs/signature.
- The bool parameter is a bit arbitrary, it could have been `isResource`, which also makes it confusing.
- The function just had an `if` based on that condition that chose either the resource schema or datasource schema.

So it seemed more right to me to have just more simple and meaningful functions `metadataEntryDatasourceSchema` and `metadataEntryResourceSchema`. This way you explicitly know what you are trying to get, without an opaque bool flag.

adambarreiro: As a side note, I also thought of using the same pattern we use across the codebase, something like `getMetadataEntrySchema("foo", "resource") and getMetadataEntrySchema("foo", "datasource")`, but again the inner body was so dumb that it didn't have much sense to have a 2x1 function.

vcd/metadata.go
Didainius: You're right, I had wrong assumption

vcd/metadata.go
```diff
@@ -4,9 +4,131 @@ import (
...
+	return &schema.Schema{
+		Type:        schema.TypeSet,
+		Optional:    true,
+		Description: "Defines a set of `metadata_entry` that need to be ignored by this provider. All filters on this attribute are computed with a logical AND",
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"resource_type": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: "Ignores metadata from the specific resource type",
```
adezxc, line 24: Would it be beneficial to add a ValidateFunc like StringInSlice here? Or is it not needed because of terraformMetadataRelation?

adezxc (approved):
Looks good to me!

website/docs/index.html.markdown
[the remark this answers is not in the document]
Didainius: I have attempted this new approach manually - I think this is much more convenient for users - thanks!

One observation to share - I think we should validate the `resource_type` field in `provider`. I can make a typo or put `sfSDFSDFSD` and then `resource_type` would be ignored (filtering applied to all resources/data source)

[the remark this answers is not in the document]
dataclouder: > @dataclouder are you OK with this being a separate metadata-centric function or should this be integrated into the vcd_resource_list/vcd_resource_schema framework?
I don't see how this could be integrated into `vcd_resource_list`. They have different purposes

vcd/metadata.go
```diff
@@ -4,9 +4,131 @@ import (
...
+		"vcd_network_isolated_v2":   "network",
+		"vcd_network_routed":        "network",
+		"vcd_network_routed_v2":     "network",
+		"vcd_org":                   "org",
+		"vcd_org_vdc":               "vdc",
+		"vcd_provider_vdc":          "providervdc",
+		"vcd_storage_profile":       "vdcStorageProfile",
+		"vcd_vapp":                  "vApp",
+		"vcd_vapp_vm":               "vApp",
+		"vcd_vm":                    "vApp",
```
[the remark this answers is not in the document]
adambarreiro: It seems they decided for `vApp` for both: https://vdc-repo.vmware.com/vmwb-repository/dcr-public/fd3d847b-e702-429f-80fa-e4bbbee25b06/d3bf90f6-56ae-49c5-8005-e068ff561aea/doc/doc/operations/GET-VAppMetadata.html

website/docs/index.html.markdown
```diff
@@ -373,6 +373,93 @@ The following arguments are used to configure the VMware Cloud Director Provider
...
+  # ...
+  ignore_metadata {
+    object_type = "org"
```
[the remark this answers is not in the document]
adambarreiro: Yes, thanks for spotting, it seems I forgot to replace it in the docs.

vcd/metadata.go
adambarreiro: I think it's a good idea, I added the validation (and removed the part in the docs that said that a wrong type would have no effect, as now it will fail)

website/docs/index.html.markdown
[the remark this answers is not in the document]
adambarreiro: @Didainius Yes, that's a good idea. I think @adezxc also had [a similar one](https://github.com/vmware/terraform-provider-vcd/pull/1057#discussion_r1253302365), so you both have a point.
I have added a ValidateFunc to `resource_type`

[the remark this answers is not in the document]
Didainius: I am closing this issue as I think we have it sorted now.

Didainius (approved):

vcd/resource_vcd_org.go
```diff
@@ -417,11 +417,11 @@ func resourceOrgUpdate(_ context.Context, d *schema.ResourceData, m interface{})
 	}
 
 	log.Printf("[TRACE] Org %s updated", orgName)
-	return nil
+	return resourceOrgRead(ctx, d, m)
```
adambarreiro, line 420: Note to reviewers: I noticed a Read was missing after Update. I assumed this is wrong, but let me know otherwise.

vcd/resource_vcd_vapp.go
```diff
@@ -152,7 +152,7 @@ func resourceVcdVAppCreate(ctx context.Context, d *schema.ResourceData, meta int
 
 	d.SetId(vapp.VApp.ID)
 
-	return resourceVcdVAppUpdate(ctx, d, meta)
+	return resourceVcdVAppRead(ctx, d, meta)
```
adambarreiro, line 155: Note to reviewers, I noticed this one was calling Update after Create. I think it should be a Read like any other resource, let me know otherwise.

adambarreiro (commented):
As this PR made me go a bit back and forth with refactoring method signatures, I noticed some things that are a bit odd. Let me know your thoughts @adezxc @dataclouder @Didainius

lvirbalas (approved):
I have reviewed the 26 commits done after my last review and it all feels right and further improved!

vcd/metadata.go
```diff
@@ -2,18 +2,163 @@ package vcd
...
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/vmware/go-vcloud-director/v2/govcd"
 	"github.com/vmware/go-vcloud-director/v2/types/v56"
+	"github.com/vmware/go-vcloud-director/v2/util"
+	"regexp"
+	"strings"
 )
 
+var (
+	// IgnoreMetadataChangesConflictResolution can hold values "error", "warn" or "none", and tells Terraform whether it should
```
dataclouder, line 17: I'd rather use `ConflictActions` instead of `ConflictResolution`.

vcd/metadata.go
```diff
@@ -2,18 +2,163 @@ package vcd
...
+		Type:        schema.TypeSet,
+		Optional:    true,
+		Description: "Defines a set of `metadata_entry` that need to be ignored by this provider. All filters on this attribute are computed with a logical AND",
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"resource_type": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: "Ignores metadata from the specific resource type",
+					ValidateDiagFunc: func(i interface{}, path cty.Path) diag.Diagnostics {
```
dataclouder, line 36: Is it possible to extract this function and put it in validate_funcs.go with some comments?

vcd/metadata.go
```diff
@@ -2,18 +2,163 @@ package vcd
...
+								break
+							}
+						}
+						if !found {
+							return diag.Errorf("can't ignore metadata of resource type '%s'", i)
+						}
+						return nil
+					},
+				},
+				"object_name": {
```
dataclouder, line 54: We have changed an early named `object_type` to `resource_type`. 
Perhaps the configuration would feel more consistent if we call this one `resource_name`

vcd/metadata.go
```diff
@@ -2,18 +2,163 @@ package vcd
...
+					Optional:    true,
+					Description: "Regular expression of the metadata entry values to ignore. Either `key_regex` or `value_regex` is required",
+					// Note: This one should have AtLeastOneOf, but it can't be used inside Sets
+				},
+				"conflict_action": {
+					Type:         schema.TypeString,
+					Optional:     true,
+					Default:      "error",
+					ValidateFunc: validation.StringInSlice([]string{"error", "warn", "none"}, false),
+					Description:  "One of 'error', 'warn' or 'none'. Configures whether a conflict between this ignored metadata block and the metadata entries set in Terraform should error, warn or do nothing. Defaults to 'error'",
```
dataclouder, line 76:
```suggestion
					Description:  "One of 'error', 'warn' or 'none'. Configures whether a conflict between this ignored metadata block and the metadata entries set in Terraform should fail, warn or do nothing. Defaults to 'error'",
```

vcd/metadata.go
```diff
@@ -2,18 +2,163 @@ package vcd
...
+				},
+			},
+		},
+	}
+}
+
+// IgnoredMetadata extends the SDK IgnoredMetadata type to be able to add also
+// the conflict resolution behavior defined as an attribute in the schema.
```
dataclouder, line 84:
```suggestion
// the conflict behavior defined as an attribute in the schema.
```

vcd/resource_vcd_vapp.go
dataclouder: The call to Update is necessary, as otherwise it would not set the vApp lease.
We are trying not to do it in new resources, but this is how it was done at the time

website/docs/index.html.markdown
```diff
@@ -373,6 +373,102 @@ The following arguments are used to configure the VMware Cloud Director Provider
...
+  See ["Ignore Metadata Changes"](#ignore-metadata-changes) for more details.
+
+## Ignore metadata changes
+
+One or more `ignore_metadata_changes` blocks can be optionally set in the provider configuration, which will allow to ignore specific `metadata_entry`
+items during `plan`, `updates-in-place` and `destroy` (**not** `create`). This is useful, for example, to avoid removing metadata entries that were created
+by an external actor or after they were created by Terraform.
+
+~> Note that this feature is only considered when using the `metadata_entry` argument in the resources and data sources that support
+it. In other words, to ignore metadata when you are using the deprecated `metadata` argument, please use the native Terraform `lifecycle` block.
```
dataclouder, line 387:
```suggestion
it. In other words, to ignore metadata when you are using the deprecated `metadata` argument, please use the native Terraform `lifecycle.ignore_changes` block.
```

website/docs/index.html.markdown
```diff
@@ -373,6 +373,102 @@ The following arguments are used to configure the VMware Cloud Director Provider
...
+results, as the metadata will be stored in state but nothing will be done in VCD. You can control what to do in this situation with
+`conflict_action`, which can be `error`, `warn` or `none`.
+
+The available sub-attributes for `ignore_metadata_changes` are:
+
+* `resource_type` - (Optional) Specifies the resource type which metadata needs to be ignored. If set, the resource type must be one of:
+  *"vcd_catalog"*, *"vcd_catalog_item"*, *"vcd_catalog_media"*, *"vcd_catalog_vapp_template"*, *"vcd_independent_disk"*, *"vcd_network_direct"*,
+  *"vcd_network_isolated"*, *"vcd_network_isolated_v2"*, *"vcd_network_routed"*, *"vcd_network_routed_v2"*, *"vcd_org"*, *"vcd_org_vdc"*, *"vcd_provider_vdc"*,
+  *"vcd_storage_profile"*, *"vcd_vapp"*, *"vcd_vapp_vm"* or *"vcd_vm"*, which are the resources compatible with `metadata_entry`.
+* `object_name`- (Optional) Specifies the name of the object which metadata needs to be ignored. All object types are supported, except for
```
dataclouder, line 399: As mentioned above, this should be `resource_name`

vcd/metadata.go
adambarreiro: Thanks for spotting, I forgot to change all variable names when I changed the attribute. Now everything should be with the "Actions" suffix.

vcd/metadata.go
adambarreiro: Done, thanks!

vcd/metadata.go
adambarreiro: Good point, changed.

vcd/metadata.go
adambarreiro: Fixed

vcd/metadata.go
adambarreiro: Fixed

website/docs/index.html.markdown
adambarreiro: Changed

vcd/resource_vcd_vapp.go
adambarreiro: Thanks for the explanation. I have reverted the change.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -661,7 +661,7 @@ func resourceVcdVAppVmCreate(_ context.Context, d *schema.ResourceData, meta int
 	timeElapsed := time.Since(startTime)
 	util.Logger.Printf("[DEBUG] [VM create] finished VM creation in vApp [took %f seconds]", timeElapsed.Seconds())
 
-	return genericVcdVmRead(d, meta, "create")
+	return genericVcdVmRead(d, meta, "resource")
```
adambarreiro, line 664: Note for reviewers, I changed this also as `genericVcdVmRead` does nothing with neither "create" or "update" values, but it does with resource (removing from tfstate if it is not found)

cc: @Didainius  @lvirbalas  @adezxc @dataclouder

lvirbalas: I don't have enough context to grasp this change. Naturally, most important is that it keeps VM operations intact. Leaving for @dataclouder @Didainius @adezxc to review.

Didainius: ACK. I have rechecked. Good to go.

Didainius (approved):
2nd approval.
I have manually tested metadata conflict situation and double checked the `origin` field change. 

To reiterate a note of caution about `create`calling `update`. We do not want to do it because of how messy and painful it becomes to deal with it (https://github.com/vmware/terraform-provider-vcd/pull/901). Yet we still have some places where it happens from long long time ago.

website/docs/index.html.markdown
```diff
@@ -373,6 +373,102 @@ The following arguments are used to configure the VMware Cloud Director Provider
...
   it is a dot (`.`).
 
+* `ignore_metadata_changes` - (Optional; *v3.10+*) Use one or more of these blocks to ignore specific metadata entries from being changed by this Terraform provider
+  after creation or when they were created outside Terraform.
+  See ["Ignore Metadata Changes"](#ignore-metadata-changes) for more details.
+
+## Ignore metadata changes
+
+One or more `ignore_metadata_changes` blocks can be optionally set in the provider configuration, which will allow to ignore specific `metadata_entry`
+items during `plan`, `updates-in-place` and `destroy` (**not** `create`). This is useful, for example, to avoid removing metadata entries that were created
```
dataclouder, line 383: How can this affect `destroy`?
Doesn't `destroy` affect the whole resource, regardless of its contents?

website/docs/index.html.markdown
```diff
@@ -373,6 +373,102 @@ The following arguments are used to configure the VMware Cloud Director Provider
...
+by an external actor, or after they were created by Terraform.
+
+~> Note that this feature is only considered when using the `metadata_entry` argument in the resources and data sources that support
+it. In other words, to ignore metadata when you are using the deprecated `metadata` argument, please use the native Terraform `lifecycle.ignore_changes` block.
+
+~> Be aware that setting a `metadata_entry` in your Terraform configuration that matches any `ignore_metadata_changes` can produce inconsistent
+results, as the metadata will be stored in state but nothing will be done in VCD. You can control what to do in this situation with
```
dataclouder, line 390:
```suggestion
results, as the metadata will be stored in state but nothing will be done in VCD. 
Using `ignore_metadata_changes` with matching metadata entries in your code is NOT recommended.
In the event that your code contains such conflict, though, ou can control the ensuing action with
```

vcd/metadata.go
```diff
@@ -146,38 +265,94 @@ func createOrUpdateMetadataEntryInVcd(d *schema.ResourceData, resource metadataC
...
+				return diag.Errorf("unknown value for 'conflict_action': %s", action)
+			}
+
+			return diag.Diagnostics{
+				diag.Diagnostic{
+					Severity: severity,
+					Summary:  "Found a conflict between 'ignore_metadata_changes' and 'metadata_entry'",
+					Detail: fmt.Sprintf("There is an 'ignored_metadata' block: %s\n"+
+						"and there is a 'metadata_entry' with key '%s' and value '%s' in your Terraform configuration that matches the criteria, hence it will be ignored.\n"+
+						"This will cause that the entry will be present in Terraform state but it won't have any effect in VCD, causing an inconsistency.\n"+
```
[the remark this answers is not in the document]
adambarreiro: Changed

website/docs/index.html.markdown
adambarreiro: This is wrong indeed. I've changed the whole sentence to avoid specifying the operations.

website/docs/index.html.markdown
adambarreiro: Changed!

website/docs/index.html.markdown
adambarreiro: Fixed

adezxc (approved):
Thanks for the feature!

PR #1061, opened 2023-05-30
Add vcd_ip_space resource and data source

Didainius:
This PR starts addressing IP space needs as requested in #1008 
It only adds resource and data source `vcd_ip_space` in this PR, while the next PRs will add an integration layer for other components.

#### other things
* `make fmt` adds `-s` flag because `make fmtcheck` expects it
* `TestAccDataSourceNotFound/vcd_nsxt_firewall` and `TestAccDataSourceNotFound/vcd_nsxt_route_advertisement` were broken and are fixed as part of this PR

PR #1062, opened 2023-06-01
IP Space integration

### LINKED ISSUE #1008: Feature Request - IP Spaces
### Community Note

Hi,

As of Cloud Directory 10.4.1 there is a feature called [IP Spaces](https://docs.vmware.com/en/VMware-Cloud-Director/10.4/VMware-Cloud-Director-Tenant-Portal-Guide/GUID-FB230D89-ACBC-4345-A11A-D099D359ED1B.html). I'm wondering if there are any plans to add  resources and/or data sources for IP Spaces in the near future? 

Thanks, Jasper

Didainius:
Closes #1008 

This PR finalizes IP space work that started in #1061.
IP Space support is available for VCD 10.4.1+.
 
In addition to the already existing resource and data source `vcd_ip_space` there are 3 more resources (and data sources) added:
* `vcd_ip_space_uplink` (maps IP Spaces to Provider Gateways `vcd_external_network_v2` resources)
* `vcd_ip_space_ip_allocation` (allocates floating IPs or IP prefixes to feed them into other resources)
* `vcd_ip_space_custom_quota` (overrides default quotas specified in `vcd_ip_space_uplink` and sets custom ones for specific Orgs)

These resources are patched to support IP Space IP allocations:
* `vcd_nsxt_edgegateway` (accepts configuration without any `subnet` blocks to inherit IP Space)
* `vcd_external_network_v2` (adds `use_ip_spaces` boolean field and accepts configuration without subnet specification, accepts `dedicated_org_id` to dedicate provider gateway to a specific Org)

Additionally, there is a **new Guide page** which should give a birds eye view to IP Space support with a complete snippet of how they can be used

#### Important notes
* Bugfix/improvement on `vcd_ip_space` `update` mechanism. In short - every `ip_range` or `ip_prefix` requires its own ID to be sent when `update` occurs. This is unfortunate as due to Terraform `TypeSet` there is no trivial way to implement it. I have added additional functions `getIpPrefixSequenceIdFromFromPreviousState` and `getIpRangeIdFromFromPreviousState` which are trying to match the IDs stored in statefile with the configuration to be applied. Due to Terraform internals it is not always possible to precisely match IDs therefore there is a risk that a certain `ip_prefix` can be recreated instead of `update`, but this is the best we can do, given Terraform technical limitations.
* When manipulating IP Space or IP Space Uplink, shortly after the operation VCD can kick in "Route Advertisement sync" task (name `ipSpaceUplinkRouteAdvertisementSync`). This task blocks other resources while it is running. To prevent hitting errors of type `The entity Ref: com.vmware.vcloud.entity.gateway:8d4ee7a3-6b42-4647-ba6a-0eb7af4f0bb4 is busy completing an operation IP_SPACE_UPLINK_ROUTE_ADVERTISEMENT_SYNC`, I have added a sleep of 3 seconds (because sometimes the operation is not immediately seen, and a func to wait until such tasks if found. Sample code snippet:
```go
	// Operations on IP Space related entities trigger a separate task
	// 'ipSpaceUplinkRouteAdvertisementSync' which is better to finish before any other operations
	// as it might cause an error: busy completing an operation IP_SPACE_UPLINK_ROUTE_ADVERTISEMENT_SYNC
	// Sleeping a few seconds because the task is not immediately seen sometimes.
	time.Sleep(3 * time.Second)
	err = vcdClient.Client.WaitForRunningTasksByName("ipSpaceUplinkRouteAdvertisementSync")
	if err != nil {
		return diag.FromErr(err)
	}
```

#### Testing
* Patch failing test `TestAccVcdOpenApiDhcpNsxtRouted`

Acceptance and binary tests run on 10.3.0 (validated test skips), 10.4.1, 10.4.2


vcd/datasource_vcd_nsxt_edgegateway.go
```diff
@@ -176,6 +176,11 @@ func datasourceVcdNsxtEdgeGateway() *schema.Resource {
 				Computed:    true,
 				Description: "Total number of IP addresses allocated for this Edge Gateway",
 			},
+			"uses_ip_spaces": {
```
lvirbalas, line 179: Is there any chance this would become writable in the future? If yes, `uses_ip_spaces` could be named `use_ip_spaces`. However, if it's read-only for the foreseeable future, then the current `uses_ip_spaces` is more descriptive.

vcd/resource_vcd_external_network_v2.go
```diff
@@ -142,9 +142,21 @@ func resourceVcdExternalNetworkV2() *schema.Resource {
 				Optional:    true,
 				Description: "Network description",
 			},
+			"dedicated_org_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "Dedicate External Network to an Org (only with IP Spaces)",
+			},
+			"use_ip_spaces": {
```
lvirbalas, line 151: This field name relates to my question above. We should consider maybe it's worth using `use_ip_spaces` (compared to `uses_ip_spaces`) across the board. Just opening up for thoughts.

website/docs/d/ip_space_ip_allocation.html.markdown
```diff
@@ -0,0 +1,51 @@
...
+page_title: "VMware Cloud Director: vcd_ip_space_ip_allocation"
+sidebar_current: "docs-vcd-data-source-ip-space-ip-allocation"
+description: |-
+  Provides a data source to read IP IP Allocations within IP Spaces. It supports both - Floating IPs 
+  (IPs from IP Ranges) and IP Prefix (subnet) allocations with manual and automatic reservations.
+---
+
+# vcd\_ip\_space\_ip\_allocation
+
+Provides a data source to read IP IP Allocations within IP Spaces. It supports both - Floating IPs
```
lvirbalas, line 12: Too many IPs: "IP IP Allocations"

website/docs/guides/ip_spaces.html.markdown
```diff
@@ -0,0 +1,291 @@
...
+blocks must be contained in. The *external scope* defines the *total span of IP addresses to which
+the IP space has access*, for example the internet or a WAN. The internal and external scopes are
+used to define default NAT rules and BGP prefixes.
+
+As a service provider, you create public, shared, or private IP spaces and assign them to provider
+gateways by creating IP space uplinks. After creating an IP space, you can assign to it IP prefixes
+for networks and floating IP addresses for network services.
+
+Organization administrators can view general information about the IP spaces in their organization
+has access, and manage the IP spaces available to them.
```
lvirbalas, line 31: "has access" - words are missing?

website/docs/guides/ip_spaces.html.markdown
```diff
@@ -0,0 +1,291 @@
...
+The following new resources for IP Space management are now available with Terraform Provider for
+VCD 3.10+:
+
+* `vcd_ip_space` - provides IP Space and default quota definition capability
+* `vcd_ip_space_uplink` - provides capability to assign IP Space Uplink for Provider Gateways
+* `vcd_ip_space_ip_allocation` - provides capability to allocation floating IPs or IP Prefix
+* `vcd_ip_space_custom_quota` - provides capability to set Org specified Custom Quotas and override
+  default ones defined in `vcd_ip_space`
+
+## Modified resources for IP Space support
```
lvirbalas, line 56: You wrote this particular section in a changelog/blog style. However, the guide will stay for years and the "modified", "amended" and "new" words will quickly become old. I suggest rephrasing this section to just reflect how to use IP spaces in these resources, but instead of highlighting that this is new just put a "*v3.10+*; *VCD 10.4.1+*" version in the parenthesis (like we do in docs) so it's clear from which version it is available.

lvirbalas (commented):
Such a powerful feature!! Please find suggestions inline.

vcd/resource_vcd_external_network_v2.go
Didainius: No idea if it can, but I think we can rename it without big doubt. It still remains `Computed` for now. The decision, if Edge Gateway is backed by IP Spaces for now, is decided by the backing Provider Gateway (`vcd_external_network_v2`) and if it has IP Space Uplinks (`vcd_ip_space_uplink`)

vcd/resource_vcd_ip_space.go
```diff
@@ -425,3 +478,97 @@ func setIpSpaceData(d *schema.ResourceData, ipSpace *types.IpSpace) error {
...
+	ipRangesOld, _ := d.GetChange("ip_range")
+	ipRangesOldSchema := ipRangesOld.(*schema.Set)
+	ipRangesSlice := ipRangesOldSchema.List()
+
+	util.Logger.Printf("[TRACE] Looking for ID of 'ip_range' with start_address '%s' and end_address '%s'\n", startAddress, endAddress)
+
+	// Looping over ip_range definitions from state which contained all values and also stored ID. It looks for this ID with 2 priority levels:
+	// 1. An exact match with the same start and end IP addresses is found - return it immediately
+	// as this particular block has not changed at all for update operation
+	// 2. A partial match - if at least the start_address or end_address remained the same - cache
```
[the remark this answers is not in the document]
Didainius: Will give another think about it, but if we leave only an exact match - the risk is that then one cannot perform `update` to the range at all, and because of Terraform schema limitations we cannot flag that to user in diff (the only option we have is to Force whole resource creation, which is not what we want.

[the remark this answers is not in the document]
Didainius: So another thought about possible outcomes for the partial match case of IP range.

Because the "_partial match"_ matches the ID to either `start_address` or the `end_address` there is a risk that if one takes a range of 
```
start_address = 10.10.10.10
end_address  = 10.10.10.20
```


And in an update, he splits into two ranges where one matches the `start_address` and the other matches the `end_address` - we will get an update error ( duplicate IDs). The workaround would be to do that split in 2 update steps.
```
start_address = 10.10.10.10 <- matches previous start_address
end_address  = 10.10.10.14

start_address = 10.10.10.18
end_address  = 10.10.10.20 <- matches previous end_address
```

Another way is - we could be more conservative and **only match on `start_address`** (and not on `end_address`) for _"partial match"_, but the chances to update become a bit more limited. OTOH, Floating IP allocation (at least to what I noticed) usually starts from the start of the range. So if one has anything allocated, he should not be able to to change `start_address` during update and if he does not have it allocated - it should be fine to recreate that range (do the _update_ without specifying previous ID)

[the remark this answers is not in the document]
lvirbalas: > And in a single update, he splits into two ranges where one matches the start_address and the other matches the end_address - we will get an update error ( duplicate IDs).

Getting an error is fine. Most importantly, is there any chance for data corruption: when user sends one thing, but as a result gets back a different one?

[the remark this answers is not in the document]
Didainius: I have looked at this particular case manually. I had a range that had 3 floating IPs allocated for services.

So here is the diff I attempted:
```
     - ip_range {
          - end_address   = "11.11.11.111" -> null
          - id            = "cd10790c-5a65-4869-aeb2-b05568d28c01" -> null
          - start_address = "11.11.11.100" -> null
        }
      + ip_range {
          + end_address   = "11.11.11.108"
          + id            = (known after apply)
          + start_address = "11.11.11.100"
        }
      + ip_range {
          + end_address   = "11.11.11.111"
          + id            = (known after apply)
          + start_address = "11.11.11.110"
        }
```

Instead of one range:
```
start_address = "11.11.11.100"
end_address   = "11.11.11.111"
```

I put two ranges 
```
start_address = "11.11.11.100" <- matches the start_address of single range I had
end_address   = "11.11.11.108"


start_address = "11.11.11.110"
end_address   = "11.11.11.111" <- matches the end_address of single range I had
```

This error occurred during apply (Update):
```
vcd_ip_space.space1: Modifying... [id=urn:vcloud:ipSpace:ab75e441-d95c-4c43-9957-d5b8f7b10955]
╷
│ Error: error updating IP Space: error updating IP Space: error in HTTP PUT request: INTERNAL_SERVER_ERROR - [ fda98c3e-e04b-4210-b49c-2f2b1d04370f ] Duplicate key cd10790c-5a65-4869-aeb2-b05568d28c01 (attempted merging values 11.11.11.110 - 11.11.11.111 and 11.11.11.100 - 11.11.11.108)
│ 
│   with vcd_ip_space.space1,
│   on vcd.TestAccVcdIpSpaceIpAllocationstep2.tf line 45, in resource "vcd_ip_space" "space1":
│   45: resource "vcd_ip_space" "space1" {
│ 
```

In the UI - it remained unchanged - meaning that no changes occurred during this error. I tried running `terraform apply` 3 times - luckily it always presents the same diff and no changes occur in UI so this error looks to be addressable with no harm to configuration.

Then, I have attempted to do this "split" in two steps:
1. Shrunk existing range from `11.11.11.100 - 11.11.11.111` to `11.11.11.100 - 11.11.11.108` (this worked - update preserved correct ID and no impact to allocated IPs occurred)
2.  I added another range `11.11.11.110-11.11.11.111` which is just an addition of a new IP range now (it wouldn't have an old ID anyway)

vcd/datasource_vcd_nsxt_edgegateway.go
Didainius: Changed to `uses_ip_spaces` here, resource, tests and in website docs. Test still passes :)

vcd/resource_vcd_external_network_v2.go
Didainius: Changed to `uses_ip_spaces` here, resource, tests and in website docs. Also ran a test to see that it still passes

website/docs/d/ip_space_ip_allocation.html.markdown
Didainius: Fixed here and and resource docs :)

website/docs/guides/ip_spaces.html.markdown
Didainius: Apparently, I copied this part from official docs. Double-checked - this error is there - improved :)

website/docs/guides/ip_spaces.html.markdown
Didainius: Changed a bit differently - have a look if it works - didn't iterate `v3.10+` everywhere - just in introductory sentence

vcd/datasource_vcd_nsxt_edgegateway.go
lvirbalas: I think you have meant `use_ip_spaces` - and it looks great now!

vcd/resource_vcd_external_network_v2.go
lvirbalas: (Trying to post a reply second time - GitHub is misbehaving).

I think you have meant `use_ip_spaces` and it looks great!

vcd/resource_vcd_ip_space.go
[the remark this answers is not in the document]
lvirbalas: Interesting. What would happen with current matching mechanism if you did such a change:

```
start_address = "11.11.11.100"
end_address   = "11.11.11.111"
```

split into two with a caveat ->

```
start_address = "11.11.11.100"
end_address   = "11.11.11.108"

start_address = "11.11.11.111" <- the old end_address becomes a new start_address
end_address   = "11.11.11.118"
```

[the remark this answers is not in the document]
Didainius: In my mind I evaluated it to create new one for the second one - because the condition for partial match is this:

```go
if ipRangeStrings["start_address"] == startAddress || ipRangeStrings["end_address"] == endAddress {
```

That is an IP address and "position" (start or end) must match.

Here is the payload excerpt it created for `PUT`:
```json
...
"ipSpaceRanges": {
    "ipRanges": [
      {
        "startIpAddress": "11.11.11.111", <-- this did not match as previously it was `end_address`, no `id` field
        "endIpAddress": "11.11.11.118"
      },
      {
        "id": "8f7ed9d5-6779-4c4e-80af-b9ce14fad3db",
        "startIpAddress": "11.11.11.100",  <--- `start_address` before and after matched - `id` is present
        "endIpAddress": "11.11.11.108"
      }
    ],
    "defaultFloatingIpQuota": 0
  },
....
```


I also tried the reverse scenario - from:
```
start_address = "11.11.11.100"
end_address   = "11.11.11.108"

start_address = "11.11.11.111"
end_address   = "11.11.11.118"
```

Change to 
```
start_address = "11.11.11.100"
end_address   = "11.11.11.111"
```

It also worked well, picked the correct ID because it matched `start_address = "11.11.11.100"`. 

Both update scenarios worked given that that this range still had 3 floating IPs allocated - `11.11.11.100`, `11.11.11.101`, `11.11.11.102`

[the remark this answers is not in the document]
lvirbalas: That's great! So, it seems the only apparent issue is how we handle when the user does hit that error:

> │ Error: error updating IP Space: error updating IP Space: error in HTTP PUT request: INTERNAL_SERVER_ERROR - [ fda98c3e-e04b-4210-b49c-2f2b1d04370f ] Duplicate key cd10790c-5a65-4869-aeb2-b05568d28c01 (attempted merging values 11.11.11.110 - 11.11.11.111 and 11.11.11.100 - 11.11.11.108)

I mean, the above will be very confusing for the user.

adezxc (approved):
Tested it out manually, works amazing and was probably more clear than the official guide. Some general comments

vcd/resource_vcd_ip_space_ip_allocation.go
```diff
@@ -0,0 +1,304 @@
...
+			"allocation_date": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Allocation date in ISO 8601 format (e.g. 2023-06-07T09:57:58.721Z)",
+			},
+			"ip": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "IP address part",
+			},
```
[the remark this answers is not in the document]
Didainius: Yes - this is a "sugar" field. `ip_address` is the value that is returned from API - it will be an IP for `type=FLOATING_IP` or CIDR for `type=IP_PREFIX`. `ip` will always contain an IP - even in `CIDR` case

vcd/resource_vcd_ip_space_ip_allocation.go
```diff
@@ -0,0 +1,304 @@
...
+	dSet(d, "ip_address", ipAllocation.IpSpaceIpAllocation.Value)
+
+	// When IP Prefix is allocated, the returned value is in CIDR format (e.g. 192.168.1.0/24), and
+	// although it can be split using Terraform native functions, we're adding a convenience layer for
+	// users by splitting this address into IP and prefix length
+	if ipAllocation.IpSpaceIpAllocation.Type == "IP_PREFIX" {
+		splitCidr := strings.Split(ipAllocation.IpSpaceIpAllocation.Value, "/")
+		if len(splitCidr) == 2 {
+			dSet(d, "ip", splitCidr[0])
+			dSet(d, "prefix_length", splitCidr[1])
```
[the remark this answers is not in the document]
Didainius: I think we do need `ip_address` which is the real representation of what the API returns. If it looks confusing, then I probably get rid of `ip` and computed `prefix_length`, which are sugar. 
The CIDR value which can be in `ip_address` can be fed directly into functions like [`cidrhost`](https://developer.hashicorp.com/terraform/language/functions/cidrhost)

vcd/resource_vcd_ip_space_test.go
```diff
@@ -258,6 +258,7 @@ resource "vcd_ip_space" "space1" {
 `
 
 const testAccVcdIpSpacePublicStep5DS = testAccVcdIpSpacePublicStep5 + `
+# skip-binary-test: Data Source test
 data "vcd_ip_space" "space1" {
   name = "{{.TestName}}"
 }
```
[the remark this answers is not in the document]
Didainius: I will check - the reason I am adding this is because:

We used this resource and data source approach in a single file for a long time, but then Terraform stopped supporting it at some time (even with `depends_on`) - all these tests started failing. It needed a lot of work to split all the test work. This happened a few years ago

vcd/resource_vcd_ip_space_ip_allocation.go
```diff
@@ -0,0 +1,304 @@
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "IP Space ID for IP Allocation",
+			},
+			"type": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				Description:  "Edge gateway name in which NAT Rule is located",
```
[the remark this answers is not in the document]
Didainius: Nope :) Fixed

website/docs/guides/ip_spaces.html.markdown
```diff
@@ -0,0 +1,286 @@
...
+
+
+## Sample end to end configuration using IP Spaces
+
+-> There are new rights for IP Space management starting with VCD 10.4.1. Some of them are [listed
+in the
+prerequisites](https://docs.vmware.com/en/VMware-Cloud-Director/10.4/VMware-Cloud-Director-Service-Provider-Admin-Portal-Guide/GUID-575513A8-9ADE-4A3D-92AB-CB0917FF8316.html)
+for IP Space management in Organizations.
+
+Each resource have their own documentation, but this snippet gives a birds eye view how all the
```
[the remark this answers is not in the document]
Didainius: Changed

vcd/resource_vcd_ip_space.go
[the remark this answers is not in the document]
Didainius: So - to reiterate. This can happen in such case:
Splitting a single range:
```
start_address = "11.11.11.100"
end_address   = "11.11.11.111"
```

Into two ranges, where one matches `start_address` and another matches `end_address`
```
start_address = "11.11.11.100" <- matches the start_address of single range I had
end_address   = "11.11.11.108"


start_address = "11.11.11.110"
end_address   = "11.11.11.111" <- matches the end_address of single range I had
```

I could catch the error (Duplicate key) and add our own message. Does such a message make sense:
> Splitting a single IP Range into two ranges with matching `start_address` and `end_address` is not supported in a single update operation. Please perform it in two steps.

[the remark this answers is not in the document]
Didainius: Made such an attempt https://github.com/vmware/terraform-provider-vcd/pull/1062/files#diff-134d08acac0e541b396ca1ac4225bbcda87e08ecbcb8772049dbc6026d52b79fR195-R212

vcd/resource_vcd_ip_space_test.go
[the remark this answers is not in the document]
Didainius: Added `depends_on` and ran tests in acceptance and binary test modes to be sure that it works

vcd/resource_vcd_ip_space.go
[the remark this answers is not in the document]
lvirbalas: Such precise error is exactly what we needed! To close this out, I'm reading new code: is there any other scenario where "duplicate" message may be returned and we may misleading the user with the "split" hint? Maybe the "if" should be more deterministic and check that it's actually a "split" use case?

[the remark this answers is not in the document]
Didainius: I have attempted to abuse partial matches to see if there are any other errors that can happen. Here is a list of what I have tried:

* **IP Prefix** (it does partial match _only on `first_ip`_ field.
To replicate a problem, I tried to create:
```hcl
    prefix {
      first_ip      = "192.168.1.100"
      prefix_length = 30
      prefix_count  = 4
    }

    prefix { <- added on update
      first_ip      = "192.168.1.100" <- duplicate `first_ip`
      prefix_length = 28
      prefix_count  = 1
    }
```

```
 Error: error updating IP Space: error updating IP Space: error in HTTP PUT request: BAD_REQUEST - [ 15a6839f-f675-4454-9cfd-fab815c9010b ] IP Space TestAccVcdIpSpaceIpAllocation has prefix length 30 mismatch in the configuration.
│ 
│   with vcd_ip_space.space1,
│   on vcd.TestAccVcdIpSpaceIpAllocationstep2.tf line 45, in resource "vcd_ip_space" "space1":
│   45: resource "vcd_ip_space" "space1" {
```


* **IP Range** (it does partial matching _either on `start_address` or `end_address`_

This is our known and tested scenario:
Splitting a single range:
```
start_address = "11.11.11.100"
end_address   = "11.11.11.111"
```

Into two ranges, where one matches `start_address` and another matches `end_address`
```
start_address = "11.11.11.100" <- matches the start_address of single range I had
end_address   = "11.11.11.108"


start_address = "11.11.11.110"
end_address   = "11.11.11.111" <- matches the end_address of single range I had
```

The error currently looks like this:
```
vcd_ip_space.space1: Modifying... [id=urn:vcloud:ipSpace:99746875-6942-4054-8030-77af980d9445]
╷
│ Error: error updating IP Space: splitting a single IP Range into two ranges with matching 'start_address' and 'end_address' is not supported in a single update operation. Please perform it in two steps: error updating IP Space: error in HTTP PUT request: INTERNAL_SERVER_ERROR - [ d20e9f9d-c985-4ef1-a60c-8b7be29d859b ] Duplicate key 7af38863-1af7-4481-9ee4-79c7ba782296 (attempted merging values 11.11.11.100 - 11.11.11.107 and 11.11.11.108 - 11.11.11.111)
│ 
│   with vcd_ip_space.space1,
│   on vcd.TestAccVcdIpSpaceIpAllocationstep2.tf line 45, in resource "vcd_ip_space" "space1":
│   45: resource "vcd_ip_space" "space1" {
```

* **Another IP Range  scenario** is adding second IP range with _duplicate `start_address` or `end_address`_ 

```hcl
  ip_range {
    start_address = "11.11.11.100"
    end_address   = "11.11.11.111"
  }

  ip_range { <- added during update
    start_address = "11.11.11.100"
    end_address   = "11.11.11.108"
  }
```

The error in both cases looks like this:
```
vcd_ip_space.space1: Modifying... [id=urn:vcloud:ipSpace:99746875-6942-4054-8030-77af980d9445]
╷
│ Error: error updating IP Space: error updating IP Space: error in HTTP PUT request: BAD_REQUEST - [ 4a62d80f-2d01-4270-b3d6-44ba13134b9e ] The following IP ranges contain an overlap: 11.11.11.100 - 11.11.11.111 and 11.11.11.100 - 11.11.11.108.
│ 
│   with vcd_ip_space.space1,
│   on vcd.TestAccVcdIpSpaceIpAllocationstep2.tf line 45, in resource "vcd_ip_space" "space1":
│   45: resource "vcd_ip_space" "space1" {
│ 
╵
```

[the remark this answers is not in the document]
lvirbalas: This is great, thank you! I'm resolving this conversation. Also, approving the PR :)

lvirbalas (approved):

website/docs/guides/ip_spaces.html.markdown
```diff
@@ -0,0 +1,286 @@
...
+  used by only one organization that is specified during the space creation. For this organization,
+  IP consumption is unlimited.
+
+
+## List of resources with IP Space support
+
+The following resources for IP Space management are available with Terraform Provider for
+VCD starting with version 3.10:
+
+* `vcd_ip_space` - provides IP Space and default quota definition capability
```
dataclouder, line 50: Shall we add the link to each resource, as we do for the roles and subscribed catalog guide?

website/docs/guides/ip_spaces.html.markdown
```diff
@@ -0,0 +1,286 @@
...
+  prefix_length = split("/", vcd_ip_space_ip_allocation.public-ip-prefix.ip_address)[1]
+
+  static_ip_pool {
+    start_address = cidrhost(vcd_ip_space_ip_allocation.public-ip-prefix.ip_address, 2)
+    end_address   = cidrhost(vcd_ip_space_ip_allocation.public-ip-prefix.ip_address, 4)
+  }
+}
+```
+
+## Sample configuration without IP Spaces (the original way)
```
dataclouder, line 225: I would like to see the material rearranged as follows:

1. a few lines of explanation of what we want to achieve
2. A sample of the old usage (this section)
3. A sample of the new usage (the previous section)

The way it is now, we only see a short way of doing things and a long way of doing things, but we can't appreciate where exactly we should apply the changes.
It will be helpful if the samples had commented labels for the corresponding operations in the old and new.

vcd/resource_vcd_external_network_v2.go
```diff
@@ -142,9 +142,21 @@ func resourceVcdExternalNetworkV2() *schema.Resource {
 				Optional:    true,
 				Description: "Network description",
 			},
+			"dedicated_org_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "Dedicate External Network to an Org (only with IP Spaces)",
```
dataclouder, line 149: description and name don't match

website/docs/r/external_network_v2.html.markdown
```diff
@@ -14,11 +14,30 @@ Provides a VMware Cloud Director External Network resource (version 2). New vers
 uses new VCD API and is capable of creating NSX-T backed external networks as well as port group
 backed ones.
 
--> **Note:** This resource uses new VMware Cloud Director
-[OpenAPI](https://code.vmware.com/docs/11982/getting-started-with-vmware-cloud-director-openapi) and
-requires at least VCD *10.0+*. It supports both NSX-T and NSX-V backed networks (NSX-T *3.0+* requires VCD *10.1.1+*)
+-> This resource manages both NSX-T and NSX-V **External Networks** and **NSX-T Provider
```
dataclouder, line 17: "both" refers to *two* elements, but we are listing three here
```suggestion
-> This resource manages NSX-T **External Networks**, NSX-V **External Networks**, and **NSX-T Provider**
```

website/docs/r/ip_space_custom_quota.html.markdown
```diff
@@ -0,0 +1,81 @@
+---
+layout: "vcd"
+page_title: "VMware Cloud Director: vcd_nsxt_alb_service_engine_group"
+sidebar_current: "docs-vcd-resource-nsxt-alb-service-engine-group"
```
dataclouder, line 4:
```suggestion
page_title: "VMware Cloud Director: vcd_ip_space_custom_quota"
sidebar_current: "docs-vcd-resource-ip-space-custom-quota"
```

website/docs/d/ip_space_custom_quota.html.markdown
```diff
@@ -0,0 +1,33 @@
+---
+layout: "vcd"
+page_title: "VMware Cloud Director: vcd_nsxt_alb_settings"
+sidebar_current: "docs-vcd-datasource-nsxt-alb-settings"
```
dataclouder, line 4:
```suggestion
page_title: "VMware Cloud Director: vcd_ip_space_custom_quota"
sidebar_current: "docs-vcd-datasource-ip-space-custom-quota"
```

vcd/resource_vcd_ip_space.go
```diff
@@ -180,9 +192,37 @@ func resourceVcdIpSpaceUpdate(ctx context.Context, d *schema.ResourceData, meta
...
+		// start_address = "11.11.11.100" <- matches the start_address of single range I had
+		// end_address   = "11.11.11.108"
+		// start_address = "11.11.11.110"
+		// end_address   = "11.11.11.111" <- matches the end_address of single range I had
+		//
+		// It does not cause any issues on VCD or Terraform state except the split is not done even
+		// after multiple apply attempts.
+		// The point here is to hint user that what he is doing could be done in two steps
+		if strings.Contains(err.Error(), "Duplicate key") {
+			return diag.Errorf("error updating IP Space: splitting a single IP Range into two ranges "+
```
adambarreiro, line 208: This is really a helpful message. I wonder whether is it possible to tell the user which IP range is problematic?
Probably not as we only have `err` (and possibly it's enough with the `Error()` message), but just asking just in case 😄

vcd/resource_vcd_ip_space.go
```diff
@@ -180,9 +192,37 @@ func resourceVcdIpSpaceUpdate(ctx context.Context, d *schema.ResourceData, meta
...
+		}
+
 		return diag.Errorf("error updating IP Space: %s", err)
 	}
 
+	// Operations on IP Space related entities trigger a separate task
+	// 'ipSpaceUplinkRouteAdvertisementSync' which is better to finish before any other operations
+	// as it might cause an error: busy completing an operation IP_SPACE_UPLINK_ROUTE_ADVERTISEMENT_SYNC
+	// Sleeping a few seconds because the task is not immediately seen sometimes.
+	time.Sleep(3 * time.Second)
```
adambarreiro, line 220: Is there any risk that 3 seconds is not enough?

adambarreiro (commented):
First scan. I really liked the way of dealing with updates in the IP Spaces, I don't think we have a better option if we don't have the ID to do an exact match.

vcd/resource_vcd_ip_space.go
Didainius: We probably can't be very precise, but VCD will hint. 

```
vcd_ip_space.space1: Modifying... [id=urn:vcloud:ipSpace:99746875-6942-4054-8030-77af980d9445]
╷
│ Error: error updating IP Space: splitting a single IP Range into two ranges with matching 'start_address' and 'end_address' is not supported in a single update operation. Please perform it in two steps: error updating IP Space: error in HTTP PUT request: INTERNAL_SERVER_ERROR - [ d20e9f9d-c985-4ef1-a60c-8b7be29d859b ] Duplicate key 7af38863-1af7-4481-9ee4-79c7ba782296 (attempted merging values 11.11.11.100 - 11.11.11.107 and 11.11.11.108 - 11.11.11.111)
│ 
│   with vcd_ip_space.space1,
│   on vcd.TestAccVcdIpSpaceIpAllocationstep2.tf line 45, in resource "vcd_ip_space" "space1":
│   45: resource "vcd_ip_space" "space1" {
```

A bit more in this discussion - https://github.com/vmware/terraform-provider-vcd/pull/1062#discussion_r1251898325

vcd/resource_vcd_ip_space.go
Didainius: I hope not - I added this as an extra precaution. In some cases, the task even finishes within this timeframe :)

vcd/resource_vcd_ip_space_ip_allocation.go
```diff
@@ -0,0 +1,304 @@
...
+	dSet(d, "allocation_date", ipAllocation.IpSpaceIpAllocation.AllocationDate)
+	dSet(d, "usage_state", ipAllocation.IpSpaceIpAllocation.UsageState)
+	dSet(d, "ip_address", ipAllocation.IpSpaceIpAllocation.Value)
+
+	// When IP Prefix is allocated, the returned value is in CIDR format (e.g. 192.168.1.0/24), and
+	// although it can be split using Terraform native functions, we're adding a convenience layer for
+	// users by splitting this address into IP and prefix length
+	if ipAllocation.IpSpaceIpAllocation.Type == "IP_PREFIX" {
+		splitCidr := strings.Split(ipAllocation.IpSpaceIpAllocation.Value, "/")
+		if len(splitCidr) == 2 {
```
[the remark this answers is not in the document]
Didainius: Added.
 We will be able to check what exactly was returned in the error code, but logging that additionally now (I don't want to "fail" just because of that as one could still use the `ip_address` field which is a value directly coming from the API. At least someone doesn't have to wait for bugfix if this happens.

vcd/datasource_vcd_nsxt_edgegateway.go
```diff
@@ -176,6 +176,11 @@ func datasourceVcdNsxtEdgeGateway() *schema.Resource {
 				Computed:    true,
 				Description: "Total number of IP addresses allocated for this Edge Gateway",
 			},
+			"use_ip_spaces": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Boolean value that hints if the Edge Gateway is using IP Spaces",
```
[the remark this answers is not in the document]
Didainius: It specified - I changed it here and in resource

vcd/datasource_vcd_ip_space_uplink.go
```diff
@@ -0,0 +1,77 @@
...
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "IP Space Uplink description",
+			},
+			"external_network_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
```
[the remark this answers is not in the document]
Didainius: Fixed

vcd/datasource_vcd_ip_space_ip_allocation.go
```diff
@@ -0,0 +1,124 @@
...
+
+func datasourceVcdIpAllocation() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: datasourceVcdIpAllocationRead,
+
+		Schema: map[string]*schema.Schema{
+			"org_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
```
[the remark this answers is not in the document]
Didainius: Fixed

vcd/datasource_vcd_ip_space_custom_quota.go
```diff
@@ -0,0 +1,80 @@
...
+
+func datasourceVcdIpSpaceCustomQuota() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: datasourceVcdIpSpaceCustomQuotaRead,
+
+		Schema: map[string]*schema.Schema{
+			"ip_space_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
```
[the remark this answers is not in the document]
Didainius: Fixed

website/docs/d/ip_space_custom_quota.html.markdown
Didainius: Fixed.

website/docs/r/ip_space_custom_quota.html.markdown
Didainius: Fixed.

website/docs/r/external_network_v2.html.markdown
Didainius: Very good catch - fixed

vcd/resource_vcd_external_network_v2.go
Didainius: Improved

website/docs/guides/ip_spaces.html.markdown
Didainius: Made links

vcd/resource_vcd_external_network_v2.go
```diff
@@ -288,13 +300,32 @@ func getExternalNetworkV2Type(vcdClient *VCDClient, d *schema.ResourceData, know
...
+	}
+
+	usingIpSpace := d.Get("use_ip_spaces").(bool)
+	if usingIpSpace {
+		newExtNet.UsingIpSpace = &usingIpSpace
+	}
+
+	// Using IP blocks
+	if !usingIpSpace {
+		subnetSlice := getSubnetsType(d)
```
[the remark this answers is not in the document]
Didainius: Good point. I think code readability is a bit better when validation is better a few lines below (added it). Here's the error:
```
vcd_external_network_v2.ext-net-nsxt: Creating...
╷
│ Error: could not get network data: 'ip_scope' must be set when 'use_ip_spaces' is not enabled
│
│   with vcd_external_network_v2.ext-net-nsxt,
│   on config.tf line 37, in resource "vcd_external_network_v2" "ext-net-nsxt":
│   37: resource "vcd_external_network_v2" "ext-net-nsxt" {
│
╵
```

website/docs/guides/ip_spaces.html.markdown
Didainius: Modified

website/docs/guides/ip_spaces.html.markdown
```diff
@@ -0,0 +1,286 @@
...
+resource "vcd_ip_space" "space1" {
+  name = "public-ip-space"
+  type = "PUBLIC"
+
+  internal_scope = ["192.168.1.0/24", "10.10.10.0/24", "11.11.11.0/24"]
+  external_scope = "0.0.0.0/24"
+
+  route_advertisement_enabled = false
+
+  ip_prefix {
```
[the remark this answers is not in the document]
Didainius: I have restructured it - you could see if it got any better

vcd/resource_vcd_ip_space.go
adambarreiro: That's great!

website/docs/guides/ip_spaces.html.markdown
```diff
@@ -0,0 +1,323 @@
...
+VCD starting with version 3.10:
+
+* [`vcd_ip_space`](/providers/vmware/vcd/latest/docs/resources/ip_space) - provides IP Space and
+  default quota definition capability
+* [`vcd_ip_space_uplink`](/providers/vmware/vcd/latest/docs/resources/ip_space_uplink) - provides
+  capability to assign IP Space Uplink for Provider Gateways (resource
+  [`vcd_external_network_v2`](/providers/vmware/vcd/latest/docs/resources/external_network_v2))
+* [`vcd_ip_space_ip_allocation`](/providers/vmware/vcd/latest/docs/resources/ip_space_ip_allocation)
+  provides capability to allocate floating IPs or IP Prefixes
+* `vcd_ip_space_custom_quota` - provides capability to set Org specific Custom Quotas and override
```
adambarreiro, line 57:
```suggestion
* [`vcd_ip_space_custom_quota`](/providers/vmware/vcd/latest/docs/resources/ip_space_custom_quota) - provides capability to set Org specific Custom Quotas and override
```

website/docs/r/external_network_v2.html.markdown
```diff
@@ -169,7 +188,11 @@ The following arguments are supported:
 
 * `name` - (Required) A unique name for the network
 * `description` - (Optional) Network friendly description
-* `ip_scope` - (Required) One or more IP scopes for the network. See [IP Scope](#ipscope) below for details.
+* `use_ip_spaces` - (Optional; *v3.10+*; *VCD 10.4.1+*) Defines if the network uses IP Spaces. Do
+* `dedicated_org_id` - (Optional; *v3.10+*; *VCD 10.4.1+*) An Org ID that this network should be
+  not specify `ip_scope` when using IP Spaces. (default `false`)
+* dedicated to. Only applicable when `use_ip_spaces=true`
```
adambarreiro, line 194: There's a strange mix there. I think it is
```suggestion
* `use_ip_spaces` - (Optional; *v3.10+*; *VCD 10.4.1+*) Defines if the network uses IP Spaces. Do not specify `ip_scope` when using IP Spaces. (default `false`)
* `dedicated_org_id` - (Optional; *v3.10+*; *VCD 10.4.1+*) An Org ID that this network should be dedicated to. Only applicable when `use_ip_spaces=true`
```

website/docs/r/ip_space_custom_quota.html.markdown
```diff
@@ -0,0 +1,81 @@
...
+  # Custom Quota can only be configured once Edge Gateway is created
+  depends_on = [vcd_nsxt_edgegateway.ip-space]
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `ip_space_id` - (Required) - IP Space ID to set Custom Quotas
```
adambarreiro, line 46:
```suggestion
* `ip_space_id` - (Required) IP Space ID to set Custom Quotas
```

website/docs/r/ip_space_custom_quota.html.markdown
```diff
@@ -0,0 +1,81 @@
...
+Gateway backed by Provider Gateway is created within the Org. An explicit `depends_on` constraint
+for an Edge Gateway to exist might be required. (See the [example](#example-1).)
+
+<a id="ip-prefix-quota"></a>
+
+## ip_prefix_quota block
+
+* `prefix_length` - (Required) - Prefix length for which the quota should be set (must be an
+  existing prefix length in parent IP Space)
+* `quota` - (Required) - Quota value for specific *prefix_length*
```
adambarreiro, line 63:
```suggestion
* `prefix_length` - (Required) Prefix length for which the quota should be set (must be an
  existing prefix length in parent IP Space)
* `quota` - (Required) Quota value for specific *prefix_length*
```

website/docs/r/ip_space_ip_allocation.html.markdown
```diff
@@ -0,0 +1,153 @@
...
+
+~> IP Allocation resources can be created only if there is a NSX-T Edge Gateway
+(`vcd_nsxt_edgegateway`) that is backed by the Provider Gateway (`vcd_external_network_v2`) with IP
+Space Uplinks (`vcd_ip_space_uplink`). Attempting to allocate IP Addresses before having an
+Edge Gateway withing VDC will return errors of type `This operation is denied`.
+
+## Attribute Reference
+
+* `ip_address` - contains either single IP when `type=FLOATING_IP` (e.g. `192.168.1.100`) or subnet
+  in CIDR format when `type=IP_PREFIX` (e.g. `192.168.1.100/30`). **Note** Terraform built-in
```
adambarreiro, line 121:
```suggestion
  in CIDR format when `type=IP_PREFIX` (e.g. `192.168.1.100/30`). **Note:** Terraform built-in
```

adambarreiro (commented):
Small things in docs. Now it is time for me to play a bit with the new resources 😄

website/docs/r/ip_space_ip_allocation.html.markdown
Didainius: Fixed.

website/docs/r/ip_space_custom_quota.html.markdown
Didainius: Fixed in 6 places.

website/docs/r/ip_space_custom_quota.html.markdown
Didainius: Fixed in 2 places.

website/docs/r/external_network_v2.html.markdown
Didainius: Yes. I moved single line accidentally it looks. Fixed.

website/docs/guides/ip_spaces.html.markdown
Didainius: fixed.

dataclouder (approved):

vcd/resource_vcd_ip_space.go
lvirbalas: Is there any way to avoid the sleep and wait for the needed state? :)

website/docs/guides/ip_spaces.html.markdown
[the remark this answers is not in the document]
adezxc: I like it now, maybe one last request, would it be too verbose to say what exactly did the IP space do/use? The introduction chapter is really nice and explains it really well, but some more comments in the snippet, like `# This allocates a single IP address from the IP space`, also IMO it would represent the real world scenario better.

vcd/remove_leftovers_test.go
```diff
@@ -501,16 +539,30 @@ func removeLeftoversNsxtAlbTenant(govcdClient *govcd.VCDClient, verbose bool) er
 	return nil
 }
 
-func removeLeftoversIpSpaces(govcdClient *govcd.VCDClient, verbose bool) error {
+func removeLeftoversIpSpacesAndAllocations(govcdClient *govcd.VCDClient, verbose bool) error {
 	allIpSpaces, err := govcdClient.GetAllIpSpaceSummaries(nil)
 	if err != nil {
 		return fmt.Errorf("error retrieving IP spaces")
```
dataclouder, line 545:
```suggestion
		return fmt.Errorf("error retrieving IP spaces: %s", err)
```

website/docs/guides/ip_spaces.html.markdown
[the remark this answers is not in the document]
Didainius: Added more details

vcd/remove_leftovers_test.go
Didainius: fixed

website/docs/guides/ip_spaces.html.markdown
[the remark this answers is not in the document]
adezxc: Thanks!

adambarreiro (approved):
LGTM! The guide looks really nice.

Issue #1058, opened 2023-05-20
Unable to use data.vcd_nsxt_app_port_profile


lpiob:
The data `data.vcd_nsxt_app_port_profile` is unusable for fetching system port profiles.
It returns an error every time.

```
│ Error: Expected exactly one NSX-T Application Port Profile. Got '3'
```

Quick debug confirms that in fact the API Request returned 3 responses. There is no way to filter them.

### Terraform Version
```
Terraform v1.4.6
on linux_amd64
+ provider registry.terraform.io/vmware/vcd v3.9.0
```

### Affected Resource(s)
- data.vcd_nsxt_app_port_profile

### Terraform Configuration Files

Whole code below:

```hcl
provider "vcd" {
  auth_type            = "integrated"
  org                  = "<removed>"
  vdc                  = "<removed>"
  url                  = "https://<removed>/api"
}

data "vcd_nsxt_app_port_profile" "custom" {
  name       = "SSH"
  scope      = "SYSTEM"
}
```

VCD_USER and VCD_PASSWORD are defined within environment variables. Other resources work correctly.

### Actual Behavior
```
│ Error: Expected exactly one NSX-T Application Port Profile. Got '3'
```

### Expected Behavior
One port profile should be returned OR I should be able to filter which one I'll get.

### Debug Output

I'm uncomfortable with sharing the debug output, as it may contain sensitive information.

I'll do so if requested.

The TF_LOG_TRACE shows that the API call returns 3 different `urn:vcloud:applicationPortProfile:uuid` values. When copied manually, only one of them is correct, the rest return the following error:
```
Application Port Profile <uuid-removed> does not exist or is invalid for Firewall Rules Allow selected outbond traffic.
```

The web interface shows only one.

### Steps to Reproduce
1. `terraform apply`

### User Access rights 
Not sure, these are my first steps with VCD envuronment. I manage an organization in a commercial cluster.

### Important Factoids
None.

### References
None.

Didainius:
Hello @lpiob ,
I saw you can't provide full log, but could you find API response in function `GetAllNsxtAppPortProfiles` (should be searchable in the log) and share only it?

It should be possible to narrow the query down, but I'd like to see these 3 responses you get.

lpiob:
@Didainius please see https://gist.github.com/lpiob/26b31115a87bcb89a4a6aeebc35dbf58

When I try to use any of the returned ids:
[...]

PR #1065, opened 2023-06-09
Fix vcd_nsxt_app_port_profile data source bug on multiple NSX-T managers

adezxc:
This PR fixes the bug mentioned in #1058. 

Before the fix, if a user tried to fetch a `SYSTEM` App Port Profile using the `vcd_nsxt_app_port_profile` data source and had more than one NSX-T Manager configured in the infrastructure, the operation would return an error as there would be duplicate entries from every NSX-T Manager. 

Fixed by providing `_context` on `SYSTEM` scope requests, so the NSX-T Managers also get filtered.

### DIFF
--- a/vcd/datasource_vcd_nsxt_app_port_profile.go
+++ b/vcd/datasource_vcd_nsxt_app_port_profile.go
@@ -58,7 +58,7 @@ func datasourceVcdNsxtAppPortProfile() *schema.Resource {
 				Type:          schema.TypeString,
 				Optional:      true,
 				Computed:      true,
-				Description:   "ID of VDC, VDC Group, or NSX-T Manager. Required if the VCD instance has more than one NSX-T manage",
+				Description:   "ID of VDC, VDC Group, or NSX-T Manager. Required if the VCD instance has more than one NSX-T manager",
 				ConflictsWith: []string{"nsxt_manager_id", "vdc"},
 			},
 			"scope": {
@@ -129,6 +129,8 @@ func datasourceVcdNsxtAppPortProfileRead(_ context.Context, d *schema.ResourceDa
 		contextId = ""
 	}
 
+	// If contextId is unset, send a request without _context query filter,
+	// Works properly only with SYSTEM scope and one NSX-T Manager configured.
 	if contextId != "" {
 		queryParams.Add("filter", fmt.Sprintf("name==%s;scope==%s;_context==%s", name, scope, contextId))
 	} else {
--- a/vcd/datasource_vcd_nsxt_app_port_profile.go
+++ b/vcd/datasource_vcd_nsxt_app_port_profile.go
@@ -4,13 +4,11 @@ import (
 	"context"
 	"fmt"
 	"net/url"
-	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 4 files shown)


vcd/datasource_vcd_nsxt_app_port_profile.go
```diff
@@ -55,11 +53,12 @@ func datasourceVcdNsxtAppPortProfile() *schema.Resource {
 				Required:    true,
 				Description: "Application Port Profile name",
 			},
+			// TODO V4 Change this to required
 			"context_id": {
 				Type:          schema.TypeString,
 				Optional:      true,
 				Computed:      true,
-				Description:   "ID of VDC, VDC Group, or NSX-T Manager",
+				Description:   "ID of VDC, VDC Group, or NSX-T Manager. Required if the VCD instance has more than one NSX-T manage",
```
lvirbalas, line 61: Last letter missing in the quotes :)

vcd/datasource_vcd_nsxt_app_port_profile.go
```diff
@@ -105,28 +105,33 @@ func datasourceVcdNsxtAppPortProfileRead(_ context.Context, d *schema.ResourceDa
...
-		// For "SYSTEM" or "PROVIDER" scoped Application Port Profiles context can be ignored.
-		// * For "SYSTEM" this is correct behavior
-		// * For "PROVIDER" it can match App Port Profiles when multiple NSX-T Managers are
-		// configured, but this is left for backwards compatibility
-		//
-		// TODO V4 - remove support for PROVIDER scope without `context_id` field
+		contextId = ""
+	}
+
+	if contextId != "" {
```
lvirbalas, line 132: Maybe add a small comment about this use case? Something in the lines of what you've added to docs' field description:
https://github.com/vmware/terraform-provider-vcd/pull/1065/files#diff-38a1cf12478b001e028f2fd38ecb2416dabe92ecd3328ee59f702407b387e9baR70

lvirbalas (approved):

adambarreiro (approved):

vcd/datasource_vcd_nsxt_app_port_profile.go
adezxc: Fixed.

vcd/datasource_vcd_nsxt_app_port_profile.go
adezxc: Updated.

vcd/datasource_vcd_nsxt_app_port_profile_test.go
```diff
@@ -49,6 +49,16 @@ func TestAccVcdNsxtAppPortProfileDsSystem(t *testing.T) {
 	postTestChecks(t)
 }
 
+const testAccVcdNsxtAppPortProfileSystemDSStep1 = `
```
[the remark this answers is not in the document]
adezxc: I think I made some changes but didn't commit them. I chose this to decouple the constants, even though maybe I could create a common one.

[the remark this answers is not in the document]
adezxc: Removed, thanks for pointing it out!

Didainius (approved):
Thanks for great investigation and fix!

dataclouder (approved):

Issue #1069, opened 2023-06-16
Changes to VPN Tunnels with "vcd_nsxt_ipsec_vpn_tunnel" sets "security_profile_customization" back to default


fgrehl:
Hello,

when an existing IPsec Tunnel with "security_profile_customization" configured is changed (eg. A Network is added to remote_networks), the Security Customizations are set back to "Default". A second `terraform apply` solves the problem.

### Terraform Version
Terraform v1.5.0
vmware/vcd v3.9.0 + 3.7.0

### Affected Resource(s)
- vcd_nsxt_ipsec_vpn_tunnel

### Terraform Configuration Files

```hcl
resource "vcd_nsxt_ipsec_vpn_tunnel" "ipsec" {
  edge_gateway_id = data.vcd_nsxt_edgegateway.egw.id
  name            = "VPN"

  pre_shared_key    = "xxxxxxxxxx"
  local_ip_address  = "1.2.3.4"
  local_networks    = ["10.10.10.0/24", "30.30.30.0/28"]
  remote_ip_address = "5.6.7.8"
  remote_networks   = ["192.168.1.0/24", "192.168.10.0/24"]

  security_profile_customization {
    ike_version                  = "IKE_V2"
    ike_encryption_algorithms    = ["AES_128"]
    ike_digest_algorithms        = ["SHA2_256"]
    ike_dh_groups                = ["GROUP14"]
    ike_sa_lifetime              = 86400
    tunnel_pfs_enabled           = true
    tunnel_df_policy             = "COPY"
    tunnel_encryption_algorithms = ["AES_256"]
    tunnel_digest_algorithms     = ["SHA2_256"]
    tunnel_dh_groups             = ["GROUP14"]
    tunnel_sa_lifetime           = 3600
    dpd_probe_internal           = "30"
  }
}

```

### Expected Behavior
Security Customizations should not be set to "Default" when making changes to the Tunnel configuration.

### Actual Behavior
Two `terraform apply` are required to make changes to IPsec Tunnels with Security Customizations

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Add Network remote_networks
2. `terraform apply` - This will Add the network but set Profile back to "Default"
3. Verify Tunnel Status in VCD UI - Tunnel Security Profile is configured to "Default"
4. `terraform apply` - This will notice that the profile is wrong and fix it

adezxc:
Hi @fgrehl, thanks for submissing this issue

[...]

PR #1072, opened 2023-06-20
Fix IPSec VPN Tunnel resource Update

adezxc:
This PR fixes the bug described in #1069.

The workflow of "update the ipsec vpn profile, but not the security settings" was untested, so it would set to default
every time there was a change in the main schema, but not `security_profile_customization`

### DIFF
--- a/vcd/resource_vcd_nsxt_ipsec_vpn_tunnel.go
+++ b/vcd/resource_vcd_nsxt_ipsec_vpn_tunnel.go
@@ -290,9 +290,13 @@ func resourceVcdNsxtIpSecVpnTunnelCreate(ctx context.Context, d *schema.Resource
 	// Check if Tunnel Profile has custom settings and apply them
 	tunnelProfileConfig := getNsxtIpSecVpnProfileTunnelConfigurationType(d)
 
-	_, err = createdIpSecVpnConfig.UpdateTunnelConnectionProperties(tunnelProfileConfig)
-	if err != nil {
-		return diag.Errorf("[nsx-t ipsec vpn tunnel create] error setting VPN Tunnel Profile: %s", err)
+	// If `security_profile_customization` is defined, update the Ipsec VPN tunnel with the new profile
+	_, customSecurityProfile := d.GetOk("security_profile_customization")
+	if customSecurityProfile {
+		_, err = createdIpSecVpnConfig.UpdateTunnelConnectionProperties(tunnelProfileConfig)
+		if err != nil {
+			return diag.Errorf("[nsx-t ipsec vpn tunnel create] error setting VPN Tunnel Profile: %s", err)
+		}
 	}
 
 	return resourceVcdNsxtIpSecVpnTunnelRead(ctx, d, meta)
@@ -328,10 +332,9 @@ func resourceVcdNsxtIpSecVpnTunnelUpdate(ctx context.Context, d *schema.Resource
 	// Inject ID for update
 	ipSecVpnConfig.ID = d.Id()
 
-	_, isSet := d.GetOk("security_profile_customization")
-
 	// If unset, it will set it back to default, so before updating the VPN config, we need to check if the `security_profile_customization`
 	// field is defined in the configuration
+	_, isSet := d.GetOk("security_profile_customization")
 	if isSet {
 		ipSecVpnConfig.SecurityType = "CUSTOM"
 	} else {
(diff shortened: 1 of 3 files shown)


vcd/resource_vcd_nsxt_ipsec_vpn_tunnel.go
```diff
@@ -334,39 +328,29 @@ func resourceVcdNsxtIpSecVpnTunnelUpdate(ctx context.Context, d *schema.Resource
...
 
-	// At first update IPsec VPN tunnel configuration
-	// It will reset Security Profile to DEFAULT at the same shot if no customization exists in 'security_profile_customization'
 	updatedIpSecVpnConfiguration, err := existingIpSecVpnConfiguration.Update(ipSecVpnConfig)
 	if err != nil {
 		return diag.Errorf("[nsx-t ipsec vpn tunnel update] error updating NSX-T IPsec VPN Tunnel configuration '%s': %s", ipSecVpnConfig.Name, err)
 	}
 
-	// If Security Profile has change and it is being customized
-	if securityProfileHasChange && newSecurityProfile != nil {
```
adambarreiro, line 357: So if I understood, this condition is the culprit as `existingIpSecVpnConfiguration.Update(ipSecVpnConfig)` restarts the security profile config?

adezxc: Yes. Basically, if there was no change to the security profile, only to the vpn tunnel itself, it would update the tunnel with a blank `SecurityType` which resulted in it getting reset to `DEFAULT`

adambarreiro: Thanks for your answer!

vcd/resource_vcd_nsxt_ipsec_vpn_tunnel.go
```diff
@@ -288,17 +288,11 @@ func resourceVcdNsxtIpSecVpnTunnelCreate(ctx context.Context, d *schema.Resource
...
-
-		_, err = createdIpSecVpnConfig.UpdateTunnelConnectionProperties(tunnelProfileConfig)
-		if err != nil {
-			return diag.Errorf("[nsx-t ipsec vpn tunnel create] error setting VPN Tunnel Profile: %s", err)
-		}
+	tunnelProfileConfig := getNsxtIpSecVpnProfileTunnelConfigurationType(d)
 
+	_, err = createdIpSecVpnConfig.UpdateTunnelConnectionProperties(tunnelProfileConfig)
+	if err != nil {
+		return diag.Errorf("[nsx-t ipsec vpn tunnel create] error setting VPN Tunnel Profile: %s", err)
```
Didainius, line 295: Should we not trigger the `TunnelConnectionProperties` if we can reliably identify that it is not set? (using `_, isSet := d.GetOk("security_profile_customization")` like in https://github.com/vmware/terraform-provider-vcd/pull/1072/files#diff-aadc80442709e37c8a86461fb33847073473b847e213ecdfb209f1c0a036eb06R331

That way we could avoid unnecessary API call at all

vcd/resource_vcd_nsxt_ipsec_vpn_tunnel.go
```diff
@@ -334,39 +328,29 @@ func resourceVcdNsxtIpSecVpnTunnelUpdate(ctx context.Context, d *schema.Resource
...
-		if ipSecTunnelProfileConfig != nil {
-			_, err = updatedIpSecVpnConfiguration.UpdateTunnelConnectionProperties(ipSecTunnelProfileConfig)
-			if err != nil {
-				return diag.Errorf("[nsx-t ipsec vpn tunnel update] error updating NSX-T IPsec VPN Tunnel Security Profile: %s", err)
-			}
+	// Get current Security Tunnel profile, if it's "DEFAULT" in VCD, returns nil
+	ipSecTunnelProfileConfig := getNsxtIpSecVpnProfileTunnelConfigurationType(d)
 
+	// To set IPsec VPN Tunnel Connection Profile - it must be updated (HTTP PUT) with all the options configured
+	if ipSecTunnelProfileConfig != nil {
```
Didainius, line 350: Wouldn't the same check `_, isSet := d.GetOk("security_profile_customization")` as in `https://github.com/vmware/terraform-provider-vcd/pull/1072/files#diff-aadc80442709e37c8a86461fb33847073473b847e213ecdfb209f1c0a036eb06R331 be simpler to read?

Didainius (approved):
Thanks for good test improvement (I tried it without the fix to see that the problem is caught now).

Missing changelog record:)

A few more inline thoughts/comments, but not insisting on it.

lvirbalas (changes requested):
Changelog is missing :)
And thanks for the tests!

vcd/resource_vcd_nsxt_ipsec_vpn_tunnel.go
adezxc: Good call. Updated.

vcd/resource_vcd_nsxt_ipsec_vpn_tunnel.go
adezxc: Updated here aswell. Changed the variables so they're more clear aswell.

adezxc:
> Changelog is missing :) And thanks for the tests!

Added :)

lvirbalas (approved):

dataclouder (approved):

dataclouder (approved):

adambarreiro (approved):

PR #1074, opened 2023-06-28
Add Behavior support to RDE Interfaces and RDE Types

adambarreiro:
## Overview

This PR complements #965, #973 and #977 as it adds support to create, read, update and delete Behaviors, both in RDE Interfaces and Types.

This PR requires https://github.com/vmware/go-vcloud-director/pull/584.

## Context

Behaviors are explained [here](https://docs.vmware.com/en/VMware-Cloud-Director/10.4/VMware-Cloud-Director-Service-Provider-Admin-Portal-Guide/GUID-0749DEA0-08A2-4F32-BDD7-D16869578F96.html), but in simplified words, they're "procedures" that belong to RDE Interfaces, are inherited by the RDE Types that use those Interfaces and are invoked on the RDEs that instantiate those types.

Also, Behaviors can be overridden at the RDE Type level. For example, if you use a RDE Interface with behavior "A", you can override what it does in a Type that uses it.

Last but not least, RDE Type Behaviors have "Access Levels", which can define what operations can be done in that behavior.

## Description

This PR adds 3 new resources and 3 new data sources:

* `vcd_rde_interface_behavior` resource to manage RDE Interface Behaviors, which can be invoked by RDEs and  overridden by RDE Types.
* `vcd_rde_interface_behavior` data source to read RDE Interface Behaviors, so they can be used
  in RDE Type Behavior overrides.
* `vcd_rde_type_behavior` resource to manage Behaviors in RDE Types, which can override those defined
  in RDE Interfaces.
* `vcd_rde_type_behavior` data source to read RDE Type Behaviors.
* `vcd_rde_type_behavior_acl` resource to manage the access to Behaviors in RDE Types and RDE Interfaces
* `vcd_rde_type_behavior_acl` data source to read Access Levels from Behaviors of RDE Types and RDE Interfaces

## Caveats

* Behaviors of a RDE Interface must be created before any RDE Type that uses that Interface is created. For that reason, if one uses `vcd_rde_interface_behavior`, she must be sure that the `vcd_rde_type` resources wait for the behaviors to be created with a `depends_on`. This is explained in `vcd_rde_type` resource and `vcd_rde_interface_behavior` resource documentation.

* `vcd_rde_type_behavior_acl` could have been a Set inside `vcd_rde_type_behavior` but it has some peculiar features that made it not possible:
  * An Access Level can be defined for a RDE Type that didn't override a Interface Behavior
  * For this reason one can use RDE Type Behavior IDs but also RDE Interface Behavior IDs.

* This PR does not implement RDE Type `Hooks`.  As one Hook is a map of keys `PostCreate`, `PostUpdate`, `PreDelete` and `PostDelete` and values <Behavior ID>, I believe the best option is yet another resource, otherwise we'd have a circular dependency RDE Type (needs behavior IDs) <-> RDE Type Behavior (needs RDE Type IDs). Probably I'll do in a separate PR


website/docs/r/rde_interface_behavior.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+* `interface_id` - (Required) The ID of the RDE Interface that owns the Behavior
+* `name` - (Required) Name of the Behavior
+* `description` - (Optional) A description specifying the contract of the Behavior
+* `execution` - (Required) A map that specifies the Behavior execution mechanism.
+  You can find more information about the different execution types, like `WebHook`, `noop`, `Activity`, `MQTT`, `VRO`, `AWSLambdaFaaS`
+  and others [in the Extensibility SDK documentation](https://vmware.github.io/vcd-ext-sdk/docs/defined_entities_api/behaviors)
+
+## Attribute Reference
+
+* `ref` - The Behavior invocation reference to be used for polymorphic behavior invocations
```
Didainius, line 72: Would this actually mean that this `ref` can be used with `vcd_rde_type`?

website/docs/r/rde_type_behavior.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+* `rde_interface_behavior_id` - (Required) The ID of the [RDE Interface Behavior](/providers/vmware/vcd/latest/docs/resources/rde_interface_behavior) to override
+* `description` - (Optional) The description of the RDE Type Behavior.
+* `execution` - (Required) A map that specifies the Behavior execution mechanism.
+  You can find more information about the different execution types, like `WebHook`, `noop`, `Activity`, `MQTT`, `VRO`, `AWSLambdaFaaS`
+  and others [in the Extensibility SDK documentation](https://vmware.github.io/vcd-ext-sdk/docs/defined_entities_api/behaviors)
+
+## Attribute Reference
+
+* `name` - Name of the overridden Behavior
+* `ref` - The Behavior invocation reference to be used for polymorphic behavior invocations
```
Didainius, line 72: Similar query to there https://github.com/vmware/terraform-provider-vcd/pull/1074/files#diff-8e094640ce4dd5e7da41399262702d7129f7443f9db76f8db95fcbd7d928cdf0R72

Would this not be a polymorphic behavior anymore as it is on exact `vcd_rde_type`?

website/docs/d/rde_type_behavior.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+layout: "vcd"
+page_title: "VMware Cloud Director: vcd_rde_type_behavior"
+sidebar_current: "docs-vcd-data-source-rde-type-behavior"
+description: |-
+  Provides the capability of fetching the Behavior overrides from an existing RDE Type in VMware Cloud Director.
+---
+
+# vcd\_rde\_type\_behavior
+
+Provides the capability of fetching the Behavior overrides from an existing RDE Type in VMware Cloud Director.
```
lvirbalas, line 11: The "Behavior overrides" sounds strange. Is there a way to rephrase the sentence? For reference, this one is clear:
https://github.com/vmware/terraform-provider-vcd/pull/1074/files#diff-0269665e1810aefda2c551dbe003dcfa3d3ddc77091875474b3e5cf905a25bd5R11

website/docs/d/rde_type_behavior.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+output "execution_id" {
+  value = data.vcd_rde_type_behavior.my_behavior.execution.id
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `rde_type_id` - (Required) The ID of the [RDE Type](/providers/vmware/vcd/latest/docs/data-sources/rde_type) that owns the Behavior override to fetch
```
lvirbalas, line 49: Seems extra?

```suggestion
* `rde_type_id` - (Required) The ID of the [RDE Type](/providers/vmware/vcd/latest/docs/data-sources/rde_type) that owns the Behavior override
```

lvirbalas (commented):
Awesome PR! Only few suggestions for docs.

website/docs/r/rde_type_behavior.html.markdown
adambarreiro: In this case, `ref` would be the Interface Behavior ID. It is like the "original" reference. I can rephrase the sentence, as I just used the official one.

website/docs/r/rde_interface_behavior.html.markdown
adambarreiro: For Interfaces I believe that `ref` and `ID` will be always the same. This field probably makes more sense in the API as one can do more of an "swap" easier, between Interface and Type behaviors.

vcd/provider.go
```diff
@@ -124,6 +124,9 @@ var globalDataSourceMap = map[string]*schema.Resource{
 	"vcd_nsxt_edgegateway_dhcp_forwarding":          datasourceVcdNsxtEdgegatewayDhcpForwarding(),    // 3.10
 	"vcd_org_saml":                                  datasourceVcdOrgSaml(),                          // 3.10
 	"vcd_org_saml_metadata":                         datasourceVcdOrgSamlMetadata(),                  // 3.10
+	"vcd_rde_interface_behavior":                    datasourceVcdRdeInterfaceBehavior(),             // 3.10
+	"vcd_rde_type_behavior":                         datasourceVcdRdeTypeBehavior(),                  // 3.10
+	"vcd_rde_type_behavior_acl":                     datasourceVcdRdeTypeBehaviorAccessLevel(),       // 3.10,
```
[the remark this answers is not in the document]
adambarreiro: Fixed, thanks!

website/docs/d/rde_type_behavior.html.markdown
adambarreiro: Fixed!

website/docs/d/rde_type_behavior.html.markdown
adambarreiro: Fixed!

lvirbalas (approved):

dataclouder (approved):

Didainius (approved):

adezxc (approved):
Approved, as Dainius said this is very meta, we might need a refresher on RDEs once again :D

PR #1076, opened 2023-07-03
DFW rule management resource vcd_nsxt_distributed_firewall_rule

### LINKED ISSUE #970: vcd_nsxt_distributed_firewall_rule resource
### Community Note

### Description
Actually, the DFW ( Distributed Firewall )  resource **vcd_nsxt_distributed_firewall** consider the Firewall as whole, all the rule are managed  within one ressource by listing all the rules...once we update the first rule on the Firewall, all the rules will change ID's and are recreated, when checking the API, it seems that a rule can be managed a part, the request is to add a new resource that manage specific rule by name 

example: 

```hcl
resource "vcd_nsxt_distributed_firewall_rule " "my-rule-1" {
  org          = "my-org"
  vdc          = "my-vdc"
  edge_gateway = "my-edge-gateway"

  source {
    ip_sets = [vcd_ipset.test-ipset_src.name]
  }

  destination {
    dynamic_groups= [vcd_dynamic_group.test-dg1.name]
   ip_sets = [vcd_ipset.test-ipset_dest.name]
  }

  service = ["ap_profile1","ap_profile2",]
}
```

Link to the API : 
https://developer.vmware.com/apis/vmware-cloud-director/latest/cloudapi/1.0.0/vdcGroups/vdcGroupId/dfwPolicies/policyId/rules/ruleId/put/

Didainius:
Closes #970

This PR introduces a new resource and data source `vcd_nsxt_distributed_firewall_rule` to manage Distributed firewall rules one by one. We already have a resource `vcd_nsxt_distributed_firewall`, but that one manages all rules at once and there is a request  #970 to manage rules one by one.

#### Notes

* There is no API to create a single distributed firewall rule, so this functionality is handled in Go SDK for VCD `VdcGroup.CreateDistributedFirewallRule`. As a consequence - firewall rules will not be created in parallel, but instead one by one (as the resource will have locks).
* There is an `above_rule_id` field in the new resource to handle the firewall rule position during create operation.
* By default, a distributed firewall gets one default rule created after it is activated in resource `vcd_vdc_group`. This would make it very inconvenient to work with firewall rules using new resources (although possible by using data source and then field `above_rule_id` to put all firewall rules above). To make things more convenient, `vcd_vdc_group` resource has a new flag `remove_default_firewall_rule` that would remove the default firewall rule.

Note for reviewers - the "internals" of resource are very similar to `vcd_nsxt_distributed_firewall`. There difference is here we don't have element nesting in a `TypeSet`

### Patch for API V38.0
* `vcd_nsxt_app_port_profile` has a patch for similar diff errors in commit [38c907b](https://github.com/vmware/terraform-provider-vcd/pull/1076/commits/38c907b82172b1348425d238e46d2947b6984048):
```
~ resource "vcd_nsxt_app_port_profile" "custom-org1" {
   id          = "urn:vcloud:applicationPortProfile:fbcf286a-4a33-40c7-8312-c5a71449a54f"
   name        = "custom_app_prof"
   # (4 unchanged attributes hidden)

 - app_port {
     - port     = [
         - "any",
       ] -> null
     - protocol = "ICMPv4" -> null
   }
 + app_port {
     + port     = []
     + protocol = "ICMPv4"
   }
```


.changes/v3.10.0/1076-features.md
```diff
@@ -0,0 +1,4 @@
+* **New Resource:** `vcd_nsxt_distributed_firewall_rule` to manage NSX-T Distributed Firewall one by
```
lvirbalas, line 1: I feel like we should state that this is above API functionality here too.

vcd/datasource_vcd_nsxt_distributed_firewall_rule.go
```diff
@@ -0,0 +1,140 @@
...
+			},
+			"comment": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Comment that is shown next to rule in UI (VCD 10.3.2+)",
+			},
+			"direction": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Direction on which Firewall Rule applies (One of 'IN', 'OUT', 'IN_OUT')",
```
lvirbalas, line 46:
```suggestion
				Description: "Direction on which Firewall Rule applies (one of 'IN', 'OUT', 'IN_OUT')",
```

vcd/datasource_vcd_nsxt_distributed_firewall_rule.go
```diff
@@ -0,0 +1,140 @@
...
+			},
+			"ip_protocol": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Firewall Rule Protocol (One of 'IPV4', 'IPV6', 'IPV4_IPV6')",
```
lvirbalas, line 51:
```suggestion
				Description: "Firewall Rule Protocol (one of 'IPV4', 'IPV6', 'IPV4_IPV6')",
```

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
```diff
@@ -0,0 +1,395 @@
...
+			},
+			"comment": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Comment that is shown next to rule in UI (VCD 10.3.2+)",
+			},
+			"direction": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "Direction on which Firewall Rule applies (One of 'IN', 'OUT', 'IN_OUT')",
```
lvirbalas, line 64:
```suggestion
				Description:  "Direction on which Firewall Rule applies (one of 'IN', 'OUT', 'IN_OUT')",
```

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
```diff
@@ -0,0 +1,395 @@
...
+				Default:      "IN_OUT",
+				ValidateFunc: validation.StringInSlice([]string{"IN", "OUT", "IN_OUT"}, false),
+			},
+			"ip_protocol": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "Firewall Rule Protocol (One of 'IPV4', 'IPV6', 'IPV4_IPV6')",
```
lvirbalas, line 71:
```suggestion
				Description:  "Firewall Rule Protocol (one of 'IPV4', 'IPV6', 'IPV4_IPV6')",
```

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+func resourceVcdNsxtDistributedFirewallRuleCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	vcdClient := meta.(*VCDClient)
+	vcdClient.lockParentVdcGroup(d)
+	defer vcdClient.unlockParentVdcGroup(d)
+
+	org, err := vcdClient.GetOrgFromResource(d)
+	if err != nil {
+		return diag.Errorf("[Distributed Firewall Rule create] error retriving Org: %s", err)
```
lvirbalas, line 148:
```suggestion
		return diag.Errorf("[Distributed Firewall Rule create] error retrieving Org: %s", err)
```

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+func resourceVcdNsxtDistributedFirewallRuleUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	vcdClient := meta.(*VCDClient)
+	vcdClient.lockParentVdcGroup(d)
+	defer vcdClient.unlockParentVdcGroup(d)
+
+	org, err := vcdClient.GetOrgFromResource(d)
+	if err != nil {
+		return diag.Errorf("[Distributed Firewall Rule update] error retriving Org: %s", err)
```
lvirbalas, line 177:
```suggestion
		return diag.Errorf("[Distributed Firewall Rule update] error retrieving Org: %s", err)
```

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
```diff
@@ -0,0 +1,395 @@
...
+
+	return resourceVcdNsxtDistributedFirewallRuleRead(ctx, d, meta)
+}
+
+func resourceVcdNsxtDistributedFirewallRuleRead(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	vcdClient := meta.(*VCDClient)
+
+	org, err := vcdClient.GetOrgFromResource(d)
+	if err != nil {
+		return diag.Errorf("[Distributed Firewall Rule read] error retriving Org: %s", err)
```
lvirbalas, line 209:
```suggestion
		return diag.Errorf("[Distributed Firewall Rule read] error retrieving Org: %s", err)
```

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+func resourceVcdNsxtDistributedFirewallRuleDelete(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	vcdClient := meta.(*VCDClient)
+	vcdClient.lockParentVdcGroup(d)
+	defer vcdClient.unlockParentVdcGroup(d)
+
+	org, err := vcdClient.GetOrgFromResource(d)
+	if err != nil {
+		return diag.Errorf("[Distributed Firewall Rule delete] error retriving Org: %s", err)
```
lvirbalas, line 240:
```suggestion
		return diag.Errorf("[Distributed Firewall Rule delete] error retrieving Org: %s", err)
```

website/docs/d/nsxt_distributed_firewall_rule.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+layout: "vcd"
+page_title: "VMware Cloud Director: vcd_nsxt_distributed_firewall_rule"
+sidebar_current: "docs-vcd-data-source-nsxt-distributed-firewall-rule"
+description: |-
+  The Distributed Firewall data source a single rules for a particular VDC Group.
+---
+
+# vcd\_nsxt\_distributed\_firewall\_rule
+
+The Distributed Firewall data source reads all defined rules for a particular VDC Group.
```
lvirbalas, line 11: I'm a bit confused here with the word "all". I thought this is for reading a single rule, as the `name` ir required?

website/docs/r/nsxt_distributed_firewall.html.markdown
```diff
@@ -12,6 +12,12 @@ description: |-
 The Distributed Firewall allows user to segment organization virtual data center entities, such as
 virtual machines, based on virtual machine names and attributes. 
 
+~> There is a different resource
```
lvirbalas, line 15: In addition to "yellow" and "blue" notes, is there a "red" note/warning? :) It's so critical not to mix the two ways of using (as it can lead to data corruption), that I'd like to highlight this more.

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
```diff
@@ -0,0 +1,117 @@
...
+  The Distributed Firewall allows user to segment organization virtual data center entities, such as
+  virtual machines, based on virtual machine names and attributes. 
+---
+
+# vcd\_nsxt\_distributed\_firewall\_rule
+
+The Distributed Firewall allows user to segment organization virtual data center entities, such as
+virtual machines, based on virtual machine names and attributes. 
+
+~> There is a different resource
```
lvirbalas, line 15: Also, feels like this should be in "red", if possible.

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
```diff
@@ -0,0 +1,117 @@
...
+  source_ids           = [vcd_nsxt_ip_set.set1.id, vcd_nsxt_ip_set.set2.id]
+  destination_ids      = [vcd_nsxt_security_group.g1-empty.id, vcd_nsxt_security_group.g2.id]
+  app_port_profile_ids = [vcd_nsxt_app_port_profile.p1.id, data.vcd_nsxt_app_port_profile.WINS.id, data.vcd_nsxt_app_port_profile.FTP.id]
+}
+
+resource "vcd_nsxt_distributed_firewall_rule" "r2" {
+  org          = "{{.Org}}"
+  vdc_group_id = vcd_vdc_group.test1.id
+
+  # Specifying a particular ID of other firewall rule will ensure that the current one is placed above
```
lvirbalas, line 52: I think this is in reverse. Current rule will go below the above_rule_id, won't it?

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
```diff
@@ -0,0 +1,117 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful
+  when connected as sysadmin working across different organisations.
+* `above_rule_id` - (Optional) ID of another `vcd_nsxt_distributed_firewall_rule` entry will ensure
+  that newly created rule is positioned above the references rule. **Note.** By default, new rule
```
lvirbalas, line 69: Oh, there seems to be some confusion here with the naming. I thought that `above_rule_id` means a pointer to the rule **above** the current one, but it seems it's not. In which case, I feel we should make the current name more deterministic (I know, it's a hard naming challenge)...

CC: @dataclouder @adambarreiro 

Also, small typo:

```suggestion
  that newly created rule is positioned above the referenced rule. **Note.** By default, new rule
```

lvirbalas (changes requested):
It is a magical feature! I have provided some "fresh eyes" thoughts.

vcd/datasource_vcd_nsxt_distributed_firewall_rule.go
Didainius: Fixed here and in resource.

vcd/datasource_vcd_nsxt_distributed_firewall_rule.go
Didainius: Fixed here and in resource.

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
Didainius: Fixed here and in data source.

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
Didainius: Fixed here and in data source.

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
Didainius: Fixed

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
Didainius: Actually, this came from other place (old DFW resource). I did a search and replace for all such occurrences.

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
Didainius: Actually, this came from other place (old DFW resource). I did a search and replace for all such occurrences.

vcd/resource_vcd_nsxt_distributed_firewall_rule.go
Didainius: Actually, this came from other place (old DFW resource). I did a search and replace for all such occurrences.

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
Didainius: No. The goal would be for it to go `above` the rule that is specified in `above_rule_id`. This is inspired by NSX-V rule, that had such field in API - https://registry.terraform.io/providers/vmware/vcd/latest/docs/resources/nsxv_firewall_rule#above_rule_id

It makes sense to put "above", because if it was places "below" - one would have no mechanism to place a new rule at the very top.

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
Didainius: Typo fixed.

As for the name itself - I chose it from NSX-V world, because this was what the NSX-V API had https://registry.terraform.io/providers/vmware/vcd/latest/docs/resources/nsxv_firewall_rule#above_rule_id

.changes/v3.10.0/1076-features.md
Didainius: Added extra line

website/docs/r/nsxt_distributed_firewall.html.markdown
Didainius: Changed. here is how it looks 
image

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
Didainius: image

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
lvirbalas: That's actually good, as it means we have lived with it and our users have survived the naming ambiguity. The issue is I can't think of an alternative (better) name yet :) ...

So, I'd like to suggest a simplifaction to the field description at least (changes in bold):

> ID of **an existing** `vcd_nsxt_distributed_firewall_rule` entry **above which**
>   **the** newly created **firewall** rule **will be positioned**.

website/docs/r/nsxt_distributed_firewall.html.markdown
lvirbalas: Serious, as it should be? To clarify: is it only "as they will override each other" or "that using both may lead to unexpected resulting rules (data corruption)"?

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
lvirbalas: p.s. I think we don't have to be fully consistent with NSX-V, so here are some alternative name suggestions for consideration. I would like at least to open the idea bucket. All with the goal of removing any ambiguity, which is especially useful when you're reading HCLs written by someone else without knowing the details. So:

1. `place_above_rule_id`
2. `position_above_rule_id`

CC: @dataclouder @adambarreiro

website/docs/r/nsxt_distributed_firewall.html.markdown
Didainius: More like the second case ("that using both may lead to unexpected resulting rules (data corruption)"). Using only multiple instances of `vcd_nsxt_distributed_firewall` would cause them to override each other.

lvirbalas: Then I'd suggest stating that. I mean, something like this:
"One should use only one of ... as using both may lead to unexpected resulting rules (data corruption)".

.changes/v3.10.0/1076-features.md
Didainius: Added to resource information as well.
image

website/docs/d/nsxt_distributed_firewall_rule.html.markdown
Didainius: Fixed.

website/docs/r/nsxt_distributed_firewall.html.markdown
Didainius: Adjusted. Or do you think we need to mention the "data corruption" explicitly?

image

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
Didainius: I have adjusted the proposed comment. Open to field name changes.

.changes/v3.10.0/1076-features.md
lvirbalas: Very useful! Here's a polishing suggestion. Changes highlighted or code-quoted:

> "Multiple rules defined with this resource will not be created in parallel, because **CD** API provides no direct endpoint to create a single rule. **To overcome this,** `vcd_nsxt_distributed_firewall_rule` **calls an "update all rules" API endpoint for each single rule, and this call is serialized to avoid data conflicts.**

website/docs/d/nsxt_distributed_firewall_rule.html.markdown
lvirbalas: There are some issues with the latest sentence :)

> The Distributed Firewall data source a single rules for a particular VDC Group.

->

> The Distributed Firewall data source **reads** a single **rule of** a particular VDC Group.

.changes/v3.10.0/1076-features.md
Didainius: I have patched and read this again. The only thing I changed in your quote is to expand CD to Cloud Director. It sounded a bit cryptic as we don't use this abbreviation too much in our docs.

website/docs/d/nsxt_distributed_firewall_rule.html.markdown
Didainius: Thank you. That wasn't good - reread both docs now :)

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
lvirbalas: A final call for comments on https://github.com/vmware/terraform-provider-vcd/pull/1076#discussion_r1252975276 from @dataclouder @adambarreiro @adezxc ^^ Which name do you prefer: leave `above_rule_id` or change to `place_above_rule_id` or `position_above_rule_id`?

lvirbalas (approved):
This looks great! The last open item I have is on naming https://github.com/vmware/terraform-provider-vcd/pull/1076#discussion_r1260887645 , but it will depend on input from a wider audience.

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
adambarreiro: IMHO, what I like from `above_rule_id` is that describes better the state, like "This rule is above the rule with ID", vs `place...` or `position...` which look more like an imperative action, "Place this rule above the rule with ID". As Terraform tends to describe a "desired state" vs actions, I'd prefer `above_rule_id` even though it may be confusing at first (I think the confusion gets cleared once the docs are read).

But let's see what others think :)

adambarreiro: It could also be that the confusion is generated by the fact that "above" in this context can mean "This rule is above the specified" or "The above rule is this one".
Maybe choosing a more exact word, like `in_preference_to_rule_id` or something similar.

lvirbalas: Yes, @adambarreiro that was exactly the confusion I hit at the start of review. To contrast, both "above rule ID" and "bellow rule ID" may be interpreted exactly the same in this context.

website/docs/d/nsxt_distributed_firewall_rule.html.markdown
```diff
@@ -0,0 +1,46 @@
+---
+layout: "vcd"
+page_title: "VMware Cloud Director: vcd_nsxt_distributed_firewall_rule"
+sidebar_current: "docs-vcd-data-source-nsxt-distributed-firewall-rule"
+description: |-
+  The Distributed Firewall data source a single rules for a particular VDC Group.
```
dataclouder, line 6:
```suggestion
  The Distributed Firewall data source reads a single rule for a particular VDC Group.
```

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
```diff
@@ -0,0 +1,122 @@
+---
+layout: "vcd"
+page_title: "VMware Cloud Director: vcd_nsxt_distributed_firewall_rule"
+sidebar_current: "docs-vcd-resource-nsxt-distributed-firewall-rule"
+description: |-
+  The Distributed Firewall allows user to segment organization virtual data center entities, such as
```
dataclouder, line 6: This description is adequate for the whole set of rules, but feels out of sort for a single rule.

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
```diff
@@ -0,0 +1,122 @@
...
+  virtual machines, based on virtual machine names and attributes. 
+---
+
+# vcd\_nsxt\_distributed\_firewall\_rule
+
+The Distributed Firewall allows user to segment organization virtual data center entities, such as
```
dataclouder, line 12: Same as the comment at line 6 above

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
adezxc: I think I prefer sticking with `above_rule_id`, all argument names will probably cause some kind of confusion, so docs will be needed anyway. I agree with @adambarreiro's point that `place_above_rule_id` is too explicit and not really declarative.

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
Didainius: Fixed.

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
Didainius: Fixed.

website/docs/d/nsxt_distributed_firewall_rule.html.markdown
Didainius: Fixed

vcd/resource_vcd_nsxt_distributed_firewall.go
```diff
@@ -157,7 +157,7 @@ func resourceVcdNsxtDistributedFirewallCreateUpdate(ctx context.Context, d *sche
 
 	org, err := vcdClient.GetOrgFromResource(d)
 	if err != nil {
-		return diag.Errorf("[Distributed Firewall Create/Update] error retriving Org: %s", err)
+		return diag.Errorf("[Distributed Firewall Create/Update] error retreiving Org: %s", err)
```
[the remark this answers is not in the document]
Didainius: fixed

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
```diff
@@ -0,0 +1,121 @@
...
+data "vcd_vdc_group" "existing" {
+  org  = "my-org" # Optional, can be inherited from Provider configuration
+  name = "main-vdc-group"
+}
+
+data "vcd_nsxt_network_context_profile" "cp1" {
+  context_id = vcd_vdc_group.test1.id
+  name       = "CTRXICA"
+  scope      = "SYSTEM"
+}
```
adezxc, line 37: Is this needed as it's not used in the example, maybe other data sources could be used?

Didainius: Indeed. I have changed the snippet

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
```diff
@@ -0,0 +1,121 @@
...
+
+data "vcd_nsxt_app_port_profile" "FTP" {
+  context_id = data.vcd_nsxt_manager.main.id
+  name       = "FTP"
+  scope      = "SYSTEM"
+}
+
+resource "vcd_nsxt_distributed_firewall_rule" "r1" {
+  org          = "my-org"
+  vdc_group_id = vcd_vdc_group.test1.id
```
adezxc, line 41: nit, but maybe you could also change this in the distributed_firewall docs?
```suggestion
  vdc_group_id = data.vcd_vdc_group.test1.id
```

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
```diff
@@ -0,0 +1,121 @@
...
+  description = "description"
+
+  source_ids           = [vcd_nsxt_ip_set.set1.id, vcd_nsxt_ip_set.set2.id]
+  destination_ids      = [vcd_nsxt_security_group.g1-empty.id, vcd_nsxt_security_group.g2.id]
+  app_port_profile_ids = [data.vcd_nsxt_app_port_profile.FTP.id]
+}
+
+resource "vcd_nsxt_distributed_firewall_rule" "r2" {
+  org          = "{{.Org}}"
+  vdc_group_id = vcd_vdc_group.test1.id
```
adezxc, line 54: nit, but maybe you could also change this in the distributed_firewall docs?
```suggestion
  vdc_group_id = data.vcd_vdc_group.test1.id
```

adezxc (approved):
Looks good to me, great PR!

dataclouder (approved):

adambarreiro (approved):
LGTM

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
Didainius: Thanks, fixed them.

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
Didainius: Thanks, fixed them.

website/docs/r/nsxt_distributed_firewall_rule.html.markdown
```diff
@@ -0,0 +1,121 @@
...
+!> There is a different resource
+[`vcd_nsxt_distributed_firewall`](/providers/vmware/vcd/latest/docs/resources/nsxt_distributed_firewall)
+that can manage all firewall rules in one resource. One should use **only one of**
+`vcd_nsxt_distributed_firewall` or `vcd_nsxt_distributed_firewall_rule` as using both will result in
+unexpected firewall configuration.
+
+## Example Usage
+
+```hcl
+data "vcd_vdc_group" "existing" {
```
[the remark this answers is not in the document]
Didainius: Thanks, fixed them.

Issue #741, opened 2021-11-11
Add support for vApp template


AlexanderB1290:
Affected resources and data sources:

- **vcd_vapp**
- **vcd_resource_list**

Adding support for vApp templates will reduce the need to create new vApps from the ground up. The proposition is to add optional parameters to the mentioned resources and data source for deploying and obtaining a vApp template:

- **catalog_name** - corresponding to the name of the catalog from which the vApp template need to be loaded.
- **template_name** - corresponding to the name of the vApp template which is loaded using data source or deployed using resource respectively.

This way a vApp could be created using the following example:
```terraform
resource "vcd_vapp" "my_vapp" {
      name = "my_vapp_name"
      catalog_name = "my_catalog_name" #Name of the catalog from which the vApp template is taken
      template_name = "my_vapp_template_from_catalog" #Name of the vApp template used to deploy a new vApp
}
```
The vApp template can be obtained in similar way using a data source:
```terraform
data "vcd_vapp" "other_vapp" {
      name = "other_vapp_template" #Name of the vApp or vApp template
      catalog_name = "other_vapp_template_catalog" #Name of the catalog from which to take the vApp template.
}
```
If the attribute **catalog_name** is not provided, the data source needs to return a vApp, not vApp template.

Using the vcd_resource_list to be able to obtain the names of the virtual machines inside an existing vApp template:
```terraform
data "vcd_resource_list" "vm_list" {
     name = "vm_list_vapp_template"
     parent = "my_vapp_template_name" #Name of the vApp or vApp template
     catalog_name = "my_catalog_name" #Name of the catalog from which vApp template is taken
}
```
If the attribute **catalog_name** is not provided, the data source needs to return virtual machine names from a vApp, not vApp template.

dataclouder:
Related to #502

vbauzysvmware:
This may be implemented only in one way, you can deploy but not manage it after that. I mean recreate, delete ok, but any change of internal details (VM, CPU, memory, disks, customization, etc.) won't be possible as this is not how terraform HCL works.

AlexanderB1290:
@vbauzysvmware I am fine with this approach! In the current Global Lab environment we are always do an identical copy. So... Such customizations are done by the users only as Day-2 operations using the vCD UI.

vbauzysvmware:
> 

@AlexanderB1290 can you share for which company this is needed? :)

carlschroder:
I'd greatly appreciate this feature set too!!
[...]

PR #1081, opened 2023-07-12
Add resource vcd_cloned_vapp

dataclouder:
Address Issue #741 
Add resource `vcd_cloned_vapp`, which allows creating a vApp from either a vApp template or another vApp.

## Testing

The needed vApp template will be uploaded during the test.

`go test -tags vapp -run TestAccVcdClonedVApp -v -timeout 0`


website/docs/r/cloned_vapp.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+description: |-
+  Provides a VMware Cloud Director Cloned vApp resource. This can be used to create vApps from either a vApp template or another vApp.
+---
+
+# vcd\_cloned\_vapp
+
+Provides a VMware Cloud Director Cloned vApp resource. This can be used to create vApps from either a vApp template or another vApp.
+This resource should be used only on creation, although deletion also works. The result of using this resource is a
+regular vApp ([`vcd_vapp`](/providers/vmware/vcd/latest/docs/resources/vapp)), with all its contents derived by either a vApp template or another vApp.
+As of this first implementation, no configuration is available: the source is simply cloned from the source vApp template
```
lvirbalas, line 14: Something's off:

> the **source** is simply cloned from the **source**

website/docs/r/cloned_vapp.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+page_title: "VMware Cloud Director: vcd_cloned_vapp"
+sidebar_current: "docs-vcd-resource-cloned-vapp"
+description: |-
+  Provides a VMware Cloud Director Cloned vApp resource. This can be used to create vApps from either a vApp template or another vApp.
+---
+
+# vcd\_cloned\_vapp
+
+Provides a VMware Cloud Director Cloned vApp resource. This can be used to create vApps from either a vApp template or another vApp.
+This resource should be used only on creation, although deletion also works. The result of using this resource is a
```
lvirbalas, line 12: Just a thought, maybe add a sentence on the main targeted use case, as it may not be instantly clear in what circumstances one would use it?

website/docs/r/cloned_vapp.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+The following arguments are supported:
+
+* `name` - (Required) A unique name for the vApp
+* `org` - (Optional) The name of organization to use, optional if defined at provider level. Useful when connected as sysadmin working across different organisations
+* `vdc` - (Optional) The name of VDC to use, optional if defined at provider level
+* `description` (Optional) An optional description for the vApp, up to 256 characters.
+* `power_on` - (Optional) A boolean value stating if this vApp should be powered on. Default is `false`.
+* `source_type` - (Required) The type of the source to use: one of `template` or `vapp`.
+* `source_id` - (Required) The ID of the source to use.
+* `delete_source` - (Optional) A boolean value stating if the source entity should be deleted after creation.
```
lvirbalas, line 68:
```suggestion
* `delete_source` - (Optional) A boolean value of `true` or `false` stating if the source entity should be deleted after creation.
```

lvirbalas (commented):
My main ask is to expand documentation explaining more about when one should use this resource.

website/docs/r/cloned_vapp.html.markdown
dataclouder: changed

website/docs/r/cloned_vapp.html.markdown
dataclouder: Added

lvirbalas: Thanks! BTW, is there a scenario where user would want to import the cloned vApp with all its VMs to Terraform management? It feels like this could be a separate section in this doc page.

dataclouder: I am sure users would love it, but cross-resource operations are not a good fit for Terraform.
The point is that this resource like `vcd_subscribed_catalog` only handles the main resource, and the byproducts can only be handled by separate use of data sources and imported simpler resources.
We may find some partial solution to this problem in a later release, where we will apply the new import infrastructure.

lvirbalas (approved):

vcd/resource_vcd_cloned_vapp_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+func TestAccVcdClonedVApp(t *testing.T) {
+	preTestChecks(t)
+	var vappFromTemplate = "TestAccClonedVAppFromTemplate"
+	var vappFromVapp = "TestAccClonedVAppFromVapp"
+	var vappDescription = "Test cloned vApp from Template"
+	vappTemplateName := "Test-small-3VM"
+	firstVmName := "firstVM"
+	//secondVmName := "secondVM"
+	//thirdVmName := "thirdVM"
```
adezxc, line 21: Is the comment still needed?

adezxc (approved):

vcd/resource_vcd_cloned_vapp_test.go
dataclouder: No. In fact, even `firstVmName` was practically unused, since I use `vm_list` to get the actual VM name. Now removed all three.
Thanks

vcd/resource_vcd_cloned_vapp.go
```diff
@@ -0,0 +1,247 @@
...
+			Deploy:      true,
+			PowerOn:     powerOn,
+			Description: vappDescription,
+			Source: &types.Reference{
+				HREF: sourceVapp.VApp.HREF,
+				ID:   sourceVapp.VApp.ID,
+				Type: sourceVapp.VApp.Type,
+				Name: sourceVapp.VApp.Name,
+			},
+			IsSourceDelete: addrOf(deleteSource),
```
Didainius, line 128: nit

```suggestion
			IsSourceDelete: &deleteSource,
```

vcd/resource_vcd_cloned_vapp.go
```diff
@@ -0,0 +1,247 @@
...
+
+	d.SetId(vapp.VApp.ID)
+
+	vappStatus, err := vapp.GetStatus()
+	if err != nil {
+		return diag.Errorf("error retrieving vApp '%s' status: %s", vappName, err)
+	}
+	// If the cloned vApp is suspended and power ON was requested, we adjust the status accordingly
+	if vappStatus == "SUSPENDED" && powerOn {
+		err = vapp.DiscardSuspendedState()
```
Didainius, line 177: is there any chance that the VM is not suspended (powered off?), but power on is requested (we'd miss it in such case)?

vcd/resource_vcd_cloned_vapp.go
```diff
@@ -0,0 +1,247 @@
...
+	return nil
+}
+
+func resourceVcdClonedVAppDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	return resourceVcdVAppDelete(ctx, d, meta)
+}
+
+func resourceVcdClonedVAppUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	return diag.Errorf("this resource doesn't support updates")
+}
```
Didainius, line 247: NIT, but you could set `ForceNew` on all non computed attributes and skip specifying `update` function at all.

Didainius (approved):

vcd/resource_vcd_cloned_vapp.go
dataclouder: The PowerOn operation is handled directly by the API. The "suspended" state is a side effect that I have seen happening often, and I added an adjustment.
I assume that the API would return an error if the powering ON fails, but I couldn't manage to trigger such occurrence.

vcd/remove_leftovers_test.go
```diff
@@ -110,29 +110,31 @@ func removeLeftovers(govcdClient *govcd.VCDClient, verbose bool) error {
...
+				if err != nil {
+					return fmt.Errorf("error disabling provider VDC '%s': %s", pvdcRec.Name, err)
+				}
+				task, err := pvdc.Delete()
+				if err != nil {
+					return fmt.Errorf("error deleting provider VDC '%s': %s", pvdcRec.Name, err)
+				}
+				err = task.WaitTaskCompletion()
+				if err != nil {
+					return fmt.Errorf("error finoshing deletion of provider VDC '%s': %s", pvdcRec.Name, err)
```
adambarreiro, line 136:
```suggestion
					return fmt.Errorf("error finishing deletion of provider VDC '%s': %s", pvdcRec.Name, err)
```

adambarreiro (approved):
LGTM

vcd/remove_leftovers_test.go
dataclouder: Thanks! Fixed

vcd/resource_vcd_cloned_vapp.go
dataclouder: I was undecided between these two strategies. I reverted to using Forcenew. Thanks

website/vcd.erb
```diff
@@ -586,6 +586,8 @@
             <li<%= sidebar_current("docs-vcd-resource-provider-vdc") %>>
               <a href="/docs/providers/vcd/r/provider_vdc.html">vcd_provider_vdc</a>
             </li>
+            <li<%= sidebar_current("docs-vcd-resource-cloned-vapp") %>>
+              <a href="/docs/providers/vcd/r/cloned_vapp.html">vcd_cloned_vapp</a>
```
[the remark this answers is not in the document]
dataclouder: I always see `Auto-merging website/vcd.erb` and I hope for the best. Apparently I shouldn't

PR #1089, opened 2023-07-19
Fix `ignore_metadata_changes` to stop ignoring deprecated `metadata`

adambarreiro:
This PR amends #1057. While in the attribute documentation we had this statement:

```
~> Note that this feature is only considered when using the `metadata_entry` argument in the resources and data sources that support
it. In other words, to ignore metadata when you are using the deprecated `metadata` argument, please use the native Terraform `lifecycle.ignore_changes` block.
```

Due to a bug this was false. This PR fixes this bug and makes the old `metadata` attribute to work like before.


lvirbalas (approved):
Nice how you found this with the help of binary tests!

dataclouder (approved):

adezxc (approved):

PR #1093, opened 2023-07-20
Make shell scripts consistent

dataclouder:
* Make shell script shebang consistent
* Add a script `.changes/init.sh` to start a new release cycle


adambarreiro (approved):

adezxc (approved):

Didainius (approved):

.changes/init.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+
```
lvirbalas, line 3: Let's add an echo or comment with an explanation of the purpose of this script?

dataclouder: (we have a very similar script in go-vcloud-director/.changes/)

lvirbalas: Thanks!

> (we have a very similar script in go-vcloud-director/.changes/)

I wouldn't mind of such comment in the govcd too :)

lvirbalas (approved):

PR #1094, opened 2023-07-21
Reduce test connections

dataclouder:
This PR tries to reduce temporary connections, by using the VCD version and API version that are placed in the test configuration file during VCD provisioning.

### DIFF
--- a/vcd/config_test.go
+++ b/vcd/config_test.go
@@ -310,7 +310,8 @@ const (
 # comment {{.Comment}}
 # date {{.Timestamp}}
 # file {{.CallerFileName}}
-#
+# VCD version {{.VcdVersion}}
+# API version {{.ApiVersion}}
 
 provider "vcd" {
   user                 = "{{.PrUser}}"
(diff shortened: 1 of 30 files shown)


adambarreiro (approved):
Nice optimization!

adezxc (approved):

Didainius (approved):

vcd/config_test.go
```diff
@@ -67,6 +67,8 @@ type TestConfig struct {
 		ApiToken                string `json:"api_token,omitempty"`
 		ApiTokenFile            string `json:"api_token_file,omitempty"`
 		ServiceAccountTokenFile string `json:"service_account_token_file,omitempty"`
+		VcdVersion              string `json:"vcdVersion,omitempty"`
+		ApiVersion              string `json:"apiVersion,omitempty"`
```
lvirbalas, line 71: I would like to ask for comments above these two fields. It would be good if they answered how did we live without them until now and why do we need to specify these versions explicitly now :)

Please also extend the PR description with such explanation.

lvirbalas: Out of curiosity, how much time does it save?

BTW, might there be a scenario where you can't fill these in? Like you just point to some automatically generated CD instance which version you don't know up-front.

dataclouder: It saves development time, and reduces the risk of excluding the test from binary runs.
Sometimes, we can't use this trick, as we need the temporary connection for more than checking the version.
This PR removes about half of the currently used temporary connections.

lvirbalas (approved):

PR #1101, opened 2023-08-01
Use minimal check with TestAccVcdSubscribedCatalog

dataclouder:
When the network is busy, `TestAccVcdSubscribedCatalog` may lag more than usual, and some checks fail, because not all the catalog items have been downloaded in time. To minimize test failures, the checks on non-automatic downloads only test if one of the items was downloaded, rather than all.


adambarreiro (approved):

vcd/resource_vcd_subscribed_catalog_test.go
```diff
@@ -177,7 +177,7 @@ func TestAccVcdSubscribedCatalog(t *testing.T) {
 						ResourceName: resourcePublisher,
 						Check: resource.ComposeAggregateTestCheckFunc(
 							testAccCheckVcdCatalogExists(resourcePublisher),
-							testCheckCatalogAndItemsExist(publisherOrg, publisherCatalog, true, numberOfVappTemplates+numberOfMediaItems, numberOfVappTemplates, numberOfMediaItems),
+							testCheckCatalogAndItemsExist(publisherOrg, publisherCatalog, true, false, numberOfVappTemplates+numberOfMediaItems, numberOfVappTemplates, numberOfMediaItems),
```
lvirbalas, line 180: Interesting optimization(s). How did you decide where to go minimal and where full? What coverage do we lose after this change?

p.s. I'm approving the PR, but just want to grasp the change better.

lvirbalas (approved):

vcd/resource_vcd_subscribed_catalog_test.go
dataclouder: The first part of the test creates a catalog (the publisher) and uploads a few vApp templates and media items.
For this part, I am using a full check, because all elements are guaranteed to be in place.
The second part of the test creates a subscribed catalog in a different organization. For this part, I use minimal check, as we want to know whether any elements were downloaded. One of the subscribed catalog configuration (non-automatic downloads) is affected by latency, and we are not sure when the full download is completed. 
There are three problems that may happen: 
* lack of storage in the subscribed organization (catalog is created, download fails eventually)
* failure to initiate the subscription (network breakage)
* code fault (what we are actually testing).

Against the first two problems, there is nothing we can do. Users will detect the failure using their own monitoring tools, and when the environment is fixed, the test will succeed.
The third case is what we are concerned about, i.e. a regression against the subscription functionality. This scenario is also covered in detail in the govcd tests. In Terraform, we test that the subscription is successful, and that the synchronization has been initiated (i.e. some of the items have been downloaded). A failure of the minimal check may be an indication of one of the first two environmental cases being involved.

lvirbalas: Thanks for an in-depth explanation!

Didainius (approved):

adezxc (approved):

PR #1107, opened 2023-08-13
Change TestAccVcdClonedVApp to use pre-installed vApp template

dataclouder:
Using the pre-installed vApp template, we can shorten the cloned vApp test times


lvirbalas (approved):
Cool!

Didainius (approved):

PR #1108, opened 2023-08-18
Bump api version usage to 37.0

dataclouder:
Two small changes were needed to adapt the code to API version 37.0 (Drops support for VCD 10.3.x):
* a `ssl_enabled` property for `vcd_nsxt_alb_pool` (Necessary when CA certificates are used)
* a change in `iops` property handling in internal disk resource. The single field `iops` was replaced by a structure in v37.0. We may need to do more work on that.

### DIFF
--- a/vcd/resource_vcd_nsxt_alb_pool.go
+++ b/vcd/resource_vcd_nsxt_alb_pool.go
@@ -141,13 +141,13 @@ func resourceVcdAlbPool() *schema.Resource {
 				Default:     true,
 				Description: "Monitors if the traffic is accepted by node (default true)",
 			},
+			// Read only information
 			"ssl_enabled": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Default:     false,
-				Description: "Enables SSL - Necessary when CA certificates are used",
+				Computed:    true,
+				Description: "Enables SSL - Must be on when CA certificates are used",
 			},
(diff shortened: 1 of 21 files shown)


Didainius (approved):

adezxc (approved):

vcd/resource_vcd_nsxt_alb_pool.go
```diff
@@ -442,6 +452,9 @@ func setNsxtAlbPoolData(d *schema.ResourceData, albPool *types.NsxtAlbPool) erro
 	dSet(d, "default_port", albPool.DefaultPort)
 	dSet(d, "graceful_timeout_period", albPool.GracefulTimeoutPeriod)
 	dSet(d, "passive_monitoring_enabled", albPool.PassiveMonitoringEnabled)
+	if albPool.SslEnabled != nil {
+		dSet(d, "ssl_enabled", *albPool.SslEnabled)
```
lvirbalas, line 456: The new Optional + Computed approach looks great! Just one question about the above line: what happens when we set it behind the scenes? How does the next plan look like?

Thanks!

dataclouder: The property is `Optional`+`Computed`. Thus its value is reported without comparison.
If we remove `Computed`, the plan fails.

PR #1109, opened 2023-08-18
Add `firmware` and `boot_options` to VM configuration

### LINKED ISSUE #1078: Add option to specify boot firmware (BIOS/EFI) and secure boot
### Community Note

### Description

VCD 10.4.1 added the option to set the boot mode of VM to EFI. Some use-cases already require the boot mode to be set in EFI like for example in case you want to use passthrough devices. Can you please add the option that allows you to specify the boot mode? Default boot mode == BIOS with an option to set it to EFI. Attached to this would be the possibility to enable Secure Boot.

vcd_vapp_vm
vcd_vm

### Terraform Configuration (if it applies)

```hcl
Copy-paste your Terraform configurations here. Please **shrink HCL replicating the bug to minimum** 
without modules. It will help us to respond and replicate the problem quicker. It may also uncover
some HCL error while doing so.

For bigger configs you may use [Gist(s)](https://gist.github.com) or a service like Dropbox and
share a link to the ZIP file.
```

### References

https://docs.vmware.com/en/VMware-Cloud-Director/10.4.1/rn/vmware-cloud-director-1041-release-notes/index.html
* Issue #0000

adezxc:
This PR introduces new fields to `vcd_vm` and `vcd_vapp_vm` resources and datasources:
* `firmware` specify either EFI or BIOS (37.1+)
* `boot_options` block:
  - `efi_secure_boot` (37.1+)
  - `boot_delay` 
  - `boot_retry_enabled`(37.1+)
  - `boot_retry_delay`(37.1+)
  - `enter_bios_setup`

Closes #1078


website/docs/r/vapp_vm.html.markdown
```diff
@@ -592,6 +594,16 @@ Changes are ignored on update. This part isn't reread on refresh. To manage inte
 * `iops` - (Optional) Specifies the IOPS for the disk. Default is 0.
 * `storage_profile` - (Optional) Storage profile which overrides the VM default one.
 
+<a id="boot-options-block"></a>
+## Customization
```
dataclouder, line 598:
```suggestion
## Boot options
```

website/docs/r/vapp_vm.html.markdown
```diff
@@ -490,6 +490,8 @@ example for usage details.
   Tools are not present on the VM.
 * `os_type` - (Optional; *v2.9+*) Operating System type. Possible values can be found in [Os Types](#os-types). Required when creating empty VM.
 * `hardware_version` - (Optional; *v2.9+*) Virtual Hardware Version (e.g.`vmx-14`, `vmx-13`, `vmx-12`, etc.). Required when creating empty VM.
+* `firmware` - (Optional; v3.11+, vCD 10.4.1+) Specify boot firmware of the VM. Can be `efi` or `bios`
```
dataclouder, line 493:
```suggestion
* `firmware` - (Optional; v3.11+, VCD 10.4.1+) Specify boot firmware of the VM. Can be `efi` or `bios`
```

website/docs/r/vapp_vm.html.markdown
```diff
@@ -711,67 +723,167 @@ resource "vcd_vapp_vm" "web2" {
...
-  `darwin12_64Guest` (Apple Mac OS X 10.8 (64-bit)), `eComStation2Guest` (Serenity Systems eComStation 2), `openServer6Guest` (SCO OpenServer 6), 
-  `solaris11_64Guest` (Oracle Solaris 11 (64-bit)), `darwin11_64Guest` (Apple Mac OS X 10.7 (64-bit)), 
-  `darwin11Guest` (Apple Mac OS X 10.7 (32-bit)), `eComStationGuest` (Serenity Systems eComStation 1), `unixWare7Guest` (SCO UnixWare 7), 
-  `openServer5Guest` (SCO OpenServer 5), `os2Guest` (IBM OS/2), `freebsd64Guest` (FreeBSD Pre-11 versions (64-bit)), 
-  `freebsdGuest` (FreeBSD Pre-11 versions (32-bit)), `darwin10_64Guest` (Apple Mac OS X 10.6 (64-bit)), 
-  `darwin10Guest` (Apple Mac OS X 10.6 (32-bit)), `otherGuest64` (Other (64-bit)), `otherGuest` (Other (32-bit)),
-   `solaris10_64Guest` (Oracle Solaris 10 (64-bit)), `solaris10Guest` (Oracle Solaris 10 (32-bit)), `netware6Guest` (Novell NetWare 6.x), 
-   `netware5Guest` (Novell NetWare 5.1)
+* **Linux**
+- `other5xLinux64Guest` - Other 5.x or later Linux (64-bit)
```
dataclouder, line 727: Using `-` instead of `*` does not guarantee a change of bullet point. To do a good differentiation, we need to indent the other lines.
```
* **Linux**
  - `other5xLinux64Guest` - Other 5.x or later Linux (64-bit)
  - `other5xLinuxGuest` - Other 5.x or later Linux (32-bit)
```
Even using the same bullet will work, provided we have the indentation
```
* **Linux**
  * `other5xLinux64Guest` - Other 5.x or later Linux (64-bit)
  * `other5xLinuxGuest` - Other 5.x or later Linux (32-bit)
```

Both will look as below
Screenshot 2023-08-28 at 16 21 43

vcd/resource_vcd_vapp_vm_boot_options_test.go
```diff
@@ -0,0 +1,246 @@
...
+		vmName   string = t.Name() + "VM"
+	)
+
+	if testConfig.VCD.ProviderVdc.StorageProfile == "" || testConfig.VCD.ProviderVdc.StorageProfile2 == "" {
+		t.Skip("Both variables testConfig.VCD.ProviderVdc.StorageProfile and testConfig.VCD.ProviderVdc.StorageProfile2 must be set")
+	}
+
+	var params = StringMap{
+		"Org":         testConfig.VCD.Org,
+		"Vdc":         testConfig.VCD.Vdc,
```
Didainius, line 27: Would be better to create test VMs in NSX-T VDC `testConfig.Nsxt.Vdc` as we're going to deprecate and remove NSX-V support at some time.

Didainius (commented):
My first scan through with some questions

vcd/resource_vcd_vapp_vm.go
```diff
@@ -834,6 +883,15 @@ func genericResourceVmCreate(d *schema.ResourceData, meta interface{}, vmType ty
...
 
+		if enterBiosSetup, ok := d.Get("boot_options.0.enter_bios_setup").(bool); ok {
+			biosSetup := &types.BootOptions{
+				EnterBiosSetup: &enterBiosSetup}
+
+			_, err = vm.UpdateBootOptions(biosSetup)
+			if err != nil {
+				return diag.Errorf("error enabling BIOS setup: %s", err)
+			}
+		}
```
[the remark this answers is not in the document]
adezxc: This is made so that after the boot, we get to the requested state. Without this piece, if a user asks for `enter_bios_setup` to be true and changes some other option that requires a VM power cycle, it would result in `enter_bios_setup` to be set to `false`

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1171,6 +1229,41 @@ func createVmEmpty(d *schema.ResourceData, meta interface{}, vmType typeOfVm) (*
 	if hardWareVersion, ok = d.GetOk("hardware_version"); !ok {
 		return nil, fmt.Errorf("`hardware_version` is required when creating empty VM")
 	}
+	// Additional checks if the hardware version and OS support new Firmware and BootOptions fields (37.1+)
```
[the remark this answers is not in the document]
adezxc: I have updated the code a bit regarding this for it to be more readable and reliable.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1171,6 +1229,41 @@ func createVmEmpty(d *schema.ResourceData, meta interface{}, vmType typeOfVm) (*
...
+		bootOptions.BootDelay = addrOf(d.Get("boot_options.0.boot_delay").(int))
+		bootOptions.EnterBiosSetup = addrOf(d.Get("boot_options.0.enter_bios_setup").(bool))
+		if vcdClient.Client.APIVCDMaxVersionIs(">=37.1") {
+			bootOptions.BootRetryDelay = addrOf(d.Get("boot_options.0.boot_retry_delay").(int))
+			bootOptions.BootRetryEnabled = addrOf(d.Get("boot_options.0.boot_retry_enabled").(bool))
+			if efiSecureBoot, ok := d.GetOk("boot_options.0.efi_secure_boot"); ok {
+				if firmware != "efi" {
+					return nil, fmt.Errorf("error: EFI secure boot can only be used with EFI firmware")
+				}
+				bootOptions.EfiSecureBootEnabled = addrOf(efiSecureBoot.(bool))
```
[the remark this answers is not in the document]
adezxc: Not that I know of. They're not required by the API, and if unset in Terraform, they're set to `false` or `0`

vcd/resource_vcd_vapp_vm_boot_options_test.go
adezxc: Agree. Updated.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1719,17 +1821,68 @@ func resourceVcdVAppVmUpdateExecute(d *schema.ResourceData, meta interface{}, ex
...
+
+				efiSecureBoot := d.Get("boot_options.0.efi_secure_boot").(bool)
+				if firmware != "efi" && efiSecureBoot {
+					return diag.Errorf("error: EFI secure boot can only be enabled with firmware set to 'efi'")
+				}
+				if firmware != "efi" && !efiSecureBoot {
+					efiSecureBootOptions := &types.BootOptions{
+						EfiSecureBootEnabled: addrOf(efiSecureBoot),
+					}
+					vm, err = vm.UpdateBootOptions(efiSecureBootOptions)
```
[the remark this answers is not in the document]
adezxc: It's a good point, although I'm not completely sure what's the best method around it. Would it be better to remove the `Refresh()` call completely or just not update the `vm` variable? 

There can be a risk of overriding other configuration, but then it's probably the other code that needs fixing because we shouldn't override the `vm` variable ourselves, only with API calls

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1789,6 +1942,17 @@ func resourceVcdVAppVmUpdateExecute(d *schema.ResourceData, meta interface{}, ex
...
+			if enterBiosSetup, ok := d.Get("boot_options.0.enter_bios_setup").(bool); ok {
+				biosSetup := &types.BootOptions{
+					EnterBiosSetup: &enterBiosSetup}
+
+				vm, err = vm.UpdateBootOptions(biosSetup)
+				if err != nil {
+					return diag.Errorf("error enabling BIOS setup: %s", err)
+				}
+			}
+
```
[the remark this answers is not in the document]
adezxc: Same as the previous comment, during `PowerOn()`, the  `EnterBiosSetup` is always set to `false`, so we set it in two places for two scenarios:
1) User wants the VM to be powered off, but wants to enter BIOS on next boot
2) User wants the VM to be powered on, updates some other value e.g `firmware` that triggers a power cycle, but they still want to enter BIOS on the power cycle that they triggered on their own.

Maybe I missed something, but I think it's the way to do it.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -257,6 +257,13 @@ func vmSchemaFunc(vmType typeOfVm) map[string]*schema.Schema {
 			Computed:    true,
 			Description: "Operating System type. Possible values can be found in documentation.",
 		},
+		"firmware": {
+			Type:         schema.TypeString,
+			Optional:     true,
+			Computed:     true,
+			ValidateFunc: validation.StringInSlice([]string{"efi", "bios"}, false),
+			Description:  "Firmware of the VM. Can be either EFI or BIOS, availability depending on the os_type argument. Can be changed only when the VM is turned off",
```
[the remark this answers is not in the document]
adezxc: It's hard to put in words, as we don't have that written in other fields. I have updated the description, not sure if it's more clear, but it will power off the VM if needed to update.

website/docs/r/vapp_vm.html.markdown
adezxc: Updated.

website/docs/r/vapp_vm.html.markdown
adezxc: Thanks for the suggestion, updated.

vcd/resource_vcd_vapp_vm.go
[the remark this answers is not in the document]
Didainius: It is somewhat safer to have an explicit `Refresh` call (easier to see when reading). There is also this comment as part of refactoring that might be useful https://github.com/Didainius/terraform-provider-vcd/blob/edge-uplinks/vcd/resource_vcd_vapp_vm.go#L53-L63

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1172,6 +1241,59 @@ func createVmEmpty(d *schema.ResourceData, meta interface{}, vmType typeOfVm) (*
...
+
+		os, err := vdc.FindOsFromId(hwVersion, osType.(string))
+		if err != nil {
+			return nil, fmt.Errorf("error finding given OS type: %s", err)
+		}
+
+		if !contains(os.SupportedFirmware, firmware.(string)) {
+			return nil, fmt.Errorf("provided `os_type` doesn't support provided firmware")
+		}
+	}
```
Didainius, line 1278: If you can remind me - these errors directly from API were ugly, right?

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1172,6 +1241,59 @@ func createVmEmpty(d *schema.ResourceData, meta interface{}, vmType typeOfVm) (*
...
+		bootOptions.EnterBiosSetup = addrOf(d.Get("boot_options.0.enter_bios_setup").(bool))
+		bootOptions.BootRetryDelay = addrOf(d.Get("boot_options.0.boot_retry_delay").(int))
+		bootOptions.BootRetryEnabled = addrOf(d.Get("boot_options.0.boot_retry_enabled").(bool))
+		if efiSecureBoot, ok := d.GetOk("boot_options.0.efi_secure_boot"); ok {
+			if firmware != "efi" {
+				return nil, fmt.Errorf("error: EFI secure boot can only be used with EFI firmware")
+			}
+			bootOptions.EfiSecureBootEnabled = addrOf(efiSecureBoot.(bool))
+		}
+		if !supportsFirmware && (bootOptions.BootRetryDelay != nil || bootOptions.BootRetryEnabled != nil) {
```
Didainius, line 1292: Won't `bootOptions.BootRetryDelay` and `bootOptions.BootRetryEnabled` always be non-nil, because a zero value (`0`` or `false``) is comming from Terraform schema?

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1607,13 +1732,19 @@ func resourceVcdVAppVmUpdateExecute(d *schema.ResourceData, meta interface{}, ex
...
 	if d.HasChanges("cpu_cores", "power_on", "disk", "expose_hardware_virtualization", "boot_image",
 		"hardware_version", "os_type", "description", "cpu_hot_add_enabled",
-		"memory_hot_add_enabled") || memoryNeedsColdChange || cpusNeedsColdChange || networksNeedsColdChange {
-
-		log.Printf("[TRACE] VM %s has changes: memory(%t), cpus(%t), cpu_cores(%t), power_on(%t), disk(%t), expose_hardware_virtualization(%t),"+
-			" boot_image(%t), hardware_version(%t), os_type(%t), description(%t), cpu_hot_add_enabled(%t), memory_hot_add_enabled(%t), network(%t)",
-			vm.VM.Name, d.HasChange("memory"), d.HasChange("cpus"), d.HasChange("cpu_cores"), d.HasChange("power_on"), d.HasChange("disk"),
-			d.HasChange("expose_hardware_virtualization"), d.HasChange("boot_image"), d.HasChange("hardware_version"),
-			d.HasChange("os_type"), d.HasChange("description"), d.HasChange("cpu_hot_add_enabled"), d.HasChange("memory_hot_add_enabled"), d.HasChange("network"))
+		"memory_hot_add_enabled", "firmware", "boot_options.0.efi_secure_boot") || memoryNeedsColdChange || cpusNeedsColdChange || networksNeedsColdChange {
```
Didainius, line 1735: Does this nested element path gets evaluated correctly? I remember there were some issues on getting that value previously.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1789,6 +1976,22 @@ func resourceVcdVAppVmUpdateExecute(d *schema.ResourceData, meta interface{}, ex
...
+			if enterBiosSetup := d.Get("boot_options.0.enter_bios_setup").(bool); enterBiosSetup {
+				biosSetup := &types.BootOptions{
+					EnterBiosSetup: &enterBiosSetup}
+
+				vm, err = vm.UpdateBootOptions(biosSetup)
+				if err != nil {
+					return diag.Errorf("error enabling BIOS setup: %s", err)
+				}
+			}
+
```
Didainius, line 1994: Will it not ignore entering bios if `customizationNeeded=true`?

vcd/resource_vcd_vapp_vm.go
```diff
@@ -834,6 +889,20 @@ func genericResourceVmCreate(d *schema.ResourceData, meta interface{}, vmType ty
...
+		if enterBiosSetup := d.Get("boot_options.0.enter_bios_setup").(bool); enterBiosSetup {
+			biosSetup := &types.BootOptions{
+				EnterBiosSetup: &enterBiosSetup}
+
+			_, err = vm.UpdateBootOptions(biosSetup)
+			if err != nil {
+				return diag.Errorf("error enabling BIOS setup: %s", err)
+			}
+		}
```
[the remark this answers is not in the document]
adezxc: We set `EnterBiosSetup` earlier as well, so it enters it here and it would do that on the next power cycle. This if branch is for the case where a user has a powered on VM, changes some options which require the VM to be powered off and then after the power on they still have `EnterBiosSetup` enabled and don't have an inconsistent plan.

vcd/resource_vcd_vapp_vm.go
adezxc: This one was alright, actually better than mine probably : `Firmware efi not supported by VM TestVM with guest operating system sles10_64Guest at hardware version HW11.` It's mostly for `Update` so we don't do anything to the VM like PowerOff or partial update but I guess I could improve the error message.

vcd/resource_vcd_vapp_vm.go
adezxc: It does, here is a sample debug output when only changing efi secure boot
```
[TRACE] VM TestAccVcdVmAdvancedComputeProperties_empty-19433 has changes: memory(false), cpus(false),
 cpu_cores(false),power_on(false), disk(false), expose_hardware_virtualization(false), boot_image(false), 
hardware_version(false), os_type(false), description(false),cpu_hot_add_enabled(false), 
memory_hot_add_enabled(false), firmware(false),efi_secure_boot(true) network(false): 
timestamp=2023-09-05T18:52:53.845+0300
```

vcd/resource_vcd_vapp_vm.go
adezxc: Yes, I missed this part, updated.

vcd/resource_vcd_vapp_vm.go
Didainius: Great, thanks!

vcd/resource_vcd_vapp_vm.go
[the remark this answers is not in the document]
Didainius: Thanks, I will attempt to do some manual checks here as well

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1789,6 +1984,12 @@ func resourceVcdVAppVmUpdateExecute(d *schema.ResourceData, meta interface{}, ex
 			if err != nil {
 				return diag.Errorf(errorCompletingTask, err)
 			}
+
+			err = vm.Refresh()
+			if err != nil {
+				return diag.Errorf("error refreshing VM: %s", err)
+			}
+
```
Didainius, line 1992: If we do a refresh here, shouldn't we also do it in the end of `if customizationNeeded {`?

vcd/resource_vcd_vapp_vm.go
```diff
@@ -257,6 +257,13 @@ func vmSchemaFunc(vmType typeOfVm) map[string]*schema.Schema {
 			Computed:    true,
 			Description: "Operating System type. Possible values can be found in documentation.",
 		},
+		"firmware": {
+			Type:         schema.TypeString,
+			Optional:     true,
+			Computed:     true,
+			ValidateFunc: validation.StringInSlice([]string{"efi", "bios"}, false),
+			Description:  "Firmware of the VM. Can be either EFI or BIOS, availability depending on the os_type argument. Changing the value when `power_on` is set to true, will cause a reboot of the VM.",
```
Didainius, line 265: Should we state that `bios` is default

website/docs/r/vapp_vm.html.markdown
```diff
@@ -490,6 +490,8 @@ example for usage details.
   Tools are not present on the VM.
 * `os_type` - (Optional; *v2.9+*) Operating System type. Possible values can be found in [Os Types](#os-types). Required when creating empty VM.
 * `hardware_version` - (Optional; *v2.9+*) Virtual Hardware Version (e.g.`vmx-14`, `vmx-13`, `vmx-12`, etc.). Required when creating empty VM.
+* `firmware` - (Optional; v3.11+, VCD 10.4.1+) Specify boot firmware of the VM. Can be `efi` or `bios`. Changing the value requires the VM to power off.
```
Didainius, line 493: Please mention default is `bios`

vcd/resource_vcd_vapp_vm.go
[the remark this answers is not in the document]
Didainius: Now I am thinking of a user workflow and if this is a good approach.
Thinking of this:
* User wants to create a VM and enter BIOS for whatever reason. He changes whatever he wants and lets the VM to boot.

What effecitevely happens is - 
* User creates a VM, enters bios AND the additional part is `this VM is configured so that next VM reboot will get stuck in BIOS`. It looks a bit risky to me - imagine a case where VM is deployed and VM should be rebooted multiple times during guest customization script (which is quite often the case). This will get stuck at bios again until someone manually exits it.



cc: @lvirbalas, @adambarreiro, @dataclouder for thoughts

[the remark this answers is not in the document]
Didainius: Some remotely similar functionality we handled in such a way https://registry.terraform.io/providers/vmware/vcd/latest/docs/resources/vapp_vm#force

vcd/resource_vcd_vapp_vm.go
adezxc: Yes, moved the refresh outside of the if statements, right before the `if enter_bios_setup` branch

vcd/resource_vcd_vapp_vm.go
adezxc: I updated the description, but didn't add the `Default:` parameter, as 37.0 and older versions don't have Firmware

website/docs/r/vapp_vm.html.markdown
adezxc: Updated.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1814,6 +2015,16 @@ func resourceVcdVAppVmUpdateExecute(d *schema.ResourceData, meta interface{}, ex
 				return diag.Errorf("failed powering on with customization: %s", err)
 			}
```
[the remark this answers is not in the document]
adezxc: Yes, moved outside the block

website/docs/r/vapp_vm.html.markdown
```diff
@@ -490,6 +490,8 @@ example for usage details.
   Tools are not present on the VM.
 * `os_type` - (Optional; *v2.9+*) Operating System type. Possible values can be found in [Os Types](#os-types). Required when creating empty VM.
 * `hardware_version` - (Optional; *v2.9+*) Virtual Hardware Version (e.g.`vmx-14`, `vmx-13`, `vmx-12`, etc.). Required when creating empty VM.
+* `firmware` - (Optional; v3.11+, VCD 10.4.1+) Specify boot firmware of the VM. Can be `efi` or `bios`. If unset, defaults to 'bios'. Changing the value requires the VM to power off.
```
Didainius, line 493: Easier on eyes:

```suggestion
* `firmware` - (Optional; v3.11+, VCD 10.4.1+) Specify boot firmware of the VM. Can be `efi` or `bios`. If unset, defaults to `bios`. Changing the value requires the VM to power off.
```

vcd/resource_vcd_vapp_vm.go
Didainius: I think I am missing this push as I still see this code and the comment doesn't report as outdated:
```
		if !supportsFirmware && (bootOptions.BootRetryDelay != nil || bootOptions.BootRetryEnabled != nil) {
			return nil, fmt.Errorf("boot retry option is only available in VCD 10.4.1+")
		}
```

adezxc: The change from d.Get to d.GetOk did the job. `d.Get` returns a default value for the type so it would be `false or `0`,  `d.GetOk` returns nil if the value is not provided in the configuration

adezxc: Updated, the code is now in `genericResourceVmCreate`

website/docs/r/vapp_vm.html.markdown
adezxc: Fixed.

vcd/resource_vcd_vapp_vm.go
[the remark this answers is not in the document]
adezxc: Changed now. Changed the argument to `enter_bios_setup_on_next_boot`, it also outputs a warning if `enter_bios_setup_on_next_boot` is set to true.
```
| Warning: TestAccVcdVmAdvancedComputeProperties_empty-19433: After VM Powers on, the enter_bios_setup_on_next_boot flag will be setback to false.
│
│   with vcd_vm.TestAccVcdVmAdvancedComputeProperties_empty-19433,
│   on vcd.TestAccVcdVmAdvancedComputeProperties.tf line 38, in resource "vcd_vm" "TestAccVcdVmAdvancedComputeProperties_empty-19433":
│   38: resource "vcd_vm" "TestAccVcdVmAdvancedComputeProperties_empty-19433" {
│
│ (and one more similar warning elsewhere)
```

vcd/resource_vcd_vapp_vm.go
```diff
@@ -713,6 +768,52 @@ func genericResourceVmCreate(d *schema.ResourceData, meta interface{}, vmType ty
...
+	if efiSecureBoot, ok := d.GetOk("boot_options.0.efi_secure_boot"); ok {
+		bootOptions.EfiSecureBootEnabled = addrOf(efiSecureBoot.(bool))
+	}
+	if enterBiosSetup, ok := d.GetOk("boot_options.0.enter_bios_setup_on_next_boot"); ok {
+		bootOptions.EnterBiosSetup = addrOf(enterBiosSetup.(bool))
+		if enterBiosSetup.(bool) && d.Get("power_on").(bool) {
+			diags = append(diags, diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary: fmt.Sprintf("%s: After VM Powers on, the enter_bios_setup_on_next_boot flag will be set"+
+					"back to false.", vm.VM.Name),
```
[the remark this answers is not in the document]
adezxc: Updated.

website/docs/r/vapp_vm.html.markdown
```diff
@@ -592,6 +594,16 @@ Changes are ignored on update. This part isn't reread on refresh. To manage inte
 * `iops` - (Optional) Specifies the IOPS for the disk. Default is 0.
 * `storage_profile` - (Optional) Storage profile which overrides the VM default one.
 
+<a id="boot-options"></a>
+## Boot options
+
+Allows to specify the boot options of a VM.
+
+* `efi_secure_boot` - (Optional, VCD 10.4.1+) Enable EFI Secure Boot on subsequent boots, requires `firmware` to be set to `efi`.
+* `enter_bios_setup_on_next_boot` - (Optional) Enter BIOS setup on subsequent boots of the VM.
```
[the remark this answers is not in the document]
adezxc: Updated with more detail

vcd/resource_vcd_vapp_vm.go
```diff
@@ -713,6 +768,52 @@ func genericResourceVmCreate(d *schema.ResourceData, meta interface{}, vmType ty
...
+	if efiSecureBoot, ok := d.GetOk("boot_options.0.efi_secure_boot"); ok {
+		bootOptions.EfiSecureBootEnabled = addrOf(efiSecureBoot.(bool))
+	}
+	if enterBiosSetup, ok := d.GetOk("boot_options.0.enter_bios_setup_on_next_boot"); ok {
+		bootOptions.EnterBiosSetup = addrOf(enterBiosSetup.(bool))
+		if enterBiosSetup.(bool) && d.Get("power_on").(bool) {
+			diags = append(diags, diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary: fmt.Sprintf("%s: After VM Powers on, the enter_bios_setup_on_next_boot flag will be set"+
+					"back to false and cause an inconsistent plan.", vm.VM.Name),
```
lvirbalas, line 794:
```suggestion
					"back to false by VCD and cause an inconsistent plan.", vm.VM.Name),
```

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1778,6 +2001,14 @@ func resourceVcdVAppVmUpdateExecute(d *schema.ResourceData, meta interface{}, ex
 			return diag.Errorf("error getting VM status before ensuring it is powered on: %s", err)
 		}
 
+		if d.Get("boot_options.0.enter_bios_setup_on_next_boot").(bool) {
+			diags = append(diags, diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary: fmt.Sprintf("%s: After VM Powers on, the enter_bios_setup_on_next_boot flag will be set"+
+					"back to false.", vm.VM.Name),
```
lvirbalas, line 2008:
```suggestion
					"back to false by VCD.", vm.VM.Name),
```

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1919,32 +2152,61 @@ func genericVcdVmRead(d *schema.ResourceData, meta interface{}, origin string) d
...
 		}
-		if vm.VM.VmSpecSection.CpuResourceMhz.Limit != nil {
-			dSet(d, "cpu_limit", vm.VM.VmSpecSection.CpuResourceMhz.Limit)
+		if vm.VM.VmSpecSection.OsType != "" {
+			dSet(d, "os_type", vm.VM.VmSpecSection.OsType)
 		}
-		if vm.VM.VmSpecSection.CpuResourceMhz.Shares != nil {
-			dSet(d, "cpu_shares", vm.VM.VmSpecSection.CpuResourceMhz.Shares)
+
+		dSet(d, "firmware", vm.VM.VmSpecSection.Firmware)
```
lvirbalas, line 2191: Just double checking, as all over dSets have an nil/empty "if" upfront. It's not needed here? And this line is not under the new bootOptions block below on purpose?

vcd/resource_vcd_vapp_vm.go
```diff
@@ -1172,6 +1279,42 @@ func createVmEmpty(d *schema.ResourceData, meta interface{}, vmType typeOfVm) (*
...
+	// 3. If firmware is set and VCD is 10.4.0 or older, we return an error
+	// 4. If firmware is unset and VCD is 10.4.0 or older, we pass
+	var firmware interface{}
+	if firmware, ok = d.GetOk("firmware"); ok {
+		if !supportsFirmware {
+			return nil, fmt.Errorf("firmware is supported in VCD 10.4.1+")
+		}
+	} else {
+		if supportsFirmware {
+			firmware = "bios"
```
lvirbalas, line 1297: Isn't here a possibility for firmware to be left nil? I.e. if firmware is unset and supportsFirmware is false.
This is remotely related to my other ask about "ifs" below.

lvirbalas (changes requested):

vcd/resource_vcd_vapp_vm.go
adezxc: We already check if `vmSpecSection` is not set to nil on line 2155. If it's not nil, then `vm.VM.VmSpecSection.Firmware` is initialized its zero value (`""`), so we can set it without checking.

 I'm not sure why there are other checks for strings as it doesn't really make a difference, only for *int or *boot types

vcd/resource_vcd_vapp_vm.go
adezxc: No, in any case firmware will be set to the value set in the `*.tf` files or `""`. Actually, the `else` statement is more cosmetic describing the wanted functionality, you can't really find out if the value is set in the SDK, as it always sets to zero value for the type, from what I understand and read in this thread: https://discuss.hashicorp.com/t/terraform-sdk-usage-which-out-of-get-getok-getokexists-with-boolean/41815

vcd/resource_vcd_vapp_vm.go
adezxc: Updated.

vcd/resource_vcd_vapp_vm.go
```diff
@@ -465,6 +473,49 @@ func vmSchemaFunc(vmType typeOfVm) map[string]*schema.Schema {
...
+					Type:        schema.TypeBool,
+					Optional:    true,
+					Computed:    true,
+					Description: "If set to true, enables EFI Secure Boot for the VM. Can only be changed when the VM is powered off.",
+				},
+				"enter_bios_setup_on_next_boot": {
+					Type:     schema.TypeBool,
+					Optional: true,
+					Description: "If set to true, the VM will enter BIOS setup on next boot. " +
+						"If a VM was powered on, the field will be set to `false` in VCD and Terraform will return a non-empty plan",
```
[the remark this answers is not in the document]
adezxc: Updated.

vcd/resource_vcd_vapp_vm_boot_options_test.go
```diff
@@ -0,0 +1,562 @@
...
+
+	if vcdShortTest {
+		t.Skip(acceptanceTestsSkipped)
+		return
+	}
+
+	vcd := createTemporaryVCDConnection(false)
+	if vcd.Client.APIVCDMaxVersionIs("<37.1") {
+		t.Skip("Most boot options are only available since 37.1")
+	}
```
Didainius, line 61: Should this check be set at the top, before any `templateFill` calls that generate binary tests? Binary tests would fail in 10.4.0

vcd/resource_vcd_vapp_vm.go
```diff
@@ -465,6 +473,49 @@ func vmSchemaFunc(vmType typeOfVm) map[string]*schema.Schema {
 				},
 			}},
 		},
+		"boot_options": {
```
Didainius, line 476: `vcd.ResourceSchema-vcd_vapp_vm.tf` and `vcd.ResourceSchema-vcd_vm.tf` binary tests will have to be added to `skip-upgrade-tests.txt`

Didainius (approved):
Thanks for all the improvements. I have a left a few more comments to address

vcd/resource_vcd_vapp_vm_boot_options_test.go
adezxc: True, updated.

vcd/resource_vcd_vapp_vm.go
adezxc: Added.

lvirbalas (approved):
Thanks for the updates! One more approval into the basket :)

vcd/resource_vcd_vapp_vm.go
```diff
@@ -713,6 +768,52 @@ func genericResourceVmCreate(d *schema.ResourceData, meta interface{}, vmType ty
...
+	}
+	if efiSecureBoot, ok := d.GetOk("boot_options.0.efi_secure_boot"); ok {
+		bootOptions.EfiSecureBootEnabled = addrOf(efiSecureBoot.(bool))
+	}
+	if enterBiosSetup, ok := d.GetOk("boot_options.0.enter_bios_setup_on_next_boot"); ok {
+		bootOptions.EnterBiosSetup = addrOf(enterBiosSetup.(bool))
+		if enterBiosSetup.(bool) && d.Get("power_on").(bool) {
+			diags = append(diags, diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary: fmt.Sprintf("%s: After VM Powers on, the enter_bios_setup_on_next_boot flag will be set"+
```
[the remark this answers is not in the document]
adezxc: Changed

vcd/resource_vcd_vapp_vm.go
```diff
@@ -713,6 +768,52 @@ func genericResourceVmCreate(d *schema.ResourceData, meta interface{}, vmType ty
...
+					"back to false by VCD and cause an inconsistent plan.", vm.VM.Name),
+			})
+		}
+	}
+	if efiSecureBoot, ok := d.GetOk("boot_options.0.efi_secure_boot"); ok {
+		if firmware != "efi" {
+			return diag.Errorf("error: EFI secure boot can only be used with EFI firmware")
```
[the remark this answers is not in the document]
adezxc: Chnaged.

vcd/resource_vcd_vapp_vm_boot_options_test.go
```diff
@@ -0,0 +1,562 @@
...
+
+	var params = StringMap{
+		"Org":                    testConfig.VCD.Org,
+		"Vdc":                    testConfig.Nsxt.Vdc,
+		"EdgeGateway":            testConfig.Networking.EdgeGateway,
```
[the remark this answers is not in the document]
adezxc: Removed.

vcd/resource_vcd_vapp_vm_boot_options_test.go
```diff
@@ -0,0 +1,562 @@
...
+  vapp_template_id = data.vcd_catalog_vapp_template.{{.VappTemplateName}}.id
+
+  firmware = "efi"
+
+  boot_options {
+    efi_secure_boot = true
+    boot_retry_delay = 2
+    boot_retry_enabled = true
+    boot_delay = 2
+    enter_bios_setup_on_next_boot = true
```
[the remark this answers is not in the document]
adezxc: Updated all the HCL in the file (also applies to other warnings)

adambarreiro (approved):
Thanks a lot for this feature 👍

PR #1110, opened 2023-08-22
Test suite partitioning

dataclouder:
Experimental method to partition the tests from the acceptance test suite.
In short, it works this way: you run the test with two parameters:
* `-vcd-partitions=3` // this is the number of partitions
* `-vcd-partition-node=1` // this is the identified of the current node

In the other two VCDs, you run the same command, but with `-vcd-partition-node=2` and `-vcd-partition-node=3`. The tests get split in three parts, and each node gets one piece.
The partitioning works by collecting the test names from the current directory, sorting them, and assigning a node number to each name, before the test starts. The operation should give the same result in every VCD, provided that all nodes work on the same branch and commit.

**Prerequisites:**

1. Every test must have a `preTestCheck` and `postTestCheck` function calls
2. All VCDs participating in the test must use the same branch and commit
3. We must not run two or more partitions in the same VCD: it will probably lead to conflicts

**Possible problems:**

* Every node in the partition is independent, and has no connection of other nodes.
* There is no completion check: we can't be sure that all nodes have run their share.
* Users who run partitioned tests must make sure that:
    * They use the same number of partitions in every test call
    * They use different node numbers in each VCD participating to the distributed run (it is possible to run the same partition on multiple nodes, which would be a total waste of resources)
    * They start as many nodes as the number of partitions declared (starting fewer nodes will leave tests out of the suite run)

**Possible enhancements:**
* Every node could produce a log of the tests it ran and the ones it skipped (DONE); 
* The results could be consolidated after completion;
* The node logs could be compared to make sure that all tests were considered by any node;
* A sentinel resource could be created at the start of the test suite to make sure that no calls to the test suite will run on the same VCD.

**Preparation for parallel handling**
This feature would work better if a centralized process were to start all the nodes, and collect the results.
For this reason, the following features were added:
* there is an option `-vcd-partition-tests-file` with the name of the file containing the tests that will run in the current node. If such an option was provided, the node will not attempt to create its own list of tests
* The names of all the test files produced by the node are standardized. If running under a Jenkins job, the names will contain the build number.
* A sentinel file is created at the end of the test suite, containing the exit code of the run.


lvirbalas (approved):
Looks great! Only there's no changelog entry.

dataclouder:
> Looks great! Only there's no changelog entry.

Added

scripts/start-multi-node-test.sh
```diff
@@ -0,0 +1,119 @@
...
+    mv terraform-provider-vcd "node$n"
+    cd "node$n/vcd"
+    pwd
+    index=$(($n-1))
+    cp ${configs[$index]} vcd_test_config.json
+
+    # TODO: this change is only needed while developing the current PR in branch
+    git checkout test-partitioning
+    # TODO: start tmux session with test
+    # go test -tags functional -v -vcd-partitions=$partitions -vcd-partition-node=$n -vcd-short 2>&1 > testacc-node$n.txt
```
Didainius, line 115: Are these for future  or your own notes?

scripts/start-multi-node-test.sh
```diff
@@ -0,0 +1,119 @@
...
+        wanted=$dir/$what
+        if [ -x $wanted ]
+        then
+            echo $wanted
+            return
+        fi
+    done
+}
+
+for needed in tmux git realpath jq
```
Didainius, line 51: Are `jq` or `tmux` needed here? I don't see them being used in the code.

vcd/config_test.go
```diff
@@ -55,7 +56,9 @@ func init() {
 	setStringFlag(&vcdSkipPattern, "vcd-skip-pattern", "VCD_SKIP_PATTERN", "Skip tests that match the pattern (implies vcd-pre-post-checks")
 	setBoolFlag(&skipLeftoversRemoval, "vcd-skip-leftovers-removal", "VCD_SKIP_LEFTOVERS_REMOVAL", "Do not attempt removal of leftovers at the end of the test suite")
 	setBoolFlag(&silentLeftoversRemoval, "vcd-silent-leftovers-removal", "VCD_SILENT_LEFTOVERS_REMOVAL", "Omit details during removal of leftovers")
-
+	setStringFlag(&testListFileName, "vcd-partition-tests-file", "VCD_PARTITION_TESTS_FILE", "Name of the file containing the tests to run in the current partition node")
```
Didainius, line 59: I think this lacks some information for the ones who don't know how to do it. Would a `TESTING.md` paragraph be possible

Didainius (commented):
First few comments on a first look. I think it would be good to have a small paragraph about how to execute core functionality with bare hands (using newly introduced flags) in `TESTING.md`.

scripts/start-multi-node-test.sh
dataclouder: In fact, I forgot to remove the whole script

scripts/start-multi-node-test.sh
dataclouder: same as previous one

vcd/config_test.go
dataclouder: Done

TESTING.md
```diff
@@ -532,6 +533,29 @@ Look at `TestResourceInfoProviders` to see a full example of how to use the meth
 once in the HCL script. If it is not, the variable `testAccProvider` may not get initialised, and if that happens,
 test checks that use the expression `conn := testAccProvider.Meta().(*VCDClient)` will panic.
 
+## Partitioned tests
+
+We can run tests in a partitioned way, by splitting the tests across several VCDs (nodes) each of which will run a given portion of the tests.
+
+To activate this modality, we use the following options:
+
+* `-vcd-partition=N` indicates the number of partitions running the test suite.
```
Didainius, line 542: typo

```suggestion
* `-vcd-partitions=N` indicates the number of partitions running the test suite.
```

TESTING.md
```diff
@@ -532,6 +533,29 @@ Look at `TestResourceInfoProviders` to see a full example of how to use the meth
...
+
+The test assignment, by default, happens by getting the list of all tests, sorting them alphabetically, and then assigning them to each node in sequence. 
+If a tests file name was provided, the node will just use such file as its assignment.
+
+Each node produces several files:
+
+* `{BUILD_NUMBER}-tests-retrieved-in-node-{NODE_NUMBER}-{VCD_VERSION}.txt` All the tests found in the current node
+* `{BUILD_NUMBER}-tests-planned-in-node-{NODE_NUMBER}-{VCD_VERSION}.txt` The tests that will run in the current node
+* `{BUILD_NUMBER}-tests-processed-in-node-{NODE_NUMBER}-{VCD_VERSION}.txt` The tests that have been processed in this node
+* `{BUILD_NUMBER}-out-{NODE_NUMBER}.txt` A sentinel file signifying that the tests in the current node are finished. It contains the error code of the run (0=success, 1=failure)
```
Didainius, line 556: We could probably match these names for `.gitignore` to avoid accidental merge

TESTING.md
dataclouder: Thanks. Fixed

TESTING.md
dataclouder: Done

Didainius (approved):
I have done a few runs on smaller set of tests and it worked - thanks!

TESTING.md
```diff
@@ -532,6 +533,33 @@ Look at `TestResourceInfoProviders` to see a full example of how to use the meth
...
+* `-vcd-partition-tests-file=FileName` (optional) provides the list of tests to run
+
+When partition mode is enabled, the current node will skip all the tests that are not assigned to it.
+
+The test assignment, by default, happens by getting the list of all tests, sorting them alphabetically, and then assigning them to each node in sequence. 
+If a tests file name was provided, the node will just use such file as its assignment.
+
+Each node produces several files:
+
+* `{BUILD_NUMBER}-tests-retrieved-in-node-{NODE_NUMBER}-{VCD_VERSION}.txt` All the tests found in the current node
```
adambarreiro, line 553: While `NODE_NUMBER` and `VCD_VERSION` meaning is trivial (imho), I think that `BUILD_NUMBER` maybe deserves a comment. Like, it is just "an ID" of the test run, or that it's optional (its value will be `LOCAL` if not set).

vcd/resource_vcd_vapp_raw_multi_test.go
```diff
@@ -1,175 +0,0 @@
...
-
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
-	"github.com/vmware/go-vcloud-director/v2/govcd"
-)
-
-// Extends TestAccVcdVAppRawM with multiple VMs
-// To execute this test, run
-// go test -v -timeout 0 -tags multivm -run TestAccVcdVAppRawMulti .
-func TestAccVcdVAppRawMulti(t *testing.T) {
```
adambarreiro, line 17: Are these not needed anymore?

TESTING.md
dataclouder: Docs updated

adambarreiro (approved):

adezxc (approved):
Ran a small subset of tests on my own machine and worked great! Thanks for this improvement

PR #1113, opened 2023-08-31
Patch VM customization tests for correct settings

Didainius:
### About

This PR attempts to fix tests (sample error at the bottom):
* TestAccVcdVAppVmCustomizationSettings
* TestAccVcdStandaloneVmCustomizationSettings

Both of these tests are identical, except that one tests vApp VM, other tests standalone VM.

### Problem

These tests look to attempt setting incorrect configuration which gets rejected by VCD.

UI does not allow setting "Auto generate password" when "Allow local administrator password" is not enabled (the option is greyed out), but currently our tests attempt to do it. As a result - the test gets inconsistent plan. This PR attempts to change tests so that they perform valid actions.

Impossible to set "Auto Generate Password"
image

"Auto Generate Password" can be set when "Allow local administrator password" is enabled
image

### Testing

Ran these tests
* Acceptance 10.4.0, 10.4.1, 10.4.2, 10.5.0
* Binary - attempted to run all 6 generated files on 10.5.0 (but not all of them are designed to be run in this mode) - no failures

#### Sample error

```
=== RUN   TestAccVcdVAppVmCustomizationSettings
    resource_vcd_vapp_vm_customization_test.go:437: Step 1/3 error: After applying this test step, the plan was not empty.
        stdout:


        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place

          # vcd_vapp_vm.test-vm will be updated in-place
          ~ resource "vcd_vapp_vm" "test-vm" {
                id                             = "urn:vcloud:vm:11999d4b-6822-4733-a0fd-75c293005b7e"
                name                           = "TestAccVcdVAppVmCustomizationSettingsVM"
                # (35 unchanged attributes hidden)

              ~ customization {
                  ~ auto_generate_password              = false -> true
                    # (8 unchanged attributes hidden)
                }
            }

        Plan: 0 to add, 1 to change, 0 to destroy.
   ...

--- FAIL: TestAccVcdVAppVmCustomizationSettings (257.85s)

```

### DIFF
--- a/.changes/v3.11.0/1113-notes.md
+++ b/.changes/v3.11.0/1113-notes.md
@@ -1,0 +1,2 @@
+* Patch tests `TestAccVcdVAppVmCustomizationSettings` and
+  `TestAccVcdStandaloneVmCustomizationSettings` to use valid guest customization settings [GH-1113]
--- a/vcd/resource_vcd_vapp_vm_customization_test.go
+++ b/vcd/resource_vcd_vapp_vm_customization_test.go
@@ -449,7 +449,8 @@ func TestAccVcdVAppVmCustomizationSettings(t *testing.T) {
 					resource.TestCheckResourceAttr("vcd_vapp_vm.test-vm", "customization.#", "1"),
 					resource.TestCheckResourceAttr("vcd_vapp_vm.test-vm", "customization.0.enabled", "true"),
 					resource.TestCheckResourceAttr("vcd_vapp_vm.test-vm", "customization.0.change_sid", "true"),
-					resource.TestCheckResourceAttr("vcd_vapp_vm.test-vm", "customization.0.allow_local_admin_password", "false"),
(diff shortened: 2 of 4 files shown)


lvirbalas (approved):
The only question is why did it work before? :)

adezxc (approved):

Didainius:
> The only question is why did it work before? :)

The answer appears to be API v37.0.

adambarreiro (approved):
LGTM!
My only thought is, should we warn the users in our documentation, or is it just too obvious?

Didainius:
> LGTM!
> My only thought is, should we warn the users in our documentation, or is it just too obvious?

Fair point. I forgot that, but added now:
https://github.com/vmware/terraform-provider-vcd/pull/1113/commits/4b617431d6091a6935fd1cd3917678ab02199aa7#diff-8cc1a991a7ce193045c19f30b8b87a91ac35b8c5efacd1ea9ec429980eae66dfR623-R625

PR #1114, opened 2023-08-31
Add SRP file to be compliant with VMware policies

adambarreiro:
This PR just adds a special file that is required to be compliant with VMware policies.


adambarreiro:
Closing as this file is not needed anymore in this repo.

PR #1117, opened 2023-09-07
Add vcd_rde_behavior_invocation data source to invoke behaviors

adambarreiro:
## Overview
This PR adds a **new data source** to invoke RDE Interface/Type Behaviors.

## Detailed description
We onboarded RDE Interface/Type Behaviors [in v3.10.0](https://github.com/vmware/terraform-provider-vcd/pull/1074) of the VCD Provider. For simplicity, let's say that a Behavior is an equivalent term for "function", so we're talking about invoking/calling functions, that are specified in RDE Interfaces/Types and invoked on RDEs. One could invoke them programatically using the [Go SDK](https://github.com/vmware/go-vcloud-director/blob/main/govcd/defined_entity.go#L662), but users may be in the need of invoking them directly using Terraform.

For that purpose, this PR onboards a new data source `vcd_rde_behavior_invocation` that behaves similarly to the [HTTP data source created by HashiCorp](https://registry.terraform.io/providers/hashicorp/http/latest/docs/data-sources/http), in the sense that it is an imperative concept (calling a function, calling a HTTP method) introduced in a declarative toolset.

That's why it was created as a data source. While it would make more sense to have it as a resource, as conceptually this would fit more with the term "invocation" (as we know, a function invocation can mutate/write things), because invoking Behaviors can modify RDEs. But the implementation of a resource for such a "imperative" concept is tricky. You can check the section "Caveats of the resource implementation" to know more.

The data source is pretty trivial, it only has a Read operation that invokes the Behavior on every refresh, with the given arguments/metadata (if given), and writes the result of the invocation in a Computed field. This field can be pretty much anything, as the invocation returns the raw result (it could be a JSON, plain text, etc).

## Caveats of the resource implementation
As we have Create, Update, Read and Delete, we would need a required argument `invoke_on_every_refresh` to decide whether we want to invoke the Behavior on every Read, or only during Create/Update. This has also an odd drawback that could be undesired, a Refresh and an Update would invoke the Behavior twice. We could make Update a no-op, but we need to be able to update arguments/metadata for the invocation. If we make Update a no-nop, we need to make everything `ForceNew: true`, but this would give us the same situation.

Also, Delete would be a no-op in any case.

While resource fits more with the concept of "invoking Behaviors can mutate RDEs", in reality the implementation is tricky and would be more complicated for end users. Also, we have the precedent of the HTTP Provider giving the `http` data source, despite one can do PUT/POST/DELETE calls with it. Probably they faced the same situation, which gets trivially solved with a data source.

## Other things included in this PR

* **Improvement:** The field `rde_type_id` from resource `vcd_rde` does not have `ForceNew: true` anymore. This was a bit inconvenient when one wanted to upgrade an RDE from one RDE Type to a newer version of the same, as it forced its deletion and re-creation, which may not suit all users needs. Now the RDE Type can be updated without any previous deletion, allowing smoother RDE Type version upgrades. **_Note:_** As stated in the updated documentation, one should only upgrade the RDE Type version. VCD API doesn't allow updating to a different type or downgrading the version.
 
* **Fixes** a bug that made impossible to delete `vcd_rde_type_behavior_acl` resources when the Access Level is the last one in the Behavior. This fix is automatically done taking the changes in https://github.com/vmware/go-vcloud-director/pull/615. The bug happened because the Provider calls `SetBehaviorAccessControls` with a nil slice of Access Levels, and the method was not considering this case as special.

* **Fixes** the resource `vcd_rde_type_behavior_acl` to avoid race conditions when creating, updating or deleting more than one Access Level. This is fixed by using a mutex whenever `SetBehaviorAccessControls` method is used. The problem with `SetBehaviorAccessControls` is that, as the name implies, sets whatever Access Levels are specified in the input, so two calls at the same time can override each other.


vcd/datasource_vcd_rde_behavior_invocation.go
```diff
@@ -0,0 +1,66 @@
...
+		return diag.Errorf("[RDE Behavior Invocation] could not retrieve the RDE with ID '%s': %s", rdeId, err)
+	}
+	result, err := rde.InvokeBehavior(behaviorId, types.BehaviorInvocation{
+		Arguments: d.Get("arguments").(map[string]interface{}),
+		Metadata:  d.Get("metadata").(map[string]interface{}),
+	})
+	if err != nil {
+		return diag.Errorf("[RDE Behavior Invocation] could not invoke the Behavior of the RDE with ID '%s': %s", rdeId, err)
+	}
+	d.SetId(rdeId + "|" + behaviorId) // Invocations are not real entities, so we make an artificial ID.
```
lvirbalas, line 63: Do we have such ID format (RDE ID + Behavior ID) anywhere else? I'm trying to understand whether this format is unique to invocations or not. Thanks!

vcd/datasource_vcd_rde_behavior_invocation_test.go
```diff
@@ -0,0 +1,121 @@
...
+  rde_id                  = vcd_rde.rde.id
+  behavior_id             = vcd_rde_interface_behavior.behavior.id
+  arguments = {
+    "arg1" : "not_used"
+  }
+  metadata = {
+    "meta1" : "not_used"
+  }
+}
+`
```
lvirbalas, line 121: Maybe it would be useful to have in the test output, which would check that the result is there too?

vcd/resource_vcd_rde_test.go
```diff
@@ -507,21 +518,21 @@ func addRightsToTenantUser(t *testing.T, vcdClient *VCDClient, vendor, nss strin
 	}
 	err = role.AddRights(rightsToAdd)
 	if err != nil {
-		t.Errorf("could not add rights '%v' to role '%s'", rightsToAdd, role.GlobalRole.Name)
+		t.Fatalf("could not add rights '%v' to role '%s'", rightsToAdd, role.GlobalRole.Name)
```
lvirbalas, line 521: Curious, why does it have to be fatal and error is not enough?

vcd/resource_vcd_rde_type_behavior_acl.go
```diff
@@ -50,6 +50,11 @@ func resourceVcdRdeTypeBehaviorAccessLevelCreate(ctx context.Context, d *schema.
 }
 
 func resourceVcdRdeTypeBehaviorAccessLevelCreateOrUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}, operation string) diag.Diagnostics {
+	// A mutex is required as we use the method SetBehaviorAccessControls which sets and overrides all Access Levels given in the
+	// input. If two or more resources are created/updated at the same time, they would clash with each other.
+	key := "vcd_rde_type_behavior_acl"
```
lvirbalas, line 55: Thinking on a use case where someone may have "a hundred" RDE resources. Do we really need to lock **all** behaviors on ACL changes? No way to narrow down the scope of the lock?

website/docs/r/rde.html.markdown
```diff
@@ -88,12 +88,39 @@ resource "vcd_rde" "my-rde" {
...
+```
+
 ## Argument Reference
 
 The following arguments are supported:
 
 * `org` - (Optional) Name of the [Organization](/providers/vmware/vcd/latest/docs/resources/org) that will own the RDE, optional if defined at provider level.
-* `rde_type_id` - (Required) The ID of the [RDE Type](/providers/vmware/vcd/latest/docs/data-sources/rde_type) to instantiate.
+* `rde_type_id` - (Required) The ID of the [RDE Type](/providers/vmware/vcd/latest/docs/data-sources/rde_type) to instantiate. It supports
+  updating to a **newer** version of the **same** RDE Type.
```
lvirbalas, line 123: I'd code-quote "version" to make it clear it's a field being referenced here.

vcd/resource_vcd_rde_test.go
adambarreiro: The Errorf marks the test as failed but continues execution, which can produce some nil pointers and undesired behavior. On the other hand Fatalf marks the test as failed and stops the execution of the test immediately.

vcd/resource_vcd_rde_type_behavior_acl.go
adambarreiro: You're right, it could be narrowed down to the RDE Type, like
```
vcdMutexKV.kvLock("vcd_rde_type_behavior_acl" + rdeTypeId)
```
I will change it and test it again.

vcd/datasource_vcd_rde_behavior_invocation.go
adambarreiro: I don't think so. The ID here is useless, it's more a requirement of Terraform to register and control the entity in the state, so I chose two of the attributes that make an invocation unique: The RDE and the Behavior itself. The `|` was just to choose an "easy to spot" separator.

But I'm open for suggestions, maybe @dataclouder @adezxc or @Didainius can jump in and suggest something else.

lvirbalas: The two IDs with a separator look good, esp. if no such combination is anywhere else, as that will aid in troubleshooting.

(I will mark this conversation as resolved, so it doesn't linger, but it's open for everyone to chime in/re-open, if needed.)

vcd/datasource_vcd_rde_behavior_invocation_test.go
adambarreiro: Actually the output (the `result` attribute) is checked on lines 48-51, do you mean a different thing?

lvirbalas: I have missed, it's all good then, thanks!

website/docs/r/rde.html.markdown
adambarreiro: Done, thanks!

adezxc:
One thing that comes to mind: could we have the `invoke_on_every_refresh` as an argument for the datasource instead? I'm not sure what could happen exactly when a behaviour is invoked, but just as a thought so users have more control instead of having to remove/add datasources.

lvirbalas (approved):

adambarreiro:
> One thing that comes to mind: could we have the `invoke_on_every_refresh` as an argument for the datasource instead? I'm not sure what could happen exactly when a behaviour is invoked, but just as a thought so users have more control instead of having to remove/add datasources.

It could be a good option indeed, and would be more optimal as one could use a variable to switch it on/off (more CI/CD friendly than adding/removing code). Also the caveat mentioned in the docs would disappear:
```
## Known caveats

* Executing a `terraform destroy` will cause a refresh that will invoke the Behavior. To avoid this situation, you can
  remove the `vcd_rde_behavior_invocation` data source from the HCL configuration before running the destroy operation,
  or use the `-refresh=false` flag (this will skip the refresh for all resources).
```

website/docs/d/rde_behavior_invocation.html.markdown
```diff
@@ -0,0 +1,97 @@
...
+
+The following arguments are supported:
+
+* `rde_id` - (Required) The ID of the [RDE](/providers/vmware/vcd/latest/docs/resources/rde) which Behavior will be invoked
+* `behavior_id` - (Required) The ID of the [RDE Interface Behavior](/providers/vmware/vcd/latest/docs/resources/rde_interface_behavior) or
+  the [RDE Interface Behavior](/providers/vmware/vcd/latest/docs/resources/rde_type_behavior) to invoke
+* `arguments` - (Optional) A map with the arguments of the invocation
+* `metadata` - (Optional) A map with the metadata of the invocation
+* `invoke_on_refresh` - (Optional) Defaults to `true`, indicates whether the Behavior should be invoked on every refresh. It can be useful
+  to suppress the invocation on the refresh step of `terraform destroy` operations, or to stop invoking a given Behavior temporarily, for example.
```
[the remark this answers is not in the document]
adambarreiro: Nice, applied.

website/docs/d/rde_behavior_invocation.html.markdown
```diff
@@ -0,0 +1,97 @@
...
+
+# vcd\_rde\_behavior\_invocation
+
+~> This feature is **experimental** and may change in future
+
+Provides the capability of invoking an existing [RDE Interface Behavior](/providers/vmware/vcd/latest/docs/resources/rde_interface_behavior)
+or [RDE Type Behavior](/providers/vmware/vcd/latest/docs/resources/rde_type_behavior) in VMware Cloud Director.
+
+The RDE Behavior is invoked on every refresh operation. To stop invoking it, one can simply remove this data source
+from the HCL code.
```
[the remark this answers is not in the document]
adambarreiro: Good catch, thanks!

vcd/resource_vcd_rde.go
```diff
@@ -40,7 +40,6 @@ func resourceVcdRde() *schema.Resource {
 			"rde_type_id": {
 				Type:        schema.TypeString,
 				Required:    true,
-				ForceNew:    true,
```
Didainius, line 43: You will have to add the relevant `vcd.ResourceSchema-*` test to `skip-upgrade-tests.txt`

Didainius (approved):

vcd/resource_vcd_rde.go
adambarreiro: Agh, I always forget about that, thanks for spotting it!

adambarreiro: After reviewing the tests, it seems that `vcd_resource_schema` is unable to read the `ForceNew` attribute from Terraform internals, making not possible to detect ForceNew changes in schemas. Hence it seems the test passes! So nothing to do here.

vcd/datasource_vcd_rde_behavior_invocation.go
```diff
@@ -0,0 +1,74 @@
+package vcd
+
+import (
+	"context"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+)
+
+// This data source is quite special, as it represents an imperative call to a function (aka invoking a behavior)
```
dataclouder, line 10:
```suggestion
// This data source is unusual, as it represents an imperative call to a function (aka invoking a behavior)
```

vcd/datasource_vcd_rde_behavior_invocation.go
```diff
@@ -0,0 +1,74 @@
...
+// in the declarative world of Terraform. Despite being a data source, whose goal is to perform Read-only operations, invocations
```
dataclouder, line 11:
```suggestion
// in the declarative world of Terraform. Despite being a data source, whose goal is to perform read-only operations, invocations
```

dataclouder (approved):

vcd/datasource_vcd_rde_behavior_invocation.go
adambarreiro: Changed as well

PR #1119, opened 2023-09-17
Fix catalog item lists

dataclouder:
Fix detection of catalog items

The list of catalog media, by mistake, was including vApp templates

Also: increase timeout for test `TestAccVcdSubscribedCatalog`


Didainius (approved):

adambarreiro (approved):

lvirbalas (approved):
Exciting! Looking forward how next test run goes :)

adezxc (approved):

PR #1122, opened 2023-09-28
Add RDE Type hooks

adambarreiro:
## Overview

This PR adds a new feature to existing `vcd_rde_type` resource and data source: The `hook` blocks.

A `hook` block maps a lifecycle event of a given Runtime Defined Entity (`vcd_rde`), one of:
- `PostCreate`: After the RDE of this type is created
- `PostUpdate`: After the RDE of this type is updated
- `PreDelete`: Before the RDE of this type is deleted
- `PostDelete`: After the RDE of this type is deleted

with an existing Behavior. This allows customers to automatically invoke the desired Behavior when these lifecycle events are triggered.

For example, let's imagine there's a Behavior that calls a webhook to send a message to a Slack channel.
If we have an RDE Type with the new block:
```
hook {
 event = "PostCreate"
 behavior_id = ... # my "send to Slack" behavior ID
}
```

Whenever a RDE **of that Type** is created, it will send a message to Slack!

## Implementation details

A new block `hook` is added to `vcd_rde_type` resource and data source. This has the sub-properties `event` (the lifecycle event) and `behavior_id`.

The Behavior referenced by the ID must exist already.

## Tests

Tests passed in VCD 10.5.0

### DIFF
--- a/vcd/resource_vcd_rde_type.go
+++ b/vcd/resource_vcd_rde_type.go
@@ -169,11 +169,11 @@ func resourceVcdRdeTypeCreate(ctx context.Context, d *schema.ResourceData, meta 
 }
 
 func getRdeTypeHooksFromSchema(d *schema.ResourceData) map[string]string {
-	if _, ok := d.GetOk("hooks"); !ok {
+	if _, ok := d.GetOk("hook"); !ok {
 		return nil
 	}
 	hooks := map[string]string{}
-	rawHooks := d.Get("hooks").(*schema.Set).List()
+	rawHooks := d.Get("hook").(*schema.Set).List()
 	for _, h := range rawHooks {
 		hookBlock := h.(map[string]interface{})
 		hooks[hookBlock["event"].(string)] = hookBlock["behavior_id"].(string)
--- a/vcd/resource_vcd_rde_type.go
+++ b/vcd/resource_vcd_rde_type.go
@@ -16,6 +16,21 @@ import (
 	"github.com/vmware/go-vcloud-director/v2/util"
 )
 
+var rdeTypeHook = &schema.Resource{
+	Schema: map[string]*schema.Schema{
+		"event": {
+			Type:        schema.TypeString,
+			Required:    true,
+			Description: "Event that will invoke the Behavior, one of PostCreate, PostUpdate, PreDelete, PostDelete",
+		},
+		"behavior_id": {
+			Type:        schema.TypeString,
+			Required:    true,
+			Description: "Existing Behavior that will be automatically invoked when the RDE of this RDE Type triggers the event",
+		},
+	},
+}
+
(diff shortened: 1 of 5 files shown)


dataclouder (approved):

website/docs/r/rde_type.html.markdown
```diff
@@ -85,6 +85,41 @@ resource "vcd_rde_type" "my_rde_type" {
...
+data "vcd_rde_interface" "my_interface" {
+  vendor  = "bigcorp"
+  ns      = "tech1"
+  version = "1.2.3"
+}
+
+resource "vcd_rde_interface_behavior" "my_behavior" {
+  interface_id = vcd_rde_interface.my_interface.id
+  name         = "MyBehavior"
+  description  = "Adds a node to the cluster.\nParameters:\n  clusterId: the ID of the cluster\n  node: The node address\n"
```
lvirbalas, line 100: This description seems out of place?

adambarreiro: Actually I got this description from [the API docs](https://developer.vmware.com/apis/vmware-cloud-director/latest/cloudapi/1.0.0/interfaces/id/behaviors/post/), as it is probably taken from a real Behavior, rather putting a made up one that could be misleading (ideally the description should define the contract of the Behavior).

But I can change it to something less specific, like 
```
"Calls an example  Behavior.\nParameters:\n  parameter1: the first param\n  parameter2: the second param\n"
```
that way it would look less odd.

lvirbalas: A! In the API docs it looks a bit less peculiar, as other fields contain information about "containerCluster". Here, we don't, so it feels better to have a generic description.

lvirbalas (approved):

vcd/resource_vcd_rde_type.go
```diff
@@ -16,6 +16,25 @@ import (
...
 )
 
+func getRdeTypeHookSchema(computed bool) *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"event": {
+				Type:        schema.TypeString,
+				Required:    !computed,
+				Computed:    computed,
+				Description: "Event that will invoke the Behavior, one of PostCreate, PostUpdate, PreDelete, PostDelete",
```
adezxc, line 26: What happens if a random string is passed here? Just checking if we should validate this field

adezxc (approved):

vcd/resource_vcd_rde_type.go
adambarreiro: Well spotted! Added a validation function there.

Didainius (approved):
Thanks. Test also passed in 10.4.0, 10.4.1, 10.4.2

PR #1123, opened 2023-10-01
Add property `upload_any_file` to `vcd_catalog_media`

dataclouder:
<s>Setting this property to true we can upload any file. The default, without using this property, only allows .ISO files</s>

Replaced by PR #1124


lvirbalas (approved):

adambarreiro (approved):
Test passed on 10.5.0 for me, thanks!

PR #1130, opened 2023-10-09
Vapp template lease handling

dataclouder:
Add lease handling to `vcd_catalog_vapp_template`


website/docs/r/catalog_vapp_template.html.markdown
```diff
@@ -62,6 +62,9 @@ The following arguments are supported:
 * `upload_piece_size` - (Optional) - Size in MB for splitting upload size. It can possibly impact upload performance. Default 1MB
 * `metadata` -  (Deprecated) Use `metadata_entry` instead. Key/value map of metadata to assign to the associated vApp Template
 * `metadata_entry` - (Optional; *v3.8+*) A set of metadata entries to assign. See [Metadata](#metadata) section for details.
+* `lease` - (Optional *v3.11+*) the information about the vApp Template lease. It includes the field below. When this section is
+  included, the field is mandatory. If lease value is higher than the one allowed for the whole Org, we get an error
```
lvirbalas, line 66: Nit, for consistency:

```suggestion
* `lease` - (Optional *v3.11+*) The information about the vApp Template lease. It includes the field below. When this section is
  included, the field is mandatory. If lease value is higher than the one allowed for the whole Org, we get an error.
```

vcd/resource_vcd_catalog_vapp_template.go
```diff
@@ -256,6 +294,35 @@ func resourceVcdCatalogVappTemplateDelete(_ context.Context, d *schema.ResourceD
...
+		// No lease block: we read the lease defaults from the Org
+		adminOrg, err := vcdClient.GetAdminOrgFromResource(d)
+		if err != nil {
+			return fmt.Errorf("error retrieving admin Org from parent Org in vApp Template %s: %s", vAppTemplate.VAppTemplate.Name, err)
+		}
+		if adminOrg.AdminOrg.OrgSettings == nil || adminOrg.AdminOrg.OrgSettings.OrgVAppTemplateSettings == nil {
+			return fmt.Errorf("error retrieving Org lease settings")
+		}
+		storageLease = *adminOrg.AdminOrg.OrgSettings.OrgVAppTemplateSettings.StorageLeaseSeconds
+	}
```
adambarreiro, line 315: I have a doubt about this `else` part. Before this PR introduced the `lease` block, we did nothing in terms of lease. Now, if `lease` is not present, we take Org's value and use it.
Does it mean that previous code was wrong due to lacking of this feature? Should this be done by VCD and it is not? Or it is just custom logic that we are introducing?

adambarreiro (commented):
Just a small doubt

vcd/resource_vcd_catalog_vapp_template.go
dataclouder: Before, VCD applied the Org defaults without asking.
Now, we have an explicit lease, and we need to match it to the Org defaults to make the plan succeed.
The default would work just as well without it, but we need to take care of the hidden values in the state.

adambarreiro: > Before, VCD applied the Org defaults without asking.
Then we don't need the `else`, do we? VCD will apply the defaults, then they will be read back when Read comes:
```
leaseInfo, err := vAppTemplate.GetLease()
...
err = d.Set("lease", leaseData)
```
Actually I did a try by removing the `else` contents, and not setting any `lease` block, and the state was correctly read:

```
            "lease": [
              {
                "storage_lease_in_sec": 64800
              }
            ],
```
(I configured the Org to have 18h lease)

website/docs/d/catalog_vapp_template.html.markdown
```diff
@@ -118,6 +118,11 @@ The following arguments are supported:
 * `metadata` - (Deprecated) Use `metadata_entry` instead. Key/value map of metadata for the associated vApp template.
 * `metadata_entry` - A set of metadata entries assigned to this vApp Template. See [Metadata](/providers/vmware/vcd/latest/docs/resources/catalog_vapp_template#metadata) section for details.
 
+* `lease` - (*v3.11+*) - The information about the vApp Template lease. It includes the following field:
+  * `storage_lease_in_sec` - How long the vApp Template is available before being automatically deleted or marked as expired. 0 means never expires (or maximum allowed by parent Org allows).
+
+
+
```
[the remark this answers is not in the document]
dataclouder: removed

website/docs/r/catalog_vapp_template.html.markdown
dataclouder: Fixed

vcd/resource_vcd_catalog_vapp_template.go
dataclouder: OK. It seems that I overthought the operation. That `else` was not needed, and I removed it.
Thanks for checking it

adambarreiro (approved):
LGTM!

website/docs/r/catalog_vapp_template.html.markdown
```diff
@@ -62,6 +62,9 @@ The following arguments are supported:
 * `upload_piece_size` - (Optional) - Size in MB for splitting upload size. It can possibly impact upload performance. Default 1MB
 * `metadata` -  (Deprecated) Use `metadata_entry` instead. Key/value map of metadata to assign to the associated vApp Template
 * `metadata_entry` - (Optional; *v3.8+*) A set of metadata entries to assign. See [Metadata](#metadata) section for details.
+* `lease` - (Optional *v3.11+*) The information about the vApp Template lease. It includes the field below. When this section is
+  included, the field is mandatory. If lease value is higher than the one allowed for the whole Org, we get an error
+   * `storage_lease_in_sec` - How long the vApp Template is available before being automatically deleted or marked as expired. 0 means never expires (or maximum allowed by parent Org). Regular values accepted from 3600+.
```
Didainius, line 67: * Could we add this section to the existing example snippet so that it is clearer?

Didainius (approved):

adezxc (approved):

lvirbalas (approved):

website/docs/r/catalog_vapp_template.html.markdown
dataclouder: done

PR #1131, opened 2023-10-11
Improve Behavior definitions to allow complex Execution properties

adambarreiro:
## Problem

With v3.10.0, the new resources `vcd_rde_interface_behavior` and `vcd_rde_type_behavior` allowed to define Behaviors for RDE Interfaces and RDE Types respectively.

The Behaviors could be defined with the generic map `execution` argument, but the problem is that Terraform only supports simple maps of string->string, where one could be able to define executions of more complex structures, like a WebHook type does:

```
{
      "type":"WebHook",
      "id":"testWebHook",
      "href":"https://hooks.slack.com:443/services/T07UZFN0N/B01EW5NC42D/rfjhHCGIwzuzQFrpPZiuLkIX",
      "key":"secretKey",
      "execution_properties": { <--- PROBLEM! This is not possible with a TypeMap!
         "template":{
            "content":"<template_content_string>"
         },
         "_secure_token":"secureToken",
         "invocation_timeout":7
      }
   }
```

## Solution

Add a new attribute `execution_json` which is a plain string that represents a JSON. Both `execution` and `execution_json` will live together, the first one being now just syntactic sugar for the latter.

The challenge here is that the WebHook behavior type uses a special `_secure_` and `_internal_` prefixes for fields that are write-only. So once created, they're never recovered from VCD. This causes dirty plans all the time, where the Provider asks for updates all the time. To avoid that, a new DiffSuppressFunc has been implemented: `hasBehaviorExecutionChanged`, which ignores secure fields if the attribute `always_update_secure_execution_properties` is disabled.

## Tests

☑️ Acceptance tests
```
go test -tags rde -v -timeout 0
```

☑️  Binary tests
```
make test-binary-prepare
cd vcd/test-artifacts
./test-binary.sh tags rde
```

☑️ Upgrade tests
```
make test-upgrade-prepare
cd vcd/test-artifacts
./test-binary.sh upgrade v3.13.0 v3.14.0
```


vcd/resource_vcd_rde_interface_behavior.go
```diff
@@ -52,6 +74,30 @@ func resourceVcdRdeInterfaceBehavior() *schema.Resource {
...
+	err := json.Unmarshal([]byte(oldValue), &unmarshaledOldJson)
+	if err != nil {
+		util.Logger.Printf("[ERROR] Could not unmarshal old value JSON: %s", oldValue)
+		return false
+	}
+	err = json.Unmarshal([]byte(newValue), &unmarshaledNewJson)
+	if err != nil {
+		util.Logger.Printf("[ERROR] Could not unmarshal new value JSON: %s", newValue)
+		return false
+	}
```
Didainius, line 91: Thought - would a failed unmarshal really mean `false` (no changes)

Didainius (approved):
Tests pass on 10.6 and 10.4

vcd/resource_vcd_rde_interface_behavior.go
adambarreiro: Yes, this got me for a second. I think the problem is the function name, which is telling the opposite of what it is really doing. I have changed it to `isBehaviorExecutionEqual` which is now more accurate.

When `hasBehaviorExecutionChanged` returns `false`, the diff is not suppressed. So yeah, it should be the other way around, hence the name change.

Now, if `isBehaviorExecutionEqual` returns `false`, it will mean that the diff is not suppressed, which makes more sense semantically.

vcd/resource_vcd_rde_type_behavior_and_acl_test.go
```diff
@@ -199,10 +225,20 @@ resource "vcd_rde_interface_behavior" "behavior2" {
   rde_interface_id = vcd_rde_interface.interface.id
   name             = "{{.Name}}2"
   description      = "{{.Description}}"
-  execution = {
```
lvirbalas, line 202: Curious, why previous way is removed from the test, we don't want to cover it anymore?

website/docs/r/rde_interface_behavior.html.markdown
```diff
@@ -47,12 +47,22 @@ resource "vcd_rde_interface_behavior" "my_behavior2" {
 resource "vcd_rde_interface_behavior" "my_behavior3" {
   rde_interface_id = vcd_rde_interface.my_interface.id
   name             = "MyBehavior3"
-  execution = {
+
```
lvirbalas, line 50: Why do we remove previous `execution` way from example? Do we prefer `execution_json` exclusively?

website/docs/r/rde_type_behavior.html.markdown
```diff
@@ -49,22 +49,88 @@ resource "vcd_rde_type_behavior" "my_rde_type_behavior" {
...
-* `execution` - (Required) A map that specifies the Behavior execution mechanism.
+* `execution_json` - (Optional; *v3.14*) A string representing a valid JSON that specifies the Behavior execution mechanism.
   You can find more information about the different execution types, like `WebHook`, `noop`, `Activity`, `MQTT`, `VRO`, `AWSLambdaFaaS`
-  and others [in the Extensibility SDK documentation](https://vmware.github.io/vcd-ext-sdk/docs/defined_entities_api/behaviors)
+  and others [in the Extensibility SDK documentation](https://vmware.github.io/vcd-ext-sdk/docs/defined_entities_api/behaviors).
+  One of `execution_json` or `execution` must be set.
+* `always_update_secure_execution_properties` - (Optional; *v3.14*) Useful to update execution properties marked with `_secure_` and `_internal_`
+  as these are not retrievable from VCD, so they are not saved in state. Setting this to `true` will make the provider
+  to ask for updates whenever there is a secure property in the execution of the Behavior
+* `execution` - (Optional) A map that specifies the Behavior execution mechanism, this is just a simplification of `execution_json` that
```
lvirbalas, line 132: So, my overall question is why we remove `execution` way from test and example, if we still leave it available.

website/docs/r/rde_interface_behavior.html.markdown
adambarreiro: It got me a few mins to understand what was going on here, as I didn't remove the `execution` example, but added a new one with `execution_json`. I think it's just GitHub diff view messing around, if you open the rendered file, you can see both are there now.

vcd/resource_vcd_rde_type_behavior_and_acl_test.go
adambarreiro: There's still the `vcd_rde_interface_behavior.behavior1` test resource which uses the "old" argument, but it doesn't show as it was not modified.

website/docs/r/rde_type_behavior.html.markdown
adambarreiro: I think both cases is diff view messing around.

website/docs/r/rde_interface_behavior.html.markdown
lvirbalas: Oh, indeed! Thanks!

lvirbalas (approved):

PR #1135, opened 2023-10-14
Fix rights container test

dataclouder:
This PR contains the improved version of `TestAccVcdRightsContainers`.

The old version of the test creates one data source for each type (right, role, global role, rights bundle)
with the risk of taking a different element in different takes, and missing some nuances.

The new version of the test creates one data source for each element of each type (except `vcd_right`, for which we only
take the ones with 3 or more implied rights)

## TESTING

Run the old and new test with verbosity (using several VCD versions) to appreciate the differences
```
go test -tags functional -run TestAccVcdRightsContainers  -vcd-verbose -v -timeout 0
```


vcd/datasource_vcd_rights_container2_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+// This file contains the improved version of TestAccVcdRightsContainers.
```
lvirbalas, line 13: Do we still need the old version?

dataclouder: Up to the team to decide. The old version takes about 30 seconds to run. If we want to keep it, it is not a big dent into the total running time.

lvirbalas (approved):

vcd/datasource_vcd_rights_container2_test.go
adambarreiro: I was going to ask the same 😅 .
I'd vote to remove it as the new one is testing a "superset" of what the old one tested, so it is a bit redundant IMHO. But as the comment is quite self-explanatory and the old one is quick to run, I won't complain if we decide to keep it.

adambarreiro (approved):

vcd/datasource_vcd_rights_container2_test.go
adezxc: I think so aswell, it tests the same functionality, I'd vote for removing it.

adezxc (approved):

vcd/datasource_vcd_rights_container2_test.go
dataclouder: Removed

PR #1140, opened 2023-10-19
Fix resource list usage of "parent" with entities depending on Org 

dataclouder:
Entities depending directly on an organization should use the `"org"` field to identify its parent.
The `"parent"` field, which is intended for intermediate entities, should not be used for this purpose.
However, if users fill the "parent" field instead of "org", they get an ambiguous error, which doesn't explain what has happened.

This change allows `vcd_resource_list` to consider both `"org"` and `"parent"` fields, as well as the `"org"`  field in the provider block.
The change is transparent for users. If the `"parent"` field was used instead of `"org`", the organization is retrieved without complaints. In case of error, they get a clear message about what is missing.

### DIFF
--- a/vcd/datasource_vcd_resource_list.go
+++ b/vcd/datasource_vcd_resource_list.go
@@ -148,7 +148,7 @@ func getPvdcList(d *schema.ResourceData, meta interface{}) (list []string, err e
 func getVdcGroups(d *schema.ResourceData, meta interface{}) (list []string, err error) {
 	client := meta.(*VCDClient)
 
-	org, err := client.GetAdminOrgByName(d.Get("org").(string))
+	org, err := client.GetAdminOrg(firstNonEmpty(d.Get("org").(string), d.Get("parent").(string)))
 	if err != nil {
 		return list, err
 	}
@@ -203,7 +203,8 @@ func externalNetworkList(d *schema.ResourceData, meta interface{}) (list []strin
 
 func rightsList(d *schema.ResourceData, meta interface{}) (list []string, err error) {
 	client := meta.(*VCDClient)
-	org, err := client.GetAdminOrg(d.Get("org").(string))
+
+	org, err := client.GetAdminOrg(firstNonEmpty(d.Get("org").(string), d.Get("parent").(string)))
 	if err != nil {
 		return list, err
 	}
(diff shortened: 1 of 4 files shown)


vcd/datasource_vcd_resource_list.go
```diff
@@ -148,7 +148,7 @@ func getPvdcList(d *schema.ResourceData, meta interface{}) (list []string, err e
 func getVdcGroups(d *schema.ResourceData, meta interface{}) (list []string, err error) {
 	client := meta.(*VCDClient)
 
-	org, err := client.GetAdminOrgByName(d.Get("org").(string))
+	org, err := client.GetAdminOrg(firstNonEmpty(d.Get("org").(string), d.Get("parent").(string)))
```
lvirbalas, line 151: Supporting both fields for
the same purpose will cause some issues down the road (like ambiguity). What about just throwing a meaningful error if 'parent' is used in a wrong context?

adambarreiro (approved):
Thanks for the fix!

vcd/datasource_vcd_resource_list.go
dataclouder: The documentation still mentions `"org"` as the necessary field.
Using `"parent"` could be a honest mistake that is difficult to diagnose when happens (it fooled me as well in a case). This change makes the error disappear without side effects.

Anyway, the main point of the fix is making sure that the `"org"` from the provider is being used when the one in the data source is missing. That was not always the case, and the fix for `"parent"` was a cheap addition.

carmine73:
Is there a reason why org defined at provider level is not used? Is it just for resource and not for datasource?
```
provider "vcd" {
  ...
  org                  = var.org.name # default for resources
  ...
}
```

dataclouder:
> Is there a reason why org defined at provider level is not used? Is it just for resource and not for datasource?

It is for both. The reason it was not used is just a plain bug.

carmine73:
> Anyway, the main point of the fix is making sure that the "org" from the provider is being used when the one in the data source is missing.

Oh yes, you already wrote it.
Thanks

.changes/v3.11.0/1140-bug-fixes.md
```diff
@@ -0,0 +1 @@
+* Fix organization retrieval in `vcd_resource_list` when users fill the "parent" field instead of "org" [GH-1140]
```
lvirbalas, line 1: Changelog is missing a note about fixing retrieval of the "org" field from the provider block.

lvirbalas: BTW, "parent" and "org" should be quoted in code-quotes :)

dataclouder: fixed

adezxc (approved):

lvirbalas (approved):

Issue #1134, opened 2023-10-13
How should I use the `ldap_mode = "SYSTEM"` and is it even possible?


cgrard:
Hi there,

I'm not sure this is a bug or missing/unclear documentation yet but I'm facing an issue with the LDAP settings and the vcd_org_ldap resource creation: https://registry.terraform.io/providers/vmware/vcd/latest/docs/resources/org_ldap#ldap_mode

I have a VMware Virtual Cloud that has a system identity provider defined (which is an Active Directory server) , so in my terraform code, according to the documentation, I should be able to use the `SYSTEM` value for the `ldap_mode` argument but then how do I set the `base_distinguished_name` value in that case?

I tried with `ldap_mode = CUSTOM` and the corresponding `ldap_settings` and it works well but it defeat the purpose of being offered to use `SYSTEM`.

So I tried setting a minimal `custom_settings` with only the base DN but of course it does not work. I don't want to provide a complete `custom_settings` argument here since I want to reuse the system's identity provider and not define a new full blown custom LDAP config each time I create an organization. I just want to say "_please use ldap_mode = SYSTEM and use ou=Foo,dc=domain,dc=local base DN_"

So all in all, how am I supposed to use the `SYSTEM` `ldap_mode`? Any idea what I'm missing here?

cgrard:
Any news?

dataclouder:
Using `SYSTEM` means that the organization will use the LDAP service that was set at system level by the provider (system administrator). If no LDAP service was set by the provider, the organization won't be able to use any.

cgrard:
That's not the problem, as I said there's a LDAP service set but I can't access the setting that would allow me to set the OU to use through the provider (which should be the `CustomUsersOu` parameter according to source code).
This setting exists but it is ignored in the source code : https://github.com/cgrard/terraform-provider-vcd/commit/a333f4db5dc391e26c6a9474290850403036dab7

cgrard:
Thanks for the fix, I can compile, of course, but testing is much more difficult for me as of now because terraform is currently run in a container through a CI in a sealed environment that I don't have direct access to, but I'll try.

lvirbalas:
We can't announce dates. However, at this point, we are striving for our traditional holiday release at the end or beginning of the new year.

PR #1142, opened 2023-10-25
Add property custom_user_ou to vcd_org_ldap

dataclouder:
Fixes Issue #1134.

The resource `vcd_org_ldap` needs a configuration property when using `ldap_mode = "SYSTEM"`. 
The property `custom_user_ou` serves this purpose

### DIFF
--- a/website/docs/r/org_ldap.html.markdown
+++ b/website/docs/r/org_ldap.html.markdown
@@ -100,7 +100,7 @@ The following arguments are supported:
 
 * `org_id` - (Required) Org ID: there is only one LDAP configuration available for an organization. Thus, the resource can be identified by the Org.
 * `ldap_mode` - (Required) One of `NONE`, `CUSTOM`, `SYSTEM`. Note that using `NONE` has the effect of removing the LDAP settings
-* `custom_user_ou` - (Optional; *v3.11+*) If `ldap_mode` is `SYSTEM`, specifies a LDAP `attribute=value` pair to use for OU (organizational unit)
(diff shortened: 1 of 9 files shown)


adambarreiro (approved):
LGTM!

Didainius (approved):

website/docs/r/org_ldap.html.markdown
```diff
@@ -80,12 +80,27 @@ resource "vcd_org_ldap" "my-org-ldap" {
...
+}
+```
+
 ## Argument Reference
 
 The following arguments are supported:
 
 * `org_id` - (Required) Org ID: there is only one LDAP configuration available for an organization. Thus, the resource can be identified by the Org.
 * `ldap_mode` - (Required) One of `NONE`, `CUSTOM`, `SYSTEM`. Note that using `NONE` has the effect of removing the LDAP settings
+* `custom_user_ou` - (Optional) If `ldap_mode` is `SYSTEM`, specifies a LDAP `attribute=value` pair to use for OU (organizational unit)
```
lvirbalas: Nit:

```suggestion
* `custom_user_ou` - (Optional) If `ldap_mode` is `SYSTEM`, specifies an LDAP `attribute=value` pair to use for OU (organizational unit)
```

p.s. "a" vs. "an" depends on the consonant/vowel **sound**:

https://dictionary.cambridge.org/grammar/british-grammar/a-an-and-the

>     /ən empi:θri: …/ an MP3 player

lvirbalas (approved):
Thanks! Just one grammar nit.

website/docs/r/org_ldap.html.markdown
dataclouder: Fixed

adezxc (approved):

PR #1146, opened 2023-10-26
Add support to metadata that is automatically added by VCD

adambarreiro:
## Context
In the most recent versions of VCD, standalone VMs that are created from a vApp Template come with three metadata entries that are created automatically by VCD: `vm.origin.id`, `vm.origin.name` and `vm.origin.type`.

This conflicts with the way metadata works in this Terraform provider, as any plan will ask to remove these "foreign" entries if one adds metadata to their VMs in their HCL configuration.

## Solution
To avoid a bad experience when dealing with this casuistic, this metadata is filtered out from `metadata` and `metadata_entry` blocks and moved to a separated block only available in `vcd_vapp_vm` and `vcd_vm`. This block is completely Computed and available for reading.

## Tests
Tests fail in the newest VCD versions as the expected metadata is n-3. After this feature, as metadata is filtered out, we get these n-3 entries and the removed 3 are in a separated block, which gets its own set of assertions.

- [x] Impacted acceptance tests pass 10.5.0
- [x] All `vm` acceptance tests pass 10.5.0
- [x] `vm` tag upgrade tests pass 10.5.0
- [x] Impacted acceptance tests pass on latest version
- [x] All `vm` acceptance tests pass on latest version
- [x] `vm` tag upgrade tests pass on latest version


vcd/metadata.go
```diff
@@ -369,6 +384,32 @@ func updateMetadataInState(d *schema.ResourceData, vcdClient *VCDClient, resourc
 	return nil
 }
 
+// filterAndSetVcdMetadataEntryToVm filters the metadata created by VCD from the input metadata, that gets updated without these.
```
lvirbalas, line 387: > , that gets updated without these.

Please rephrase this comment. It's hard to grasp what exactly it means :)

Also, would be good to emphasize that the inherited metadata entries are actually "removed" from the metadata.

lvirbalas (commented):
Looks to be a tidy approach to work around the change in API behavior!

vcd/metadata.go
adambarreiro: Yep, now that I re-read it after a few days I can say it was a bit odd. I had a few iterations of rephrasing, I think it's better now.

website/docs/r/vapp_vm.html.markdown
```diff
@@ -518,9 +518,14 @@ example for usage details.
 
 ## Attribute reference
 
-* `vm_type` (*3.2+*) - type of the VM (either `vcd_vapp_vm` or `vcd_vm`).
+* `vm_type` - (*3.2+*) Type of the VM (either `vcd_vapp_vm` or `vcd_vm`).
 * `status` - (*v3.8+*) The vApp status as a numeric code.
 * `status_text` - (*v3.8+*) The vApp status as text.
+* `inherited_metadata` - (*v3.11+*) A block that contains read-only metadata that is automatically added by VCD and provides
+  details of the origin of the VM. The sub-attributes are:
```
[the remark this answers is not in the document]
adambarreiro: Fixed, thanks!

vcd/metadata.go
lvirbalas: Thanks! I noticed a typo, but I would even simplify it further:

> The filtered out entries **will not be longer present** in the original metadata input parameter.

->

> Note: the filtered out entries **are removed from** the original metadata input parameter.

adambarreiro: Changed!

lvirbalas (approved):

.changes/v3.11.0/1146-improvements.md
```diff
@@ -0,0 +1,4 @@
+* Add support to the metadata that gets automatically created on `vcd_vapp_vm` and `vcd_vm` when they are created by a VM from a vApp Template,
+  with the new `inherited_metadata` block. Example of metadata entries of this kind: `vm.origin.id`, `vm.origin.name`, `vm.origin.type` [GH-1146]
+* Add support to the metadata that gets automatically created on `vcd_vapp` when they are created by a vApp Template,
```
dataclouder, line 3:
```suggestion
* Add support to the metadata that gets automatically created on `vcd_vapp` when it is created by a vApp Template or another vApp,
```

vcd/datasource_vcd_vapp.go
```diff
@@ -78,6 +78,30 @@ func datasourceVcdVApp() *schema.Resource {
 					},
 				},
 			},
+			"inherited_metadata": {
```
dataclouder, line 81: I suggest making a variable of this structure, using
`origin_id`, `origin_name`, and `origin_type` instead of the current strings.
This way, it could be used by both the vApp and the VM

vcd/datasource_vcd_vapp_test.go
```diff
@@ -119,3 +120,81 @@ output "status_text" {
...
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(
+						"data.vcd_vapp.fetched_vapp", "metadata.#", "0"),
+					resource.TestCheckResourceAttr(
+						"data.vcd_vapp.fetched_vapp", "metadata.%", "0"),
+					resource.TestMatchResourceAttr(
+						"data.vcd_vapp.fetched_vapp", "inherited_metadata.0.vapp_origin_id", regexp.MustCompile(`^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$`)),
```
dataclouder, line 161: We have used this regexp many times in our tests. Perhaps it's time to make a constant out of it

vcd/datasource_vcd_vapp_vm.go
```diff
@@ -437,6 +437,30 @@ func vcdVmDS(vmType typeOfVm) map[string]*schema.Schema {
 			Computed:    true,
 			Description: "Shows the status of the VM",
 		},
+		"inherited_metadata": {
```
dataclouder, line 440: See comment for `inherited_metadata` in vApp

vcd/datasource_vcd_vapp.go
lvirbalas: The abstract names would be too confusing. Right now it's very clear what these fields represent, comparison:
- `vapp_origin_name` ("Name of the **vApp** from which this vApp was generated")
- `vm_origin_name` ("Name of the **VM** from which this VM was generated")

Without these prefixes:
- `origin_name` - it's very vague and user has to know a lot up front.

So, let's leave the names as is.

vcd/datasource_vcd_vapp_test.go
adambarreiro: I have noted it down, I will do this in another PR so this one doesn't get polluted with a lot of changed files.

.changes/v3.11.0/1146-improvements.md
adambarreiro: Fixed, thanks!

vcd/datasource_vcd_vapp.go
adambarreiro: I've changed the block into a `TypeMap`, so the metadata is added as it comes from VCD, instead of transforming it.

lvirbalas (approved):
The new "map" approach looks great and versatile!

Didainius (approved):
LGTM

dataclouder (approved):

adezxc (approved):
Thanks for this improvement! Also really like the new TestCheckFuncs with version constraint!

PR #1147, opened 2023-10-26
Add IP Space IP Allocation support for custom value

### LINKED ISSUE #1141: ip_space_ip_allocation the argument  __value__ is marked as optionnal in the documentation but it's not implemented
Hello,

### Terraform Version
Terraform v1.6.2
on linux_amd64
+ provider registry.terraform.io/vmware/vcd v3.10.0

### Cloud Director Version
VDC 10.4.2 (Build 21595985)

### Affected Resource(s)
- vcd_ip_space_ip_allocation

```hcl
resource "vcd_ip_space_ip_allocation" "IPSPACE-PREFIX-ALLOC_VM-NETWORK" {
  org_id        = data.vcd_org.TENANT.id
  ip_space_id   = vcd_ip_space.PRIVATE-NET-192.id
  type          = "IP_PREFIX"

  value = "192.168.0.0/24"  ## <- Unsupported agrument bit it's in the Documentation

  depends_on = [vcd_nsxt_edgegateway.MY-EDGE]
}
```

### Expected Behavior
Using attribut __value__ to request a specific IP or subnet in my case from IP Space.

### Actual Behavior
The documentation of ip_space_ip_allocation talk about an optionnal argument named __value__.
https://registry.terraform.io/providers/vmware/vcd/latest/docs/resources/ip_space_ip_allocation#argument-reference

__value__ - (Optional; VCD 10.4.2+) An option to request a specific IP or subnet from IP Space

This argumenent doens't seems to be implemented in the code..
Source code: https://github.com/vmware/terraform-provider-vcd/blob/v3.10.0/vcd/resource_vcd_ip_space_ip_allocation.go

As a result I have this :
│ Error: Unsupported argument
│
│   on networks.tf line 53, in resource "vcd_ip_space_ip_allocation" "IPSPACE-PREFIX-ALLOC_DMO-192-168-0-0_24":
│   53:   value = "192.168.0.0/24"
│
│ An argument named "value" is not expected here.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Add a resource type ip_space_ip_allocation with the argument "value" 
2. `terraform plan`

Didainius:
Closes #1141

This PR adds the missing `value` field to `vcd_ip_space_ip_allocation` resource that provides the capability to explicitly specify IP or Prefix value on VCD 10.4.2+

**Important note:** The `value` field supports IP ranges in API (e.g. 192.168.1.1-192.168.1.3), but this would result in 3 separate allocations with 3 separate IDs. This goes against Terraform's approach of mapping a single resource to a single entity, and we have the `quantity` field disabled (and defaulted to 1) for that same reason. I have added an extra validation function to check if a `-` is present in the string. If so - it will return an error with a hint to the user:
```shell
$ terraform validate
╷
│ Error: invalid value for value (This resource does not support allocating ranges due to Terraform resources map to single entity. Please use multiple resource instances to allocate multiple IP addresses)
│ 
│   with vcd_ip_space_ip_allocation.public-floating-ip,
│   on vcd.TestAccVcdIpSpaceIpAllocationCustomValuestep1.tf line 123, in resource "vcd_ip_space_ip_allocation" "public-floating-ip":
│  123:   value = "11.11.11.101-11.11.11.102"
```

**Note:** Tests starting with `TestAccVcdIpSpace` passed on 10.4.1, 10.4.2, 10.5.0


website/docs/r/ip_space_ip_allocation.html.markdown
```diff
@@ -105,8 +126,10 @@ The following arguments are supported:
   * `IP_PREFIX` - allocates subnets. **Note** field `prefix_length` is required to allocate IP
     Prefix
 * `prefix_length` (Optional) Required when `type=IP_PREFIX`
-* `value` - (Optional; VCD *10.4.2+*) An option to request a specific IP or subnet from IP Space
-* `usage_state` - (Optiona) Not required unless manual IP reservation is required which can be
+* `value` - (Optional; VCD *10.4.2+*) An option to request a specific IP or subnet from IP Space.
+  **Note:** This field does not support IP ranges because it would cause multiple allocations
+  created in one resource. Please use multiple resource instances to allocate IP ranges.
+* `usage_state` - (Optional) Not required unless manual IP reservation is required which can be
   enabled `USED_MANUAL`. Value `UNUSED` must be set to release manual allocation of IP.
```
lvirbalas, line 133: Though it's not from this PR, I just noticed that this sentence is funky. Would be good to fix it. Something between the lines (this is just an example to make a point, I'm not sure it's correct):

```suggestion
* `usage_state` - (Optional) Value  `USED_MANUAL` enables manual IP reservation. Value `UNUSED` must be set to release manual allocation of IP.
```

lvirbalas (changes requested):

website/docs/r/ip_space_ip_allocation.html.markdown
Didainius: Indeed it could be clearer. What if I prepend such a sentence to your proposal as I don't want to confuse users. 
> Only required for manual reservations.

In the end it could be:

* `usage_state` - (Optional) Only required for manual reservations. Value `USED_MANUAL` enables manual IP reservation. Value `UNUSED` must be set to release manual allocation of IP.

lvirbalas: Working on the above further:

* `usage_state` - (Optional) Only used with manual reservations. Value `USED_MANUAL` enables manual IP reservation. Value `UNUSED` is set to release manual allocation of IP.

Didainius: I like this last suggestion - applied.

lvirbalas (approved):

dataclouder (approved):

adezxc (approved):

adambarreiro (approved):
LGTM! The validation hint is really useful.

My only concern is what would happen if a user needs to allocate something like `11.11.11.1-11.11.11.255`. I guess it would be trivial with `count`, but still one could end with a lot of resources there. I guess this is a rare use case though.

PR #1139, opened 2023-10-18
Update CSE cluster management guide to use CSE v4.1

adambarreiro:
~This PR requires #1063, review and merge that one first.~

This PR updates the CSE cluster management guide to use v4.1.

PR #1164, opened 2023-11-23
Improve ignore_metadata_changes by adding support to OpenAPI entries

adambarreiro:
## Description

While PR #1018 added the new `metadata_entry` blocks to `vcd_rde` to manage OpenAPI metadata, these were not compatible with `ignore_metadata_changes` provider configuration block. This PR extends this compatibility by adding OpenAPI support for ignoring metadata.

## Fixes

This PR also fixes a bug when `conflict_action = "warn"` inside a `ignore_metadata_changes` block, that caused that any conflict ended the operation prematurely with an error, instead of continuing and returning a mere warning at the end.

To fix this, one must consider the Diagnostics objects and append them one after the other, to return them at the end if there's any.

This PR also completes #1139 bugfix, detected by TestAccVcdRde.


lvirbalas (approved):
Thanks!

vcd/metadata.go
```diff
@@ -358,7 +359,7 @@ func checkIgnoredMetadataConflicts(d *schema.ResourceData, vcdClient *VCDClient,
 					Severity: severity,
 					Summary:  "Found a conflict between 'ignore_metadata_changes' and 'metadata_entry'",
 					Detail: fmt.Sprintf("There is an 'ignored_metadata' block: %s\n"+
-						"and there is a 'metadata_entry' with key '%s' and value '%s' in your Terraform configuration that matches the criteria, hence it will be ignored.\n"+
+						"and there is a 'metadata_entry' with key '%s' and value '%s' in your Terraform configuration or state that matches that criteria, hence it will be ignored.\n"+
```
dataclouder, line 362:
```suggestion
						"and there is a 'metadata_entry' with key '%s' and value '%s' in your Terraform configuration or state that matches such criteria, hence it will be ignored.\n"+
```

vcd/metadata_common_test.go
```diff
@@ -339,7 +339,7 @@ func testMetadataEntryCRUD(t *testing.T, resourceTemplate, resourceAddress, data
 // - Step 3: Add a data source that fetches the created resource.
 //
 // The different ignore_metadata_changes sub-tests check what happens if the filter matches or doesn't match the metadata entry
-// added in Pre-Step 2. If it doesn't match, Terraform will delete it from VCD. If it match, it gets ignored as it doesn't exist.
+// added in Pre-Step 2. If it doesn't match, Terraform will delete it from VCD. If it matches, it gets ignored as it doesn't exist.
```
dataclouder, line 342:
```suggestion
// added in Pre-Step 2. If it doesn't match, Terraform will delete it from VCD. If it matches, it gets ignored as if it didn't exist.
```

vcd/resource_vcd_org.go
```diff
@@ -610,9 +622,9 @@ func resourceVcdOrgImport(ctx context.Context, d *schema.ResourceData, meta inte
 		return nil, fmt.Errorf(errorRetrievingOrg, err)
 	}
 
-	diagErr := setOrgData(d, vcdClient, adminOrg)
-	if diagErr != nil {
-		return []*schema.ResourceData{}, fmt.Errorf("error setting Org data: %v", diagErr)
+	diags := setOrgData(d, vcdClient, adminOrg)
+	if diags != nil && diags.HasError() {
+		return []*schema.ResourceData{}, fmt.Errorf("error setting Org data: %v", diags)
```
Didainius, line 627: I guess we don't care for ignore, but just to not that warnings are lost. OTOH, import requires `error` only so that is fine

vcd/metadata_common_test.go
adambarreiro: Fixed, thanks

vcd/resource_vcd_network_routed_v2.go
```diff
@@ -257,20 +256,19 @@ func resourceVcdNetworkRoutedV2Read(_ context.Context, d *schema.ResourceData, m
...
 		}
 	}
-	if diagErr != nil {
-		log.Printf("[DEBUG] Unable to set routed network v2 metadata: %s", err)
-		return diagErr
+	if diags != nil && diags.HasError() {
+		return diags
 	}
 
 	return nil
```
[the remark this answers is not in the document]
adambarreiro: Good find, fixed.

vcd/resource_vcd_org.go
adambarreiro: Yes, that is unfortunate, that Import does not support Diagnostics.

dataclouder (approved):

Didainius (approved):

adezxc (approved):

PR #1172, opened 2023-11-29
Fix missing test field in TestAccVcdNsxtEdgeGatewayVdcGroupExternalUplink

Didainius:
This PR patches test config for `TestAccVcdNsxtEdgeGatewayVdcGroupExternalUplink` so that binary test fields are correctly rendered.

Previously it would cause such field names (It worked, but it is not optimal.) - `"*** MISSING FIELD [ExternalNetworkName] from func vcd.TestAccVcdNsxtEdgeGatewayVdcGroupExternalUplink-2"`


adambarreiro (approved):

lvirbalas (approved):
Thanks!

adezxc (approved):

PR #1175, opened 2023-12-06
Pre-release test cleanups

Didainius:
* Added missing `upgrade` test skips for resources that have changed something in their schema
* Split acceptance test `TestAccVcdVmPlacementPolicyInVdc` that used multiple `provider` definitions to test provider and tenant capabilities. This is not a good approach as Terraform testing framework causes errors. Now `TestAccVcdVmPlacementPolicyInVdc` will only test in provider while a new test `TestAccVcdVmPlacementPolicyInVdcTenant` will use SDK to create prerequisites and then works in both provider and tenant modes.
* Skip import tests `vcd.ResourceList-ldap-server-vcd_vm.tf` and `vcd.ResourceList-testVm-vcd_vapp_vm.tf` in binary mode
* Fix text file path lookup for `vcd.TestAccVcdCatalogMediaBasic.tf` and `vcd.TestAccVcdCatalogMediaBasic-Update.tf` binary tests


lvirbalas (approved):
Thank you!

PR #1185, opened 2023-12-19
Drop legacy API versions

adambarreiro:
This PR bumps Go SDK so we won't get "not supported" errors when running the Terraform Provider in the newest VCD versions.


Didainius:
I suspect we don't need it anymore as we have bumped SDK with other PRs in main

adambarreiro:
> I suspect we don't need it anymore as we have bumped SDK with other PRs in main

True, I completely forgot about this PR.

PR #1194, opened 2024-01-09
Attempt to make flake test TestAccVcdAnyAccessControlGroups more reliable

Didainius:
This PR attempts to make test `TestAccVcdAnyAccessControlGroups` more reliable.

It tends to sometimes fail with error given below.
The hypothesis here is that LDAP is being configured and the test tries to use it immediately while it is not initialized by VCD. 

```
resource_vcd_any_access_control_test.go:39: Step 1/1 error: Error running apply: exit status 1

Error: error creating group admin_staff: error creating group: API Error: 500: [ 05ebc1f3-2aa1-45bb-bfb3-63e97b945d2c ] LDAP context not initialized. Error connecting to LDAP.
```

### DIFF
--- a/vcd/resource_vcd_any_access_control_test.go
+++ b/vcd/resource_vcd_any_access_control_test.go
@@ -5,6 +5,7 @@ package vcd
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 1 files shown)


lvirbalas (approved):
Thanks!

adambarreiro (approved):

PR #1196, opened 2024-01-12
Improve flaky test TestAccVcdVAppMultiVmInTemplate

Didainius:
Diff got very ugly here, but what happened is that I moved inherited metadata validation checks from "step1" to "step2" of the test.


Didainius:
This approach didn't help. Discarding.

PR #1273, opened 2024-05-31
Improve media item upload stability

adambarreiro:
This PR modifies vcd_catalog_media code to prevent it from hanging when the upload Task completes successfully but for some reason the progress is not 100% (could happen with some glitch in VCD)

### Tests

Run `TestAccVcdCatalogMediaBasic` in a VCD version that has this problem. Without this fix, it never ends.

### DIFF
--- a/.changes/v3.13.0/1273-bug-fixes.md
+++ b/.changes/v3.13.0/1273-bug-fixes.md
@@ -1,2 +1,2 @@
 * Fix `vcd_catalog_media` resource so it doesn't wait indefinitely to the upload task to reach 100% progress,
-  by checking also its status, to decide that the upload is complete [GH-1273]
+  by checking also its status, to decide that the upload is complete or aborted [GH-1273]
--- a/.changes/v3.13.0/1273-bug-fixes.md
+++ b/.changes/v3.13.0/1273-bug-fixes.md
@@ -1,0 +1,1 @@
+* Fix `vcd_catalog_media` resource so it doesn't hang on successfully completed tasks [GH-1273]
--- a/.changes/v3.13.0/1273-bug-fixes.md
+++ b/.changes/v3.13.0/1273-bug-fixes.md
@@ -1,1 +1,2 @@
-* Fix `vcd_catalog_media` resource so it doesn't hang on successfully completed tasks [GH-1273]
+* Fix `vcd_catalog_media` resource so it doesn't wait indefinitely to the upload task to reach 100% progress,
+  by checking also its status, to decide that the upload is complete [GH-1273]
--- a/vcd/resource_vcd_catalog_media.go
+++ b/vcd/resource_vcd_catalog_media.go
@@ -184,9 +184,8 @@ func resourceVcdMediaCreate(ctx context.Context, d *schema.ResourceData, meta in
 				return diag.FromErr(err)
 			}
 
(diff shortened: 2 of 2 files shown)


vcd/resource_vcd_catalog_media.go
```diff
@@ -196,11 +196,12 @@ func resourceVcdMediaCreate(ctx context.Context, d *schema.ResourceData, meta in
...
 			progress, err := task.GetTaskProgress()
 			if err != nil {
-				log.Printf("vCD Error importing new catalog item: %s", err)
-				return diag.Errorf("vCD Error importing new catalog item: %s", err)
+				log.Printf("VCD Error importing new catalog item: %s", err)
+				return diag.Errorf("VCD Error importing new catalog item: %s", err)
 			}
-			logForScreen("vcd_catalog_media", fmt.Sprintf("vcd_catalog_media."+mediaName+": vCD import catalog item progress "+progress+"%%\n"))
-			if progress == "100" {
+			logForScreen("vcd_catalog_media", fmt.Sprintf("vcd_catalog_media.%s: VCD import catalog item progress %s%%\n", mediaName, progress))
```
dataclouder, line 202: This loop exits in two cases:
* when the task status is "error" (returned from `task.GetTaskProgress') 
* when the status is "success".

What about the other status values? Probably we should considered `"aborted"` in this context, as other statuses refer to transient operations ("queued", "running", "preRunning")

Didainius: We could probably factor in aborted to catch the error if the user presses "cancel" in UI.

dataclouder: We could do more than caching the error: an interruption at this stage will leave the media item empty in the catalog, preventing a re-run of the HCL script (the item is already in the catalog)

dataclouder: But maybe this is beyond the scope of this change

adambarreiro: Indeed! I have added `aborted` to the exit condition of the loop.

I checked that `task.WaitTaskCompletion()` (after the modified loop) finishes without error if the task is aborted, but since it was also the original behavior, I'm not changing anything there.

adambarreiro: Ah, I didn't see the other replies to this thread 🙂 

I thought the same @dataclouder (adding a condition that if it's aborted it should fail), but I'm also concerned about changing too much.

Didainius (approved):

dataclouder (approved):

vcd/resource_vcd_catalog_media.go
adambarreiro: Finally added the check of `aborted` to throw an error in that case. Test passes now.

.changes/v3.13.0/1273-bug-fixes.md
```diff
@@ -0,0 +1 @@
+* Fix `vcd_catalog_media` resource so it doesn't hang on successfully completed tasks [GH-1273]
```
lvirbalas, line 1: Please amend comment to contain the details. I mean that now we're adding wait for "success" as an alternative to wait for "100%".

lvirbalas (approved):
Looks great! One ask to clarify changelog.

.changes/v3.13.0/1273-bug-fixes.md
adambarreiro: Done, let me know how it looks now, if I should change it.

Issue #1170, opened 2023-11-28
vapp_template_id missing after import vcd_vapp_vm, triggers recreate of the vcd_vapp_vm


cjswart:
Hello,

After changing from catalog_name/template_name to vapp_catalog_id in the vcd_vapp_vm resource the vm is destroyed and recreated 
I removed the resource from the state file and import the resource back in the state file
After the import the vapp_template_id is not added to the state file still forcing a recreate of the vcd_vapp_vm resource

### Terraform Version
Terraform v1.6.4
on windows_amd64
+ provider registry.terraform.io/hashicorp/template v2.2.0
+ provider registry.terraform.io/vmware/vcd v3.10.0

### Affected Resource(s)
Please list the resources as a list, for example:
- vcd_vapp_vm

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files

```hcl
data "vcd_catalog" "my-catalog" {
  name = var.vapp_vm_template.catalog_name
  org  = var.vapp_vm_template.org
}
data "vcd_catalog_vapp_template" "osboot" {
  catalog_id = data.vcd_catalog.my-catalog.id
  name       = var.vapp_vm_template.template_name
  org  = var.vapp_vm_template.org
}

resource "vcd_vapp_vm" "vm" {
  org                    = var.org
  vdc                    = lookup(var.vapp_vm_settings, "vdc", var.vdc)
  vapp_name              = var.vapp_vm_settings.vapp_name
  name                   = var.vapp_vm_settings.hostname
  vapp_template_id       = try(var.vapp_vm_settings.vapp_template_id, data.vcd_catalog_vapp_template.osboot.id)
  memory                 = try(var.vapp_vm_settings.memory, null)
  cpus                   = try(var.vapp_vm_settings.cpus, null)
  cpu_cores              = try(var.vapp_vm_settings.cpu_cores, null)
  os_type                = try(var.vapp_vm_settings.os_type, null)
  hardware_version       = try(var.vapp_vm_settings.hardware_version, null)
  computer_name          = try(var.vapp_vm_settings.hostname, null)
  cpu_hot_add_enabled    = try(var.vapp_vm_settings.cpu_hot_add_enabled, true)
  memory_hot_add_enabled = try(var.vapp_vm_settings.memory_hot_add_enabled, true)
  storage_profile        = try(var.vapp_vm_settings.storage_profile, null)

  customization {
[...]

PR #1274, opened 2024-06-03
Fix issue-1170

dataclouder:
Fix Issue #1170, where an imported VM complains about missing or altered fields and Terraform tries to re-create the resource.

### DIFF
--- a/website/docs/r/vapp_vm.html.markdown
+++ b/website/docs/r/vapp_vm.html.markdown
@@ -570,6 +570,7 @@ example for usage details.
 * `inherited_metadata` - (*v3.11+*; *VCD 10.5.1+*) A map that contains read-only metadata that is automatically added by VCD (10.5.1+) and provides
   details on the origin of the VM (e.g. `vm.origin.id`, `vm.origin.name`, `vm.origin.type`).
 * `extra_config` - (*v3.13.+*) The VM extra configuration. See [Extra Configuration](#extra-configuration) for more detail.
+* `imported` - (*v3.13.+*) A true/false value telling whether the resource was imported.
 
 <a id="disk"></a>
 ## Disk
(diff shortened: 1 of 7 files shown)


website/docs/r/vapp_vm.html.markdown
```diff
@@ -570,6 +570,7 @@ example for usage details.
 * `inherited_metadata` - (*v3.11+*; *VCD 10.5.1+*) A map that contains read-only metadata that is automatically added by VCD (10.5.1+) and provides
   details on the origin of the VM (e.g. `vm.origin.id`, `vm.origin.name`, `vm.origin.type`).
 * `extra_config` - (*v3.13.+*) The VM extra configuration. See [Extra Configuration](#extra-configuration) for more detail.
+* `imported` - (*v3.13.+*) A true/false value telling whether the resource was imported.
```
lvirbalas, line 573: Would be great to have a description about in which cases the user should use this.

dataclouder: I think it's self-explanatory. If you have something in mind, please explain.

lvirbalas: I'm referring to the following logic:
https://github.com/vmware/terraform-provider-vcd/pull/1274/files#diff-e39bfe24504d3209d5395389ca0a05af237416e91c4f16a0e3fe56f299cf9b49R49-R51

We should expose that if user turns on `imported`, that it's not just an empty switch, but that the above logic gets triggered.

dataclouder: `imported` is a read-only attribute, not an argument. Users won't be able to change it.

lvirbalas: Oops! My bad, missed that it's computed. Thanks.

lvirbalas (approved):

adambarreiro (approved):

PR #1275, opened 2024-06-04
Improve vcd_nsxt_edgegateway large IPv6 address count behavior

### LINKED ISSUE #1270: (Datasource) NSX-T Edge gateway Plugin chrashes when reading Egde gateway with allocated IPv6 addresses. V2
Hi, We are unfortunately still experiencing an issue with the datasource 'vcd_edgegateway' when the edge has IPv6 addresses allocated.
When using this datasource on an edge gateway that has ipv6 IP allocations, we get the following error:

image

This is a different error now that we have updated to 3.12.1, but it still doesn't work... 
(https://github.com/vmware/terraform-provider-vcd/issues/1242)

When removing the IPv6 allocation from the edge, the plugin works fine again, as before.

Terraform vcd provider version v3.12.1
Terraform version v1.8.3

Didainius:
This PR closes #1270

The problem is that having huge IPv6 prefixes assigned can cause compute resource exhaustion while counting them.
To prevent this, both - resource and data source `vcd_nsxt_edgegateway` now have new field `read_limit_unused_ip_count` (with default of `1000000` - 1 million). This field will limit counting `used_ip_count` and `unused_ip_count` to the defined limit. 

It is unrealistic that a single Edge Gateway can manage more than 1 million IPs effectively so counting up to `18,446,744,073,709,551,616` (IPv6 /64) or more has no meaning at all


website/docs/r/nsxt_edgegateway.html.markdown
```diff
@@ -351,6 +351,10 @@ can be used to lookup ID by name.
 * `external_network` - (Optional, *VCD 10.4.1+*, *v3.11+*) attaches NSX-T Segment backed External
   Networks with a given [configuration block](#edgegateway-subnet-external-network). It *does not
   support IP Spaces*.
+* `read_limit_unused_ip_count` - (Optional, *v3.13+*) Sets a limit of IPs to count for
+  `used_ip_count` and `unused_ip_count` attributes to avoid exhausting compute resource while
+  counting IPs in large IPv6 subnets. Defaults to `1000000`, update is a no-op, but will affect
+  newly read data.
```
lvirbalas, line 357: Though description is already very helpful, I think implications of using this field are still unclear. Maybe add a sentence (or example separately somewhere below) on the actual drawback of using it?

website/docs/d/nsxt_edgegateway.html.markdown
```diff
@@ -57,6 +57,10 @@ The following arguments are supported:
 definition at provider level.
 
 * `name` - (Required) NSX-T Edge Gateway name.
+* `read_limit_unused_ip_count` - (Optional, *v3.13+*) Sets a limit of IPs to count for
+  `used_ip_count` and `unused_ip_count` attributes to avoid exhausting compute resource while
+  counting IPs in large IPv6 subnets. Defaults to `1000000`, update is a no-op, but will affect
+  newly read data.
```
adambarreiro, line 63: As there's no update here
```suggestion
  counting IPs in large IPv6 subnets. Defaults to `1000000`
```

adambarreiro (approved):

.changes/v3.13.0/1275-bug-fixes.md
```diff
@@ -0,0 +1,3 @@
+* Resource and data source `vcd_nsxt_edgegateway` may crash due to exhausting memory while counting
+  huge IPv6 subnets (issue [#1202](https://github.com/vmware/terraform-provider-vcd/issues/1270))
```
dataclouder, line 2:
```suggestion
* Fix [Issue #1202](https://github.com/vmware/terraform-provider-vcd/issues/1270) - Resource and data source `vcd_nsxt_edgegateway` may crash due to exhausting memory while counting huge IPv6 subnets
```

go.mod
```diff
@@ -70,3 +70,5 @@ require (
 	google.golang.org/grpc v1.63.2 // indirect
 	google.golang.org/protobuf v1.34.0 // indirect
 )
+
+replace github.com/vmware/go-vcloud-director/v2 => github.com/Didainius/go-vcloud-director/v2 v2.17.0-alpha.2.0.20240604105409-0e2e3bab9442
```
dataclouder, line 74: reminder to remove this replacement before merging

vcd/resource_vcd_nsxt_edgegateway.go
```diff
@@ -214,6 +216,12 @@ func resourceVcdNsxtEdgeGateway() *schema.Resource {
 				Computed:    true,
 				Description: "Number of unused IP addresses",
 			},
+			"read_limit_unused_ip_count": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     defaultReadLimitOfUnusedIps,
```
dataclouder, line 222: What does a value of "0" do? Unlimited? No counting at all?

website/docs/d/nsxt_edgegateway.html.markdown
dataclouder: Probably worth mentioning what a value of "0" will do, as in many cases we use it to signify "unlimited"

.changes/v3.13.0/1275-bug-fixes.md
Didainius: Changed.

website/docs/d/nsxt_edgegateway.html.markdown
Didainius: Good point!
0 is `unlimited` as per SDK code https://github.com/vmware/go-vcloud-director/pull/682/files#diff-11389a99316769bddb45effb145f62650d74c0f8711c34ef9b22eb719d0d7ca5R976-R979

Updated docs.

website/docs/r/nsxt_edgegateway.html.markdown
Didainius: Please check once more as I have addressed other comments and hopefully improved this as well.
The point is - nothing in operations except values of `used_ip_count` and `unused_ip_count` are affected.

dataclouder (approved):

website/docs/r/nsxt_edgegateway.html.markdown
```diff
@@ -353,8 +353,10 @@ can be used to lookup ID by name.
   support IP Spaces*.
 * `read_limit_unused_ip_count` - (Optional, *v3.13+*) Sets a limit of IPs to count for
   `used_ip_count` and `unused_ip_count` attributes to avoid exhausting compute resource while
-  counting IPs in large IPv6 subnets. Defaults to `1000000`, update is a no-op, but will affect
-  newly read data.
+  counting IPs in large IPv6 subnets. It does not affect operation of Edge Gateway configuration,
```
lvirbalas, line 356: Sorry for a late thought. I was re-reading the PR. I think it would be clearer if this field was named (if it's not too late to change, of course):
`ip_count_read_limit`

Witht this suggestion I'm trying to address two things:
1. Right now it says "..._unused_ip_count", but description says it's actual for both
  `used_ip_count` and `unused_ip_count` attributes.
2. Putting the "_read_limit" at the end makes it clearer, IMHO.

Additional thought: is "read" really precise here? It sounds like API GET call limit. Maybe just name the field `ip_count_limit`?

Didainius: > Sorry for a late thought. I was re-reading the PR. I think it would be clearer if this field was named (if it's not too late to change, of course):
> `ip_count_read_limit`

Not an issue at all. The heavy lifting is in SDK anyway.

> Additional thought: is "read" really precise here? It sounds like API GET call limit. Maybe just name the field `ip_count_limit`?

What happens is that when there is a huge subnet, VCD API does not return IP count for unused IPs anymore (with small subnets it does). Then, to count them our code has to iterate over all IPs in defined subnets (this is what can overload compute with large subnets). So I would say this is our read that can exhaust resources. It does return the `used_ips` as it has a dedicated endpoint that tells where each IP is used.

`ip_count_limit` sounds too generic for me, but `ip_count_read_limit` sounds OK to me

lvirbalas: `ip_count_read_limit` it is then! :)

Didainius: Changed. Test `TestAccVcdIpv6SupportLargeSubnet` still passes.

PR #1277, opened 2024-06-07
Improve RDE Behavior ACL documentation and fix test

adambarreiro:
This PR fixes the `TestAccVcdRdeTypeBehaviorAndAcl` test that fails on latest VCD, as the API now doesn't return redundant Access Level IDs.

What happened is that the first step of the test was setting both
```
urn:vcloud:accessLevel:ReadOnly
urn:vcloud:accessLevel:FullControl
```

But, indeed, `urn:vcloud:accessLevel:ReadOnly` is implicitly inside of `urn:vcloud:accessLevel:FullControl`, so the new version of the API simplifies the returned access levels, causing undesired updates-in-place when doing a Terraform plan after creation.

For that reason, and to avoid confusion, documentation has also been updated.

_Extra content:_ Renamed two files that were wrongly named.


Didainius (approved):

.changes/v3.13.0/1277-improvements.md
```diff
@@ -0,0 +1,2 @@
+* Improve `rde_type_behavior_acl` documentation to state that redundant access levels should be avoided to 
+  prevent undesired updates-in-place during plans [GH-1277]
```
lvirbalas, line 2: Let's state from which API version this changed?

lvirbalas (approved):

.changes/v3.13.0/1277-improvements.md
```diff
@@ -0,0 +1,2 @@
+* Improve `rde_type_behavior_acl` documentation to state that redundant access levels should be avoided, specially
```
dataclouder, line 1:
```suggestion
* Improve `rde_type_behavior_acl` documentation to state that redundant access levels should be avoided, especially
```

dataclouder (approved):

.changes/v3.13.0/1277-improvements.md
adambarreiro: Fixed, thanks

PR #1296, opened 2024-07-24
Add vcd_nsxt_alb_edgegateway_service_engine_group and vcd_nsxt_alb_service_engine_group to vcd_resource_list data source

### LINKED ISSUE #1145: new resource_type vcd_nsxt_alb_edgegateway_service_engine_group and vcd_nsxt_alb_service_engine_group in data source vcd_resource_list
### Description

it would be nice to have `vcd_nsxt_alb_edgegateway_service_engine_group` and `vcd_nsxt_alb_service_engine_group` in data source  `vcd_resource_list` as `resource_type`
https://registry.terraform.io/providers/vmware/vcd/latest/docs/data-sources/resource_list#resource_type

* data source `vcd_resource_list`

I had a look in the 3.11.0 code but I did't find them in `datasource_vcd_resource_list.go`

adambarreiro:
## Description

Closes #1145

This PR adds values `vcd_nsxt_alb_edgegateway_service_engine_group` and `vcd_nsxt_alb_service_engine_group` to the `resource_type`.

## Tests

Test `TestAccVcdDatasourceResourceList` passed on 10.6.0 with the newest changes.

To test manually, add a AVI Controller, then a Cloud, then create a Service Engine Group. Then assign this Service Engine Group to an Edge gateway. The data source should then list the SEGs.

Examples:
```hcl
data "vcd_resource_list" "rs" {
  name          = "foo"
  resource_type = "vcd_nsxt_alb_service_engine_group"
  list_mode     = "import"
}

data "vcd_resource_list" "rs2" {
  name          = "foo"
  resource_type = "vcd_nsxt_alb_edgegateway_service_engine_group"
  parent        = "nsxt-gw"
  list_mode     = "import"
}
```

Then you can import the resources with the proposed import outputs, example:

```
terraform import vcd_nsxt_alb_edgegateway_service_engine_group.example my-org.nsxt-vdc.nsxt-gw.seg
```


vcd/datasource_vcd_resource_list.go
```diff
@@ -1110,6 +1110,56 @@ func vdcTemplateList(d *schema.ResourceData, meta interface{}) (list []string, e
...
+	return genericResourceList(d, "vcd_nsxt_alb_service_engine_group", nil, items)
+}
+
+func nsxtAlbServiceEngineGroupAssignment(d *schema.ResourceData, meta interface{}) (list []string, err error) {
+	client := meta.(*VCDClient)
+
+	orgName, vdcName, _, _, egw, err := getEdgeGatewayDetails(d, meta)
+	if err != nil {
+		return list, fmt.Errorf("error retrieving edge gateway '%s': %s ", d.Get("parent").(string), err)
+	}
```
Didainius, line 1137: There is a bit of a problem that this function uses "NSX-V" Edge Gateway function which can cause some unexpected error. 
IMO would be good to validate that this is an NSX-T Edge Gateway (we don't have ALB for NSX-V). Either on the edge gateway, or even checking if the VDC is an NSX-T one.

Both `AnyTypeEdgeGateway` and `Vdc` have `IsNsxt` methods

website/docs/d/resource_list.html.markdown
```diff
@@ -391,6 +426,8 @@ The following arguments are supported:
     * `vcd_network_routed_v2`
     * `vcd_network_isolated_v2`
     * `vcd_nsxt_network_imported`
+    * `vcd_nsxt_alb_service_engine_group`
+    * `vcd_nsxt_alb_edgegateway_service_engine_group`
```
Didainius, line 430: This one remains untested and prone to breaking because it needs a lot of prerequisites in this resource list. Maybe we could pull in the data source to relevant ALB test where all the prerequisites are created anyway.

Didainius (commented):
Two comments from me inline. Other than that - I have tried it and also filtering - it worked.

vcd/datasource_vcd_resource_list.go
adambarreiro: Good point, I tried to reuse this function to avoid adding more code but turned out to be a bad idea 😄 
Now it should be good, it checks NSX-T and uses NSX-T Gateway types/methods.

website/docs/d/resource_list.html.markdown
adambarreiro: Modified `TestAccVcdNsxtEdgeGatewayServiceEngineGroupDedicated` so it uses also `vcd_resource_list` data source.
Acceptance and binary test passed

Didainius (approved):
LGTM. Validated both fixes and also ran ALB test.

lvirbalas (approved):
Nice!

PR #1303, opened 2024-08-08
Allow unlimited CPU limit in vcd_vm_sizing_policy resource

### LINKED ISSUE #1243: vcd_vm_sizing_policy limits don't allow for unlimited
Hello,

The current code is incorrect in requiring the VM Sizing Policy CPU and Memory limits need to be at least 0, as this completely removes the ability to create policies with unlimited CPU/Memory as per the API.

You can see the section of code here for CPU Limits
https://github.com/vmware/terraform-provider-vcd/blob/1f61f1eb157271cbe9cdd98698503f45148b22c7/vcd/resource_vcd_vm_sizing_policy.go#L94C3-L99C37

```go
"limit_in_mhz": {
			Type:         schema.TypeString,
			ForceNew:     true,
			Optional:     true,
			Description:  "Defines the CPU limit in MHz for a VM. If not defined in the VDC compute policy, CPU limit is equal to the vCPU speed multiplied by the number of vCPUs.",
			ValidateFunc: IsIntAndAtLeast(0),
		}
```

And here is the API showing that `-1` is a valid value

```swagger
cpuLimit	integer($int64)
The limit of CPU in MHz for the VM on which this policy is applied. -1 means unlimited. This
is optional. If not provided, limit would be equal to the vCPUSpeed multiplied by numCPU.

memoryLimit	integer($int64)
The limit of memory in MB for the VM on which this policy is applied. -1 means unlimited. This
is optional. If not provided, limit would be equal to the configured memory of the VM.
```

### Terraform Version
```shell
% tofu -v
OpenTofu v1.6.2
on darwin_amd64
+ provider registry.opentofu.org/vmware/vcd v3.12.0
```

### Affected Resource(s)
- vcd_vm_sizing_policy

```hcl
resource "vcd_vm_sizing_policy" "vmsizingpolicy1" {
    description = "General Compute | 2 vCPU"
    name        = "General.2vCPU"

    cpu {
        count                 = "2"
        limit_in_mhz          = "-1"
        reservation_guarantee = "0.1"
        shares                = "4000"
        speed_in_mhz          = "2500"
    }

    memory {
        reservation_guarantee = "0.5"
    }
}
```

### Expected Behavior
Using the above config, I'd expect a new VM Sizing policy to be created with `CPU Limit` set to `Unlimited`

image

### Actual Behavior
An error is printed that `-1` is not a valid value as it's less than 0.

```shell
│ Error: expected cpu.0.limit_in_mhz to be at least (0), got -1
│ 
│   with vcd_vm_sizing_policy.vmsizingpolicy1,
│   on main.tf line 171, in resource "vcd_vm_sizing_policy" "vmsizingpolicy1":
│  171:         limit_in_mhz          = "-1"
│ 
```

### Steps to Reproduce
1. `terraform plan`

adambarreiro:
This PR amends `vcd_vm_sizing_policy` resource to allow the value -1 in the `limit_in_mhz` argument of the `cpu` block, which means **Unlimited** (or **No limit**).

Before this change, the resource would output an error as the minimum allowed value was 0.

Closes #1243

## Tests

- `TestAccVcdVmSizingPolicy` in VCD 10.6.0 passed
- `vcd.ResourceSchema-vcd_vm_sizing_policy.tf` Upgrade test passed (no need to skip)

### DIFF
--- a/.changes/v3.14.0/1303-bug-fixes.md
+++ b/.changes/v3.14.0/1303-bug-fixes.md
@@ -1,0 +1,2 @@
+* Fix [Issue 1243](https://github.com/vmware/terraform-provider-vcd/issues/1243):
+  Allow unlimited `limit_in_mhz` in `vcd_vm_sizing_policy` resource [GH-1303]
--- a/vcd/resource_vcd_vm_sizing_policy.go
+++ b/vcd/resource_vcd_vm_sizing_policy.go
@@ -95,7 +95,7 @@ var sizingPolicyCpu = &schema.Resource{
 			Type:         schema.TypeString,
 			ForceNew:     true,
 			Optional:     true,
-			Description:  "Defines the CPU limit in MHz for a VM. If not defined in the VDC compute policy, CPU limit is equal to the vCPU speed multiplied by the number of vCPUs.",
+			Description:  "Defines the CPU limit in MHz for a VM. If not defined in the VDC compute policy, CPU limit is equal to the vCPU speed multiplied by the number of vCPUs. -1 means unlimited",
 			ValidateFunc: IsIntAndAtLeast(-1),
 		},
 		"shares": {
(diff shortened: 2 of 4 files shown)


Didainius (approved):

Didainius:
Tests worked

lvirbalas (approved):
Awesome!

PR #1308, opened 2024-08-21
Override firmware for template based VMs on creation

### LINKED ISSUE #1307: 'firmware' field is not applied when creating VM from template or VM copy
Hello,
I am unable to set `firmware = "efi"`. It is automatically changed to `firmware = "bios"` (see below). Only when I also set `efi_secure_boot = true` firmware will also be set to "efi".

### VMware Cloud Director version
10.5.1.23400185

### Terraform Version
Terraform v1.9.4
on linux_amd6

### Affected Resource(s)
- vcd_vapp_vm

```hcl
resource "vcd_vapp_vm" "vm" {
  name             = var.VM_NAME
  org              = var.VCD_ORG
  vdc              = var.VCD_ZONE
  vapp_name        = vcd_vapp.vapp.name

  memory                 = 2048
  cpus                   = 2
  cpu_cores              = 1
  cpu_hot_add_enabled    = true
  memory_hot_add_enabled = true
  hardware_version       = "vmx-17"
  firmware               = "efi"
  boot_options {
    efi_secure_boot = false
  }

  network {
    type               = "org"
    name               = vcd_vapp_org_network.network.org_network_name
    ip_allocation_mode = "POOL"
    is_primary         = true
  }
}
```

### Debug Output
```
2024-08-13T10:35:56.611+0200 [WARN]  Provider "provider[\"registry.terraform.io/vmware/vcd\"]" produced an unexpected new value for vcd_vapp_vm.vm, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .firmware: was cty.StringVal("efi"), but now cty.StringVal("bios")
      - .boot_options[0].enter_bios_setup_on_next_boot: was null, but now cty.False
```
### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
`firmware = "efi"`

### Actual Behavior
`firmware = "bios"`

### Steps to Reproduce
1. `terraform apply`
2. See that firmware is set to BIOS and not EFI.

### User Access rights 
Information about user used. Role and/or more exact rights if it is customized.
Organization Operator

### Important Factoids
Is there anything atypical about your accounts that we should know?
I am using ubuntu ova cloud image uploaded to VMware Cloud Director

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:

Didainius:
Closes #1307 

There is a case where creating a VM from template with `bios` `firmware` does not change it to `efi` if set. The reason is that VM create operation did not do that in time. This PR attempts to solve it with a relevant test. More details in #1307.


lvirbalas (approved):
Looks great!

adambarreiro (approved):

PR #1312, opened 2024-08-29
Fix issue #1262

### LINKED ISSUE #1262: iops view rights
In Cloud Director 10.5 and higher, you can disable IOPS viewing rights for tenants
iops policy is enabled and inherited from provider vdc
If you push a role to tenants without "iops view", then terraform will not be able to create a VM

adambarreiro:
This PR closes #1262

## Problem

When a user creates a VM with a VM template, and she does not have the "Organization vDC Disk: View IOPS" right, the VCD Provider crashes with a panic:

```
goroutine 786 [running]:
github.com/vmware/terraform-provider-vcd/v3/vcd.updateTemplateInternalDisks(0xc000054680, {0xcd4e980, 0xc000c8e870}, {0xc0003902c0, 0xc00017f088})
	/Users/adambarreiro/terraform-provider-vcd/vcd/resource_vcd_vapp_vm_tools.go:700 +0x945
github.com/vmware/terraform-provider-vcd/v3/vcd.createVmFromImage(0xc000054680, {0xcd4e980, 0xc000c8e870}, {0xc6a4f44, 0x6}, {0xc6b662b, 0x10})
	/Users/adambarreiro/terraform-provider-vcd/vcd/resource_vcd_vapp_vm.go:1347 +0x3886
github.com/vmware/terraform-provider-vcd/v3/vcd.genericResourceVmCreate(0xc000054680, {0xcd4e980, 0xc000c8e870}, {0xc6a4f44, 0x6})
	/Users/adambarreiro/terraform-provider-vcd/vcd/resource_vcd_vapp_vm.go:893 +0x395
```

HCL to reproduce:

```hcl
provider "vcd" {
  user                 = "abarreiro-admin" # User without Organization vDC Disk: View IOPS right!!
  password             = "******"
  token                = ""
  api_token            = ""
  saml_adfs_rpt_id     = ""
  url                  = "https://vcd-url/api"
  sysorg               = "abarreiro"
  org                  = "abarreiro"
  vdc                  = "nsxt-vdc-abarreiro"
  allow_unverified_ssl = "true"
  max_retry_timeout    = 600
  logging              = true
  logging_file         = "go-vcloud-director.log"
}

data "vcd_catalog" "my-catalog" {
  name = "cat-abarreiro-nsxt-backed"
}

data "vcd_catalog_vapp_template" "template" {
  catalog_id = data.vcd_catalog.my-catalog.id
  name       = "nsxt-photon-hw11"
}

resource "vcd_vm" "my_vm" {
  name             = "my_vm_test_01"
  memory           = 512
  cpus             = 1

  override_template_disk {
    bus_type        = "paravirtual"
    size_in_mb      = "16384"
    bus_number      = 0
    unit_number     = 0
    storage_profile = "*"
  }
}
```

## Solution

In code, it was assumed that the IOPS information was always present. The fix consists of safely checking for `nil` pointers.

## Tests

A new test is provided: `TestAccVcdVm_WithoutOrganizationVdcDiskIopsRights`. This test will panic without this fix (hence will remove the `Organization vDC Disk: View IOPS` from your VCD testing org user, be careful and restore it afterwards).
With this patch, the test obviously should pass.

What it does, it removes the mentioned rights, applies the HCL, then restores the same rights.

### DIFF
--- a/vcd/resource_vcd_vapp_vm_test.go
+++ b/vcd/resource_vcd_vapp_vm_test.go
@@ -476,6 +476,7 @@ func TestAccVcdVm_WithoutIopsRights(t *testing.T) {
 	var params = StringMap{
 		"ProviderVcdOrg": providerVcdOrg1,
 		"Org":            testConfig.VCD.Org,
+		"Name":           t.Name(),
 		"Vdc":            testConfig.Nsxt.Vdc,
 		"Catalog":        testConfig.VCD.Catalog.NsxtBackedCatalogName,
 		"CatalogItem":    testConfig.VCD.Catalog.NsxtCatalogItem,
--- a/vcd/resource_vcd_vapp_vm_test.go
+++ b/vcd/resource_vcd_vapp_vm_test.go
@@ -543,6 +543,7 @@ func TestAccVcdVm_WithoutIopsRights(t *testing.T) {
 					}
 				},
 				Check: resource.ComposeTestCheckFunc(
+					// Nothing to test here, we just want to assert that no panic is thrown anymore
 					resource.TestCheckResourceAttr("vcd_vm.my_vm", "name", t.Name()),
 				),
 			},
--- a/vcd/resource_vcd_vapp_vm_test.go
+++ b/vcd/resource_vcd_vapp_vm_test.go
@@ -4,6 +4,8 @@ package vcd
 
 import (
 	"fmt"
+	"github.com/vmware/go-vcloud-director/v2/types/v56"
+	"net/url"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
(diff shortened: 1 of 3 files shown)


vcd/resource_vcd_vapp_vm_test.go
```diff
@@ -466,6 +468,133 @@ resource "vcd_vm" "boot-image-vm" {
...
+				PreConfig: func() {
+					role, err := vcdClient.Client.GetGlobalRoleByName("Organization Administrator")
+					if err != nil {
+						if govcd.ContainsNotFound(err) {
+							t.Skipf("could not find required global role Organization Administrator")
+						}
+						t.Fatal(err)
+					}
+					filter := make(url.Values)
+					filter.Set("filter", "name==Organization vDC Disk*")
```
lvirbalas, line 529: Nit/request to clarify. The right here is called "Organization vDC Disk", yet the function comment doesn't mention that - only "IOPS rights". Hence, it's not very easy to follow the function through when reading without context.

Maybe you could extend/align the right name in function description more?

vcd/resource_vcd_vapp_vm_test.go
```diff
@@ -466,6 +468,133 @@ resource "vcd_vm" "boot-image-vm" {
...
+					existingRights, err := role.GetRights(filter)
+					if err != nil {
+						t.Fatal(err)
+					}
+					if len(existingRights) == 0 {
+						return
+					}
+					for _, right := range existingRights {
```
lvirbalas, line 537: Small comment about why we're looping here may be helpful when reading later (I mean, that it's, probably, more than one "IOPS" right).

lvirbalas (commented):
Some comment polishing/clarification asks.

vcd/resource_vcd_vapp_vm_test.go
adambarreiro: Done, changed the test name with full right name

vcd/resource_vcd_vapp_vm_test.go
adambarreiro: Done, added a few useful comments.

vcd/resource_vcd_vapp_vm_tools.go
```diff
@@ -695,7 +695,7 @@ func updateTemplateInternalDisks(d *schema.ResourceData, meta interface{}, vm go
 		}
 
 		// Update details of internal disk for disk existing in template
-		if value, ok := internalDiskProvidedConfig["iops"]; ok {
+		if value, ok := internalDiskProvidedConfig["iops"]; ok && diskCreatedByTemplate.IopsAllocation != nil {
```
Didainius, line 698: I have a few thoughts for discussion

The way it checks now is - if VM that was spun from template has `diskCreatedByTemplate.IopsAllocation` defined - we're setting the `iops` field. Otherwise we ignore.

On one hand - it validates if the customer is able to possibly read (right `View Disk IOPS`) the disk structure. However we'd still not be sure if he has right `Edit Disk IOPS` and in the end it still might cause panic.

Additionally, given how complicated the APIs may be in some cases - it could be that the VM template simply didn't have this `diskCreatedByTemplate.IopsAllocation` structure and VCD doesn't return it? If such case happens (I don't know if it can happen) - we might start ignoring `iops` even if the user set it that could lead to another bug.

The issue that user has (#1262) - he is not setting the `iops` field (at least in his sample snippet). But the condition `if value, ok := internalDiskProvidedConfig["iops"]; ok {` looks to be always `true` because the field is `schema.TypeInt` (which is `0`, when unset). 

My thought is - should we additionally validate another condition - that the user has set the field `iops` explicitly to be non zero (`value.(int) != 0`). If so - even if he didn't have rights (`diskCreatedByTemplate.IopsAllocation == nil`, to be precise)  - we could initialise the structure and send it. We'd prevent panic and the user would get API error that he doesn't have rights. He would just then know that he shouldn't set `iops` field.

_My thoughts in code (unvalidated)_

``` go
if value := internalDiskProvidedConfig["iops"]; value.(int) != 0 || diskCreatedByTemplate.IopsAllocation != nil {
	if diskCreatedByTemplate.IopsAllocation == nil {
		diskCreatedByTemplate.IopsAllocation = &types.IopsResource{}
	}
	iops := int64(value.(int))
	diskCreatedByTemplate.IopsAllocation.Reservation = iops
}
```

adambarreiro: Actually that's a good point, I'll reiterate the implementation with this in mind and push an update.

adambarreiro: Changed. Test passed! I'll run `vm` tag tests just in case, before merging.

vcd/resource_vcd_vapp_vm_test.go
```diff
@@ -466,6 +468,138 @@ resource "vcd_vm" "boot-image-vm" {
...
+					if err != nil {
+						t.Fatal(err)
+					}
+					// If the rights are not present in that role, there's nothing else to do
+					if len(existingRights) == 0 {
+						return
+					}
+					// Convert the obtained rights to OpenApiReference to be able to remove them.
+					// We save them in the outer-scoped "modifiedRights" so they can be re-added later
+					for _, right := range existingRights {
```
Didainius, line 542: NIT, but you might find useful `convertSliceOfStringsToOpenApiReferenceIds, `extractIdsFromOpenApiReferences` and `extractNamesFromOpenApiReferences`

adambarreiro: Yeah, I had a look at `convertSliceOfStringsToOpenApiReferenceIds` but unfortunately I don't have a plain `[]string`, but `[]*types.Right`. Maybe I could think about a generic function and reflection, but probably too much.

adambarreiro:
`vm` tests passed

lvirbalas (approved):

PR #1317, opened 2024-09-03
Make ip_allocation_mode required in vcd_vapp_vm / vcd_vm

### LINKED ISSUE #1205: terraform apply returns error 500 if ip_allocation_mode not specified, however plan works fine.
### Terraform Version

Terraform v1.4.7
on linux_amd64
+ provider registry.terraform.io/vmware/vcd v3.11.0

### Affected Resource(s)

- vcd_vapp_vm

```hcl
resource "vcd_vapp_vm" "vm" {
  count = var.vm_count
  vapp_name = "${var.env}-vm"
  name      = "${var.env}-${count.index}"
  memory = 4096
  cpus = 4
  cpu_cores = 4
  
 
 dynamic "network" {
  for_each = values(var.vlan_names)
  content {
    type = "org"
    name = data.vcd_network_direct.hrs-vlan.name
    #ip_allocation_mode = "MANUAL"
    ip = network.value[count.index]
  }
 }
 
 
}
```

### Debug Output

https://gist.github.com/zeol/a6e532dd0e72f4e01a453048f7fb1c99

### Expected Behavior

Plan should fail, with some descriptive error

### Actual Behavior
API returned 500 

│ Error: error creating VM from template: [VM creation] error getting VM <VM_NAME> : error instantiating a new VM: API Error: 500: [ f5956605-ac3b-4540-8fa2-665489c2ad86 ] Unable to perform this action. Contact your cloud administrator.

### Steps to Reproduce

try to create a resource vcd_vapp_vm with network and ip assigned, but no ip_allocation_mode (as in the code above)

### User Access rights 
Organization administrator

adambarreiro:
Closes #1205

## Description

VCD does not allow to specify a NIC without setting an IP Allocation mode option, so it doesn't make sense that `ip_allocation_mode` (inside a `network` block) is Optional in the schema.

Despite this should be a breaking change (the change from `Optional -> Required` could force existing configurations to be updated to set the `ip_allocation_mode` argument on each `network` block), the configurations that don't have `ip_allocation_mode` will always fail (see #1205), so this can't break any configuration as they would be already broken.

## Test

There are no changes in tests as this is just a change in schema. However, it can be tested with the following script.

- Without the changes from this PR, it fails at runtime with a 500 API error.
- With the changes, it fails during planning with a more human-friendly error.

```hcl
terraform {
  required_providers {
    vcd = {
      source  = "vmware/vcd"
      version = ">= 3.14.0"
    }
  }
}

provider "vcd" {
  password             = "******"
  url                  = "https://vcd.my-company.com/api"
  org                  = "ab1060"
  vdc                  = "nsxt-vdc-ab1060"
  allow_unverified_ssl = "true"
  max_retry_timeout    = 600
  logging              = true
  logging_file         = "go-vcloud-director.log"
}

data "vcd_catalog" "catalog" {
  name = "cat-ab1060-nsxt-backed"
}

data "vcd_catalog_vapp_template" "template" {
  catalog_id = data.vcd_catalog.catalog.id
  name       = "nsxt-photon-hw11"
}

resource "vcd_vm" "vm" {
  name             = "vm"
  memory           = 1024
  cpus             = 1
  cpu_cores        = 1

  network {
    type = "org"
    name = "foo"
    #ip_allocation_mode = "MANUAL"
    ip = "1.2.3.4"
  }
}
```


scripts/skip-upgrade-tests.txt
```diff
@@ -356,3 +356,4 @@ vcd.ResourceSchema-vcd_ip_space_uplink.tf v3.13.0 "Added 'associated_interface_i
 vcd.ResourceSchema-vcd_org.tf v3.13.0 "Added account_lockout block"
 vcd.ResourceSchema-vcd_rde_interface_behavior.tf v3.13.0 "Changed fields to computed and added new arguments"
 vcd.ResourceSchema-vcd_rde_type_behavior.tf v3.13.0 "Changed fields to computed and added new arguments"
+vcd.ResourceSchema-vcd_vm_sizing_policy.tf v3.13.0 "Changed description"
```
adambarreiro, line 359: Not coming from this PR, but spotted it

vcd/resource_vcd_vapp_vm.go
```diff
@@ -329,7 +329,7 @@ func vmSchemaFunc(vmType typeOfVm) map[string]*schema.Schema {
 						Description:  "Network type to use: 'vapp', 'org' or 'none'. Use 'vapp' for vApp network, 'org' to attach Org VDC network. 'none' for empty NIC.",
 					},
 					"ip_allocation_mode": {
-						Optional:     true,
+						Required:     true,
```
Didainius, line 332: Mainly one query if you could check if there is a test of a VM that does not have `network` declared at all and it passes.

I think part of the reason that all fields where optional is that if one had a `Required` field within nested list - it would automatically force the root component `network` to be required. This was especially true when this resource had (long time ago)only two plain fields to specify a `network_name` and `ip`. The new `network` block had to coexist until we released 3.0 and removed the deprecated fields.

adambarreiro: Good point, will run all `vm` tests and get back to this comment

Didainius (approved):

lvirbalas (approved):

PR #1337, opened 2024-10-01
Patch panic and improve doc clarity

Didainius:
This PR patches 2 things:
* A possible panic in `vcd_org_group` panic
* Clarifies in docs that `network_dhcp_wait_seconds` in `vcd_vapp_vm` and `vcd_vm` would only be able to find IP address in DHCP lease pool in NSX-V (NSX-T only works with guest IP reporting)


adambarreiro (approved):
Thanks!

lvirbalas (approved):

PR #1338, opened 2024-10-01
Add org, region resources and  org, vdc, region, supervisor data sources

Didainius:
Changelog contains the list of implemented resources. 
`make testtm` works

PR #1339, opened 2024-10-03
Add vcd_tm_region_storage_policy + vcd_tm_content_library resource and data source

adambarreiro:
Adds data source `vcd_tm_region_storage_policy` Region Storage Policies and `vcd_tm_content_library` data source and resource for Content Libraries.

Some changes are copied from https://github.com/vmware/terraform-provider-vcd/pull/1338 (not pulled), so it should not give conflicts after merged.

**Disclaimer**: There are a lot of unresolved topics marked with `TODO: TM:`, so this implementation will need a revisit.


.github/workflows/check-docs.yml
```diff
@@ -21,6 +21,9 @@ jobs:
       with:
         fetch-depth: 0  # Required to have tag information available
 
+    - name: Install Terraform binary
+      uses: hashicorp/setup-terraform@v3
+
```
Didainius, line 26: There are a few docs that need fixing

go.mod
```diff
@@ -69,3 +69,5 @@ require (
 	google.golang.org/grpc v1.63.2 // indirect
 	google.golang.org/protobuf v1.34.0 // indirect
 )
+
+replace github.com/vmware/go-vcloud-director/v2 => github.com/adambarreiro/go-vcloud-director/v2 v2.17.0-alpha.1.0.20241011135537-f92ca39ded9c
```
Didainius, line 73: Should this be removed now as your PR is merged for govcd

vcd/resource_vcd_tm_content_library.go
```diff
@@ -0,0 +1,269 @@
...
+		UpdateContext: resourceVcdTmContentLibraryUpdate,
+		DeleteContext: resourceVcdTmContentLibraryDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceVcdTmContentLibraryImport,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true, // TODO: Update not supported
```
Didainius, line 26: There are many more missed updates, but only mentioning here.

```suggestion
				ForceNew:    true, // TODO: TM: Update not supported
```

go.mod
adambarreiro: Indeed, saving this comment as unresolved as reminder.

vcd/resource_vcd_tm_content_library.go
adambarreiro: Fixed, thanks!

Didainius (approved):
🎉

vcd/resource_vcd_tm_content_library_test.go
```diff
@@ -0,0 +1,93 @@
...
+				Config: configText1,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "name", t.Name()),
+					resource.TestCheckResourceAttr(resourceName, "description", t.Name()),
+					resource.TestCheckResourceAttr(resourceName, "storage_policy_ids.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "auto_attach", "true"), // TODO: TM: Test with false
+					resource.TestCheckResourceAttrSet(resourceName, "creation_date"),
+					resource.TestCheckResourceAttr(resourceName, "is_shared", "true"),        // TODO: TM: Test with false
+					resource.TestCheckResourceAttr(resourceName, "is_subscribed", "false"),   // TODO: TM: Test with true
+					resource.TestCheckResourceAttr(resourceName, "library_type", "PROVIDER"), // TODO: TM: Test with tenant catalog
```
lvirbalas, line 51: Would it be possible to add these TODO tests in this PR still?

adambarreiro: I don't think so, for publishing and subscribe we need the option to be able to publish a catalog, which does not exist yet (only in VCD).
For _tenant_ catalogs I'm not 100% sure we can do it now, but I'll investigate on next PRs.

lvirbalas: Got it. Thanks!

lvirbalas (approved):

PR #1348, opened 2024-10-30
Region, Supervisor and Supervisor Zones

Didainius:
This PR adds:
* **New Resource:** `vcd_tm_region` to manage Regions [GH-1348]
* **New Data Source:** `vcd_tm_region` to read Regions [GH-1348]
* **New Data Source:** `vcd_tm_supervisor` to read Supervisors [GH-1348]
* **New Data Source:** `vcd_tm_supervisor_zone` to read Supervisor Zones [GH-1348]

Notes:
* `vcd_tm_region` does not support update (underlying API does not support it)
* `vcd_tm_region` read does not return storage profiles and is_enabled values.

Other notes:
* `make testtm-binary-prepare` to generate binary test configs for `tm` tag
* Resource `vcd_vcenter` adds additional flag `refresh_policies_on_read` to refresh policies (it works the same way that `refresh_vcenter_on_read` works)
* Tune `getTypeFunc` in `crudConfig` to also include `*VCDClient`


adambarreiro (commented):
First scan

website/docs/d/tm_region.html.markdown
```diff
@@ -0,0 +1,30 @@
...
+layout: "vcd"
+page_title: "VMware Cloud Director: vcd_tm_region"
+sidebar_current: "docs-vcd-datasource-tm-region"
+description: |-
+  Provides a data source to read Regions.
+---
+
+# vcd\_tm\_region
+
+Provides a data source to read Regions.
```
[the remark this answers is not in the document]
Didainius: Adjusted all new things in this PR

adambarreiro (approved):
Tests passed. All ok from my side, probably this will need a revisit for update, so :)

lvirbalas:
> vcd_tm_region does not support update (underlying API does not support it)

So, currently Terraform doesn't complain if you, for example, update description. Instead, it forces replacement. That's handy while the region is empty.

Didainius:
> > vcd_tm_region does not support update (underlying API does not support it)
> 
> So, currently Terraform doesn't complain if you, for example, update description. Instead, it forces replacement. That's handy while the region is empty.

Yes, because all schema fields have `ForceNew: true`  for now

vcd/resource_vcd_tm_region_test.go
```diff
@@ -0,0 +1,153 @@
...
+package vcd
+
+import (
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+// TODO: TM: adjust the test for testing update operation once available
```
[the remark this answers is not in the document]
Didainius: I can do this, but every update operation attempt will return you an error for now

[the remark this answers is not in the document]
Didainius: I adjusted the test to perform modification of description field in additional step and also added a check to ensure that the Region ID becomes different (that ensures it was recreated).

website/docs/d/tm_supervisor.html.markdown
```diff
@@ -0,0 +1,32 @@
...
+
+  depends_on = [vcd_vcenter.one]
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of Supervisor.
```
lvirbalas, line 27: Very similar question to the one I had about storage policies. Aren't we missing vcenter_id here?

Also, as it stands right now, you need to have depends_on vCenter clause to make it work in a fresh setup:

```
data "vcd_tm_supervisor" "test" {
  name = var.supervisor_name

  depends_on = [vcd_vcenter.test]
}
```

website/docs/r/tm_region.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+```hcl
+data "vcd_tm_supervisor" "one" {
+  name = "first-supervisor"
+
+  depends_on = [vcd_vcenter.one]
+}
+
+resource "vcd_tm_region" "one" {
+  name                 = "region-one"
+  is_enabled           = true
```
lvirbalas, line 26: I can't seem to disable the region. It still shows Status = READY after successful apply.

```
resource "vcd_tm_region" "test" {
  name                 = "region-lino-tf"
  description          = "Region made by Terraform"
  is_enabled           = false
  nsx_manager_id       = vcd_nsxt_manager.test.id
  supervisor_ids       = [data.vcd_tm_supervisor.test.id]
  storage_policy_names = [var.storage_policy_name]
}
```

Didainius: You cannot, the field doesn't work yet https://github.com/vmware/terraform-provider-vcd/pull/1348/files#diff-b102b402cd712cd79b740b0e8ee9394aab65d13e0c703e4500ce5217cfea1619R195

website/docs/d/tm_supervisor.html.markdown
Didainius: This is now patched to require `vcenter_id`

website/docs/r/vcenter.html.markdown
```diff
@@ -37,6 +37,9 @@ The following arguments are supported:
 * `refresh_vcenter_on_read` - (Optional) An optional flag to trigger refresh operation on the
   underlying vCenter on every read. This might take some time, but can help to load up new artifacts
   from vCenter (e.g. Supervisors). Update is a no-op. Default `false`.
```
lvirbalas, line 39: I'm thinking whether we should add a note when it is not useful to do so, as it is a heavy operation and it even produces a visible task in UI. Or at least mention that this produces that task, so users know where it is coming from.

Didainius: Made an attempt to rephrase both - this and `refresh_policies_on_read`

website/docs/r/vcenter.html.markdown
```diff
@@ -35,8 +35,15 @@ The following arguments are supported:
 * `username` - (Required) A username for authenticating to vCenter server
 * `password` - (Required) A password for authenticating to vCenter server
 * `refresh_vcenter_on_read` - (Optional) An optional flag to trigger refresh operation on the
-  underlying vCenter on every read. This might take some time, but can help to load up new artifacts
-  from vCenter (e.g. Supervisors). Update is a no-op. Default `false`.
+  underlying vCenter (that is visible as a separate task in UI) on every read. This might take some
+  time, but can help to load up new artifacts from vCenter (e.g. Supervisors). Update is a no-op.
+  Default `false`. It may be useful after adding vCenter or if new infrastructure is added to
```
lvirbalas, line 40:
```suggestion
  underlying vCenter on every read. This might take some
  time, but can help to load up new artifacts from vCenter (e.g. Supervisors).
  This operation is visible as a new task in UI. Update is a no-op.
  Default `false`. It may be useful after adding vCenter or if new infrastructure is added to
```

lvirbalas (approved):
Fantastic job!

PR #1349, opened 2024-11-06
Fix Content Library resource and Region Storage Policy data source

adambarreiro:
- `vcd_tm_content_library` has changed the name of `storage_profiles` to `storage_classes`
- `vcd_tm_region_storage_policy` data source requires `region_id` now

Created testing functions `getVCenterHcl`, `getNsxManagerHcl`, `getRegionHcl` that allow to fetch existing items or create them on the fly.


website/docs/r/tm_content_library.html.markdown
```diff
@@ -37,7 +37,7 @@ resource "vcd_tm_content_library" "cl" {
 The following arguments are supported:
 
 * `name` - (Required) The name of the Content Library
-* `storage_policy_ids` - (Required) A set of [Region Storage Policy](/providers/vmware/vcd/latest/docs/resources/tm_region_storage_policy) IDs or VDC Storage Policy IDs used by this Content Library
+* `storage_class_ids` - (Required) A set of [Region Storage Policy](/providers/vmware/vcd/latest/docs/resources/tm_region_storage_policy) IDs or VDC Storage Policy IDs used by this Content Library
```
Didainius, line 40: Double checking. Is this really still true. Is it not only region storage policy IDs?

adambarreiro: Yep, forgot to change. Put here the same description as API docs, with a reference to the data source

Didainius (approved):

website/docs/r/tm_content_library.html.markdown
```diff
@@ -37,7 +37,7 @@ resource "vcd_tm_content_library" "cl" {
 The following arguments are supported:
 
 * `name` - (Required) The name of the Content Library
-* `storage_policy_ids` - (Required) A set of [Region Storage Policy](/providers/vmware/vcd/latest/docs/resources/tm_region_storage_policy) IDs or VDC Storage Policy IDs used by this Content Library
+* `storage_class_ids` - (Required) A set of [Storage Class IDs](/providers/vmware/vcd/latest/docs/data-sources/tm_region_storage_policy) used by this Content Library
```
lvirbalas, line 40: If it's renamed to "class", why are we still using "policy" in data source?

adambarreiro: The name in its API didn't change for Region Storage Policies (yet?). I guess it will be done "soon" to have both aligned and we will be able to change the data source name then

lvirbalas: Let's add a TODO: TM: here as a reminder to check for API change.

adambarreiro: Done

lvirbalas (approved):

PR #1351, opened 2024-11-14
Improve docs and validation wording for `vcd_tm_org` resource

Didainius:
This PR does two things:
* Improves docs and validation message for when the `name` can be changed (requires the Org to transition from or to disabled state as was discussed in https://github.com/vmware/terraform-provider-vcd/pull/1345#discussion_r1812748530
* Adds `make testtm-coverage` command that would storage test coverage profile `tm.cover`


lvirbalas (approved):
Thanks!

adambarreiro (approved):

PR #1369, opened 2024-12-12
Add support of classic orgs to vcd_tm_org

adambarreiro:
This PR changes `is_classic_tenant` from Computed to Optional, as it can be set to create soft/classic tenants in TM.

While in UI Classic tenants are limited to 1, there's no such constraint in API.

Classic tenants cannot be updated to regular tenants.

### DIFF
--- a/.github/workflows/check-security.yml
+++ b/.github/workflows/check-security.yml
@@ -1,5 +1,13 @@
 name: Run Gosec
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
+      - '**.md'
+      - 'website/**'
+  pull_request:
+    paths-ignore:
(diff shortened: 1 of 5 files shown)


.github/workflows/check-security.yml
```diff
@@ -1,5 +1,13 @@
 name: Run Gosec
-on: [push, pull_request]
+on:
```
adambarreiro, line 2: Reminder to create PR in `release/v3.x`

Didainius (approved):
All Org tests (new and old) and also Org VDC test still passes, thanks for fixing typos!

lvirbalas (approved):

PR #1361, opened 2024-12-04
TM Provider Gateway 

Didainius:
This PR adds resources and data source `vcd_tm_provider_gateway` for managing TM Provider Gateways. In addition, there is a new data source `vcd_tm_tier0_gateway` for lookup of NSX-T Tier 0 Gateways

PR #1370, opened 2024-12-12
Adjust internal generic function signatures to include *VCDClient

Didainius:
Add `*VCDClient` parameter to `crudConfig.stateStoreFunc` and `dsReadConfig.stateStoreFunc`
It will be used #1361, but splitting it off so that #1361 is less crowded.

### DIFF
--- a/vcd/datasource_vcd_tm_region_zone.go
+++ b/vcd/datasource_vcd_tm_region_zone.go
@@ -79,7 +79,7 @@ func resourceVcdTmRegionZoneRead(ctx context.Context, d *schema.ResourceData, me
 	return readDatasource(ctx, d, meta, c)
 }
 
-func setZoneData(d *schema.ResourceData, z *govcd.Zone) error {
+func setZoneData(_ *VCDClient, d *schema.ResourceData, z *govcd.Zone) error {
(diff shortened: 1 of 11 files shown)


adambarreiro (approved):

vcd/datasource_vcd_tm_region_zone.go
```diff
@@ -79,7 +79,7 @@ func resourceVcdTmRegionZoneRead(ctx context.Context, d *schema.ResourceData, me
 	return readDatasource(ctx, d, meta, c)
 }
 
-func setZoneData(d *schema.ResourceData, z *govcd.Zone) error {
+func setZoneData(_ *VCDClient, d *schema.ResourceData, z *govcd.Zone) error {
```
lvirbalas, line 82: Sorry, a bit confused. Why are we adding a blank VCDClient identifier which is not used afterwards?

Didainius: We're not in need it in this particular PR, but generic function signature requires it now. It is to be used in #1361 and possibly other more complicated entries

lvirbalas (approved):

PR #1416, opened 2026-02-04
Update install path to schubergphilis

schans:


schans:
Sorry for the noise... this was not intended.

